PR #6, opened 2021-07-04
Docs fixes for shell compatibility

Nitive:
### Problem

`docker run` fails with error `Error response from daemon: create ./dhctl-tmp: "./dhctl-tmp" includes i
nvalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a h
ost directory, use absolute path.
See 'docker run --help'.`. I have reproduced the problem on Linux and macOS

### Solution

Use absolute path like error message suggests

### Problem

Fish shell uses different syntax for inline commands so command can't be copy-pasted

bash: `echo "$(pwd)"`
fish: `echo "(pwd)"`

It's not a big deal because all fish users know it but it would be nice to use $PWD instead of $(pwd) to make fish users life a litter easier

### Solution

Use $PWD instead of $(pwd)

### Problem

Weird error message when current directory or $HOME contains space characters

```sh
$ pwd
/Users/user/Desktop/deckhouse test

$ docker run -it -v $(pwd)/config.yml:/config.yml -v $HOME/.ssh/:/tmp/.ssh/ -v $(pwd)/dhctl-tmp:/tmp registry.de
ckhouse.io/deckhouse/ee/install:alpha bash
Unable to find image 'with:latest' locally
```

### Solution

Enclose directories names in quotation marks

### DIFF
--- a/dhctl/README.md
+++ b/dhctl/README.md
@@ -119,8 +119,8 @@ For example, use a docker image from the Flant docker registry:
   * `config.yaml` - configuration file for the cluster bootstrap as described above.
      ```bash
      docker run -it \
-       -v $(pwd)/config.yaml:/config.yaml \
-       -v $HOME/.ssh/:/tmp/.ssh/ \
+       -v "$(pwd)/config.yaml:/config.yaml" \
+       -v "$HOME/.ssh/:/tmp/.ssh/" \
        registry.deckhouse.io/fe/install:alpha \
        bash
      ```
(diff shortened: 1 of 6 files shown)


z9r5:
Thank you for contributing, @Nitive !

Merged in https://github.com/deckhouse/deckhouse/commit/cfc6c98645b249c9a5cc8e4033dce9e8ed6026e6.

This PR will be closed.

PR #8, opened 2021-07-07
Fix NodeGroup apiVersion conversion on bare metal clusters

Nitive:
### Problem

Error `Invalid character in input stream.` during v1alpha → v1 convervion for NodeGroup. The issue can be reproduced only on bare metal clusters where no `kube-system/d8-provider-cluster-configuration` secret exists.

### Reason

`jq '.snapshots.cluster_config[0].filterResult'` returns `""` which can't be base64 decoded and script fails.

### Solution

Use `-r` flag to return empty string as no symbols instead fo `""`

### DIFF
--- a/modules/040-node-manager/webhooks/conversion/node_group
+++ b/modules/040-node-manager/webhooks/conversion/node_group
@@ -99,7 +99,7 @@ EOF
 
 
 function __on_conversion::alpha2_to_v1() {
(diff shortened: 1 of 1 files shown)


nabokihms (approved):
Thank you @Nitive. Good catch!

z9r5:
Merged in https://github.com/deckhouse/deckhouse/commit/10021f3b260a88b8fbebe6aa1cc3dc6a9cf61d73

@Nitive Thank you.

PR #36, opened 2021-08-08
Increase pod deletion timeout in Upmeter scheduler monitoring

### LINKED ISSUE #35: Upmeter scheduler monitoring shouldn't fail on pod deletion timeout
Hello! I have a bad upmeter statistic for scheduler — about 30% of scheduler tests are failing. I've dug into upmeter source code and found out this happening not because something is wrong with a scheduler but because removal of pod is taking more then [specified timeout of 5 seconds](https://github.com/deckhouse/deckhouse/blob/main/modules/500-upmeter/images/upmeter/pkg/probe/controlplane.go#L91) sometimes.

Failing to remove pod in 5 seconds is probably not the best behavior but it definitely doesn't indicate problem with scheduler so we need to fix it. The easiest way is to increase pod deletion timeout but maybe there shouldn't be error at all even when pod deletion takes a lot time

Nitive:
Failing to delete pod doesn't indicate that there is a problem with scheduler so test shouldn't fail in this case.

I've increased deletion timeout to 20 seconds because that what we have left before next round of test is executed in worst case scenario:

5s of creation + 20s of scheduling + 20s of deletion + 10s of GC + 5s just in case = 60s

Fixes #35

### DIFF
--- a/modules/500-upmeter/images/upmeter/pkg/probe/checker/k8s_pod.go
+++ b/modules/500-upmeter/images/upmeter/pkg/probe/checker/k8s_pod.go
@@ -80,7 +80,7 @@ func (c *podLifecycleChecker) Check() check.Error {
 /*
 1. check control-plane
 2. collect garbage
-2. crate pod                    (podCreationTimeout)
+2. create pod                   (podCreationTimeout)
 3. see the pod is scheduled     (podScheduledTimeout)
 4. delete the pod               (podDeletionTimeout)
 	+ensure the pod is not listed
--- a/modules/500-upmeter/images/upmeter/pkg/probe/controlplane.go
+++ b/modules/500-upmeter/images/upmeter/pkg/probe/controlplane.go
@@ -88,7 +88,7 @@ func ControlPlane(access kubernetes.Access) []runnerConfig {
 				Node:                     os.Getenv("NODE_NAME"),
 				CreationTimeout:          5 * time.Second,
 				SchedulingTimeout:        20 * time.Second,
-				DeletionTimeout:          5 * time.Second,
+				DeletionTimeout:          20 * time.Second,
 				GarbageCollectionTimeout: gcTimeout,
 			},
 		},


modules/500-upmeter/images/upmeter/pkg/probe/checker/k8s_pod.go
```diff
@@ -107,7 +107,7 @@ func (c *podLifecycleChecker) new(pod *v1.Pod) check.Checker {
 
 	collectGarbage := newGarbageCollectorCheckerByLabels(c.access, pod.Kind, c.namespace, pod.Labels, c.garbageCollectionTimeout)
 
-	timeout := c.creationTimeout + c.schedulingTimeout + c.deletionTimeout
+	timeout := controlPlaneChecker.Timeout + c.garbageCollectionTimeout + c.creationTimeout + c.schedulingTimeout + c.deletionTimeout
```
Nitive, line 110: I'm not sure if control plane checker and garbage collection timeouts were left out on purpose or not

modules/500-upmeter/images/upmeter/pkg/probe/controlplane.go
```diff
@@ -88,7 +88,7 @@ func ControlPlane(access kubernetes.Access) []runnerConfig {
 				Node:                     os.Getenv("NODE_NAME"),
 				CreationTimeout:          5 * time.Second,
 				SchedulingTimeout:        20 * time.Second,
-				DeletionTimeout:          5 * time.Second,
+				DeletionTimeout:          20 * time.Second,
```
shvgn, line 91: Looks reasonable.

shvgn (changes requested):
I suggest reverting the timeout calculation in probe for the time being. Nothing else blocks this PR from merging.

modules/500-upmeter/images/upmeter/pkg/probe/checker/k8s_pod.go
Nitive: Ok, I've reverted the change back

shvgn (approved):
LGTM

PR #37, opened 2021-08-08
[candi] Fix templating of proxy uri

alexey-gusarov:
specification: https://github.com/deckhouse/deckhouse/blob/main/candi/bashible/openapi.yaml#L28


nabokihms (approved):
Looks good, thanks.

PR #296, opened 2021-11-16
Changelog: fix github syntax quotes

shvgn:
## Description

There is a double quote issue that we encounter when PRs are merged. 

See https://github.com/actions/runner/issues/866

Fixes the changelog workflow results


shvgn:
merged elsewhere

PR #364, opened 2021-12-10
[chrony] fix rollout restart time

RomanenkoDenys:
## Description

1. Fixed rollout restart time  for chrony daemonset (now it is about one minute per pod).
2. Added probe to check if NTP port on node is used.

1. If we changes chrony config, changes applies takes large amount of time.
2. We don't know if NTP port on node is used by another software.

## Changelog entries

```changes
module: d8-chrony
type: fix
description: "Fixed rollout restart time fo chrony daemonset."
note: d8-chrony module will be restarted.
```


nabokihms (approved):
Looks good, thanks.

konstantin-axenov (approved):

konstantin-axenov (approved):

Issue #268, opened 2021-11-03
Implement Userspace OOM killer


zuzzas:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

`kswapd` starts thrashing memory subsystem (pages) around and stops giving CPU time to userspace processes, leading to a complete freeze of the whole system.

### Proposed Solution

Implementation could be as simple as [this](https://gist.github.com/stewartpark/14a5690afbf717be04857d2cda1d8620) gist. It is crude but avoids deadlocking an instance until it is rebooted.

Understand that this is only a stopgap solution until kubelet swap support is in beta. After that, we should use `oomd` or other Kubernetes-centric solutions to perform userspace OOM kills.

### Additional Information

Consider implementing [ebpf_exporter](https://github.com/cloudflare/ebpf_exporter) to efficiently track OOM kills.

I've done almost all the required [work](https://github.com/cloudflare/ebpf_exporter/pull/102) to track OOMs via eBPF.

PR #387, opened 2021-12-13
[monitoring-kubernetes][node-manager] Early OOM killer and ebpf exporter

zuzzas:
## Description

1. ebpf-exporter that monitors global and per-cgroup OOMs. With recording rules and dashboard.
2. Primitive early OOM that prevents nodes from getting stuck in out-of-memory conditions.

Provide a simple early oom killer which prevents Nodes from getting stuck, while kernel tries to free up memory pages and not trigger OOM killer.

kubelet generates OOMKilled Events only when PID 1 of a container dies. It skips Node-level OOMs and non-pid 1 oomkills.

Previously, we've had based on klog parsing. It was inefficient and consumed lots of memory.

The proposed ebpf-exporter based solution works on Ubuntu with kernel version >5.4. It has a hook that schedules ebpf-exporter DaemonSet only on those Nodes.

## Changelog entries

```changes
module: monitoring-kubernetes
type: feature
description: Added ebpf-exporter
note: |
  ebpf-exporter that monitors global and per-cgroup OOMs. With recording rules and dashboard.
---
module: node-manager
type: feature
description: Added Early OOM killer
note: |
  Primitive early OOM that prevents nodes from getting stuck in out-of-memory conditions. Triggers when MemAvailable becomes less than 500 MiB.
```


EvgenySamoylov:
Fixes https://github.com/deckhouse/deckhouse/issues/268

mac2000:
Not sure if understand everything well, just want to clarify a few things

- swap intended to help with memory overcommit but usually is turned off because of performance
- by default system allows memory overcommit because there is a swap for such cases
- if we disable swap but did not disable memory overcommit - we definitely will hit ourselves

What I'm trying to understand - maybe with this early oom killer we are fixing dependant issues but not the root cause (just my guess, not saying that it is the case, just want to clarify this one for myself)

In our case, it seems that the nodes have swap turned off, and what is interesting system oom killer is also turned off which definitely may behave in non expected ways

PS: Also there is a chance that we want `MIN_MEMORY_KB` to be configurable per node role, e.g. frontend nodes are usually small ones with just few gigabytes of ram

zuzzas:
@mac2000 

First of all, kernel OOM killer is never disabled since most programs that allocate memory are ill-prepared to deal with the negative response of a malloc() attempt. It's called overcommit.

This PR is a bandaid that aims to remove permanently hung-up Linux instances stuck on a thrashing memory subsystem. There is [a KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2400-node-swap) underway, and we'll see how we can leverage this new functionality along with more competent early oom killers, such as `oomd` from Facebook.

konstantin-axenov (approved):

PR #406, opened 2021-12-14
[prometheus] fixed FAQ about Lens

RomanenkoDenys:
## Description

Improved Prometheus FAQ about Lens access.
## Changelog entries

```changes
module: prometheus
type: feature
description: "Improved Prometheus FAQ about Lens access."
```

### DIFF
--- a/modules/300-prometheus/docs/FAQ.md
+++ b/modules/300-prometheus/docs/FAQ.md
@@ -257,15 +257,22 @@ metadata:
   name: prometheus-lens-proxy-conf
   namespace: lens-proxy
 data:
-  prometheus.conf.template: |
+  "40-prometheus-proxy-conf.sh": |
+    #!/bin/sh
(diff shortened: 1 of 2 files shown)


modules/300-prometheus/docs/FAQ.md
```diff
@@ -260,10 +260,12 @@ data:
   prometheus.conf.template: |
     server {
       listen 80 default_server;
+      resolver kube-dns.kube-system.svc.cluster.local valid=30s;
```
nabokihms, line 263: Cluster domain may differ for Deckhouse installations. Can we avoid providing the full domain name somehow?

RomanenkoDenys: We don't have the right way to do this. Issue about downward api - https://github.com/kubernetes/dns/issues/421.
So i added manual fill of cluster domain to yaml file.

nabokihms (approved):

PR #408, opened 2021-12-14
[monitoring-kubernetes] Set label `tier: cluster` on node alerts

konstantin-axenov:
## Description

Alerts that are tightly coupled with node state should have tier cluster. Also refactored impact and likelihood labels into severity_level.

Correct labels for alerts should fix their routing.


nabokihms (approved):

nabokihms (approved):
This is great that you have fixed the severity annotations too. 👍

PR #416, opened 2021-12-15
[deckhouse] fix stale cache for disabled modules

diafour:
## Description

Update dependencies for deckhouse-controller binary.

Clean caches for disabled modules to prevent usage of stale cache when module is enabled again.

See https://github.com/flant/addon-operator/pull/280

## Changelog entries

```changes
module: deckhouse
type: fix
description: clear values cache when module disabled
```


nabokihms (commented):
@diafour, could you please fill in the description template so we can start a review? There is also some problem with tests.

diafour:
`Map()` is not safe for recursive matching, change it to `Value()`.

nabokihms (approved):

PR #417, opened 2021-12-16
[cert-manager] Fix. Service account for acme solver pod cannot create

name212:
## Description
If acmesolver registry secret already exists in namespace when challenge was created, then service account for acme solver will not create

Certificates cannot created because acmesolver pod cannot be pulled.

## Changelog entries

```changes
module: cert-manager
type: fix
description: Service account will be created if registry secret already located in namespace
```


nabokihms (commented):
Just a small spelling fix, everything else looks good 👍

nabokihms:
@z9r5, it is better to backport it to `v1.28`.

nabokihms (approved):

Issue #304, opened 2021-11-18
[ceph-csi] New module


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

Ceph CSI plugins implement an interface between CSI enabled Container Orchestrator (CO) and Ceph cluster. It allows dynamically provisioning Ceph volumes and attaching them to workloads.

https://github.com/ceph/ceph-csi

### Proposed Solution

Create a new module based on the ceph-csi helm chart.

PR #426, opened 2021-12-16
[ceph-csi] new module

vitaliy-sn:
## Description
Added new module CephCSI

CephCSI allows dynamically provisioning Ceph volumes (RBD and CephFS) and attaching them to workloads.

## Changelog entries

```changes
module: ceph-csi
type: feature
description: Added new module ceph-csi
note: CephCSI allows dynamically provisioning Ceph volumes (RBD and CephFS) and attaching them to workloads.
```


EvgenySamoylov:
Closes #304

modules/099-ceph-csi/crds/cephcsi.yaml
```diff
@@ -0,0 +1,131 @@
...
+  preserveUnknownFields: false
+  versions:
+    - name: v1alpha1
+      served: true
+      storage: true
+      schema: &schema
+        openAPIV3Schema:
+          type: object
+          description: |
+            Ceph cluster connection parameters and StorageClasses configuration
```
nabokihms, line 24:
```suggestion
            Ceph cluster connection parameters and StorageClasses configuration.
```
and everywhere in the doc.

modules/099-ceph-csi/crds/doc-ru-cephcsi.yaml
```diff
@@ -0,0 +1,119 @@
+spec:
+  versions:
+    - name: v1alpha1
+      schema: &schema
+        openAPIV3Schema:
+          type: object
```
nabokihms, line 6: There should be only descriptions in a doc-ru file without `type`, `required`, and other properties.

modules/099-ceph-csi/crds/cephcsi.yaml
```diff
@@ -0,0 +1,131 @@
...
+  labels:
+    heritage: deckhouse
+    module: ceph-csi
+    app: ceph-csi
+spec:
+  group: deckhouse.io
+  scope: Cluster
+  names:
+    plural: cephcsi
+    kind: CephCSI
```
nabokihms, line 14: This means Ceph Container Storage Interface. Maybe we should call it CephCSIDriver or CephCSIInstance?

modules/099-ceph-csi/docs/USAGE.md
```diff
@@ -0,0 +1,34 @@
...
+apiVersion: deckhouse.io/v1alpha1
+kind: CephCSI
+metadata:
+  name: example
+spec:
+  clusterID: 2bf085fc-5119-404f-bb19-820ca6a1b07e
+  monitors:
+  - 10.0.0.10:6789
+  userID: admin
+  userKey: AQDbc7phl+eeGRAAaWL9y71mnUiRHKRFOWMPCQ==
```
nabokihms, line 17: Is this encrypted? Looks like a plain text keyring to me.

modules/099-ceph-csi/crds/cephcsi.yaml
```diff
@@ -0,0 +1,131 @@
...
+                    type: string
+                rbd:
+                  type: object
+                  properties:
+                    storageClasses:
+                      description: |
+                        Description of StorageClasses for Rados Block Device (RBD)
+                      type: array
+                      items:
+                        type: object
```
nabokihms, line 62: Are there any required fields?

modules/099-ceph-csi/crds/cephcsi.yaml
```diff
@@ -0,0 +1,131 @@
...
+                        properties:
+                          name:
+                            description: |
+                              StorageClass name
+                            type: string
+                          pool:
```
nabokihms, line 68: Maybe we need to use a default pool name and default reclaimPolicy?

modules/099-ceph-csi/enabled
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Copyright 2021 Flant JSC
+# Licensed under the Deckhouse Platform Enterprise Edition (EE) license. See https://github.com/deckhouse/deckhouse/blob/main/ee/LICENSE
+
+source /deckhouse/shell_lib.sh
+
+function __main__() {
+  enabled::disable_module_if_cluster_is_not_bootstraped
```
nabokihms, line 9: We may want to use the CSI driver even if a cluster is not bootstrapped, e.g., deploying a network gateway.

modules/099-ceph-csi/crds/doc-ru-cephcsi.yaml
```diff
@@ -0,0 +1,119 @@
...
+                        properties:
+                          name:
+                            description: |
+                              Имя StorageClass
+                            type: string
+                          pool:
+                            description: |
+                              Название пула в котором будут создаваться RBD-образы
+                            type: string
+                          reclaimPolicy:
```
nabokihms, line 102: If the reclaim policy is changed, we do not have a hook to delete the storage class. Helm upgrade will fail in this case.

modules/099-ceph-csi/crds/cephcsi.yaml
```diff
@@ -0,0 +1,131 @@
...
+                            description: |
+                              Pool name for creating RBD images
+                            type: string
+                          reclaimPolicy:
+                            description: |
+                              The reclaim policy for a PersistentVolume
+                              https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming
+                            type: string
+                          allowVolumeExpansion:
```
nabokihms, line 77: I propose `true` by default.

modules/099-ceph-csi/crds/cephcsi.yaml
```diff
@@ -0,0 +1,131 @@
...
+                            description: |
+                              Allows the users to resize the volume by editing the corresponding PVC object
+                            type: boolean
+                          mountOptions:
+                            description: |
+                              List of mount options
+                            type: array
+                            items:
+                              type: string
```
nabokihms, line 86: Do we need to validate these options?

testing/openapi_validation/validators/enum.go
```diff
@@ -102,6 +102,16 @@ var (
 			// GeoIP base constants: GeoIP2-ISP, GeoIP2-ASN, ...
 			"spec.versions[*].schema.openAPIV3Schema.properties.spec.properties.geoIP2.properties.maxmindEditionIDs.items",
 		},
+		"modules/099-ceph-csi/crds/cephcsi.yaml": {
+			// ignore
```
nabokihms, line 106: Could you please elaborate more on why we need to ignore these errors?

nabokihms (changes requested):

modules/099-ceph-csi/crds/cephcsi.yaml
vitaliy-sn: Done.

modules/099-ceph-csi/crds/doc-ru-cephcsi.yaml
vitaliy-sn: Unnecessary properties have been removed.

modules/099-ceph-csi/crds/cephcsi.yaml
vitaliy-sn: Let it be CephCSIDriver.

modules/099-ceph-csi/docs/USAGE.md
vitaliy-sn: No, it's a value obtained with `ceph auth get client.admin` . Base64 contains binary data.

modules/099-ceph-csi/crds/cephcsi.yaml
vitaliy-sn: Yes, there should be required field. Added required section.

modules/099-ceph-csi/enabled
vitaliy-sn: I didn't find any reason why we couldn't do that. Removed the if_cluster_is_not_bootstraped check.

modules/099-ceph-csi/crds/doc-ru-cephcsi.yaml
vitaliy-sn: Removing storage class added to get_crds.

modules/099-ceph-csi/crds/cephcsi.yaml
vitaliy-sn: I accept your proposal.

modules/099-ceph-csi/crds/cephcsi.yaml
vitaliy-sn: I don't see how we can validate this. It's just a list of mount options.

testing/openapi_validation/validators/enum.go
vitaliy-sn: Done.

modules/099-ceph-csi/crds/cephcsi.yaml
nabokihms: I checked, and it seems Kubernetes also skips validation for mountOptions, and so will we.

modules/099-ceph-csi/docs/USAGE.md
```diff
@@ -0,0 +1,34 @@
+---
+title: "The ceph-csi module: usage"
+---
+
+## An example of the `CephCSI` configuration
```
nabokihms, line 5: You changed the kind name to `CephCSIDriver` but forgot to change examples.

modules/099-ceph-csi/docs/USAGE.md
nabokihms: Data in this object looks too sensitive to leave it there in plain text. What do you think?

modules/099-ceph-csi/hooks/get_crds.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
+	"github.com/flant/addon-operator/sdk"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/deckhouse/deckhouse/go_lib/dependency"
+)
+
+type CephCSI struct {
```
nabokihms, line 32: CephCSIDriver?

modules/099-ceph-csi/crds/cephcsi.yaml
nabokihms: What's about this comment?

modules/099-ceph-csi/crds/cephcsi.yaml
```diff
@@ -0,0 +1,141 @@
...
+                              Part of the StorageClass name after `-`. The name from the CustomResource CephCSIDriver is used as the first part.
+                            type: string
+                          pool:
+                            description: |
+                              Pool name for creating RBD images.
+                            type: string
+                          reclaimPolicy:
+                            description: |
+                              The reclaim policy for a PersistentVolume.
+                            type: string
```
nabokihms, line 127: I assume it should be the same schema as for the rbd section:
```
enum:
- Delete
- Retain
default: Retain
```

modules/099-ceph-csi/crds/cephcsi.yaml
```diff
@@ -0,0 +1,141 @@
...
+                            description: |
+                              Allows the users to resize the volume by editing the corresponding PVC object.
+                            type: boolean
+                          mountOptions:
+                            description: |
+                              List of mount options.
+                            type: array
+                            items:
+                              type: string
+                          fsName:
```
nabokihms, line 138: If this is not required, which filesystem will be used?

ee/fe/modules/values-default.yaml
```diff
@@ -3,3 +3,4 @@ monitoringApplicationsEnabled: true
 extendedMonitoringEnabled: true
 secretCopierEnabled: true
 flantIntegrationEnabled: true
+cephCsiEnabled: true
```
nabokihms, line 6: I don't think we want to enable this by default because it's too specific, do we?

modules/099-ceph-csi/hooks/get_crds.go
```diff
@@ -0,0 +1,197 @@
...
+		Name:          sc.Name,
+		ReclaimPolicy: string(*sc.ReclaimPolicy),
+	}, nil
+}
+
+func setInternalValues(input *go_hook.HookInput, dc dependency.Container) error {
+	values := []InternalValues{}
+	csiConfig := []CSIConfig{}
+
+	kubeClient, err := dc.GetK8sClient()
```
nabokihms, line 138: You should avoid using kubeclient if you don't need it. In your case, it's better to stick to the patches and patch collector.

nabokihms (changes requested):

modules/099-ceph-csi/crds/cephcsi.yaml
vitaliy-sn: This is required, I missed it. Added to required.

ee/fe/modules/values-default.yaml
vitaliy-sn: You're right, we don't want to enable by default.

modules/099-ceph-csi/hooks/get_crds.go
vitaliy-sn: I want to be sure that by the time helm starts, the storage class will be removed.

modules/099-ceph-csi/crds/cephcsi.yaml
vitaliy-sn: Fixed

modules/099-ceph-csi/docs/USAGE.md
vitaliy-sn: Fixed.

modules/099-ceph-csi/hooks/get_crds.go
vitaliy-sn: Yes. Fixed.

modules/099-ceph-csi/docs/USAGE.md
vitaliy-sn: If we place them in a dedicated secret, then it will not become safer.

modules/099-ceph-csi/hooks/get_crds.go
nabokihms: Then it is ok to use patch collector.

vitaliy-sn: Replaced kubeclient with patchCollector.

nabokihms (approved):

Issue #332, opened 2021-12-01
[deckhouse] Canary updating


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

With the growth of the popularity of Deckhouse, now we have plenty of clusters on the same release channel. To avoid overloading or mass problems, we need to integrate the canary deployment solution into our release process.

### Proposed Solution

The idea is to add updating time variations. 

When the new release is found in the release channel, do the following:
1. Calculate the murmur hash (or other numeric hash) of the concatenated cluster UUID and the release version.
2. Find the remainder of the division for five (or more?) The value means the number of the canary update.
3. For every cluster, which remainder equals zero, execute the update immediately.
4. For others, calculate the update execution time by formula `reminder * 30min + current time`.
5. Write the time to the release status.
6. Execute the update for such releases according to their statuses.

### Additional Information

* https://en.wikipedia.org/wiki/MurmurHash
* We also need to think about removing the new version from the release channel in case of problems. I assume that in this case releases should be marked with a special status and skipped.

nabokihms:
1. If there are update windows in the config, do not follow this logic.
2. We need to calculate the update execution time only once, so we have to store it somewhere in the cluster. Storing it in the DeckhouseRelease custom resource seems not as good as for the first glance, because of the complicated logic around it.

nabokihms:
It seems reasonable to skip canary deployments for patch releases.

PR #429, opened 2021-12-16
[deckhouse] Add canary deployment for releases

yalosev:
## Description
For a large number of clusters on a release channel, we want to have the ability to spread release in time

Spreaded release will generate less load and become more stable

Closes https://github.com/deckhouse/deckhouse/issues/332

## Changelog entries

```changes
module: deckhouse
type: feature
description: Add canary deckhouse release update
```

### DIFF
--- a/canary.json
+++ b/canary.json
@@ -1,0 +1,7 @@
+{
+  "alpha": {"enabled": true, "waves": 2, "interval": "5m"},
+  "beta": {"enabled": false, "waves": 1, "interval": "1m"},
+  "early-access": {"enabled": true, "waves": 6, "interval": "30m"},
+  "stable": {"enabled": true, "waves": 6, "interval": "20m"},
+  "rock-solid": {"enabled": false, "waves": 5, "interval": "5m"}
+}
--- a/modules/020-deckhouse/hooks/update_deckhouse_image.go
+++ b/modules/020-deckhouse/hooks/update_deckhouse_image.go
@@ -159,21 +159,33 @@ func filterDeckhouseRelease(unstructured *unstructured.Unstructured) (go_hook.Fi
 		return nil, err
 	}
 
+	var hasSuspendAnnotation bool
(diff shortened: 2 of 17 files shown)


canary.json
```diff
@@ -0,0 +1,7 @@
+{
+  "alpha": {"enabled": false, "waves": 1, "interval": 1},
+  "beta": {"enabled": false, "waves": 1, "interval": 1},
+  "early-access": {"enabled": true, "waves": 6, "interval": 10},
```
nabokihms, line 4: 1. It looks like 10 minutes may not be enough to react to release problems.
2. What do you think about using the `time.Duration` Golang format as a value for the interval?

modules/020-deckhouse/hooks/update_deckhouse_image.go
```diff
@@ -301,10 +313,20 @@ func releaseChannelUpdate(input *go_hook.HookInput, releases []deckhouseReleaseU
...
 			// pass
 
 		case v1alpha1.PhasePending:
+			// skip suspended releases
+			if rl.HasSuspendAnnotation {
+				sp := statusPatch{
+					Phase:          v1alpha1.PhaseSuspended,
+					TransitionTime: now,
+				}
+				input.PatchCollector.MergePatch(sp, "deckhouse.io/v1alpha1", "DeckhouseRelease", "", rl.Name, object_patch.WithSubresource("/status"))
```
nabokihms, line 326: Looked through briefly and I think I missed something. Could you elaborate more on why we need both the annotation and the status field?

yalosev: Annotation is placed by a hook which checks the release image. It marks a release as "should be suspended".
And this hook takes marked release and move it to a suspended state to make it more obvious for user experience. Also it makes sense for a release loop

canary.json
yalosev: makes sense

nabokihms (approved):

nabokihms:
I have created the necessary labels.

PR #430, opened 2021-12-16
[deckhouse] Add alert for invalid config

### LINKED ISSUE #87: [deckhouse] alert if config validation fails
We have a metric `deckhouse_config_values_errors_total` which represents validation error count. 

It would be nice to have an alert if Deckhouse fails to start in case that someone put wrong values in `cm/deckhouse`.

yalosev:
## Description
Add alert for broken deckhouse config

(Close #87)

We don't have in place validation for deckhouse config values cm. If someone put wrong values into it we can miss a problem and will be surprised by broken/abnormally working Deckhouse

## Changelog entries

```changes
module: deckhouse
type: feature
description: Add alert if deckhouse config is broken
```

### DIFF
--- a/modules/020-deckhouse/monitoring/prometheus-rules/deckhouse.yaml
+++ b/modules/020-deckhouse/monitoring/prometheus-rules/deckhouse.yaml
@@ -322,3 +322,23 @@
         Deckhouse does not work as expected.
 
         You can find out the exact problem in the relevant alerts.
+
+  - alert: D8DeckhouseConfigInvalid
+    expr: increase(deckhouse_config_values_errors_total[30s]) > 0
+    for: 1m
+    labels:
+      severity_level: "5"
+      d8_module: deckhouse
+      d8_component: deckhouse
+      tier: cluster
+    annotations:
+      plk_markup_format: "markdown"
+      plk_protocol_version: "1"
+      plk_grouped_by__main: "D8DeckhouseMalfunctioning,tier=cluster,prometheus=deckhouse"
+      description: |
+        Deckhouse config contains errors.
+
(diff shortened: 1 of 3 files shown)


modules/020-deckhouse/monitoring/prometheus-rules/deckhouse.yaml
```diff
@@ -322,3 +322,23 @@
         Deckhouse does not work as expected.
 
         You can find out the exact problem in the relevant alerts.
+
+  - alert: D8DeckhouseConfigInvalid
+    expr: deckhouse_config_values_errors_total > 0
```
nabokihms, line 327: Are you sure about this promql expression?

Per spec, `deckhouse_config_values_errors_total` is a counter, so it can only go upward.
image

nabokihms (commented):
It looks like a good PR, but by diving a little deeper, we realized a problem lying inside of the addon-operator code. 

It only increases the `deckhouse_config_values_errors_total ` counter when invalid values are added to the config map (the actual Kubernetes event occurred). Such behavior will never allow us to find the problem because it is impossible to determine whether the invalid values are still in the config map and how long they were.

In its current conditions, the `deckhouse_config_values_errors_total` metrics does not look helpful.

modules/020-deckhouse/monitoring/prometheus-rules/deckhouse.yaml
yalosev: Added new metric behavior in [addon-operator](https://github.com/flant/addon-operator/commit/d86a065d5763685c6e8ed93610009eb6405ce63d) and fix it here

nabokihms (approved):

Issue #340, opened 2021-12-07
[deckhouse] Increase pending releases alert severity based on their quantity


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

As for now, we have the alert about pending releases with severity_level `9`.
https://github.com/deckhouse/deckhouse/blob/683ae738464b442dec8e516fd6d3e34082b18dc2/modules/020-deckhouse/monitoring/prometheus-rules/alerting.yaml#L23-L39

This notification is not urgent enough and can be ignored (which is fine). However, ignoring two or even three releases seems critical and can signify the real problem.

### Proposed Solution

Add more alerts with the increasing severity, as we do for [the target down alert](https://github.com/deckhouse/deckhouse/blob/main/modules/300-prometheus/monitoring/prometheus-rules/target-down.yaml). Make alerts severity equal `5`  if there are more than two pending minor releases and `3` if there are more than three.

PR #439, opened 2021-12-17
[deckhouse] severity level based on releases count

yalosev:
## Description
Change alert severity level based on pending manual releases

(Close https://github.com/deckhouse/deckhouse/issues/340)

We can miss the low severity alert but if we have a few pending releases - something is going wrong.
We want to increase the level and pay more attention for the alert

## Changelog entries

```changes
module: deckhouse
type: feature
description: Different severity level based on pending DeckhouseReleases count
```

### DIFF
--- a/modules/020-deckhouse/hooks/update_deckhouse_image.go
+++ b/modules/020-deckhouse/hooks/update_deckhouse_image.go
@@ -86,31 +86,105 @@ type deckhousePodInfo struct {
 	Ready     bool   `json:"ready"`
 }
 
-// isNextReleasePatch check SORTED array of DeckhouseReleases.
-// If the next release after CURRENT Deployed release is a patch release - returns true
-// else returns false
-func isNextReleasePatch(releases []deckhouseReleaseUpdate) bool {
-	var currentReleaseIndex = -1
-	var currentRelease *semver.Version
+type deckhouseUpdater struct {
+	manualMode bool // Manual update mode activated
 
-	for i, r := range releases {
-		if r.Phase == v1alpha1.PhaseDeployed {
-			currentReleaseIndex = i
-			currentRelease = r.Version
-			continue
-		}
+	releases                   []deckhouseRelease
+	hasPatchRelease            bool // next release is a patch
+	totalPendingManualReleases uint
+}
 
-		if currentRelease != nil && i == currentReleaseIndex+1 {
-			// check next release
-			if r.Version.Major() == currentRelease.Major() && r.Version.Minor() == currentRelease.Minor() {
-				return true
+func newDeckhouseUpdater(input *go_hook.HookInput) *deckhouseUpdater {
+	approvalMode := input.Values.Get("deckhouse.update.mode").String()
+
+	return &deckhouseUpdater{
+		manualMode: approvalMode == "Manual",
+	}
+}
+
+func (du *deckhouseUpdater) patchReleaseStatus(input *go_hook.HookInput, release deckhouseRelease, ts time.Time) deckhouseRelease {
+	var statusChanged bool
+
+	statusPatch := statusPatch{
+		Phase:          release.Phase,
+		Approved:       release.StatusApproved,
+		TransitionTime: ts,
+	}
+
+	// when DeckhouseRelease just created and doesn't have any status
+	if release.Phase == "" {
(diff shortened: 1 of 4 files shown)


modules/020-deckhouse/hooks/update_deckhouse_image.go
```diff
@@ -321,9 +419,9 @@ func releaseChannelUpdate(input *go_hook.HookInput, releases []deckhouseReleaseU
 			}
 
 			// always deploy Patch releases
-			if !rl.StatusApproved && !isPatch {
+			if !rl.StatusApproved && !du.hasPatchRelease {
 				input.LogEntry.Infof("Release %s is waiting for manual approval", rl.Name)
-				input.MetricsCollector.Set("d8_release_waiting_manual", 1, map[string]string{"name": rl.Name}, metrics.WithGroup("d8_releases"))
+				input.MetricsCollector.Set("d8_release_waiting_manual", float64(du.totalPendingManualReleases), map[string]string{"name": rl.Name}, metrics.WithGroup("d8_releases"))
```
nabokihms, line 424: So if I get it right, we expose the metric for all names with the same value?
```
d8_release_waiting_manual{name="v1-28-0"} 3
d8_release_waiting_manual{name="v1-29-0"} 3
d8_release_waiting_manual{name="v1-30-0"} 3
```

I am puzzled a little because it does not seem obvious what exactly these values represent.
Maybe if we look into the code, we will be able to get the meaning. However, it is impossible if we only see these metrics on Prometheus.

modules/020-deckhouse/monitoring/prometheus-rules/alerting.yaml
```diff
@@ -34,6 +34,40 @@
       description: |
         Deckhouse release is waiting for manual approval.
 
+        Please run `kubectl patch DeckhouseRelease {{ $labels.name }} --type=merge -p='{"approved": true}'` for confirmation.
+      summary: |
+        Deckhouse release is waiting for manual approval.
+  - alert: DeckhouseReleaseIsWaitingManualApproval
+    expr: max by (name) (d8_release_waiting_manual) == 2
```
nabokihms, line 41: If I understand this right, a user will receive three alerts, one for each waiting version. It would be better to group them then.

modules/020-deckhouse/monitoring/prometheus-rules/alerting.yaml
```diff
@@ -34,6 +34,40 @@
...
+      d8_component: deckhouse
+      tier: cluster
+    annotations:
+      plk_markup_format: "markdown"
+      plk_protocol_version: "1"
+      plk_incident_initial_status: "todo"
+      description: |
+        Deckhouse release is waiting for manual approval.
+
+        Please run `kubectl patch DeckhouseRelease {{ $labels.name }} --type=merge -p='{"approved": true}'` for confirmation.
```
nabokihms, line 54: Does the order of executing this operation matter? If so, maybe we do not want to send users the alert for the next version before they approve the first one (so as to save consistency).

modules/020-deckhouse/hooks/update_deckhouse_image.go
yalosev: this metric will be created only for the first pending release
```
d8_release_waiting_manual{name="v1-28-0"} 3
```

after this release is approved, the metric will be deleted
```
Expire("d8_releases")
```
and we will have a new metric

```
d8_release_waiting_manual{name="v1-29-0"} 2
```

That has been done because we have to apply releases in order. And we have instructions in this alert only for the next release. So we export the total count of manual pending releases but put labels only for the first one

yalosev: Also, we have a `return` after this exporting metric. So we place it only one time

modules/020-deckhouse/monitoring/prometheus-rules/alerting.yaml
yalosev: We will have only one metric here. See the upper comment

modules/020-deckhouse/monitoring/prometheus-rules/alerting.yaml
yalosev: It is. And we have that behavior

nabokihms (approved):

Issue #365, opened 2021-12-10
[extended-monitoring] Alert about certificate expiration


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

We do not have an alert on the expiration of:
* Self-signed certificates
* Orphaned certificates
* Certificates bought from third-party providers

### Proposed Solution

1. Deploy exporter, which will subscribe to all the `kubernetes.io/tls` certificates in the cluster and check their expiration.
2. Fire an alert in the case of the presence of extended monitoring annotation on the namespace and certificate expiration.

### Additional Information

1. https://github.com/giantswarm/cert-exporter
2. We need to group alerts. There is the case when the same wildcard certificate is copied to all namespaces by secret copier.

PR #479, opened 2021-12-22
[extended-monitoring] Add cert-exporter

kvaps:
## Description
This PR adds cert-exporter to the extended-monitoring module

cert-exporter allows to track certificates expiration.

fixes https://github.com/deckhouse/deckhouse/issues/365

## Changelog entries

```changes
module: extended-monitoring
type: feature
description: Add cert-exporter
note: |
  Added cert-exporter to track certificates expiration
```

### DIFF
--- a/ee/fe/modules/340-extended-monitoring/templates/cert-exporter/rbac-for-us.yaml
+++ b/ee/fe/modules/340-extended-monitoring/templates/cert-exporter/rbac-for-us.yaml
@@ -1,0 +1,47 @@
+{{- if .Values.extendedMonitoring.certificates.exporterEnabled }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cert-exporter
+  namespace: d8-monitoring
+  {{- include "helm_lib_module_labels" (list .) | nindent 2 }}
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: d8:extended-monitoring:cert-exporter
+  {{- include "helm_lib_module_labels" (list .) | nindent 2 }}
+rules:
+- apiGroups: [""]
+  resources: ["namespaces", "secrets"]
+  verbs: ["get", "list"]
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: d8:extended-monitoring:cert-exporter
+  {{- include "helm_lib_module_labels" (list .) | nindent 2 }}
+subjects:
+- kind: ServiceAccount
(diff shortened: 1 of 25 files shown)


ee/fe/modules/340-extended-monitoring/templates/cert-exporter/rbac-for-us.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.extendedMonitoring.certificates.exporterEnabled }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cert-exporter
+  namespace: d8-monitoring
+  {{- include "helm_lib_module_labels" (list .) | nindent 2 }}
```
nabokihms, line 8:
```suggestion
{{- include "helm_lib_module_labels" (list .) | indent 2 }}
```
Just to keep the style. Or, we could switch to using `nindent` everywhere, but then there are a lot of changes to make.

ee/fe/modules/340-extended-monitoring/openapi/config-values.yaml
```diff
@@ -52,6 +52,12 @@ properties:
         default: false
         description: |
           Enables certExporter.
+      namespaces:
```
nabokihms, line 55: I have a proposal to stick to k8s-image-availability-exporter here:
1. Always collect metrics from all namespaces.
2. Alert about expiring certificates only if the extended monitoring enabled label present on the namespace.

Thus we could avoid adding the `namespaces` option to the config.

nabokihms (changes requested):
Everything according to the code looks good, thank you. The only concern left is the namespaces option.

ee/fe/modules/340-extended-monitoring/openapi/config-values.yaml
kvaps: OK, solved in https://github.com/deckhouse/deckhouse/pull/479/commits/6402266d7eb82226ee1f84925321b277a965e095

ee/fe/modules/340-extended-monitoring/templates/cert-exporter/rbac-for-us.yaml
kvaps: I believe that using `nindent` will make the code more readable.
If you want to, I can prepare a PR to update it in other places.

kvaps: solved in d9c4aa5b2dc24fa6f15b1ee5fb064c3ae6148dbc

nabokihms (approved):
Looks good, thank you.

PR #486, opened 2021-12-23
[flant-integration] Remove the plan parameter from the openAPI specification

z9r5:
## Description
The plan parameter was removed from the OpenAPI specification

The plan parameter was removed from the configuration in 1.29 and now we can remove it from the OpenAPI specification.

## Changelog entries

```changes
module: flant-integration
type: fix
description: Remove the plan parameter from the OpenAPI specification
```


shvgn (approved):
LGTM

Issue #58, opened 2021-08-24
[candi] Add ability to flexible nodes configuration


nabokihms:
As for now, we have only limited hardcoded options in the bashible configuration. We thought this is not right to permit users to change anything related to the os/kernel/packages/interfaces. 

Apparently, we were wrong.
The first idea for implementation is adding some CRD with an ansible playbook, which can be attached to the node group. Bashible will run this playbook on all nodes.

nabokihms:
The alpha version of the generic plan of adding Bashible API server steps customization:

1. Create the `BashibleStep` CRD.
2. Bashible API server deploys steps into the cluster.
3. Steps contain two fields - spec.shell, spec.created_by and spec.weight.
4. The `shell` field contains the base64 encoded content of the step.
5. The `created_by` field value equals `user` by default.
6. (?) Add the validating webhook. Nobody is allowed to use the created_by field other than bashible-apiserver service account.
7. The weight field contains the 4-digits string value, e.g. `0100`, `0042`, `1001`.
8. Put labels onto all steps - bundle.deckhouse.io/centos-7: "", bundle.deckhouse.io/ubuntu-lts: "", node-group.deckhouse.io/master: "", node-group.deckhouse.io/worker: "".
9. If it is a common step, it can be labeled with more than one label of the same kind, e.g. two node-groups, two bundles.
10. Bashible on nodes.
11. Create the second CRD - BashibleBundleChecksum.
12. With two fields - `checksum` and `lastTransitionTime`.
13. (?) Write to the BashibleBundleChecksum status field which step has triggered the transition.
14. Bashible API server maintains checksums.
15. Bashible API servers subscribes to all steps changes. 
16. If the step created by the Bashible API server changed (we can determine it by the `created_by` field value), it reverts the changes (or validate it with validating webhook).
17. If the step was changed - recalculate the checksum.
18. Bashible API server continues to render the `bashible` script.

nabokihms:
So, it turned out that the plan is impossible to execute (we edited it and added more concerns to comments).

Let's go with the more simple and realistic idea:
1. Create CRD - `NodeGroupConfiguration`.
2. Fields: 
    * `content`(_string_) - the content of the step, can be either go template or plain bash
    * `weight` (_int_) - affects the order of the script execution
    * `nodeGroups` (_list_) - the list of node groups to apply the step
    * `bundle` (_string_) _optional_ - bundle to apply the step (empty bundle means all bundles)
[...]

PR #492, opened 2021-12-24
[node-manager] Add custom NodeGroupConfiguration

yalosev:
## Description
Feature: add ability to create user scripts for node configuration
(Close https://github.com/deckhouse/deckhouse/issues/58)

Sometimes we need to configure nodes with some custom scripts. This feature creates CRD for this purpose.

## Changelog entries

```changes
module: node-manager
type: feature
description: NodeGroupConfiguration CRD for handling user-defined scripts for Node configuration
```


modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
```diff
@@ -17,48 +17,195 @@ limitations under the License.
...
+	ginformer := ngConfigFactory.ForResource(schema.GroupVersionResource{
+		Group:    "deckhouse.io",
+		Version:  "v1alpha1",
+		Resource: "nodegroupconfigurations",
+	})
+
+	informer := ginformer.Informer()
+
+	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
```
nabokihms, line 136: It is recommended to avoid writing logic right in the `AddFunc`, `UpdateFunc`, `DeleteFunc`. As far as I remember, it badly affects performance. Instead, you need to use some queue.

See this example from Prometheus https://github.com/prometheus/prometheus/blob/c0c22ed04200a8d24d1d5719f605c85710f0d008/discovery/kubernetes/pod.go#L60-L73

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
```diff
@@ -17,48 +17,195 @@ limitations under the License.
...
+			}
+			s.RemoveNodeGroupConfiguration(&ngc)
+			s.emitChanges()
+		},
+	})
+
+	go informer.Run(ctx.Done())
+
+	// Wait for the first sync of the informer cache, should not take long
+	for !informer.HasSynced() {
```
nabokihms, line 189: There is a function in client-go especially for this case, wich is also provides a circuit breaker:
```go
if !cache.WaitForCacheSync(ctx.Done(), informer.HasSynced) {
    return nil, fmt.Errorf("unable to sync caches: %v", ctx.Err())
}
```

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
```diff
@@ -138,3 +289,99 @@ func readTemplates(rootDir string, templates map[string][]byte) error {
...
+	configurations = append(configurations, s.nodeGroupConfigurations[wildcardBundle]...)
+	configurations = append(configurations, s.nodeGroupConfigurations[wildcardNG]...)
+	configurations = append(configurations, s.nodeGroupConfigurations[totalWildcard]...)
+	s.m.RUnlock()
+
+	steps := make(map[string]string, len(configurations))
+	for _, sc := range configurations {
+		step, err := RenderTemplate(sc.Name, []byte(sc.Content), templateContext)
+		if err != nil {
+			return nil, fmt.Errorf("cannot render node configuration \"%s\" for bundle \"%s\": %v", sc.Name, bundle, err)
```
nabokihms, line 377: NIT `\"%s\"` -> `%q`

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
```diff
@@ -138,3 +289,99 @@ func readTemplates(rootDir string, templates map[string][]byte) error {
...
+
+	sc := nodeConfigurationScript{
+		Name:    name,
+		Content: nc.Spec.Content,
+	}
+
+	s.m.Lock()
+	defer s.m.Unlock()
+	for _, ng := range nodeGroups {
+		key := fmt.Sprintf("%s:%s", bundle, ng)
```
nabokihms, line 314: Instead of concatenating strings, what do you think about using struct with two fields (`bundle` and `ng`) for `nodeGroupConfigurations` keys (and the same for system scripts)?

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
```diff
@@ -17,48 +17,195 @@ limitations under the License.
...
+	return steps, nil
+}
+
+func (s *StepsStorage) OnNodeConfigurationsChanged() chan struct{} {
+	return s.configurationsChanged
+}
+
+func (s *StepsStorage) runBufferedEmitter() chan struct{} {
+	for {
+		time.Sleep(500 * time.Millisecond)
```
nabokihms, line 103: Didn't get the purpose of this sleep.

modules/040-node-manager/crds/nodegroupconfiguration.yaml
```diff
@@ -0,0 +1,57 @@
...
+                  description: The content of the step, can be either go template or plain bash.
+                weight:
+                  type: integer
+                  default: 100
+                  description: Order of the step execution.
+                nodeGroups:
+                  type: array
+                  description: List of NodeGroups to apply the step for.
+                  default:
+                    - "*"
```
nabokihms, line 42: All node groups are affected by default? Maybe it would be more explicit to make this field required? If this field were required, users would think about which node groups to apply the custom step.

nabokihms (changes requested):

RomanenkoDenys (changes requested):

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
yalosev: Well. It makes sense on i/o operations. But I guest putting value into a memory map is not more complicated then put this value into a queue

modules/040-node-manager/crds/nodegroupconfiguration.yaml
```diff
@@ -0,0 +1,57 @@
...
+                  items:
+                    type: string
+                bundle:
+                  type: string
+                  description: Limit step execution only for specified bundle such as centos-7 / ubuntu-lts.
```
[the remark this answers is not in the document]
yalosev: As I remember we decided to split configurations for different bundles. Usually, one NodeGroup has only one bundle needed. If we make a bundle as list of strings - we will have many if/else switches inside the logic.

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
yalosev: To not emit a few time on batch updates, for example on start or on applying a few configurations. Because every configuration emitted will rerun content rendering

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
yalosev: Nope. Atm I want to see string keys here :)

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
nabokihms: Ok, it makes sense. Could you please comment in the code on why we need this sleep?

nabokihms (approved):
Approve in advance. Only one NIT left.

modules/040-node-manager/crds/nodegroupconfiguration.yaml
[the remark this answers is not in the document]
yalosev: Refactor to the list

modules/040-node-manager/images/bashible-apiserver/pkg/template/steps_storage.go
yalosev: Done

nabokihms (dismissed):

modules/040-node-manager/crds/nodegroupconfiguration.yaml
```diff
@@ -0,0 +1,67 @@
...
+                    - ["master", "worker"]
+                  items:
+                    type: string
+                bundles:
+                  type: array
+                  minItems: 1
+                  x-examples:
+                    - ["ubuntu-lts"]
+                    - ["ubuntu-lts", "centos-7"]
+                  description: Bundles for step execution. Ex. ubuntu-lts / centos-7. You can set '*' for selecting all bundles.
```
RomanenkoDenys, line 55: May be better 
```
x-examples:
  - ["ubuntu-lts", "centos-7"]
  - ["ubuntu-lts"]
  - ["*"]
```

modules/040-node-manager/crds/nodegroupconfiguration.yaml
```diff
@@ -0,0 +1,67 @@
...
+                  type: integer
+                  default: 100
+                  description: Order of the step execution.
+                nodeGroups:
+                  type: array
+                  minItems: 1
+                  description: List of NodeGroups to apply the step for. To select all NodeGroups you can set '*'.
+                  x-examples:
+                    - ["*"]
+                    - ["master", "worker"]
```
RomanenkoDenys, line 46: May be better 
```
x-examples:
  - ["master", "worker"]
  - ["worker"]
  - ["*"]
```

RomanenkoDenys (dismissed):

modules/040-node-manager/crds/nodegroupconfiguration.yaml
yalosev: fixed

RomanenkoDenys (approved):

nabokihms (approved):

PR #504, opened 2021-12-27
[cloud-provider-aws::docs] add IAM policies for creating peering connection

vitaliy-sn:
## Description
The necessary IAM policies for creating a peering connection have been added to the documentation.

The ability to create a peering connection was previously added, but the list of required IAM policies has not been updated.

## Changelog entries

```changes
module: cloud-provider-aws
type: fix
description: The necessary IAM policies for creating a peering connection have been added to the documentation.
```


apolovov (dismissed):
Added clarification about IAM.

nabokihms (approved):

PR #507, opened 2021-12-27
[github] fix(ci): make curl error more verbose

diafour:
## Description

- handle circular structures when dump js errors

Github API has limits, but there is no message about problems.


diafour:
Merged in #483

PR #509, opened 2021-12-27
[extended-monitoring] Add cert-exporter alerts

### LINKED ISSUE #365: [extended-monitoring] Alert about certificate expiration

kvaps:
## Description

This PR adds cert-exporter alerts to the extended-monitoring module

follows-up #479
fixes #365

## Changelog entries

```changes
module: extended-monitoring
type: feature
description: Add cert-exporter alerts
note: |
  Added alerts to track certificates expiration and cert-exporter health
```

### DIFF
--- a/ee/fe/modules/340-extended-monitoring/images/cert-exporter/Dockerfile
+++ b/ee/fe/modules/340-extended-monitoring/images/cert-exporter/Dockerfile
@@ -1,0 +1,19 @@
+ARG BASE_ALPINE
+FROM $BASE_ALPINE as artifact
+
+ARG VERSION=2.0.1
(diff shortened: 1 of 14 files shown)


kvaps:
Closed in favor of #512

PR #523, opened 2021-12-28
[global] copy custom certificate: serialize with struct, not map

shvgn:
## Description

Refactor to use structs instead of maps on serialization

When serializing empty slices in `map[string][]byte`, they are omitted in values.

## Changelog entries

```changes
module: global
type: fix
description: Fix serialization of empty strings in secrets
```


nabokihms (approved):

shvgn:
Follows #518

PR #524, opened 2021-12-28
[global] delete empty ca.crt entry in cert map

shvgn:
## Description

Refactor to use structs instead of maps on serialization

When serializing empty slices in `map[string][]byte`, they are omitted in values.

## Changelog entries

```changes
module: global
type: fix
description: Fix serialization of empty strings in secrets
```


shvgn:
Could be preferable to #523

shvgn:
close in favor of #524 523

PR #532, opened 2021-12-29
[monitoring][grafana] Migrate old tables to new and replace from __cell variable to __value and add time interval to url

name212:
## Description
Migrate old tables to new. 
Replace from `__cell` variable to `__value`.
Add time interval to urls.

- `__cell` variable does not work for new table
- some tables did not migrate to 'new table'. Old tables are deprecated
- some urls did not contain time interval. It is not conveniently

## Changelog entries

```changes
module: grafana
type: fix
description: Migrate old tables to new and replace from __cell variable to __value and add time interval to url
```


nabokihms (approved):

z9r5:

PR #533, opened 2021-12-29
[docs] Fix instructions for switching registry and image copier

name212:
## Description
Fix instructions for switching registry and image copier

Instructions had errors and users cannot switch registry

## Changelog entries

```changes
module: docs
type: fix
description: Fix instructions for switching registry and image copier
```


name212:
document validation was skipped, because image copier doc was changed and it does not have RU version

nabokihms (approved):

PR #538, opened 2021-12-30
build and test authz templates

name212:
## Changelog entries

```changes
module: <kebab-case>
type: fix | feature
description: <what effectively changes>
note: <what to expect>
```


name212:
some test in ci

PR #546, opened 2022-01-06
[general] new bundle for Debian 9, 10, 11 and AstraLinux 2.12 (Orel)

RomanenkoDenys:
## Description

New bundle supporting Debian 9, 10, 11 and AstraLinux 2.12(Orel).
Based on https://github.com/deckhouse/deckhouse/pull/397.

We need support of wide range of linux distros.
## Changelog entries

```changes
module: general
type: feature
description: "New bundle supporting Debian 9, 10, 11 and AstraLinux 2.12(Orel)"
note:
```


modules/040-node-manager/openapi/config-values.yaml
```diff
@@ -29,11 +29,12 @@ properties:
     default:
       - "ubuntu-lts"
       - "centos-7"
+      - "debian"
     x-examples:
       - ["ubuntu-lts"]
     items:
       type: string
-      enum: [ubuntu-lts, centos-7]
+      enum: [ubuntu-lts, centos-7, debian]
```
apolovov, line 37: astra?

modules/040-node-manager/template_tests/module_test.go
```diff
@@ -72,6 +72,7 @@ const nodeManagerConfigValues = `
 allowedBundles:
   - "ubuntu-lts"
   - "centos-7"
+  - "debian"
```
apolovov, line 75: astra?

apolovov (dismissed):

modules/040-node-manager/openapi/config-values.yaml
RomanenkoDenys: Astra discovered as debian, so bundle isn't needed

modules/040-node-manager/template_tests/module_test.go
RomanenkoDenys: Astra discovered as debian, so bundle isn't needed

nabokihms:
@konstantin-axenov promised to take a look soon.

konstantin-axenov:
Everything is fine. After rebase and squash, it can be merged.

konstantin-axenov (approved):

PR #555, opened 2022-01-11
[CI] Skip checks for automation PRs

shvgn:
## Description

Skip CI checks for Prs made by our bot.

Validations fail on changelog PRs. We want to avoid the inconvenience.


diafour (approved):
LGTM.
The `workflow_dispatch` trigger is tested in diafour/deckhouse.

PR #561, opened 2022-01-11
[local-path-provisioner] add reclaimPolicy field.

RomanenkoDenys:
## Description

1. Added reclaimPolicy selector.
2. Changed default reclaimPolicy from Delete to Retain.

Users asks us to have ability to set reclaim policy for local-path-provisioner volumes.

## Changelog entries

```changes
module: local-path-provisioner
type: feature
description: "Added reclaimPolicy selector, set default reclaimPolicy to Retain"
```


nabokihms (approved):
Approved in advance, but please fill in the description template.

z9r5:

PR #571, opened 2022-01-12
[node-manager][bashible-apiserver] Fix: no context for node groups in static cluster

name212:
## Description
If cloud provider does not present in bashible-apiserver input context will not return error for calculating checksum.

In dhctl, first will check deckhouse pod is ready and then will get and print log. It needs for add master node group. Without master node group we can get case in which control-plane manager change kublet tls settings, and dhctl cannot get logs, because kubelet certificate is become not valid.

Deleted `test_data` dir with context render test.

Static cluster does not bootstrap.

## Changelog entries

```changes
module: bashible-apiserver
type: fix
description: Fix do not render bashible bundle in static clusters .
```

```changes
module: dhctl
type: fix
description: Check deckhouse pod readiness before get logs. It fixes static cluster bootstrap.
```


yalosev (approved):
LGTM

PR #575, opened 2022-01-13
[istio][ingressgateway] static port for NodePort inlet

apolovov:
## Description
The module option `alliance.ingressGateway.nodePort.port` to set a static port for NodePort-type ingressgateway Service.

In some environments, we need to explicitly set the port number. So we will be able to configure some firewall rules for example.

## Changelog entries

```changes
module: istio
type: feature
description: `alliance.ingressGateway.nodePort.port` option to set a static port for NodePort-type ingressgateway Service.
```

### DIFF
--- a/ee/modules/110-istio/openapi/config-values.yaml
+++ b/ee/modules/110-istio/openapi/config-values.yaml
@@ -61,6 +61,17 @@ properties:
               - `LoadBalancer` — is a recommended method if you have a cloud-based cluster and it supports Load Balancing.
               - `NodePort` — for installations that do not have the LB.
             default: LoadBalancer
+          nodePort:
+            type: object
+            description: Special settings for NodePort inlet.
+            default: {}
+            x-examples: [{}, {"port": 30001}]
+            properties:
+              port:
+                type: integer
+                description: Static port number for NodePort-type Service. Must be in range, set by kube-apiserver --service-node-port-range argument (default is 30000-32767).
+                minimum: 1024
+                maximum: 65535
           serviceAnnotations:
             type: object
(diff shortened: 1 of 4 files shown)


ee/modules/110-istio/openapi/config-values.yaml
```diff
@@ -61,6 +61,17 @@ properties:
               - `LoadBalancer` — is a recommended method if you have a cloud-based cluster and it supports Load Balancing.
               - `NodePort` — for installations that do not have the LB.
             default: LoadBalancer
+          nodePort:
+            type: object
+            description: Special settings for NodePort inlet.
+            default: {}
```
name212, line 67: Can we remove default for nodePort and add `port` field into required section? I think it'll look better.

name212: We can use `oneOf` for requiring field based on enum value.
For example, we use this technique in cluster configuration for choice required fields between static and cloud params. See it [here](https://github.com/deckhouse/deckhouse/blob/6822fac856e94ff500878d120b2362ccef70b047/candi/openapi/cluster_configuration.yaml#L84)

Could we use it here for requiring nodePort object if `inlet` is `NodePort`?

Additional link https://stackoverflow.com/a/38781027

name212 (approved):

ee/modules/110-istio/openapi/config-values.yaml
apolovov: We decided to keep the original behavior.

nabokihms (approved):

PR #580, opened 2022-01-13
[node-manager] consider globally defined zones on standby calculations

31337Ghost:
## Description
When calculating maximum instanses for particular NodeGroup without zones defined — use global zones count from CloudProvider configuration.

There is a bug on calculation of standby instances desired amount. If NodeGroup has no zones defined, it's treated that it has only one zone.

## Changelog entries

```changes
module: node-manager
type: fix
description: |
  When calculating maximum instanses for particular NodeGroup without zones defined — use global zones count from CloudProvider configuration.
```


diafour (approved):
LGTM

PR #582, opened 2022-01-13
[site] Don't show a parameter's default value in an OpenAPI spec

z9r5:
Don't show a parameter's default value in an OpenAPI specification on the site if it is an empty array or hash.


z9r5:

PR #586, opened 2022-01-14
[ci] checkout sources before deploy site and docs

diafour:
## Description

werf converge requires checkout.

Checkout is a must for self-hosted runners. We end up deploying site and documentation to production environment using images built 2 days ago from PR.


z9r5:

PR #591, opened 2022-01-14
[deckhouse][docs] Fix images copier docs

name212:
## Description
Fix images copier docs.
Add commands for patch deckhouse-registry secret and deployment

`How it works` section is confuse


apolovov (changes requested):
I realized we need some tool to generate the new registry secret with new creds.
It is quite painful to generate it manually.

apolovov (approved):

nabokihms (approved):

z9r5:

Issue #107, opened 2021-09-01
[node-manager] nodeSelector for NodeUser


nabokihms:
Bashible adds users to all nodes by default. We need to add the nodeSelector property to the NodeUser spec to add users only to specific nodes.

RomanenkoDenys:
Fixed by https://github.com/deckhouse/deckhouse/pull/595

Issue #85, opened 2021-08-25
[node-manager] NodeUser multiple ssh keys


nabokihms:
NodeUser CR adds system users on nodes, and only one ssh key per user is allowed.

RomanenkoDenys:
Fixed by #595

PR #595, opened 2022-01-16
[node-manager] update nodeUser resource

RomanenkoDenys:
## Description

This PR adds:
1. Support of node-group selector for node-users.
2. Support of multiple ssh keys per user.
3. Immediate change of corresponding system user, if nodeuser crd is changed.

## Changelog entries

```changes
module: node-manager
type: feature
description: "Update nodeuser resource to support node-group selector and multiple ssh keys."
note:
```


modules/040-node-manager/images/bashible-apiserver/pkg/template/context.go
```diff
@@ -422,3 +454,188 @@ func (c *BashibleContext) saveChecksum(name, newChecksum string) {
...
+
+	go informer.Run(ctx.Done())
+
+	// Wait for the first sync of the informer cache, should not take long
+	if !cache.WaitForCacheSync(ctx.Done(), informer.HasSynced) {
+		klog.Fatalf("unable to sync caches: %v", ctx.Err())
+	}
+}
+
+func (c *BashibleContext) runBufferedEmitter() chan struct{} {
```
yalosev, line 529: Maybe we can move this emitter to some common code to not copy the code between different packages

EvgenySamoylov:
Closes #107 and #85

modules/040-node-manager/images/bashible-apiserver/pkg/template/context.go
RomanenkoDenys: its a good point, i refactor this code.

RomanenkoDenys: good point, i refactor them

modules/040-node-manager/images/bashible-apiserver/pkg/template/node_user.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if nu.PasswordHash != newSpec.PasswordHash {
+		return false
+	}
+
+	if nu.IsSudoer != newSpec.IsSudoer {
+		return false
+	}
+
+	if !slicesIsEqual(nu.NodeGroups[:], newSpec.NodeGroups[:]) {
```
yalosev, line 62: why do you use [:] here? Don't think that it's necessary.

RomanenkoDenys: fixed

yalosev (approved):

nabokihms (approved):

PR #598, opened 2022-01-17
[deckhouse] Deckhouse release requirements

yalosev:
## Description
Check requirements before applying a DeckhouseRelease

DeckhouseRelease can have incompatible changes. We can check requirements for the desired Release and block it if requirements are not met

## Changelog entries

```changes
module: deckhouse
type: feature
description: Check requirements before applying a DeckhouseRelease
```


go_lib/dependency/requirements/requirements.go
```diff
@@ -0,0 +1,85 @@
...
+	Get(path string) gjson.Result
+}
+
+type requirementsResolver interface {
+	Register(key string, f CheckFunc)
+	GetByKey(key string) (CheckFunc, error)
+}
+
+type requirementsRegistry struct {
+	funcs map[string]CheckFunc
```
name212, line 65: `checkers` maybe?

name212: not important

go_lib/dependency/requirements/requirements.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func newRegistry() *requirementsRegistry {
+	return &requirementsRegistry{
+		funcs: make(map[string]CheckFunc),
+	}
+}
+
+func (r *requirementsRegistry) Register(key string, f CheckFunc) {
+	r.funcs[key] = f
```
name212, line 75: Do we need protect `funcs` field with mutex?

modules/040-control-plane-manager/hooks/requirements.go
```diff
@@ -0,0 +1,48 @@
...
+			return false, err
+		}
+
+		currentVersionStr := getter.Get("global.discovery.kubernetesVersion").String()
+		currentVersion, err := semver.Parse(currentVersionStr)
+		if err != nil {
+			return false, err
+		}
+
+		if currentVersion.GE(desiredVersion) {
```
name212, line 40: For k8s version we have two requirements: for minimal version and for max version.

For example, release 1.29.0 does not support k8s version >= 1.22, because it uses apis were removed in 1.22 (ingress for example)

modules/020-deckhouse/hooks/update_deckhouse_image.go
```diff
@@ -404,14 +414,41 @@ func (du *deckhouseUpdater) releaseChannelUpdate(input *go_hook.HookInput) error
...
+	}
+
+	return true
+}
+
+func applyRelease(input *go_hook.HookInput, rl deckhouseRelease, ts time.Time) (applied bool) {
+	// check release dependencies
+	passed := checkReleaseRequirements(input, rl, ts)
+	if !passed {
+		input.MetricsCollector.Set("d8_release_blocked", 1, map[string]string{"name": rl.Name}, metrics.WithGroup("d8_releases"))
```
name212, line 440: Do we need to clear metric if release will be accepted?

name212 (changes requested):
max k8s version

and clear metric

modules/020-deckhouse/hooks/update_deckhouse_image.go
yalosev: In the beginning of this hook we have group cleanup
```
input.MetricsCollector.Expire("d8_releases")
```

name212 (changes requested):
max k8s version

and clear metric

modules/020-deckhouse/hooks/update_deckhouse_image.go
name212: ok. I saw it

go_lib/dependency/requirements/requirements.go
yalosev: Currently, we use this method only in `init` functions. They are sequential, that's why mutex here is not required. And we also don't have even concurrent read here

go_lib/dependency/requirements/requirements.go
yalosev: got it

modules/040-control-plane-manager/hooks/requirements.go
name212: It is wrong. I'm sorry.

We need to check all components when update k8s and itself updating deckhouse not affect this case

go_lib/dependency/requirements/requirements.go
name212: ok, close it

name212 (approved):

z9r5:

PR #600, opened 2022-01-17
[site] Update resource generator

z9r5:
Don't render OpenAPI schema if it has no fields.


z9r5:

PR #608, opened 2022-01-18
[bashible] Increase kubelet log rotation chunk size

zuzzas:
## Description

Increase kubelet log rotation size.

This change will increase log buffer and effective logline count while viewing logs via kubectl or collecting them externally.

## Changelog entries

```changes
module: bashible
type: feature
description: Log rotation changed from 10 Mi / 10 files to 50 Mi / 4 files.
note: kubelet will restart, neither docker nor containerd won't restart. The change affects **only** containerd CRI.
```


konstantin-axenov (approved):

z9r5:

PR #609, opened 2022-01-18
[user-authz] Doc: fix subject access groups option name

nabokihms:

## Description
Fix doc

The right option is the `groups`


z9r5 (approved):

z9r5:

PR #612, opened 2022-01-18
[ci] Fix google cloud platform e2e test running

name212:
## Description
Fix google cloud platform e2e test running

Google cloud platform e2e tests don't run


diafour (approved):
LGTM! The label in the repository is already renamed to `e2e/run/gcp`.

Issue #508, opened 2021-12-27
[node-manager] NodeGroup must show errors on machine creation process


konstantin-axenov:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.28

### Expected Behavior

Execution of `kubectl get ng` shows errors correctly.

### Actual Behavior

When machines can't be created because of quota limit exceeded, we can see a lot of them creating and then deleted
```
main-sentry-fc46ed96-765d4-99fw6                 Running       104m
main-sentry-fc46ed96-765d4-9xfbz                 Creating      2s
main-sentry-fc46ed96-765d4-bk5bd                 Failed        4s
main-sentry-fc46ed96-765d4-kp6dv                 Terminating   6s
main-sentry-fc46ed96-765d4-mhm9r                 Terminating   11s
main-sentry-fc46ed96-765d4-nr8p7                 Terminating   12s
main-sentry-fc46ed96-765d4-nv4f7                 Terminating   8s
main-sentry-fc46ed96-765d4-vzn6q                 Terminating   12s
```
But in status of NodeGroup we can see only
```
status:
  conditionSummary:
    ready: "False"
    statusMessage: Started Machine creation process
  desired: 3
  error: ""
  instances: 3
  lastMachineFailures:
  - lastOperation:
      description: Started Machine creation process
      lastUpdateTime: "2021-12-27T14:28:09Z"
      state: Failed
      type: Create
    name: main-sentry-fc46ed96-765d4-9xfbz
    ownerRef: main-sentry-fc46ed96-765d4
```
Machine controller manager shows error correctly
```
E1227 14:24:05.728762       1 machine.go:548] Error while creating machine main-sentry-fc46ed96-765d4-hpt9j: rpc error: code = ResourceExhausted desc = The limit on maximum number of disks has exceeded.
```

### Steps To Reproduce

Was seen in Yandex Cloud but most likely it can be reproduced in any cloud.

PR #621, opened 2022-01-19
[node-manager] NodeGroup machine creation process details into events

yalosev:
## Description
Create `events` with detailed logs when machine-controller-manager creates Machines
(Close https://github.com/deckhouse/deckhouse/issues/508)

If we have some errors on a Machine creation process (ex. quota exceeded or some limit reached), we have a description on ng status but it removes quickly because machine-controller-manager changes the state of the Machine and tries to recreate it. Instead of showing status in a NodeGroup's description we put it into the events and attach these events to a NodeGroup. It allows users to transparently observe a problem

## Changelog entries

```changes
module: node-manager
type: feature
description: Create events with detailed logs when Node provisioning fails
```


modules/040-node-manager/hooks/update_node_group_status.go
```diff
@@ -356,6 +373,49 @@ func handleUpdateNGStatus(input *go_hook.HookInput) error {
...
+
+	now := time.Now()
+
+	return &corev1.Event{
+		TypeMeta: v1.TypeMeta{
+			Kind:       "Event",
+			APIVersion: "v1",
+		},
+		ObjectMeta: v1.ObjectMeta{
+			Namespace:    "default",
```
name212, line 394: Events should be in `d8-cloud-instance-manager` namespace, where located `machine-controller-manager` `machine-deploloyments`, etc...

I think

name212: Oh, Event will be generated from NodeGroup, Ng is not namespaced object. I do not have any question.

But this event I would looking for in `d8-cloud-instance-manager` namespace :-)

name212 (dismissed):

nabokihms (changes requested):

modules/040-node-manager/hooks/update_node_group_status.go
yalosev: if you place this events into `d8-cloud-instance-manager` they will not be attached to a NodeGroup object. And when you make `kubectl describe ng worker` command, you will see no events

name212 (approved):

nabokihms (approved):

PR #624, opened 2022-01-19
[azure] Add notice that only regions with availability zones are supported

konstantin-axenov:
## Description

Dechouse supports only regions with availability zones.

There are strict requirements for Azure regions. Users must be aware of them.


deckhouse-BOaTswain:
Run workflow "deploy/web/test"...

z9r5:
Cherry-picked into release-1.29, release-1.30

Issue #410, opened 2021-12-15
[vertical-pod-autoscaler] Hook set_max_allowed triggers evictions on Deckhouse restart


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.28.0

### Expected Behavior

The effect of pods eviction should not occur on Deckhouse restarts.

### Actual Behavior

Pods should not be restarted.

### Steps To Reproduce

1. Wait till pods lifetime exceeds the required limit.
2. Restart Deckhouse.

### Additional Information

1. Deckhouse is restarted:
![image](https://user-images.githubusercontent.com/32434187/146141925-718326d6-7ffb-489c-ac73-eeac1ef55e26.png)

2. Set max allowed values of the VPA object changes:
![image](https://user-images.githubusercontent.com/32434187/146142015-931eaa40-57b9-4042-b3af-7b61763f4a4e.png)

3. Recommendations changes:
![image](https://user-images.githubusercontent.com/32434187/146142186-4b51d938-1c0a-4678-93c1-abc0d8f45f0b.png)

4. Pods are restarted:
![image](https://user-images.githubusercontent.com/32434187/146142226-75654468-d374-4432-bbf3-c15f1236b3ba.png)

nabokihms:
Fixed by https://github.com/deckhouse/deckhouse/pull/627

PR #627, opened 2022-01-20
[vertical-pod-autoscaler] vpa improvements

RomanenkoDenys:
## Description

This PR improves hook max_allowed work.
1. If the new calculated max_allowed values for pods are less than 10% less than the old values, the values are not changed.
2. Hook starts only when Deckhouse pod becomes ready.

Also this PR fixes VPA linter and fixes some errors in vpa in istio module.
 
Vpa hook which set max_allowed values causes to massive pod restart on deckhouse version update.
## Changelog entries

```changes
module: vertical-pod-autoscaler
type: fix
description: |
  1. If the new calculated max_allowed values for pods are less than 10% of old values, the values are not changed.
  2. Hook starts only when Deckhouse pod becomes ready.
```

### DIFF
--- a/modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
+++ b/modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
@@ -344,7 +344,6 @@ func containerPoliciesTreshold(newPolicies []autoscaler.ContainerResourcePolicy,
 			}
 		}
 	}
-	return
 }
 
 func isInfinity(value float64) bool {
--- a/modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
+++ b/modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
@@ -323,20 +323,20 @@ func getPathFloat64(input *go_hook.HookInput, path string) (float64, error) {
 }
 
 func containerPoliciesTreshold(newPolicies []autoscaler.ContainerResourcePolicy, oldPolicies []autoscaler.ContainerResourcePolicy) {
-	for i, newPolicy := range newPolicies {
-		cpu := newPolicy.MaxAllowed.Cpu().MilliValue()
-		memory := newPolicy.MaxAllowed.Memory().Value()
-		for _, oldPolicy := range oldPolicies {
-			if oldPolicy.ContainerName != newPolicy.ContainerName {
+	for i := range newPolicies {
+		cpu := newPolicies[i].MaxAllowed.Cpu().MilliValue()
+		memory := newPolicies[i].MaxAllowed.Memory().Value()
+		for j := range oldPolicies {
+			if oldPolicies[j].ContainerName != newPolicies[i].ContainerName {
 				continue
 			}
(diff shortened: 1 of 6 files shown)


modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
```diff
@@ -43,20 +44,21 @@ We have 3 groups of resources:
...
 const (
-	groupLabelKey  = "workload-resource-policy.deckhouse.io"
-	everyNodeLabel = "every-node"
-	masterLabel    = "master"
-	vpaAPIVersion  = "autoscaling.k8s.io/v1"
+	groupLabelKey   = "workload-resource-policy.deckhouse.io"
+	everyNodeLabel  = "every-node"
+	masterLabel     = "master"
+	vpaAPIVersion   = "autoscaling.k8s.io/v1"
+	tresholdPercent = 10
```
name212, line 61: typo

`threshold` should be here

modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
```diff
@@ -66,6 +68,27 @@ type VPA struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse pod object from unstructured: %v", err)
+	}
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == v1.PodReady && cond.Status == v1.ConditionTrue {
+			isReady = true
+			break
+		}
+	}
+	return &DeckhousePod{IsReady: isReady}, nil
```
name212, line 89: We can return boolean variable immediately.
We do not need wrap boolean var with struct.

modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
RomanenkoDenys: yep

modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
RomanenkoDenys: yep

modules/302-vertical-pod-autoscaler/hooks/set_maxallowed.go
```diff
@@ -246,3 +321,39 @@ func getPathFloat64(input *go_hook.HookInput, path string) (float64, error) {
...
+
+func isInfinity(value float64) bool {
+	return math.IsInf(value, 1) || math.IsInf(value, -1)
+}
+
+func calculatePercent(value1, value2 int64) int64 {
+	return int64(float64(value1) / float64(value2) * 100)
+}
+
+func inTreshold(valuePercent int64) bool {
```
name212: typo

RomanenkoDenys: yep

name212 (approved):

EvgenySamoylov:
Closes #410

PR #628, opened 2022-01-20
[ci] fix: remove exit 0

diafour:
## Description

Remove debug lines.

`exit 0` line makes deploy to channel jobs stop immediately.


z9r5:
Cherry-picked into release-1.29, release-1.30

PR #635, opened 2022-01-20
[ci] fix: use prNUM as docker tag for deckhouse dev images

### LINKED ISSUE #543: [ci] Use slugified branch name for container image tag
### Preflight Checklist

### Version

v1.29.0

### Expected Behavior

Expected build to be succesfull.

### Actual Behavior

Build fails for branches that contain slashes in its name because on CI, we are using [${{needs.git_info.outputs.ci_commit_tag}}](https://github.com/deckhouse/deckhouse/blob/e7356489eddcae23d6a156ecd4c00ff03c2ab7b3/.github/ci_templates/build.yml#L31) variable, which can contain forbidden characters.

### Steps To Reproduce

1. Create a branch with slashes in its name.
2. See CI logs.

### Additional Information

https://github.com/deckhouse/deckhouse/runs/4698878948

### Logs

```shell
Digest: sha256:3a42eb351b3d8301a599fe89f4db7648f4a56f1dca5df1b9785acf3c39b29f3c
Status: Downloaded newer image for ***/sys/deckhouse-oss:ae7f3a3a1ec499a7d9e8471df0e0f3dc424e7e6af826543c53e8cd81-1640932478279
***/sys/deckhouse-oss:ae7f3a3a1ec499a7d9e8471df0e0f3dc424e7e6af826543c53e8cd81-1640932478279
Publish images for branch dependabot/go_modules/github.com/onsi/ginkgo-1.16.5
Error parsing reference: "***/sys/deckhouse-oss/dev:dependabot/go_modules/github.com/onsi/ginkgo-1.16.5" is not a valid repository/tag: invalid reference format
```

diafour:
## Description

Slugify docker tag for dev images:
- use prNUM as tag for pull request branches
- use `werf slugify` for tags to deal with `+` sign

Fixes #543. It is possible now to run e2e tests for dependabot PRs.


.github/workflows/deploy-beta.yml
```diff
@@ -97,17 +104,20 @@ jobs:
               const prRepo = context.payload.pull_request.head.repo.full_name
               const prRef = context.payload.pull_request.head.ref
 
-              refName = (prRepo === targetRepo) ? prRef : `pr${context.issue.number}`
+              refSlug = `pr${context.issue.number}`;
+              refName = (prRepo === targetRepo) ? prRef : refSlug;
               githubBranch = refName
```
shvgn, line 109: Since `githubBranch` can become to "pr123", `output.ci_commit_branch` will be assigned it as well. And then we'll refer to it as `CI_COMMIT_BRANCH` in later code, that looks a little confusing since it is an image tag that is passing, not branch. Otherwise looks like it does not break things.

shvgn (changes requested):
I wish we would change naming to avoid confusion.

.github/workflows/deploy-beta.yml
diafour: The value of CI_COMMIT_BRANCH was used only in build-and-test_dev.yml as the tag for the dev image. This PR makes CI_COMMIT_BRANCH effectively a flag to detect if workflow triggers for the branch.

shvgn: Does `CI_COMMIT_BRANCH=pr123` mean the workflow is triggered for a branch? If yes, there is no "commit branch" named "pr123". Why it is called "commit branch" then? Probably, it is "build tag" or "dev build tag".

shvgn: Got reasons to simultaneous support of Gitlab and Github flows, hence the naming weirdness. Will torerate it for the time being.

shvgn (approved):

PR #637, opened 2022-01-21
[ingress-ngnix] proper validating webhook configuration for 1.22 k8s

31337Ghost:
## Description
Provide proper validating webhook configuration for Ingress Nginx Controller v1.0.

Ingress Nginx Controller supports only modern versions of ingresses and different validation path.

## Changelog entries

```changes
module: ingress-ngnix
type: fix
description: |
  Proper validating webhook configuration for k8s 1.22+.
```


yalosev (approved):
I have added one more commit. Because we have validation enabled only for controller >= 0.48. This version of the controller supports networking/v1 so we don't need extra conditions here

nabokihms (approved):

PR #639, opened 2022-01-21
[bashible] Fix remove of docker

konstantin-axenov:
## Description

There can be a situation when docker-ce is installed not from registry packages. So bashible fails on the next step because it assumes that containerd is installed correctly.

Fix switch from docker to containerd.

## Changelog entries

```changes
module: bashible
type: fix
description: Fix remove of docker
```


z9r5:

PR #640, opened 2022-01-21
[istio][multicluster] multiclustersNeedIngressGateway fix

apolovov:
## Description
Correct decision to deploy ingressgateway for multiclusters.

It was impossible to set up ingressgateways for multicluster from scratch. Loop dependency.

## Changelog entries

```changes
module: istio
type: fix
description: Correct decision to deploy ingressgateway for multiclusters.
```


nabokihms (approved):

z9r5:

PR #655, opened 2022-01-25
[dhctl] Do not print error about not exists bastion host key for abort command

name212:
## Description
If we do not have bastion host, we will see error `no such file or directory` for `abort` command.
It is not error, because we can not have bastion host.
This PR check on exists key in cache before get it from cache and output error in error cases only.

Confusing error message in `abort` command

## Changelog entries

```changes
module: dhctl
type: fix
description: Do not print error about not exists bastion host key for abort command.
```

### DIFF
--- a/werf.yaml
+++ b/werf.yaml
@@ -455,10 +455,15 @@ ansible:
         chdir: /dhctl
 ---
 image: dhctl-tests
+fromCacheVersion: "20220125"
 fromImage: base-for-go
 git:
 - add: /dhctl
   to: /dhctl
+  stageDependencies:
+    install:
+      - go.sum
+      - Makefile
 ansible:
   beforeInstall:
     - apk:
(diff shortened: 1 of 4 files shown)


werf.yaml
```diff
@@ -455,10 +455,14 @@ ansible:
...
 ---
 image: dhctl-tests
+fromCacheVersion: "20220125"
 fromImage: base-for-go
 git:
 - add: /dhctl
   to: /dhctl
+  stageDependencies:
+    install:
+    - go.sum
```
nabokihms, line 465: Please add Makefile as a stage dependency as well. According to the `make deps`, it downloads liner and other test requirements with curl, which is not dependent on the `go.sum` file.

name212: done

nabokihms (approved):
Looks good, thank you!

PR #666, opened 2022-01-26
[monitoring-kubernetes] Match OOM kills only on systemd cgroup driver

zuzzas:
## Description

Match only systemd cgroup driver CRIs.

`label_replace` regexes are not suitable for the kubelet cgroup driver.


konstantin-axenov (approved):

konstantin-axenov (approved):

z9r5:

PR #677, opened 2022-01-26
[ci] Fix extract pull request number

name212:
## Description
Remove `/head` suffix from pull request ref

E2E tests don't work.


diafour (approved):
Ok

PR #683, opened 2022-01-27
[registry-packages] fix containerd-fe build

RomanenkoDenys:
## Description

Fixed build of containerd-fe registry package.

After release of registry packages with content tagging, we lost patch in containerd-fe package 
## Changelog entries

```changes
module: registry-packages
type: fix
description: "Fixed build of containerd-fe registry package."
```

### DIFF
--- a/modules/007-registrypackages/images/containerd-fe/werf.inc.yaml
+++ b/modules/007-registrypackages/images/containerd-fe/werf.inc.yaml
@@ -23,6 +23,11 @@ git:
     stageDependencies:
       setup:
       - '**/*'
+  - add: /modules/007-{{ $.ModuleName }}/images/{{ $.ImageName }}/patches
+    to: /patches
+    stageDependencies:
+      install:
+      - '**/*'
 shell:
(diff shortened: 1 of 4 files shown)


modules/007-registrypackages/images/containerd-fe/werf.inc.yaml
```diff
@@ -23,6 +23,11 @@ git:
     stageDependencies:
       setup:
       - '**/*'
+  - add: /modules/007-{{ $.ModuleName }}/images/{{ $.ImageName }}/patches
+    to: /patches
+    stageDependencies:
+      setup:
```
konstantin-axenov, line 29: It seems that stageDependencies should be `install`. Because patches are used during `install`.

RomanenkoDenys: yes you right.

konstantin-axenov (approved):

z9r5:

PR #542, opened 2022-01-03
[general] new bundle for Centos 8

RomanenkoDenys:
## Description

New bundle supporting Centos 8.
Based on https://github.com/deckhouse/deckhouse/pull/397.
## Why do we need it, and what problem does it solve?

We need support of wide range of linux distros.
## Changelog entries

```changes
module: general
type: feature
description: "New bundle supporting Centos 8"
note:
```

PR #709, opened 2022-01-31
[bashible] remove selinux enable step

RomanenkoDenys:
## Description

 Remove selinux enable step.

If Selinux is disabled on node, step fails.
## Changelog entries

```changes
module: bashible
type: fix
description:  "Remove selinux enable step."
note:
```


konstantin-axenov (approved):

konstantin-axenov:
This is already fixed in https://github.com/deckhouse/deckhouse/pull/542, but can't be back ported.

PR #715, opened 2022-01-31
[build] fix werf cleanup step

RomanenkoDenys:
## Description

Fix werf cleanup step.

We are trying to start cleaning up the dev registry.
## Changelog entries

```changes
module: build
type: feature
description: "Fix werf cleanup step."
note:
```


diafour (approved):
lgtm

PR #722, opened 2022-02-01
[ci] fix: pull release-channel-version image before tagging

diafour:
## Description

Pull release-channel-version image before tagging.

'Build EDITION' jobs fail with 'No such image' error if release-channel-version image not available on the runner.

## Changelog entries

```changes
module: ci
type: fix
description: |
  Pull release-channel-version image before tagging.
```


z9r5:
Cherry-picked into release1-30

PR #739, opened 2022-02-02
[bashible] fix redhat version detection

RomanenkoDenys:
## Description

Fixed redhat version detection in bashible.

Bashible detection of distro version fails in case of Redhat distro due to VERSION_ID from /etc/os-release has minor part (for example 7.7 instead of 7 in Centos).

## Changelog entries

```changes
module: bashible
type: fix
description:"Fixed redhat version detection in bashible."
note:
```


konstantin-axenov (approved):

z9r5:

PR #757, opened 2022-02-02
[node-manager] handle k8s <1.20 in NG error events creation process

yalosev:

## Description
Fix event creation for kubernetes <= 1.19

`events.k8s.io/v1` API was opened for publicity in k8 1.20, before it was available only for scheduler. K8s tries to convert `events.k8s.io/v1` to `core.io/v1` but it can't convert events that bind to ClusterScope resources properly.
We can figure out a version of the k8s and create an event of the necessary type.

## Changelog entries

```changes
module: node-group
type: fix
description: Fix event creation for NG when new Machine provisioning process is failed
```

### DIFF
--- a/modules/040-node-manager/hooks/update_node_group_status.go
+++ b/modules/040-node-manager/hooks/update_node_group_status.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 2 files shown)


yalosev:
@z9r5 have to backport in 1.30

nabokihms (approved):

z9r5:

PR #766, opened 2022-02-03
[candi] Configurable kubelet log rotation

zuzzas:
## Description

Allow users to configure kubelet log rotation.

Based on a choice of a log collector and volume of logging it is desirable to configure kubelet log rotation.

## Changelog entries

```changes
module: node-manager
type: feature
description: Allows changing kubelet log rotation via new NodeGroup parameters: `containerLogMaxSize` and `containerLogMaxFiles`.
```

### DIFF
--- a/candi/openapi/node_group.yaml
+++ b/candi/openapi/node_group.yaml
@@ -501,6 +501,15 @@ spec:
                       type: string
                       x-doc-default: '/var/lib/kubelet'
                       description: Directory path for managing kubelet files (volume mounts,etc).
+                    containerLogMaxSize:
+                      type: string
+                      default: 50Mi
+                      description: Maximum log file size before it is rotated.
+                    containerLogMaxFiles:
+                      type: integer
+                      minimum: 1
+                      default: 4
+                      description: How many rotated log files to store before deleting them.
     - name: v1alpha2
       served: true
       storage: true
--- a/candi/openapi/node_group.yaml
+++ b/candi/openapi/node_group.yaml
@@ -504,12 +504,18 @@ spec:
                     containerLogMaxSize:
                       type: string
                       default: 50Mi
-                      description: Maximum log file size before it is rotated.
+                      pattern: '\d+[Ei|Pi|Ti|Gi|Mi|Ki|E|P|T|G|M|k|m]'
+                      description: |
+                        Maximum log file size before it is rotated.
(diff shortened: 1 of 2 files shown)


konstantin-axenov (approved):

candi/openapi/node_group.yaml
```diff
@@ -501,6 +501,15 @@ spec:
                       type: string
                       x-doc-default: '/var/lib/kubelet'
                       description: Directory path for managing kubelet files (volume mounts,etc).
+                    containerLogMaxSize:
```
nabokihms, line 504: How does it work with docker? I mean, if rotation is enabled in both places: docker and kubelet.

candi/openapi/node_group.yaml
```diff
@@ -501,6 +501,15 @@ spec:
...
+                      type: string
```
nabokihms, line 505: We need to add the pattern field here for validation.

candi/openapi/node_group.yaml
```diff
@@ -501,6 +501,15 @@ spec:
...
+                      default: 50Mi
+                      description: Maximum log file size before it is rotated.
+                    containerLogMaxFiles:
+                      type: integer
+                      minimum: 1
```
nabokihms, line 510: Let's also set the maximum here.

nabokihms (changes requested):

candi/openapi/node_group.yaml
zuzzas: Docker is deprecated. This change focuses the mind of an engineer on a crucial task: migrating away from docker ASAP.

nabokihms: I wonder about the impact. How setting both docker and kubelet log rotation options will affect the system? 

I assume that kubelet logs rotation will have no effect in the case of having Docker as a CRI. If so, the only thing we need to do is add a warning about it to the description.

zuzzas: Here's the relevant [check](https://github.com/kubernetes/kubernetes/blob/8d456f2b9138771b90bb17ec5cdcb4f9a11c4243/pkg/kubelet/kubelet.go#L632). It won't even start rotating logs on docker.

I'll update the description ASAP.

nabokihms (approved):

PR #770, opened 2022-02-03
Update kubernetes patch versions

konstantin-axenov:
## Description

Update Kubernetes components to the latest patch versions.
Fix discover new patch versions script.

Using the latest stable upstream versions helps to avoid debugging bugs that are already fixed.

## Changelog entries

```changes
module: global
type: fix
description: Update Kubernetes components to the latest patch versions
note: Kubernetes control-plane components and kubelet will restart for 1.20, 1.21 and 1.22 minor versions.
```

### DIFF
--- a/candi/tools/discover_new_kubernetes_patchversions.sh
+++ b/candi/tools/discover_new_kubernetes_patchversions.sh
@@ -42,21 +42,8 @@ function update_version_map() {
   fi
 }
 
-# Updates registry images to new patch versions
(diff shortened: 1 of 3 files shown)


RomanenkoDenys (approved):

shvgn:
Should we use module "candi" instead of "global"? Control-plane version changes are global for the codebase. While it affects cluster fundamentals

PR #773, opened 2022-02-03
[control-plane-manager] Support bound service account tokens and TokenRequest API

zuzzas:
## Description

1. Support bound service account tokens in Kubernetes >=1.21
2. Support TokenRequest API in all support Kubernetes versions

Currently, we do not support the aforementioned API nor [Bound Service Account Tokens](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md).

## Changelog entries

```changes
module: control-plane-manager
type: feature
description: |
  1. Support bound service account tokens in Kubernetes >=1.21
  2. Support TokenRequest API in all supported Kubernetes versions
```

### DIFF
--- a/candi/control-plane-kubeadm/config.yaml.tpl
+++ b/candi/control-plane-kubeadm/config.yaml.tpl
@@ -28,6 +28,10 @@ apiServer:
   {{- end }}
 {{- end }}
   extraArgs:
+    api-audiences: https://kubernetes.default.svc{{ if .apiserver.saTokenAPIAudiences }},{{ .apiserver.saTokenAPIAudiences | join "," }}{{ end }}
+    service-account-issuer: https://kubernetes.default.svc
+    service-account-key-file: /etc/kubernetes/pki/sa.pub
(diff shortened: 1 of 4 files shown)


apolovov:
There should be warning:
* The kube-apiserver will restart.
* All working controllers SA tokens will temporarily (up to five minutes) lose their legitimacy due to `aud` and `iss` changing. In case of continuing "Unauthorized"-errors please restart the controller's pod.

zuzzas:
@apolovov 
I'd like to implement an auto-restart procedure for all Pods with older tokens. Along with mandatory restart for Pods older than a year that use new service accounts with limited validity.

WDYT?

nabokihms (approved):

nabokihms:
Kubernetes >=v1.22 allows specifying the `--service-account-issuer` flag multiple times. The first value will be used for token generation and validation, other values are only for validation. With this Deckhouse will be able to migrate issuer with no downtime. 

Let's keep this PR simple for now without diving into migrations.

konstantin-axenov (approved):

PR #783, opened 2022-02-04
[control-plane-manager] Allow enabling additional admission plugins as well as defaulting some of them

zuzzas:
## Description

Allow enabling additional admission plugins as well as defaulting some of them

There are some useful admission controllers. Some of them are worth force-enabling.

## Changelog entries

```changes
module: control-plane-manager
type: feature
description: |
  Allow changing a list of active admission plugins via `controlPlaneManager.apiserver.admissionPlugins` configuration.
  ExtendedResourceToleration and EventRateLimit are always enabled.
```

### DIFF
--- a/modules/040-control-plane-manager/openapi/config-values.yaml
+++ b/modules/040-control-plane-manager/openapi/config-values.yaml
@@ -6,6 +6,32 @@ properties:
     description: |
       `kube-apiserver` parameters.
     properties:
+      admissionPlugins:
+        type: array
+        description: |
+          Enable additional [admission plugins](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers).
+
+          Always enabled admission plugins in addition to the ones enabled by Kubernetes itself:
+          1. ExtendedResourceToleration
+          2. EventRateLimit with the following config:
+
+          apiVersion: eventratelimit.admission.k8s.io/v1alpha1
+          kind: Configuration
+          limits:
+          - type: Namespace
+            qps: 50
+            burst: 100
+            cacheSize: 2000
+        x-examples:
+          - [ "AlwaysPullImages", "NamespaceAutoProvision" ]
+        items:
+          type: string
+          enum:
+          - AlwaysPullImages
+          - NamespaceAutoProvision
+          - OwnerReferencesPermissionEnforcement
+          - PodNodeSelector
+          - PodTolerationRestriction
       bindToWildcard:
         type: boolean
         default: false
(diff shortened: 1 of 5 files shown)


modules/040-control-plane-manager/openapi/config-values.yaml
```diff
@@ -6,6 +6,32 @@ properties:
...
+            cacheSize: 2000
+        x-examples:
+          - [ "AlwaysPullImages", "NamespaceAutoProvision" ]
+        items:
+          type: string
+          enum:
+          - AlwaysPullImages
+          - NamespaceAutoProvision
+          - OwnerReferencesPermissionEnforcement
+          - PodNodeSelector
```
nabokihms, line 33: As said in the doc, the PodNodeSelector admission controller also demands config.
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector

nabokihms (changes requested):

modules/040-control-plane-manager/openapi/config-values.yaml
zuzzas: Nope, the doc is ambiguous. And I've verified that this is not the case by reading the source [code](https://github.com/kubernetes/kubernetes/blob/f0ea54070bec90dd829b7054117d670f9f90839f/plugin/pkg/admission/podnodeselector/admission.go#L74-L97) (it get an element from a nil map, but that's fine, it just returns an empty string). It'll work just with annotations.

nabokihms: Thus you could not provide defaults, but annotations work just fine? That's good news!

Could you please add a comment to the open API spec with the link and the description you have written?

zuzzas: Done.

modules/040-control-plane-manager/openapi/doc-ru-config-values.yaml
```diff
@@ -3,6 +3,21 @@ properties:
     description: |
       Параметры `kube-apiserver`.
     properties:
+      admissionPlugins:
+        description: |
+          Включить дополнительные [admission plugin'ы](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers).
```
nabokihms, line 8: Could you please synchronize English and Russian docs?

nabokihms (changes requested):
There is one comment left, and you also need to resolve conflicts before we can merge it.

modules/040-control-plane-manager/openapi/doc-ru-config-values.yaml
zuzzas: Done.

nabokihms (approved):

PR #784, opened 2022-02-04
[control-plane-manager] Default scheduler config

zuzzas:
## Description

Specify a scheduler config.

Define a scheduler config that spreads Pods between zones with finer granularity than before.

## Changelog entries

```changes
module: control-plane-manager
type: feature
description: Define default config that spreads Pods between zones with finer granularity than before.
```


deckhouse-BOaTswain:
Run workflow "e2e/run/aws"...

nabokihms (approved):

PR #812, opened 2022-02-08
[ci] fix: use tagged dev/install image to run in e2e tests

diafour:
## Description

Save 'dev/install' image name and use `docker run` to start e2e tests.

`werf run` could rebuild some stages and it makes e2e test last more than 30 minutes.

## Changelog entries

```changes
module: ci
type: fix
description: use docker run to start e2e tests
```

### DIFF
--- a/.github/ci_templates/e2e_tests.yml
+++ b/.github/ci_templates/e2e_tests.yml
@@ -24,15 +24,7 @@
   LAYOUT_VSPHERE_BASE_DOMAIN: ${{ secrets.LAYOUT_VSPHERE_BASE_DOMAIN }}
 {!{- end }!}
 run: |
-  echo "Workflow url: ${{github.server_url}}/${{github.repository}}/actions/runs/${{github.run_id}}"
-
-  werfRunProg=$(cat <<'END_WERF_RUN'
-
-  # WERF_REPO should be set by caller.
-  type werf && source $(werf ci-env github --verbose --as-file)
-
-  echo "Execute '{!{ $script }!} {!{ $script_arg }!}' via 'werf run' ...
-  Use environment:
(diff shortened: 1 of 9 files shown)


.github/ci_templates/build.yml
```diff
@@ -157,5 +162,13 @@ steps:
         fi;
       fi
 {!{- end }!}
+
+  - name: Save dev/install image name to cache after build
+    if: ${{ env.WERF_ENV == 'FE' }}
+    uses: {!{ index (ds "actions") "actions/cache" }!}
```
nabokihms, line 168: I did not get the part about caching. AFAIK, you can use the image `${registry}/sys/deckhouse-oss/dev/install:prXXX`, and it should be in the registry after build. 

E2E should not require running build again if the image is in the registry.

diafour: This caching is an assurance that E2E test uses the image built for the latest commit on the branch. It is not a requirement to build image again, it is a mechanism to prevent running e2e tests before "Build FE" job finishes for latest commit.

On the other hand, using prXXX image is more convenient and caching is not a reliable tool, as it expire in 2 weeks. We already have problems with expired modules_images_json files. Also, werf sets a `werf-project-repo-commit` label on the image, so we can check if the image is related to the current commit and fail the E2E job.

There is one dangerous situation, when "Build FE" and "E2E: provider" jobs run on the same runner and interfere with each other, but we have dedicated runners for e2e.

diafour:
I've changed workflows to use images tagged as 'prXXX' or 'main' or with a Git tag.

> Also, werf sets a werf-project-repo-commit label on the image, so we can check if the image is related to the current commit and fail the E2E job.

This assumption is wrong. The build job uses the stage image from cache, so commit in the label is not always a current commit.

nabokihms (approved):

Issue #830, opened 2022-02-09
[helmlib] Update CSI templates


kvaps:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

- The CSI manifests are outdated, we need to update them.
- LINSTOR module should reuse these manifests https://github.com/deckhouse/deckhouse/pull/746

### Proposed Solution

- Update CSI-images for Kubernetes >1.21.0 
- Add `csi-snapshotter` and `livenessprobe`
- Update permissions in `helm_lib_csi_controller_rbac`
- Replace StatefulSet by Deployment and enable `leader-election`

PR #831, opened 2022-02-09
[common] Update CSI

kvaps:
## Description

- [ ] Update CSI-images for Kubernetes >1.21.0 
- [ ] Add `csi-snapshotter` and `livenessprobe`
- [ ] Update permissions in `helm_lib_csi_controller_rbac`
- [ ] Replace StatefulSet by Deployment and enable `leader-election`

fixes https://github.com/deckhouse/deckhouse/issues/830
- The CSI manifests are outdated, we need to update them.
- LINSTOR module should reuse these manifests https://github.com/deckhouse/deckhouse/pull/746

## Changelog entries

```changes
module: common
type: feature
description: Update csi images and manifests
```

### DIFF
--- a/candi/version_map.yml
+++ b/candi/version_map.yml
@@ -188,7 +188,7 @@ k8s:
       kubeProxy: sha256:092f9526686d27964d17be772c42cde086690209cc8aea10c49991456eb879c2
   '1.20':
     status: available
-    patch: 15
+    patch: 14
     cniVersion: 0.8.7
     bashible: *bashible_k8s_ge_1_19
     ccm:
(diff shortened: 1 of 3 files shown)


helm_lib/templates/_csi_controller.tpl
```diff
@@ -8,6 +8,11 @@
   {{- $provisionerTimeout := $config.provisionerTimeout | default "600s" }}
   {{- $attacherTimeout := $config.attacherTimeout | default "600s" }}
   {{- $resizerTimeout := $config.resizerTimeout | default "600s" }}
+  {{- $snapshotterTimeout := $config.snapshotterTimeout | default "600s" }}
+  {{- $provisionerWorkers := $config.provisionerWorkers | default "10" }}
```
kvaps, line 12: It's up to consider, the default value 100, but I found that not every previsioner is capable to handle that, so 10

candi/version_map.yml
```diff
@@ -188,7 +188,7 @@ k8s:
       kubeProxy: sha256:092f9526686d27964d17be772c42cde086690209cc8aea10c49991456eb879c2
   '1.20':
     status: available
-    patch: 15
```
zuzzas, line 191: What's happening here and later on? Downgrade?

zuzzas (changes requested):

candi/version_map.yml
kvaps: I got these values from the old fork, fixed

zuzzas (approved):

PR #833, opened 2022-02-09
[dhctl] Create additional kube resources in order

name212:
## Description
Create additional kube resurces in order instead create group by GVK

If you need to create namespace and resource in this namespace, you can get error because dhctl can create resource before namespace. It happens because we group resources by GVK into map and maps in go are unordered.

## Changelog entries

```changes
module: dhctl
type: feature
description: Create additional kube resources in order
```


deckhouse-BOaTswain:
Run workflow "e2e/run/openstack"...

nabokihms:
It is probably a good idea to use the same resources order as helm
https://github.com/helm/helm/blob/9ad53aac42165a5fadc6c87be0dea6b115f93090/pkg/tiller/kind_sorter.go#L29

As we already use helm in ductal, you can just change the rendering function to the one which helm uses, and it will work just fine.

name212:
@nabokihms 
We are not using helm in `dhctl`. But your proposition is interesting. I think about sorts resources after rendering. Thank you!

nabokihms (approved):

Issue #461, opened 2021-12-20
[deckhouse] 'Cluster is being updated' alert


EvgenySamoylov:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

We need to notify teams somehow about cluster updates.
When a release is published on the release channel the clusters with 'automatic' update type are going to update right away. The clusters with 'update windows' in place will be updated somewhen later. Moreover, when the canary release will come out we will have all the clusters updated at different times even on the single release channel. All that causes the misunderstanding among the users if Deckhouse is processing something in the cluster right now or not?

### Proposed Solution

It would be nice to have an alert that starts firing when the updating begins and resolves after it finishes. We also think about the cluster-specific description of what components will be updated during the release in the cluster.

Issue #235, opened 2021-10-18
[deckhouse] mute alerts about deckhouse pod during the release upgrade


31337Ghost:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

During the deckhouse release upgrade, we are receiving alerts about its Pod like `D8DeckhouseUnavailable`, `D8DeckhousePodIsNotRunning`, `D8DeckhousePodIsNotReady`. 

But it's an expected behavior so we should consider muting those alerts during the update is processing.

#### Explicit
We might have a special CR for handling deckhouse module update status and a special exporter for those CR's statuses. Then consider this metric in deckhouse Prometheus rules expression.

#### Simple
Export a metric "deckhouse_is_updating: (0|1)" in a special deckhouse global hook and consider it in rules expression.

For both cases, we should provide an additional alert `D8UpdatingTooLong` to be sure the update process is not stuck.

PR #845, opened 2022-02-10
[deckhouse] suppress alerts on update

yalosev:

## Description
Create alert on Deckhouse update
Add possibility to ignore alerts annotated with `d8_ignore_on_update` when Deckhouse is updating

1. After adding canary releases and update windows - we don't know the exact time of deploying a Deckhouse release. But we should notify users when the update process is really being started.
2. We are not interested in some alerts during the update process. Now we can ignore them by adding an annotation

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Adding alert about the beginning of the Deckhouse update
impact_level: low
```


EvgenySamoylov:
Closes #461 and #235

nabokihms (approved):

PR #850, opened 2022-02-10
[node-manager] add error file when bashible context failed

yalosev:

## Description
Add file with error on failed context building

We can lose errors in bashible-apiserver logs and it becomes complicated to debug some corner-case situations

## Changelog entries

```changes
module: node-manager
type: feature
description: Add file with context-building error on failure
```

### DIFF
--- a/modules/040-node-manager/images/bashible-apiserver/pkg/template/context.go
+++ b/modules/040-node-manager/images/bashible-apiserver/pkg/template/context.go
@@ -360,8 +360,11 @@ func (c *BashibleContext) update() {
 			_, _ = errStr.WriteString(fmt.Sprintf("\t%s: %s\n", bundle, err))
 		}
 		klog.Warningf("bundles checksums have errors:\n%s", errStr.String())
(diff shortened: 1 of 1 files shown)


modules/040-node-manager/images/bashible-apiserver/pkg/template/context.go
```diff
@@ -360,9 +360,12 @@ func (c *BashibleContext) update() {
 			_, _ = errStr.WriteString(fmt.Sprintf("\t%s: %s\n", bundle, err))
 		}
 		klog.Warningf("bundles checksums have errors:\n%s", errStr.String())
+		_ = ioutil.WriteFile("/tmp/context.error", []byte(errStr.String()), 0444)
```
name212, line 363: 644

name212 (changes requested):
mode

modules/040-node-manager/images/bashible-apiserver/pkg/template/context.go
yalosev: Indeed

RomanenkoDenys (approved):

PR #861, opened 2022-02-11
[docs] Fix article text in «The deckhouse-web module: usage» page

Zhbert:
## Description

The "The deckhouse-web module: usage" article reviewing and fixing.

## Changelog entries

```changes
module: docs
type: fix
description: "Review and fix the 'The deckhouse-web module: usage' article."
```

### DIFF
--- a/modules/810-deckhouse-web/docs/USAGE.md
+++ b/modules/810-deckhouse-web/docs/USAGE.md
@@ -4,6 +4,8 @@ title: "The deckhouse-web module: usage"
 
 ## An example of the module configuration
 
+Below is a simple example of the module configuration:
+
 ```yaml
 deckhouseWeb: |
   nodeSelector:
@@ -17,3 +19,5 @@ deckhouseWeb: |
     authSignInURL: "https://<applicationDomain>/sign-in"
     authResponseHeaders: "Authorization"
 ```
+
+You can also configure additional parameters. More information about this [in the next chapter](./configuration.html).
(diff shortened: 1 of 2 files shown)


modules/810-deckhouse-web/docs/USAGE.md
```diff
@@ -4,6 +4,8 @@ title: "The deckhouse-web module: usage"
 
 ## An example of the module configuration
 
+Below is a simple example of the module configuration:
```
z9r5, line 7: @Zhbert no need to repeat the same text I think. The header contains the same text.

modules/810-deckhouse-web/docs/USAGE.md
```diff
@@ -17,3 +19,5 @@ deckhouseWeb: |
     authSignInURL: "https://<applicationDomain>/sign-in"
     authResponseHeaders: "Authorization"
 ```
+
+You can also configure additional parameters. More information about this [in the next chapter](./configuration.html).
```
z9r5, line 23: @Zhbert I think it is unnecessary to link to the configuration page here. It is not a guide or something like that.

modules/810-deckhouse-web/docs/USAGE.md
Zhbert: @z9r5 Maybe you're right. But I've always thought that a headline is a headline. After it, you need some kind of introduction. And the black square just behind the title looks ugly :)

modules/810-deckhouse-web/docs/USAGE.md
Zhbert: @z9r5 OK.

PR #864, opened 2022-02-11
Add enviroment variables to CSI

kvaps:
## Description
New CSI-controller requires NAMESPACE environment variable set.
Official templates are always include NAMESPACE and ADDRESS environment variables set while our templates are not.
This PR updates our template to use environment variables too.

## Changelog entries

```changes
module: common
type: fix
description: fix csi-controller requires NAMESPACE environment variable
```

### DIFF
--- a/helm_lib/templates/_csi_controller.tpl
+++ b/helm_lib/templates/_csi_controller.tpl
@@ -106,7 +106,7 @@ spec:
         args:
         - "--timeout={{ $provisionerTimeout }}"
         - "--v=5"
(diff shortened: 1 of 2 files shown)


zuzzas (approved):

deckhouse-BOaTswain:
Running workflow "e2e/run/aws"...

shvgn:
@kvaps May I ask you not to use "common" in the module field of changes? We have not so big number of modules changed in the diff. It is better to be more precise in this case

kvaps:
OK. Got it!

PR #868, opened 2022-02-11
[candi] proper work with astra bundle in EE/FE

RomanenkoDenys:
## Description

[candi] Proper work with astra bundle in EE/FE.

Because Astra Linux bundle exists only in EE/FE edition, we remove `astra` from allowedBundles.
## Changelog entries

```changes
module: candi
type: fix
description: "Proper work with astra bundle in EE/FE."
```


nabokihms (approved):

z9r5:

PR #869, opened 2022-02-11
[ingress-nginx] Expose status class label values with the xx suffix

nabokihms:

## Description
Add the `xx` suffix to the `status_class` label of the `ingress_nginx_detail_backend_responses_total` metric.

This PR fixes panels for ingress-nginx controllers in Grafana. As for now, users an only see zeros in some tables, e.g.

image

## Changelog entries
```changes
section: ingress-nginx
type: fix
summary: "Expose status class label values with the xx suffix. Fix info about backends on dashboards."
impact_level: high
impact: |
  All ingress controllers >= 0.33 will restart.
```


nabokihms:
@konstantin-axenov, let's not backport this PR because it will trigger ingress controller restarts.

PR #876, opened 2022-02-14
[node-manager] Fix constant OOM on small-sized nodes

zuzzas:
## Description

The previous configuration ensured almost constant OOMs on small-sized Nodes.

The amount of available memory is calculated as a minimum of 512 MB and 5% of the total node memory. If the amount of free node memory is less than the calculated value, the oom is triggered.
## Changelog entries

```changes
section: node-manager
type: fix
summary: Fix constant OOM on small-sized nodes.
impact_level: low
```


RomanenkoDenys (approved):

z9r5:

PR #888, opened 2022-02-14
[node-manager] sort node users by name

yalosev:

## Description
Sort node-users output in bashible-apiserver

Users are not sorted and their order could be different => we can have different checksum for bashible

## Changelog entries

```changes
section: node-manager
type: fix
summary: Sort node users for persistance
impact_level: low
```


RomanenkoDenys (approved):

nabokihms (approved):

z9r5:

PR #890, opened 2022-02-15
Add post bootstrap script exec

name212:
## Changelog entries

```changes
section: <kebab-case of a modules/*> | <1st level dir in the repo>
type: fix | feature
summary: <what effectively changes in a single line>
impact_level: low | high*
impact: <what to expect, possibly multi-line>, required if impact_level is high
```


deckhouse-BOaTswain:
Running workflow "e2e/run/openstack"...

dhctl/cmd/dhctl/commands/bootstrap/bootstrap.go
```diff
@@ -18,6 +18,8 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/deckhouse/deckhouse/dhctl/pkg/operations/bootstrap"
```
nabokihms, line 21: It looks like imports ain't sorted. How did our linter miss it?

dhctl/pkg/operations/bootstrap.go
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO structure this functions into classes and move to operations/bootstrap module
```
nabokihms, line 15:
```suggestion
// TODO structure these functions into classes and move to the operations/bootstrap module
```

dhctl/pkg/system/ssh/frontend/upload-script.go
```diff
@@ -71,17 +84,35 @@ func (u *UploadScript) Execute() (stdout []byte, err error) {
...
-		cmd = NewCommand(u.Session, "/tmp/"+scriptName, u.Args...).Sudo()
+		scriptFullPath = "/tmp/" + scriptName
+		cmd = NewCommand(u.Session, scriptFullPath, u.Args...).Sudo()
 	} else {
-		cmd = NewCommand(u.Session, "./"+scriptName, u.Args...).Cmd()
+		scriptFullPath = "./" + scriptName
+		cmd = NewCommand(u.Session, scriptFullPath, u.Args...).Cmd()
+	}
+
+	if u.chmod {
```
nabokihms, line 96: It doesn't look like modifying permission is dhctl job. Our soft should either set permissions automatically or forget about that idea. WDHY? Is there a reason behind adding chmod?

dhctl/pkg/operations/bootstrap/state.go
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO move all states from operations/bootstrap to here
+
+package bootstrap
+
+import "github.com/deckhouse/deckhouse/dhctl/pkg/state"
+
+const PostBootstrapResultCacheKey = "post-bootstrap-result"
```
nabokihms, line 21: I completely missed the part about caching bootstrap results. Could you please explain to me why do we need to cache it? 

(From my point of view, users can write any script and output whenever they want. It should not be idempotent or reproducible).

dhctl/pkg/operations/bootstrap/post-script.go
```diff
@@ -0,0 +1,102 @@
...
+var resultPattern = regexp.MustCompile("(?m)^Result of post-bootstrap script:(.+)$")
+
+func (e *PostBootstrapScriptExecutor) run() (string, error) {
+	var result string
+	cmd := e.sshClient.UploadScript(e.path).
+		WithSetExecuteModeBefore(true).
+		WithTimeout(e.timeout).
+		Sudo()
+
+	out, err := cmd.Execute()
```
nabokihms, line 87: I assume that live output should be the default for post bootstrap scripts, or one of the options at least.

dhctl/pkg/app/bootstrap.go
```diff
@@ -53,6 +57,24 @@ func DefineDeckhouseFlags(cmd *kingpin.CmdClause) {
...
+Experimental. This feature may be deleted in the future.`).
+		Envar(configEnvName("POST_BOOTSTRAP_SCRIPT_PATH")).
+		StringVar(&PostBootstrapScriptPath)
+
+	cmd.Flag("post-bootstrap-script-timeout", "Timeout to execute after bootstrap resources script. Experimental. This feature may be deleted in the future.").
+		Envar(configEnvName("POST_BOOTSTRAP_SCRIPT_TIMEOUT")).
+		Default(PostBootstrapScriptTimeout.String()).
+		DurationVar(&PostBootstrapScriptTimeout)
+
+	cmd.Flag("post-bootstrap-script-exit-if-failed", "Exit with error if post bootstrap script was failed. Experimental. This feature may be deleted in the future.").
```
nabokihms, line 73: I have a concern about this option. Should out soft think about it, or should the user write a script the way it will skip errors?

nabokihms (changes requested):
Good job! I have opened a couple of threads to discuss implementation nuances.

dhctl/pkg/operations/bootstrap/state.go
name212: We can use result from script in another applications. Share information between another applications via cache (state).

dhctl/cmd/dhctl/commands/bootstrap/bootstrap.go
name212: I'ill see it

dhctl/pkg/system/ssh/frontend/upload-script.go
name212: Ok. Now we check permissions of script before execution

dhctl/pkg/operations/bootstrap/post-script.go
name212: 1. We need output for logging.
2. We need output for save results. See [this comment](https://github.com/deckhouse/deckhouse/pull/890#discussion_r816927089)

And we need it for prevent `merged` output from another jobs.

deckhouse-BOaTswain:

dhctl/pkg/operations/bootstrap/state.go
nabokihms: Let's keep things simple. I do not see cases when we need to share specific information through cache. From my point of view, it is a complication without benefits.

dhctl/pkg/operations/bootstrap/post-script.go
nabokihms: 1. This is ok. The live output presumes that all entries will be in a log.
2. Yes, that's why we need to abstain from caching and make the output live. Users should see what's going on in the cluster. I assume that the postscript will be used to discover some data from the cluster, and if I were a user, I would like to see discovery attempts results.

nabokihms (changes requested):

dhctl/pkg/operations/bootstrap/state.go
name212: Load balancer IP, for example

dhctl/pkg/system/ssh/frontend/upload-script.go
name212: done

dhctl/pkg/operations/bootstrap/state.go
nabokihms: We discussed it internally and agreed to implement this, yet slightly change the logic of how we parse the output.

```bash
#! /bin/bash

# $OUTPUT="/tmp/dhctl-exec-script-adasdasda"
kubectl get pods -A > $OUTPUT
```

* $OUTPUT is the env variable that points to the tmp file.
* At the end of the script, we read the data from this file and save it to the cache.

dhctl/pkg/operations/bootstrap/post-script.go
nabokihms: Nikolay will debug the problem a little more. If there is no success, he will leave a to-do message, and we will merge it as is.

nabokihms (approved):

PR #913, opened 2022-02-16
[user-authz] Webhook probe

nabokihms:

## Description
* Implement the check method
* Add it to liveness probe
* Increase liveness probe thresholds

This will help us restart a pod if the API becomes unavailable for it (we faced a problem once).

## Changelog entries

```changes
section: user-authz
type: fix
summary: Add proper liveness probe which checks that webhook can access Kubernetes API
impact_level: low
```


konstantin-axenov (approved):

z9r5:

PR #915, opened 2022-02-16
Changelog: validate sections

shvgn:
## Description

Validate changelog section

Avoid typos and reduce mis-using

## Changelog entries

```changes
section: ci
type: feature
summary: Added validation of changelogs sections
```

### DIFF
--- a/.github/actions/milestone-changelog/action.yml
+++ b/.github/actions/milestone-changelog/action.yml
@@ -36,6 +36,78 @@ runs:
       with:
         token: ${{ inputs.token }}
         pull_requests: ${{ steps.merged_milestone.outputs.prs }}
+        # section:forced_impact_level
+        allowed_sections: |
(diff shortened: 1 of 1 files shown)


z9r5 (approved):
I would only add the linstor module

.github/actions/milestone-changelog/action.yml
```diff
@@ -36,6 +36,78 @@ runs:
       with:
         token: ${{ inputs.token }}
         pull_requests: ${{ steps.merged_milestone.outputs.prs }}
+        # section:forced_impact_level
+        allowed_sections: |
```
nabokihms, line 40: I assume that it would be more convenient for the developers to autogenerate this list to ensure that all modules are present. Anyway, let's merge it as is.

nabokihms (approved):

PR #921, opened 2022-02-17
[ingress-nginx] max surge and manual rollout

yalosev:

## Description
Add maxSurge==1 for clusters >=1.21 k8s version.
Add hook for manual Daemonset rollout for <1.21 clusters

Ingress controller pod terminating and only then creating a new pod during rollout. This could be a problem in some heavy-loaded clusters. Adding maxSurge changes this behavior.
Sometime native DaemonSet rollout removes two pods concurrently instead of sequential update. We can handle this situation by manual rollout control via the hook. It seems like a bug in old k8s version, guess we can disabled it after everyone will upgrade to 1.21+

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: Manual update for ingress controllers
impact_level: high
impact: ingress nginx controllers will be removed manually by hook
```

### DIFF
--- a/candi/control-plane-kubeadm/config.yaml.tpl
+++ b/candi/control-plane-kubeadm/config.yaml.tpl
@@ -35,7 +35,7 @@ apiServer:
 {{- end }}
     anonymous-auth: "false"
 {{- if semverCompare ">= 1.21" .clusterConfiguration.kubernetesVersion }}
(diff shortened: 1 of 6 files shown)


nabokihms (approved):

z9r5:

PR #946, opened 2022-02-24
[global] added global proxy setup.

RomanenkoDenys:
## Description

Added global proxy setup for modules.

If we work in the restricted environment, when all external requests are doing thru proxy server, we need config option to set global proxy to use.

## Changelog entries

```changes
section: global
type: feature
summary: Added global proxy setup for modules.
impact_level: low
```


ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
```diff
@@ -0,0 +1,122 @@
...
+
+func newMadisonProxy(madisonScheme, madisonBackend, madisonAuthKey string) http.Handler {
+	transport := http.DefaultTransport
+	transport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	return &httputil.ReverseProxy{
+		Transport: LoggingRoundTripper{transport},
+		Director: func(req *http.Request) {
+			req.URL.Scheme = madisonScheme
+			req.URL.Host = madisonBackend
+			if req.URL.Path == "/api/v1/alerts" || req.URL.Path == "/api/v2/alerts" {
```
nabokihms, line 112: Just an idea, your code is also good.
```go
switch req.URL.Path {
case "/api/v1/alerts", "/api/v2/alerts":
  req.URL.Path = "/api/events/prometheus/" + madisonAuthKey
case "/readyz":
  req.URL.Path = "/healthz"
default:
  log.Fatalf("path %q is not allowed", req.URL.Path)
}
```

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
```diff
@@ -0,0 +1,122 @@
...
+
+	mux.HandleFunc("/healthz", readyHandler)
+	mux.HandleFunc("/", proxy.ServeHTTP)
+
+	s := &http.Server{
+		Addr:    listenHost + ":" + listenPort,
+		Handler: mux,
+	}
+	go func() {
+		err := s.ListenAndServe()
```
nabokihms, line 75: How would the main goroutine exit if there was an error in this goroutine? I assume that the kubernetes liveness probe will help us there, yet it is not always right to rely on external systems behavior.

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
```diff
@@ -0,0 +1,122 @@
...
+	"net/http/httputil"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const logFormat = `request:%s %s %s, status:%s, body:%s`
```
nabokihms, line 21: Previously, the proxy logs were in JSON format. Logrus natively supports JSON.

Anyway, let's leave logs recording as it is right now in this PR. It is not a big deal.

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
```diff
@@ -0,0 +1,122 @@
...
+	res, e = lrt.Proxied.RoundTrip(req)
+
+	log.Infof("%s", string(dump))
+	if res != nil {
+		log.Infof("response: %s", res.Status)
+	}
+	return
+}
+
+var (
```
nabokihms, line 43: This would be better to configure these options with environment variables, but global variables are also ok for our case anyway.

nabokihms (dismissed):

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
RomanenkoDenys: yep

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
RomanenkoDenys: yes, it's more clean code

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
RomanenkoDenys: it can be set from envs, but have default values now

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
RomanenkoDenys: now log.SetFormatter is set to log.JSONFormatter{}

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
RomanenkoDenys: But log.Fatalf is not a good option, because when URL is wrong, pod is restarted

ee/modules/600-flant-integration/images/madison-proxy/madison-proxy.go
RomanenkoDenys: added second channel

nabokihms (approved):

PR #947, opened 2022-02-24
[deckhouse-web] Do not set external authentication by default

nabokihms:

## Description
Delete default value for externalAuthentication

It is impossible to disable external authentication with current default values. The `auth.password` field is always removed by the hook because the external authentication field always exists.

## Changelog entries

```changes
section: deckhouse-web
type: fix
summary: Do not set external authentication by default.
impact_level: low
```


konstantin-axenov (approved):

z9r5:

PR #954, opened 2022-02-24
[node-manager] Fix alpha2_to_v1 conversion for CloudPermanent NodeGroup type

konstantin-axenov:
## Description

Fix jq

If there is more than one Hybrid NodeGroup in the cluster then conversion works only for one of NodeGroups.


yalosev (approved):

z9r5:

PR #955, opened 2022-02-24
[ingress-nginx] background remove of a pod

yalosev:

## Description
Remove pod in background. Also use ControllerRevision as point of truth about controller generation

Ingress pod can be in a terminating state for some time and hook fails with timeout error
Also DS can write a wrong generation

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: remove pod in a background
impact_level: low 
```


nabokihms (approved):

z9r5:

PR #958, opened 2022-02-25
[control-plane-manager] Encryption at rest

zuzzas:
## Description

Enables use of encryption at rest mode for apiserver (https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/)

Encryption at rest solves problem with unencrypted secrets in etcd storage.
## Changelog entries

```changes
type: feature
summary: Encryption at rest.
impact_level: low
```

### DIFF
--- a/modules/040-control-plane-manager/openapi/config-values.yaml
+++ b/modules/040-control-plane-manager/openapi/config-values.yaml
@@ -171,7 +171,7 @@ properties:
             # Avoid trailing slash
             pattern: ^[a-zA-Z0-9_/.-]+[a-zA-Z0-9_.-]$
             default: /var/log/kube-audit
-      encryptingAtRestEnabled:
+      encryptionEnabled:
         type: boolean
         default: false
         description: |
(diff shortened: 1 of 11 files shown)


modules/040-control-plane-manager/openapi/config-values.yaml
```diff
@@ -171,7 +171,13 @@ properties:
             # Avoid trailing slash
             pattern: ^[a-zA-Z0-9_/.-]+[a-zA-Z0-9_.-]$
             default: /var/log/kube-audit
-
+      encryptingAtRestEnabled:
```
konstantin-axenov, line 174: What about renaming this parameter to `encryptionEnabled`?

RomanenkoDenys: yep, it is good naming

RomanenkoDenys (approved):

konstantin-axenov (approved):

PR #964, opened 2022-02-25
[node-manager] use step number with leading zeroes for custom steps

RomanenkoDenys:
## Description

 Use step number with leading zeroes for custom steps.

When we use NodeGroupConfigurations, we set number for custom step. But in system steps numbers with leading zeroes, so we cannot fit in our custom step in system steps.
## Changelog entries

```changes
section: node-manager
type: fix
summary: Use step number with leading zeroes for custom steps.
impact_level: low
```


nabokihms (approved):

z9r5:

PR #1001, opened 2022-03-03
[candi] Suspend Astra Linux support

konstantin-axenov:
## Description

Astra Linux support is suspended.

There are several known issues and we are considering bundles refactoring. Probably it's better to provide better Debian support.


z9r5:

PR #1028, opened 2022-03-09
[user-authz] Validate service account names

nabokihms:

## Description
Add validation webhook

* Validate ServiceAccount name using regex
* Validate that namespace of the ServiceAccount is provided
* Validate that namespace is not provided for the Group and UIser subjects

## Changelog entries

```changes
section: user-authz
type: fix
summary: Add more validations for the ClusterAuthorizationRule subjects section
impact_level: low
```


RomanenkoDenys (approved):
looks good

PR #1044, opened 2022-03-14
[candi] Fix docker-stuck-containers-cleaner unit

konstantin-axenov:
## Description

- escape curly braces to prevent faulty bash script rendering
- get repo tag for comparison with pattern

```
if ! image="$(docker container inspect "$container_id" -f '<no value>' 2> /dev/null)"; then
      echo_err "Container $container_id was not found or it was remove on previous run"
      continue
fi
```

## Changelog entries

```changes
section: candi
type: fix
summary: Fix docker-stuck-containers-cleaner unit
impact_level: low
```


name212 (approved):

z9r5:

PR #1060, opened 2022-03-17
[helm_lib][istio] control-plane resources management

apolovov:
## Description
New helm lib for resource management and it's application to istiod.

We need a way to fine tune the istiod resources. Instead of standard VPA for requests we need limits for example.

## Changelog entries

```changes
section: istio
type: feature
summary: New helm lib for resource management and it's application to istiod.
impact_level: low
```


ee/modules/110-istio/openapi/config-values.yaml
```diff
@@ -271,6 +271,122 @@ properties:
...
+                  The VPA usage mode.
+                enum: ['Initial', 'Auto']
+                default: 'Auto'
+              cpu:
+                type: object
+                default: {}
+                description: |
+                  CPU-related parameters.
+                properties:
+                  max:
```
yalosev, line 323: As I remember we use the next line to describe CPU:
```
  max:
    oneOf:
      - type: string
        pattern: "^[0-9]+m?$"
      - type: number
 ```

because it could be set in 2 ways: `1` or `1000m`

File: ee/modules/110-istio/openapi/config-values.yaml [hunk 2]
```diff
@@ -271,6 +271,122 @@ properties:
...
+                    examples: [1.5]
+                    description: |
+                      The coefficient to calculate cpu limits. It is optionally used to calculate initial limits for Pod. VPA for its part keeps the initial limits/requests ratio during further resource tunings.
+              memory:
+                type: object
+                default: {}
+                description: |
+                  The amount of memory requested.
+                properties:
+                  max:
```
yalosev, line 344:
```
max:
    oneOf:
      - type: string
        pattern: "^[0-9]+(E|P|T|G|M|K|Ei|Pi|Ti|Gi|Mi|Ki)?$"
      - type: number
```
same as CPU

ee/modules/110-istio/openapi/config-values.yaml
```diff
@@ -271,6 +271,122 @@ properties:
...
+                    description: |
+                      Maximum allowed memory requests.
+                    default: '2Gi'
+                    type: string
+                  min:
+                    description: |
+                      Minimum allowed memory requests.
+                    default: '256Mi'
+                    type: string
+                  limitRatio:
```
yalosev, line 354: don't we need some default value here?

yalosev (changes requested):

ee/modules/110-istio/openapi/config-values.yaml
apolovov: no, it is a meaningful choice.

yalosev (approved):

PR #1064, opened 2022-03-17
[dhctl] Add `node.kubernetes.io/exclude-from-external-load-balancers` label on control-plane nodes

name212:
## Description
Add node.kubernetes.io/exclude-from-external-load-balancers label on control-plane nodes with dhctl for new clusters.
Add migration for exists clusters.

In kubeadm v1.21 label node.kubernetes.io/exclude-from-external-load-balancers is added to all control-plane nodes. Deckhouse clusters have this label only on the first control-plane node that was bootstrapped by dhctl.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Add node.kubernetes.io/exclude-from-external-load-balancers label on control-plane nodes
impact_level: high
impact: IngressNginxController Pods with the LoadBalancer.* inlets on master nodes will be excluded from load balancing
```

### DIFF
--- a/modules/040-node-manager/hooks/add_control_plane_role_to_master_ng.go
+++ b/modules/040-node-manager/hooks/add_control_plane_role_to_master_ng.go
@@ -14,8 +14,9 @@
 
 package hooks
 
-// TODO remove after 1.32 release
-// add control-plane role for all master nodes over master node group
+// TODO remove after 1.33 release
+// add control-plane role and "node.kubernetes.io/exclude-from-external-load-balancers" label
+// for all master nodes over master node group
 // At current moment, first bootstrapped master get 'control-plane' role,
 // but other master nodes don't get this role, because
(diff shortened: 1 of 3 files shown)


deckhouse-BOaTswain:

modules/040-node-manager/hooks/add_control_plane_role_to_master_ng.go
```diff
@@ -39,7 +43,8 @@ func addControlPlaneRoleToMasterNodeGroup(input *go_hook.HookInput) error {
 		"spec": map[string]interface{}{
 			"nodeTemplate": map[string]interface{}{
 				"labels": map[string]interface{}{
-					"node-role.kubernetes.io/control-plane": "",
+					controlPlaneRoleLabel:    "",
```
name212, line 46: i do not remove adding control plane role label in migration.

nabokihms (approved):

Issue #909, opened 2022-02-16
[snapshot-controller] New module


kvaps:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

After merging https://github.com/deckhouse/deckhouse/pull/831 all our csi-modules started supporting making snapshots. But this opportunity will not work without additional [snapshot-controller](https://github.com/kubernetes-csi/external-snapshotter) component and it's CRDs.

### Proposed Solution

We need to deploy one common snapshot-controller for whole cluster
https://github.com/kubernetes-csi/external-snapshotter

PR #1068, opened 2022-03-17
[snapshot-controller] New module

kvaps:
## Description

This module enables snapshots functionality in Kubernetes cluster

fixes https://github.com/deckhouse/deckhouse/issues/909
linked issue: https://github.com/deckhouse/deckhouse/pull/1121

## Changelog entries

```changes
section: snapshot-controller
type: feature
summary: New module: snapshot-controller
impact_level: high
impact: This module enables snapshots functionality in Kubernetes cluster 
```


modules/045-snapshot-controller/docs/USAGE.md
```diff
@@ -0,0 +1,90 @@
...
+To use snapshots, you first need to create a `VolumeSnapshotClass`:
+
+```yaml
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotClass
+metadata:
+  name: my-first-linstor-snapshot-class
+driver: linstor.csi.linbit.com
+deletionPolicy: Delete
+```
```
kvaps, line 16: It's up to consideration. I think we can create `VolumeSnapshotClass` automatically for every supported driver

zuzzas: We should do just that.

modules/045-snapshot-controller/hooks/generate_webhook_certs.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func applyCertsFilter(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
+	secret := &v1.Secret{}
+	err := sdk.FromUnstructured(obj, secret)
+	if err != nil {
+		return nil, fmt.Errorf("cannot convert selfsigned certificate secret to secret: %v", err)
+	}
+
+	return CertSnapshot{
```
zuzzas, line 52: Prefer a pointer here and later on when type asserting. Lots of stuff to copy lots of time for no reason at all.

modules/045-snapshot-controller/hooks/generate_webhook_certs.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	if len(cert.CA) == 0 {
+		var err error
+		caCert, err = certificate.GenerateCA(input.LogEntry, "snapshot-validation-webhook-ca")
+		if err != nil {
+			return fmt.Errorf("cannot generate selfsigned ca: %v", err)
+		}
+
+		// serviceFQDN := fmt.Sprintf(
```
zuzzas, line 99: Do these comments add anything of value?

modules/045-snapshot-controller/hooks/generate_webhook_certs.go
```diff
@@ -0,0 +1,124 @@
...
+	var cert certificate.Certificate
+	var s CertSnapshot
+
+	snaps := input.Snapshots["certs"]
+	for _, snap := range snaps {
+		s = snap.(CertSnapshot)
+		cert = s.Cert
+	}
+
+	if len(cert.CA) == 0 {
```
zuzzas, line 92: How can a single CA field be empty, while others... are not? I am not sure that I understand the conditional logic here.

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+    name: d8-snapshot-controller
+  updatePolicy:
+    updateMode: "Auto"
+{{- end }}
+---
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  annotations:
+    helm.sh/hook: post-upgrade, post-install
```
zuzzas, line 22: What?

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
```diff
@@ -0,0 +1,72 @@
+{{- if (.Values.global.enabledModules | has "vertical-pod-autoscaler-crd") }}
+---
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  name: d8-snapshot-controller
+  namespace: kube-system
+  {{- include "helm_lib_module_labels" (list . (dict "app" "d8-snapshot-controller" "workload-resource-policy.deckhouse.io" "master")) | nindent 2 }}
```
zuzzas, line 8: Perhaps, a system node scheduling and workload-policy will do better?
You haven't specified any node scheduling constraints later on? Why do you specify `workload-resource-policy: master`?

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+  namespace: kube-system
+  {{- include "helm_lib_module_labels" (list . (dict "app" "d8-snapshot-controller" )) | nindent 2 }}
+spec:
+  minAvailable: {{ include "helm_lib_is_ha_to_value" (list . 1 0) }}
+  selector:
+    matchLabels:
+      app: d8-snapshot-controller
+---
+apiVersion: apps/v1
+kind: Deployment
```
zuzzas, line 34: Can't see anything related to monitoring. Does the controller and webhook export any metrics at all?

zuzzas (changes requested):

modules/045-snapshot-controller/hooks/generate_webhook_certs.go
kvaps: Possible in the future we would add FQDN name with `cluster.local` to the certificate, this is just the implementation. Should I remove it?

modules/045-snapshot-controller/hooks/generate_webhook_certs.go
kvaps: Here we're just checking if provided certificate is not zero-initiated but has some data.

modules/045-snapshot-controller/hooks/generate_webhook_certs.go
zuzzas: Here is my list in this situation:

- [ ] Commented code block is sufficiently small. So that it's wont compromise the comprehension of the surrounding code that actually does the work.
- [ ] You are 100% sure that this code will become useful at some defined point in the future.
- [ ] The code block has a documenting comment above, that clearly specifies conditions after which you can uncomment this piece of code.

If you believe that all points hold true, leave the commented code.

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
kvaps: This was introduced in all modules by https://github.com/deckhouse/deckhouse/commit/6fb39beb1e45a2668f98cbd87afb75c62affff53.

I'm assuming the PDB violates the deploying process in some cases when it is processed in the standard way.

apolovov: I can't remember why :(.

modules/045-snapshot-controller/crds/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
```diff
@@ -0,0 +1,119 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.0
```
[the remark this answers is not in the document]
zuzzas: Consider applying the [Occam's razor](https://en.wikipedia.org/wiki/Occam%27s_razor).

Kubernetes has no use for these annotations, deckhouse has no use for there annotations. If a deckhouse developer gazes upon this labels at some point in the future, their head would be filled with all manner of uncertainties.

Are you sure that these annotation serve actual purpose in the `deckhouse`?

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
zuzzas: Ok, leave it be then.

zuzzas (approved):

kvaps:
@zuzzas plz, check volumesnapshotclasses names, do you agree on them?

https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/ee/modules/030-cloud-provider-openstack/templates/csi/volume-snapshot-class.yaml#L8

https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/ee/modules/030-cloud-provider-vsphere/templates/csi/volume-snapshot-class.yaml#L8

https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/030-cloud-provider-aws/templates/csi/volume-snapshot-class.yaml#L8

https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/030-cloud-provider-azure/templates/csi/volume-snapshot-class.yaml#L8

https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/030-cloud-provider-gcp/templates/csi/volume-snapshot-class.yaml#L8

https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/030-cloud-provider-yandex/templates/csi/volume-snapshot-class.yaml#L7

https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/031-linstor/templates/csi/volume-snapshot-class.yaml#L8

kvaps:
PR rebased to solve conflicts

zuzzas:
> @zuzzas plz check volumesnapshotclasses names, do you agree on them? They are up to consideration, eg. we can use `cinder` instead of `openstack`, `ec2` instead of `aws` and etc.
> 
> https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/ee/modules/030-cloud-provider-openstack/templates/csi/volume-snapshot-class.yaml#L8
> 
> https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/ee/modules/030-cloud-provider-vsphere/templates/csi/volume-snapshot-class.yaml#L8
> 
> https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/030-cloud-provider-aws/templates/csi/volume-snapshot-class.yaml#L8
> 
> https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/030-cloud-provider-azure/templates/csi/volume-snapshot-class.yaml#L8
> 
> https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/030-cloud-provider-gcp/templates/csi/volume-snapshot-class.yaml#L8
> 
> https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/030-cloud-provider-yandex/templates/csi/volume-snapshot-class.yaml#L7
> 
> https://github.com/deckhouse/deckhouse/blob/7041fc41e31c35ae74144cdd6c4f357a1b2777f3/modules/031-linstor/templates/csi/volume-snapshot-class.yaml#L8

`openstack` -> `cinder`
`aws` -> `ebs`
`azure` -> `disks`
`yandex` -> `disks`
`gcp` -> `persistend-disks`

This works well because it conforms with each cloud's internal naming and allows to differentiate between possible future snapshots for file or object storage.

kvaps:
> `openstack` -> `cinder`
> `aws` -> `ebs`
> `azure` -> `disks`
> `yandex` -> `disks`
> `gcp` -> `persistent-disks`
> `vsphere` -> `block`


Some of names are too general. Don't you afraid of name collisions in hybrid clusters?

zuzzas:
Agreed. Let's prefix everything (even unique ones) with the cloud's name. `openstack-cinder`, for example. That should alleviate all problems.

kvaps:
I just checked the default examples provided by the cloud providers, they are usually have nothing meaningful:

```
openstack-cinder: standard-csi
vmware-vsphere: example-vanilla-rwo-filesystem-sc
aws-ebs: test-snapclass
azure-disk: csi-azuredisk-vsc
gcp-pd: my-snapshotclass
linstor: my-first-linstor-snapshot-class
```

None of cloud providers except azure does not provide any options (yet?) for volumeSnapshotClassess, so usually they are means just to specify the csi driver name.

> Let's prefix everything (even unique ones) with the cloud's name

OK so:

```
openstack-cinder
vmware-vsphere
aws-ebs
azure-disk
yandex-disk
gcp-pd
linstor
```

Alternatively we can also add `csi-` prefix:

```
csi-openstack-cinder
csi-vmware-vsphere
csi-aws-ebs
csi-azure-disk
csi-yandex-disk
csi-gcp-pd
csi-linstor
```


image

zuzzas:
And how would that fix the issue of possible concurrent snapshot classes in the same cloud? There are three possible storage classes:
1. block
2. file
3. object

kvaps:
The VolumeSnapshotClass for block devices (RWO) and filesystem (RWX) will not be different as both of them are provisioned by the same csi-driver.

The configuration parameters are possible only for [azure driver](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/73d4faf47c05ffd8561c45fce0f098901ca9c3e1/deploy/example/snapshot/storageclass-azuredisk-snapshot.yaml#L8-L9):

```yaml
parameters:
  incremental: "true"  # available values: "true", "false" ("true" by default for Azure Public Cloud, and "false" by default for Azure Stack Cloud)
```

AFAIK, S3 is not provisioned by the standard PersistentVolumes abstractions at all

zuzzas:
And what if they will be? What if they add new configuration parameters? Your current proposition seems extremely non future-proofed.

kvaps:
I think we would need that in the future, we can add new snashotvolumeclasses later, and left the old ones without changes for the compatibility with existing snapshots.

zuzzas:
Leave it as you wish then. This is a pointless discussion about a pointless thing.

modules/045-snapshot-controller/crds/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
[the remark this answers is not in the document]
kvaps: Removing these annotations makes snapshot-controller not working:

```
ModuleRun failed in phase 'Startup'. Requeue task to retry after delay. Failed count is 28. Error: 3 errors occurred:
	* Create object: apiextensions.k8s.io/v1/CustomResourceDefinition//volumesnapshotclasses.snapshot.storage.k8s.io: CustomResourceDefinition.apiextensions.k8s.io "volumesnapshotclasses.snapshot.storage.k8s.io" is invalid: metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111
	* Create object: apiextensions.k8s.io/v1/CustomResourceDefinition//volumesnapshotcontents.snapshot.storage.k8s.io: CustomResourceDefinition.apiextensions.k8s.io "volumesnapshotcontents.snapshot.storage.k8s.io" is invalid: metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111
	* Create object: apiextensions.k8s.io/v1/CustomResourceDefinition//volumesnapshots.snapshot.storage.k8s.io: CustomResourceDefinition.apiextensions.k8s.io "volumesnapshots.snapshot.storage.k8s.io" is invalid: metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111
```

Take a look at this issue: https://github.com/kubernetes/enhancements/pull/1111

I would not modify them with no reason, both returned back in 5aec79cd980c71f7b317bd2cfdef650cdd70862a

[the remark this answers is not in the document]
zuzzas: Wow. That's new. Thanks!

zuzzas (approved):

modules/045-snapshot-controller/enabled
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source /deckhouse/shell_lib.sh
+
+function __main__() {
+  enabled::disable_module_in_kubernetes_versions_less_than 1.20.0
```
nabokihms, line 20: It seems like we do not need snapshots if a cluster is not bootstrapped.

modules/045-snapshot-controller/templates/snapshot-validation-webhook/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+    matchLabels:
+      app: d8-snapshot-validation-webhook
+  template:
+    metadata:
+      labels:
+        app: d8-snapshot-validation-webhook
+    spec:
+      imagePullSecrets:
+        - name: deckhouse-registry
+      {{- include "helm_lib_priority_class" (tuple . "cluster-low") | nindent 6 }}
```
nabokihms, line 51: Everything which affects API interactions should have the cluster high priority class?

modules/045-snapshot-controller/templates/snapshot-validation-webhook/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+      app: d8-snapshot-validation-webhook
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: d8-snapshot-validation-webhook
+  namespace: kube-system
+  {{- include "helm_lib_module_labels" (list . (dict "app" "d8-snapshot-validation-webhook" )) | nindent 2 }}
+spec:
+  replicas: {{ include "helm_lib_is_ha_to_value" (list . 2 1) }}
```
nabokihms, line 40: Should use a different helper to run these pods on each master node.

modules/045-snapshot-controller/templates/snapshot-validation-webhook/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+  minAvailable: {{ include "helm_lib_is_ha_to_value" (list . 1 0) }}
+  selector:
+    matchLabels:
+      app: d8-snapshot-validation-webhook
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: d8-snapshot-validation-webhook
+  namespace: kube-system
```
nabokihms, line 37: Do not like deploying this to the kube-system namespaces. Is there something which presents us to deploy it to a dedicated namespace?

modules/045-snapshot-controller/templates/snapshot-validation-webhook/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+          requests:
+            {{- include "helm_lib_module_ephemeral_storage_only_logs" . | nindent 12 }}
+        volumeMounts:
+          - name: snapshot-validation-webhook-certs
+            mountPath: /etc/snapshot-validation-webhook/certs
+            readOnly: true
+      volumes:
+        - name: snapshot-validation-webhook-certs
+          secret:
+            secretName: d8-snapshot-validation-webhook-certs
```
nabokihms, line 77: Do we need to add the checksum annotation to reload pods on every secret change?

modules/045-snapshot-controller/templates/snapshot-validation-webhook/webhook.yaml
```diff
@@ -0,0 +1,24 @@
...
+    service:
+      namespace: "kube-system"
+      name: "d8-snapshot-validation-webhook"
+      path: "/volumesnapshot"
+      port: 4443
+    caBundle: {{ .Values.snapshotController.internal.webhookCert.ca | b64enc }}
+  admissionReviewVersions: ["v1", "v1beta1"]
+  sideEffects: None
+  failurePolicy: Fail
+  timeoutSeconds: 2
```
nabokihms, line 24: Let's make it something like 10 seconds.

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+      app: d8-snapshot-controller
+  template:
+    metadata:
+      labels:
+        app: d8-snapshot-controller
+    spec:
+      imagePullSecrets:
+        - name: deckhouse-registry
+      {{- include "helm_lib_priority_class" (tuple . "cluster-low") | nindent 6 }}
+      {{- include "helm_lib_node_selector" (tuple . "master") | nindent 6 }}
```
nabokihms, line 53: Why do we need to schedule it on master nodes?

modules/045-snapshot-controller/openapi/config-values.yaml
```diff
@@ -0,0 +1,2 @@
+type: object
+properties: {}
```
nabokihms, line 2: There should be basic configuration options like the HA mode.

nabokihms (changes requested):

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
kvaps: Because it is the same controller as any others.
Eg. if we deploy csi-controller on masters, why is snapshot-controller should be exception?

modules/045-snapshot-controller/templates/snapshot-validation-webhook/deployment.yaml
kvaps: Actually not, I was inspired by other cluster-wide services, eg: `d8-kube-dns` `d8-kube-proxy` and `d8-control-plane-manager`. `snapshot-controller` is extending Kubernetes API by adding general APIs for all CSI drivers in cluster. Should I move it?

nabokihms: We also have extensions like bashibile-apiserver, and authorization webhook.

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
nabokihms: We schedule the CSI controller on master nodes because we need to order disks before the cluster becomes bootstrapped. I doubt that we need to make snapshots of these disks at the same time.

modules/045-snapshot-controller/templates/snapshot-validation-webhook/deployment.yaml
kvaps: priority class switched to `system-cluster-critical`

modules/045-snapshot-controller/templates/snapshot-validation-webhook/deployment.yaml
kvaps: Which one do you mean?

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
kvaps: Ok, should I remove taint as well?

modules/045-snapshot-controller/templates/snapshot-validation-webhook/deployment.yaml
nabokihms: `  {{- include "helm_lib_deployment_on_master_strategy_and_replicas_for_ha" . | nindent 2 }}`

modules/045-snapshot-controller/templates/snapshot-controller/deployment.yaml
nabokihms: There are should be tolerations and a node selector for system nodes.

kvaps: ok, got it!

nabokihms (approved):

PR #1086, opened 2022-03-19
[tests] feat: support x-required-for-helm, add x-test-focus

### LINKED ISSUE #186: Openapi case tests don't work with x-required-for-helm
### Preflight Checklist

### Version

v.1.24.17

### Expected Behavior

I want to write negative openapi case test for testing skipped properties which contains in `x-required-for-helm` object spec.
These tests should pass.

### Actual Behavior

Tests fail

### Steps To Reproduce

1. Values file:
```
x-extend:
  schema: config-values.yaml
type: object
x-required-for-helm:
- internal
properties:
  internal:
    type: object
    additionalProperties: false
    default: {}
    x-required-for-helm:
    - adapterPem
    - adapterKey
    - adapterCA
    - prometheusAPIClientTLS
    properties:
      adapterPem:
        x-examples: ["YjY0ZW5jX3N0cmluZwo="]
        type: string

      adapterCA:
        x-examples: ["YjY0ZW5jX3N0cmluZwo="]
        type: string

      adapterKey:
        x-examples: [ "YjY0ZW5jX3N0cmluZwo=" ]
        type: string

      prometheusAPIClientTLS:
        default: {}
        additionalProperties: false
        x-required-for-helm:
          - certificate
          - key
        type: object
        properties:
          certificate:
            x-examples: [ "YjY0ZW5jX3N0cmluZwo=" ]
            type: string

          key:
            x-examples: [ "YjY0ZW5jX3N0cmluZwo=" ]
            type: string

      customMetrics:
        type: object
        default: {}
        additionalProperties: false
        properties:
          daemonset:
            type: object
            default: { }
            x-examples: [ {"name":{"cluster":"query", "namespaced": {"ns", "query"}}} ]
            additionalProperties: true

          ## skip similar objects (for pods, service...) here
```

2. Openapi cases tests file:
```
positive:
  values: []

negative:
  values:
  - internal:
      prometheusAPIClientTLS:
        certificate: cert
        key: key
      customMetrics:
        daemonset: { }
	# skip similar objects
```
3. Add files (or replaces) in some module
4. Run openapi test cases

### Additional Information

_No response_

### Logs

_No response_

diafour:
## Description

- Use "helmValues" in positive or negative cases if values.yaml to test effect of x-required-from.
- Use "x-test-focus: true" to run only specific cases.

## TODO

1. Delete `modules/000-prom-test` before merging.
1. Documentation validation failed because of this example module.

Fix #186

## Changelog entries

```changes
section: tests
type: feature
summary: Support x-required-for-helm, add x-test-focus
impact_level: low
```

### DIFF
--- a/modules/301-prometheus-metrics-adapter/openapi/openapi-case-tests.yaml
+++ b/modules/301-prometheus-metrics-adapter/openapi/openapi-case-tests.yaml
@@ -73,45 +73,32 @@ negative:
 # next case don't work because deckhouse testes not support x-required-for-helm
 # need to fix it and uncomment cases after fix
 
-##  # does not have internal pma tls
-#  - internal:
-#      prometheusAPIClientTLS:
-#        certificate: cert
-#        key: key
-#      customMetrics:
-#        daemonset: { }
-#        deployment: { }
-#        ingress: { }
-#        namespace: { }
-#        pod: { }
-#        service: { }
-#        statefulset: { }
-#
-#  # does not have prom client tls object
-#  - internal:
-#      adapterPem: cert
-#      adapterCA: cert
-#      adapterKey: key
-#      customMetrics:
-#        daemonset: { }
-#        deployment: { }
-#        ingress: { }
-#        namespace: { }
-#        pod: { }
-#        service: { }
-#        statefulset: { }
-#
-#  # does not have prom client tls
-#  - internal:
-#      adapterPem: cert
-#      adapterCA: cert
-#      adapterKey: key
-#      prometheusAPIClientTLS: { }
-#      customMetrics:
-#        daemonset: { }
-#        deployment: { }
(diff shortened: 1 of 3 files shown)


modules/000-prom-test/openapi/openapi-case-tests.yaml
```diff
@@ -0,0 +1,16 @@
+positive:
+  values: []
+
+negative:
+  helmValues:
```
nabokihms, line 5: You should delete this test example and add tests to the actual module.

nabokihms (changes requested):

modules/000-prom-test/openapi/openapi-case-tests.yaml
diafour: Done!

PR #1104, opened 2022-03-21
[candi] Automatically discover zone for volumes in OpenStack

konstantin-axenov:
## Description

Add terraform module volume-zone that helps to discover correct zone for OpenStack volumes. This module checks if provided compute zone exists for volumes. If it doesn't exist than no availability zone is set, and we assume that the default zone will be used.

Some private OpenStack clouds have different availability zones for volumes and computes.

## Changelog entries

```changes
section: candi
type: fix
summary: Automatically discover zone for volumes in OpenStack
```


z9r5:

PR #1116, opened 2022-03-22
[Changelog] fix quoting

shvgn:
## Description
Fix quoting backticks in changelog files

Fix quoting backticks in changelog files

## Changelog entries

```changes
section: ci
type: fix 
summary: Fix quoting backticks in changelog files
```


z9r5:

PR #1122, opened 2022-03-22
[deckhouse] use object instead of string in changelog

yalosev:

## Description
Fix changelog struct

Changelog struct is map of objects not strings

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Fix DeckhouseRelease CRD
impact_level: low 
```


z9r5:

PR #1128, opened 2022-03-22
[user-authn] Fix DexAuthenticator object template for https mode OnlyInURI

nabokihms:
Signed-off-by: m.nabokikh <maksim.nabokikh@flant.com>

## Description
* Add a clause to set authenticator ingress certificate secret name

## Why do we need it, and what problem does it solve?
Urgent fix

## Changelog entries

```changes
section: user-authn
type: fix
summary: Do not set ingress TLS certificate secret name if HTTPS mode is the OnlyInURI
impact_level: high
impact: Fix deploy of DexAuthenticators to clusters with the OnlyInURI HTTPS mode
```

PR #1134, opened 2022-03-22
[user-authn] UnlyInURI support in authenticator.yaml

apolovov:
## Description
UnlyInURI https mode support for standart authenticator.yaml template.

There wasn't a check for tls switched off.

## Changelog entries

```changes
section: user-authn
type: fix
summary: UnlyInURI https mode support for standart authenticator.yaml template.
impact_level: low
```


nabokihms:
Looks like a duplicate of https://github.com/deckhouse/deckhouse/pull/1128

PR #1141, opened 2022-03-23
[user-authn] Move authenticator migration to the user-authn-crd module

nabokihms:

It turned out that on startup hook it not actually executes on deckhouse startup

## Changelog entries

```changes
section: user-authn
type: fix
summary: Execute authenticator migration before all modules
impact_level: low
```


konstantin-axenov (approved):

z9r5:

PR #1145, opened 2022-03-23
[deckhouse] disable release event

yalosev:

## Description
Disable reaction on release events

We run hook every 15 seconds. I don't think we need this reaction which can lead to some artifacts

## Changelog entries

```changes
section: deckhouse
type: fix
summary: disable hook reaction on DeckhouseRelease events
impact_level: low
```


nabokihms (changes requested):
There are failed tests. It looks like a problem after setting execution on events to false.

z9r5:

PR #1154, opened 2022-03-24
[prometheus-metrics-adapter] remove namespace enforcing

yalosev:

## Description
Fix prometheus metrics adapter external metrics

This behavior was broken at November by applying 0.9.1 PMA version.
By default from version 0.9.0 PMA enforces to pass namespace label to a query.
By setting `namespaced:  false` we allow PMA to accept external metrics from all namespaces.
Then, if the metrics needed by your HPA need to be selected with a specific namespace label value, you should set it directly in a HPA's label selector.
Example: 
```
  - type: External
    external:
      metric:
        name: rmq_queue_length
        selector:
          labelSelector:
            queue: my-queue
            namespace: example
```

## Changelog entries

```changes
type: fix
summary:  Restore HPA external metrics behavior
impact_level: high
impact: Disable enforced namespace passing to a query. If your metric need to be selected with a specific namespace label value, you should set it directly in an HPA's label selector
```

### DIFF
--- a/modules/301-prometheus-metrics-adapter/templates/config-map.yaml
+++ b/modules/301-prometheus-metrics-adapter/templates/config-map.yaml
@@ -9,6 +9,8 @@ data:
   config.yaml: |
     externalRules:
     - seriesQuery: '{__name__=~"kube_adapter_metric_.+"}'
+      resources:
+        namespaced: false
       name:
         matches: "^kube_adapter_metric_(.+)$"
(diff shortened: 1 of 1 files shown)


shvgn:
There no impact for user apps. We should remove `impact_level` from the change description.

nabokihms (approved):

z9r5:

PR #1156, opened 2022-03-24
[cloud-provider-vsphere] Fix handle of compatibilityFlag in Deckhouse config

konstantin-axenov:
## Description

Get correct value from config

Compatibility mode doesn't work

## Changelog entries

```changes
section: cloud-provider-vsphere
type: fix
summary: Fix handle of compatibilityFlag in Deckhouse config
impact_level: medium
```


z9r5:

PR #1168, opened 2022-03-25
[deckhouse] Fix cleanup hook

yalosev:
## Description
Fix hook which cleanup old Releases and also run it only at startup

Hook as invalid logic and sometimes in a race DeckhouseRelease can become Outdated

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Fix cleanup DeckhouseReleases hook
impact_level: low
```


z9r5:

PR #1189, opened 2022-03-29
[candi] Set reasonable image GC thresholds

konstantin-axenov:
## Description

Parameter `evictionSoftThresholdImagefsInodesFree` is set to 10% or even smaller amount in gigabytes if node has large volume capacity. So it seems logical to run soft image GC first when there is 20% free space left and hard image GC when it is only 15% left. If GC can't free disk space, and it continues to reduce, then evictions start.

Low thresholds lead to constant launch of GC on nodes with small image fs size. And it usually leads to great amount of error events because disk space can't be reclaimed.

## Changelog entries

```changes
section: candi
type: fix
summary: Set reasonable image GC thresholds
impact_level: low
impact: Kubelet restart
```


nabokihms (approved):

z9r5:

PR #1192, opened 2022-03-29
[candi] Fix update password hash for node user

name212:
## Description

Fix update password hash for node user.

If hash exists for another user, we does not update hash 

## Changelog entries

```changes
section: candi
type: fix
summary: Fix update password hash for node user.
```

### DIFF
--- a/candi/bashible/common-steps/node-group/090_node_users.sh.tpl
+++ b/candi/bashible/common-steps/node-group/090_node_users.sh.tpl
@@ -31,32 +31,13 @@ function modify_user() {
 
   usermod -G "$extra_groups" "$user_name"
 
-  update_password=''
-  # with grep we decrease checks count
-  # why we can get multiple line with grep?
-  # for example, we have users 'test' and 'testuser'
-  # for 'test' user we get 2 lines and we should iterate through lines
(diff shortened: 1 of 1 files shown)


candi/bashible/common-steps/node-group/090_node_users.sh.tpl
```diff
@@ -31,8 +31,13 @@ function modify_user() {
 
   usermod -G "$extra_groups" "$user_name"
 
-  if ! grep -q -F "$password_hash" /etc/shadow; then
-      usermod -p "$password_hash" "$user_name"
+  local shadow_line="$(getent shadow "$user_name")"
```
RomanenkoDenys, line 34: local shadow_line="$(getent shadow "$user_name" | awk -F ":" '{print $2}')"
тогда 36-37 не нужно

RomanenkoDenys (approved):

candi/bashible/common-steps/node-group/090_node_users.sh.tpl
name212: done

z9r5:

PR #1205, opened 2022-03-29
[candi] Fix docker stuck containers cleaner pattern check

konstantin-axenov:
## Description

Ignore signal check because it can be 3 or 15 or some other.

This container was ignored but should have been considered stuck.
```
Mar 29 19:43:59 kube-master-1 dockerd[1848]: time="2022-03-29T19:43:59.088143054Z" level=info msg="Container 43e187a13c0a6d5e6115a578c7b853a0f883dc0b880d4b91753e4c9eaeec08e8 failed to exit within 30 seconds of signal 3 - using the force"
```

## Changelog entries

```changes
section: candi
type: fix
summary: Fix docker stuck containers cleaner pattern check
impact_level: low
```


z9r5:

Issue #1188, opened 2022-03-29
[Prometheus] The amount of metrics in Grafana exceeds the default sample limit of 1000


shvgn:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.31

### Expected Behavior

Prometheus should be able to scrape all Grafana metrics

### Actual Behavior

Prometheus scrapes at most 1000 metrics as defined in default sample limit. Grafana grows metrics cardinality per handler (from plugin or datasource) and per response status, that can add dozens of new timeseries with time.

```
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="0.005"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="0.01"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="0.025"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="0.05"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="0.1"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="0.25"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="0.5"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="1"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="2.5"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="5"}
grafana_http_request_duration_seconds_bucket{handler="/public/plugins/:pluginId/*",method="GET",status_code="304",le="+Inf"}
[...]

PR #1215, opened 2022-03-30
[Prometheus] Add bigger sample limit to Grafana service

shvgn:
## Description

Add bigger sample limit to Grafana service

The number of Grafana metrics grows with time and can exceed the default limit of 1000.

## Changelog entries

```changes
section: prometheus
type: fix 
summary: Set Grafana sample limit to 5000
```

### DIFF
--- a/modules/300-prometheus/templates/grafana/service.yaml
+++ b/modules/300-prometheus/templates/grafana/service.yaml
@@ -8,6 +8,7 @@ metadata:
   annotations:
     prometheus.deckhouse.io/port: "8443"
     prometheus.deckhouse.io/tls: "true"
+    prometheus.deckhouse.io/sample-limit: "5000"
 spec:
   type: ClusterIP
   clusterIP: None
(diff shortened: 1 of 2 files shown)


shvgn:
Fixes #1188

modules/300-prometheus/templates/grafana/service.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
   annotations:
     prometheus.deckhouse.io/port: "8443"
     prometheus.deckhouse.io/tls: "true"
+    prometheus.deckhouse.io/sample-limit: "5000"
```
nabokihms, line 11: This is not the only place. We have another monitor for Grafana right here
https://github.com/deckhouse/deckhouse/blob/main/ee/fe/modules/340-monitoring-applications/templates/service-monitors/grafana.yaml

shvgn: Changed it as well!

nabokihms (approved):

PR #1219, opened 2022-03-30
[ci] fix: avoid secret masking for tests_image_name output

diafour:
## Description

Use gzip+base64 encoding to pass tests image name containing non-secret secrets between jobs.

tests_image_name may contain rw registry host. Github runner masks this host in base64 encoded output tests fail.  This host is non-secret, so it should be passed.

See [ValueEncoders.cs](https://github.com/actions/runner/blob/main/src/Sdk/DTLogging/Logging/ValueEncoders.cs).

## Changelog entries

```changes
section: ci
type: fix
summary: Pass tests_image_name output with rw registry host
impact_level: low
```


z9r5:

PR #1247, opened 2022-04-01
[candi] added metrics endpoint to containerd

RomanenkoDenys:
## Description

Added metrics endpoint to containerd.
## Changelog entries

```changes
section: candi
type: feature
summary: "Added metrics endpoint to containerd."
impact_level: low
```


candi/bashible/common-steps/node-group/050_configure_and_start_containerd.sh.tpl
```diff
@@ -58,7 +60,7 @@ oom_score = 0
   gid = 0
   level = ""
 [metrics]
-  address = ""
+  address = "${discovered_node_ip}:1338"
```
zuzzas, line 63: Without any authentication whatsoever?

RomanenkoDenys: Containerd cannot support any authentication on metrics endpoint.

zuzzas: Of course. But what if we deploy an intermediate kube-rbac-proxy and bind containerd metrics to a locahost?

PR #1248, opened 2022-04-01
[cloud-provider-openstack] Remove default podNetworkMode from config-values

konstantin-axenov:
## Description

Value of `podNetworkMode` must be discovered from cloud-discovery-data and can be overwritten from Deckhouse config.

Default value broke automatic discovery that worked previously.

## Changelog entries

```changes
type: fix
summary: Remove default podNetworkMode from config-values
impact_level: low
```


nabokihms (approved):

yalosev (approved):

z9r5:
Cherry-picked into release-1.31.

PR #1254, opened 2022-04-04
disable double execution for cert hooks

yalosev:

Disable hook execution on sync.


zuzzas:
@yalosev tests are failing.

Issue #314, opened 2021-11-23
[Upmeter] Web pages upmeter.* and status.* load slowly


shvgn:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.26

### Expected Behavior

They should load in seconds

### Actual Behavior

Sometimes it takes ~20s to load

PR #1257, opened 2022-04-04
[upmeter] fix slowing response of probe list

shvgn:
## Description

Pick groups and probes from runtime for the initial response for UIs

- Fixes the slow response on web UI's
- Cleans code for clarity

## Changelog entries

```changes
section: upmeter
type: fix
summary: Fixed slow data loading in
```


modules/500-upmeter/images/upmeter/pkg/constant/constant.go
```diff
@@ -0,0 +1,7 @@
+package constant
+
+// TODO move to config
```
shvgn, line 3: move User-Agent values to config

shvgn: done

modules/500-upmeter/images/upmeter/pkg/server/api/public_status.go
```diff
@@ -122,16 +119,16 @@ func getStatusSummary(dbCtx *dbcontext.DbContext, monitor *crd.DowntimeMonitor)
 			muteDowntimeTypes: muteTypes,
 		}
 
-		resp, err := getStatus(dbCtx, monitor, filter)
+		resp, err := getStatus(h.DbCtx, h.DowntimeMonitor, filter)
 		if err != nil {
 			log.Errorf("cannot calculate status for group %s: %v", group, err)
-			continue
+			return groupStatuses, StatusOutage, err
```
diafour, line 125: Just a note: it returns partial groupStatuses, may be `nil` is better?

diafour (approved):

modules/500-upmeter/images/upmeter/pkg/server/api/public_status.go
shvgn, line 125:
```suggestion
			return nil, StatusOutage, err
```

modules/500-upmeter/images/upmeter/pkg/server/api/public_status.go
```diff
@@ -122,16 +119,16 @@ func getStatusSummary(dbCtx *dbcontext.DbContext, monitor *crd.DowntimeMonitor)
...
 		}
 
 		summary, err := pickSummary(ref, resp.Statuses)
 		if err != nil {
 			log.Errorf("cannot parse status for group %s: %v", group, err)
-			continue
+			return groupStatuses, StatusOutage, err
```
shvgn, line 131:
```suggestion
			return nil, StatusOutage, err
```

modules/500-upmeter/images/upmeter/pkg/server/api/public_status.go
shvgn: Returning nil

shvgn:
Closes #314

nabokihms (approved):

PR #1259, opened 2022-04-04
[prometheus-metrics-adapter] Restore config file for proxy

yalosev:

## Description
Restore VolumeMount for prometheus-metrics-adapter proxy. Also check file existence on startup and logging fatal if file is absent.

VolumeMount was removed during the last update. Without this file custom-metrics-handler panics on requests. But it was hidden inside logs. 

## Changelog entries

```changes
type: fix
summary: Fix custom metrics workability
```


konstantin-axenov (approved):

z9r5:
Cherry-picked into release-1.31.

Issue #1258, opened 2022-04-04
Unclear relation between enableMultiTenancy and allowAccessToSystemNamespaces


diafour:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.30

### Expected Behavior

- Relation between `enableMultiTenancy` and `allowAccessToSystemNamespaces` is easy to understand.
- I can predict consequences of 4 possible combinations after reading documentation.

### Actual Behavior

We got several places in documentation that describe these parameters:

1. [enableMultiTenancy](https://deckhouse.io/en/documentation/v1/modules/140-user-authz/configuration.html#parameters-enablemultitenancy) - no mentions of allowAccessToSystemNamespaces.
2. [allowAccessToSystemNamespaces](https://deckhouse.io/en/documentation/v1/modules/140-user-authz/cr.html#clusterauthorizationrule-v1-spec-allowaccesstosystemnamespaces) - description may be interpreted as "`allowAccessToSystemNamespaces: true` will be ignored when `enableMultiTenancy: false`"
3. [multi-tenancy mode](https://deckhouse.io/en/documentation/v1/modules/140-user-authz/usage.html#configuring-kube-apiserver-for-multi-tenancy-mode) - no mentions of allowAccessToSystemNamespaces.

Issue #231, opened 2021-10-14
[user-authz] ClusterAuthorizationRules validation webhook


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

Deckhouse validates ClusterAuthorizationRules within helm templates. In that case, users can save invalid resources to the cluster and receive no feedback about their invalidity.

### Proposed Solution

Use a webhook from the `deckhouse` module to check whether multitenancy is enabled and decline saving attempts of invalid resources.

PR #1260, opened 2022-04-04
[user-authz] add validation webhook for multitenancy feature

yalosev:

## Description
Add validation webhook which prevents CR creation if `limitNamespaces` or `allowAccessToSystemNamespaces` features are used without enabled multitenancy

As described in #1258, it could be hard to determine what is going wrong when you forgot to enable multitenancy

## Changelog entries

```changes
section: user-authz
type: feature
summary: Add validation for ClusterAuthorizationRule with `limitNamespaces` or `allowAccessToSystemNamespaces` features 
impact_level: low
```

### DIFF
--- a/ee/modules/140-user-authz/.build.yaml
+++ b/ee/modules/140-user-authz/.build.yaml
@@ -1,1 +1,2 @@
 templates/webhook
+webhooks/validating
--- a/ee/modules/140-user-authz/.build.yaml
+++ b/ee/modules/140-user-authz/.build.yaml
@@ -1,2 +1,2 @@
 templates/webhook
-webhooks/validating
+webhooks/validating/*
--- a/ee/modules/140-user-authz/.build.yaml
+++ b/ee/modules/140-user-authz/.build.yaml
@@ -1,2 +1,1 @@
 templates/webhook
-webhooks/validating/*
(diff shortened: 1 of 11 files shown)


nabokihms:
There was a reason why such a hook was not implemented earlier in Deckhouse. With the hook from this PR, it is impossible to prevent the following sequence of actions:
1. Enable multitenancy
2. Add cluster authorization rules with the limit namespaces option enabled
3. Disable multitenancy

We were not satisfied with any solution to this problem, so we decided to keep things as is until multitenancy refactoring.

nabokihms:
related https://github.com/deckhouse/deckhouse/issues/231

yalosev:
> There was a reason why such a hook was not implemented earlier in Deckhouse. With the hook from this PR, it is impossible to prevent the following sequence of actions:
> 
> 1. Enable multitenancy
> 2. Add cluster authorization rules with the limit namespaces option enabled
> 3. Disable multitenancy
> 
> We were not satisfied with any solution to this problem, so we decided to keep things as is until multitenancy refactoring.

This case is handled by [this code](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/140-user-authz/templates/webhook/configmap.yaml#L15). But it left error only in Deckhouse logs. I don't see any reason to avoid improvement of user experience and warn user on CR creation if multi tenancy is disabled

nabokihms (approved):

PR #1266, opened 2022-04-05
[cloud-provider-azure] Enable accelerated networking for new machine-controller-manager instances

vitaliy-sn:
## Description

- Accelerated networking is enabled by default for new machine-controller-manager instances.
  - If you want your existing nodes to have accelerated networking enabled, you have to [roll out update](https://deckhouse.io/en/documentation/v1/modules/040-node-manager/faq.html#how-do-i-redeploy-ephemeral-machines-in-the-cloud-with-a-new-configuration) them or just delete.
  - To disable set `.spec.acceleratedNetworking` to `false` in AzureInstanceClass.
- The master and static nodes are set to `acceleratedNetworking=false` by default to automatically add it to AzureClusterConfiguration, which will allow us to change the default value later without any problems.
  - To enable set `.spec.acceleratedNetworking` to `true` and do `dhctl converge`.

[Accelerated networking](https://azure.microsoft.com/fr-fr/blog/maximize-your-vm-s-performance-with-accelerated-networking-now-generally-available-for-both-windows-and-linux/) is a stable feature that helps to improve performance of vm in Azure cloud.

## Changelog entries

```changes
section: cloud-provider-azure
type: feature
summary: Enable accelerated networking for new machine-controller-manager instances.
```

### DIFF
--- a/candi/cloud-providers/azure/openapi/cluster_configuration.yaml
+++ b/candi/cloud-providers/azure/openapi/cluster_configuration.yaml
@@ -159,7 +159,7 @@ apiVersions:
                   type: string
               acceleratedNetworking:
                 type: boolean
-                x-doc-default: true
+                default: false
                 description: |
                   Accelerated Networking providing up to 30Gbps in networking throughput.
       nodeGroups:
--- a/candi/cloud-providers/azure/openapi/cluster_configuration.yaml
+++ b/candi/cloud-providers/azure/openapi/cluster_configuration.yaml
@@ -161,7 +161,7 @@ apiVersions:
                 type: boolean
                 default: false
                 description: |
-                  Accelerated Networking providing up to 30Gbps in networking throughput.
+                  Accelerated Networking provides up to 30Gbps in networking throughput.
       nodeGroups:
         description: |
           An array of additional NodeGroups for creating static nodes (e.g., for dedicated front nodes or gateways).
--- a/candi/cloud-providers/azure/openapi/cluster_configuration.yaml
+++ b/candi/cloud-providers/azure/openapi/cluster_configuration.yaml
@@ -157,6 +157,11 @@ apiVersions:
                 type: object
                 additionalProperties:
                   type: string
+              acceleratedNetworking:
+                type: boolean
+                x-doc-default: true
+                description: |
+                  Accelerated Networking providing up to 30Gbps in networking throughput.
       nodeGroups:
         description: |
           An array of additional NodeGroups for creating static nodes (e.g., for dedicated front nodes or gateways).
--- a/candi/cloud-providers/azure/openapi/instance_class.yaml
+++ b/candi/cloud-providers/azure/openapi/instance_class.yaml
@@ -74,6 +74,11 @@ spec:
                   additionalProperties:
                     type: string
                   x-kubernetes-preserve-unknown-fields: true
+                acceleratedNetworking:
+                  type: boolean
+                  default: true
+                  description: |
+                    Accelerated Networking providing up to 30Gbps in networking throughput.
     - name: v1
       served: true
       storage: false
--- a/candi/cloud-providers/azure/openapi/instance_class.yaml
+++ b/candi/cloud-providers/azure/openapi/instance_class.yaml
@@ -77,7 +77,7 @@ spec:
                 acceleratedNetworking:
                   type: boolean
                   description: |
-                    Accelerated Networking providing up to 30Gbps in networking throughput.
+                    Accelerated Networking provides up to 30Gbps in networking throughput.
                   x-doc-default: true
     - name: v1
       served: true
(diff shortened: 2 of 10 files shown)


nabokihms (commented):
From my point of view, it looks good, but I'd like someone to see it in addition.

modules/030-cloud-provider-azure/cloud-instance-manager/machine-class.yaml
```diff
@@ -43,3 +43,5 @@ spec:
             path: /home/azureuser/.ssh/authorized_keys
             keyData: {{ .Values.nodeManager.internal.cloudProvider.azure.sshPublicKey }}
     zone: {{ .zoneName }}
+    networkProfile:
+      acceleratedNetworking: {{ .nodeGroup.instanceClass.acceleratedNetworking }}
```
konstantin-axenov, line 47: `acceleratedNetworking` must be added to MachineClass checksum

candi/cloud-providers/azure/terraform-modules/master-node/variables.tf
```diff
@@ -40,19 +40,20 @@ variable "clusterUUID" {
...
+  enable_external_ip     = lookup(var.providerClusterConfiguration.masterNodeGroup.instanceClass, "enableExternalIP", false)
+  urn                    = split(":", var.providerClusterConfiguration.masterNodeGroup.instanceClass.urn)
+  image_publisher        = local.urn[0]
+  image_offer            = local.urn[1]
+  image_sku              = local.urn[2]
+  image_version          = local.urn[3]
+  actual_zones           = lookup(var.providerClusterConfiguration, "zones", null) != null ? tolist(setintersection(["1", "2", "3"], var.providerClusterConfiguration.zones)) : ["1", "2", "3"]
+  zones                  = lookup(var.providerClusterConfiguration.masterNodeGroup, "zones", null) != null ? tolist(setintersection(local.actual_zones, var.providerClusterConfiguration.masterNodeGroup["zones"])) : local.actual_zones
+  additional_tags        = merge(lookup(var.providerClusterConfiguration, "tags", {}), lookup(var.providerClusterConfiguration.masterNodeGroup.instanceClass, "additionalTags", {}))
+  accelerated_networking = lookup(var.providerClusterConfiguration.masterNodeGroup.instanceClass, "acceleratedNetworking", true)
```
konstantin-axenov, line 58: This is not good. We don't want to change defaults, so all masters nodes have to be converged. I suggest using the same scheme as for yandex platform migration. First, we set default to `false` for acceleratedNetworking and then on the next release we change it to `true`. The same is about static nodes.

konstantin-axenov (changes requested):

konstantin-axenov:
@vitaliy-sn Very bad description of pull request. It's not clear what is going on. We are changing default behavior, and this is important. You must write accurate information. I expected something like this:

> New machine-controller-manager machines will be ordered with accelerated networking enabled by default. If you want your existing machines to have accelerated networking enabled, you have to roll out update them or just delete. Control-plane nodes and static nodes won't be affected by these changes because blah-blah-blah. If you want master nodes to be updated, you have to blah-blah-blah.

modules/030-cloud-provider-azure/cloud-instance-manager/machine-class.yaml
vitaliy-sn: done

candi/cloud-providers/azure/terraform-modules/master-node/variables.tf
vitaliy-sn: The default is specified in the spec, left here just in case to avoid errors. I set a value similar to that specified in the spec, so as not to confuse.

candi/cloud-providers/azure/openapi/cluster_configuration.yaml
```diff
@@ -157,6 +157,11 @@ apiVersions:
                 type: object
                 additionalProperties:
                   type: string
+              acceleratedNetworking:
+                type: boolean
+                default: false
+                description: |
+                  Accelerated Networking providing up to 30Gbps in networking throughput.
```
konstantin-axenov, line 164:
```suggestion
                  Accelerated Networking provides up to 30Gbps in networking throughput.
```

candi/cloud-providers/azure/openapi/instance_class.yaml
```diff
@@ -74,6 +74,11 @@ spec:
                   additionalProperties:
                     type: string
                   x-kubernetes-preserve-unknown-fields: true
+                acceleratedNetworking:
+                  type: boolean
+                  description: |
+                    Accelerated Networking providing up to 30Gbps in networking throughput.
```
konstantin-axenov, line 80:
```suggestion
                    Accelerated Networking provides up to 30Gbps in networking throughput.
```

konstantin-axenov (approved):

PR #1294, opened 2022-04-07
[node-mananger] fix panic in apiserver endpoints discovery

shvgn:
## Description

Fix panic when we get nil instead of a string

When pod is nt ready, the hooks panics and blocks deckhouse main queue

## Changelog entries

```changes
section: node-manager
type: fix
summary: Fix panic
impact_level: low
```


RomanenkoDenys (approved):
Looks very good

nabokihms (approved):

PR #1193, opened 2022-03-29
[candi] refactor kubernetes-api-proxy

RomanenkoDenys:
## Description

Refactor kubernetes-api-proxy.
1. Use discovered apiserver endpoints to set up kubernetes api proxy on nodes.
2. Use kubernetes api proxy static pod instead of system package.
## Why do we need it, and what problem does it solve?

This PR prevents situations, when all apiservers are going down and kubernetes-api-proxy cannot discover live endpoints.
## Changelog entries

```changes
section: candi
type: feature
summary: refactor kubernetes-api-proxy
impact_level: low
```

PR #1297, opened 2022-04-07
[candi] fixed kubernetes-api-proxy run in docker envs

RomanenkoDenys:
## Description

1. Fixed kubernetes-api-proxy run in docker envs.
2. Added imagePullPolicy: IfNotPresent to kubernetes-api-proxy static pod.

When we use docker as a CRI, we cannot get control plane images from the private registry because of the lack of registry credentials on the node.

### DIFF
--- a/candi/bashible/common-steps/node-group/050_configure_and_start_docker.sh.tpl
+++ b/candi/bashible/common-steps/node-group/050_configure_and_start_docker.sh.tpl
@@ -48,16 +48,9 @@ EOF
 {{- end }}
 
 {{- if .registry.auth }}
-mkdir -p /.docker
-bb-sync-file /.docker/config.json - << "EOF"
-{
-	"auths": {
-		"{{ .registry.address }}": {
-			"auth": "{{ .registry.auth }}"
-		}
-	}
-}
-EOF
+username="$(base64 -d <<< "{{ .registry.auth }}" | awk -F ":" '{print $1}')"
+password="$(base64 -d <<< "{{ .registry.auth }}" | awk -F ":" '{print $2}')"
+HOME=/ docker login --username "${username}" --password "${password}" {{ .registry.address }}
 {{- end }}
 
 {{- end }}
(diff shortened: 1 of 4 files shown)


candi/bashible/common-steps/node-group/050_configure_and_start_docker.sh.tpl
```diff
@@ -47,4 +47,17 @@ bb-sync-file /etc/docker/certs.d/{{ .registry.address }}/ca.crt  - << "EOF"
 EOF
 {{- end }}
 
+{{- if .registry.auth }}
+mkdir -p /.docker
+bb-sync-file /.docker/config.json - << "EOF"
```
konstantin-axenov, line 52: Maybe we should check that files exists and append content to it using jq?

konstantin-axenov (dismissed):

konstantin-axenov:
@nabokihms Take a look please. This is a fix for https://github.com/deckhouse/deckhouse/pull/1193.

modules/040-control-plane-manager/images/kubernetes-api-proxy/werf.inc.yaml
```diff
@@ -1,6 +1,11 @@
 ---
 image: {{ $.ModuleName }}/{{ $.ImageName }}
 from: {{ env "BASE_NGINX_ALPINE" }}
+import:
```
nabokihms, line 4: I did not get this part. Why do we need the pause inside the Nginx image?

RomanenkoDenys: I did it because i added kubernetes-api-proxy to image-holder. I think it is an important part of control-plane.

konstantin-axenov: We need to pull kubernetes-api-proxy image before imagePullSecret exists. So we are using the same scheme as for control-plane images.

RomanenkoDenys: It can be excessive, so i removed pause binary from container and removed kubernetes-api-proxy from image-holder.

nabokihms (approved):

PR #1301, opened 2022-04-08
[candi] added ssh allow list to cloud providers

RomanenkoDenys:
## Description

Added ssh allow list to cloud providers terraform security groups.
Providers list:
Azure (standard layout).
GCP (standard, without-nat layouts).
AWS (standard, without-nat, with-nat layouts).
Openstack (simple, simple-with-internal-network, standard, standard-with-no-router layouts).

We want to limit ssh access to cluster nodes only to the allowed list of CIDRs.
## Changelog entries

```changes
section: candi
type: feature
summary: Added ssh allow list to cloud providers terraform security groups.
impact_level: low
```


RomanenkoDenys:
- [ ] Azure test
- [ ] GCP test
- [ ] AWS test
- [ ] Openstack test

ee/candi/cloud-providers/openstack/layouts/standard/base-infrastructure/main.tf
```diff
@@ -5,6 +5,7 @@ module "network_security" {
   source           = "../../../terraform-modules/network-security"
   prefix           = local.prefix
   remote_ip_prefix = "0.0.0.0/0"
```
konstantin-axenov, line 7: Can we get rid of `remote_ip_prefix` variable? I think it is confusing if we have `ssh_allow_list`.

RomanenkoDenys: Yes, that's a good point.

konstantin-axenov (approved):

konstantin-axenov:
@nabokihms Have a look, please.

PR #1302, opened 2022-04-08
[dhctl] chore: print kube client errors

shvgn:
## Description

Handle errors better for kube client.

Based on #1297

For now, the error of the listing of deckhouse pod is muted. It makes debugging dhctl converge harder

## Changelog entries

```changes
section: dhctl
type: chore
summary: Make error handling more explicit
impact_level: low
```

### DIFF
--- a/dhctl/pkg/kubernetes/actions/deckhouse/log.go
+++ b/dhctl/pkg/kubernetes/actions/deckhouse/log.go
@@ -33,10 +33,12 @@ import (
 )
 
 var (
-	ErrListPods      = errors.New("No Deckhouse pod found.")
-	ErrTimedOut      = errors.New("Time is out waiting for Deckhouse readiness.")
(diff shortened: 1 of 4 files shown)


nabokihms:
If you want to see Kubernetes errors, now it is possibly by exposing the DHCTL_DEBUG env variable. The idea was to hide them by default.

shvgn:
Blocked by https://github.com/deckhouse/deckhouse/pull/1297

dhctl/pkg/kubernetes/actions/deckhouse/log.go
```diff
@@ -33,10 +33,12 @@ import (
 )
 
 var (
-	ErrListPods      = errors.New("No Deckhouse pod found.")
```
name212, line 36: We use upper letter because these errors output to log "as is" and they will look ugly for client.

Please revert it

PR #1308, opened 2022-04-08
[prometheus] handle custom alertmanager path

yalosev:

## Description
Handle path prefix in CustomAlertmanager CR

Path prefix was forgotten in templates

## Changelog entries

```changes
section: prometheus
type: fix
summary: Handle path prefix for CustomAlertmanager
impact_level: low
```


nabokihms (approved):

z9r5:

PR #1309, opened 2022-04-08
[openvpn] fix dex applicationDomain

vitaliy-sn:
## Description

Fixed DexAuthenticator applicationDomain.

This fixes a bug that makes it impossible to login to the web interface when the user-auth module is enabled

## Changelog entries

```changes
section: openvpn
type: fix
summary: Fixed DexAuthenticator applicationDomain.
impact_level: low
```


nabokihms (approved):

z9r5:

PR #1315, opened 2022-04-11
[control-plane-manager] Etcd automatic defragmentation

name212:
## Description
Etcd stops working if it exceeds space quota.

Check etcd storage usage and if it less than 10% free space left then trigger defrag operation.

## Changelog entries

```changes
type: feature
summary: Etcd automatic defragmentation.
impact_level: low
```


modules/040-control-plane-manager/openapi/config-values.yaml
```diff
@@ -205,6 +205,12 @@ properties:
     description: |
       `etcd` parameters.
     properties:
+      disableAutoDefragmentation:
```
konstantin-axenov, line 208: We decided that usage of `enable` is preferable than `disable`. Change this option to `enableAutoDefragmentation`.

modules/040-control-plane-manager/hooks/etcd_defragmentation.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		input.LogEntry.Warnf("Start defrag etcd instance '%s' %d/%d", instance.PodName, status.DbSize, instance.MaxDbSize)
+		_, err = etcdClient.Defragment(context.TODO(), instance.Endpoint)
+		if err != nil {
+			input.MetricsCollector.Inc("d8_etcd_defragmentation_failed_total", map[string]string{
+				"pod_name": instance.PodName,
+				"node":     instance.Node,
+			})
+			input.LogEntry.Errorf("Defrag etcd '%s' instance finished with err: %v", instance.PodName, err)
```
konstantin-axenov, line 81: Don't we need to stop defrag operation on an etcd cluster if defrag of one of the members failed? For example, an etcd member was killed by oom.

name212: yep

name212: I forgot about it. Thank you

name212:
We decided to give up automatic defragmentation

PR #1353, opened 2022-04-13
[openvpn] Set default value for loadbalancer object in the OpenAPI schema

vitaliy-sn:
## Description

Set default value for loadbalancer object in the OpenAPI schema

This fixes a bug in the OpenAPI schema that occurs when using the LoadBalancer inlet.

## Changelog entries

```changes
section: openvpn
type: fix
summary: Set default value for loadbalancer object in the OpenAPI schema.
```

### DIFF
--- a/ee/fe/modules/500-openvpn/openapi/config-values.yaml
+++ b/ee/fe/modules/500-openvpn/openapi/config-values.yaml
@@ -18,6 +18,7 @@ properties:
     type: object
     description: |
       A section of optional parameters of the `LoadBalancer` inlet.
+    default: { }
     properties:
       annotations:
         type: object
--- a/ee/fe/modules/500-openvpn/openapi/config-values.yaml
+++ b/ee/fe/modules/500-openvpn/openapi/config-values.yaml
@@ -18,7 +18,7 @@ properties:
     type: object
     description: |
       A section of optional parameters of the `LoadBalancer` inlet.
-    default: { }
+    default: {}
     properties:
       annotations:
         type: object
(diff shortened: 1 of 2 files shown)


ee/fe/modules/500-openvpn/openapi/config-values.yaml
```diff
@@ -18,6 +18,7 @@ properties:
     type: object
     description: |
       A section of optional parameters of the `LoadBalancer` inlet.
+    default: { }
```
zuzzas, line 21: Will work fine, but breaks consistency with `{}` elsewhere. Even in the same file: https://github.com/deckhouse/deckhouse/blob/e83377c1233c8bc7299cfce7e352e5805fe71856/ee/fe/modules/500-openvpn/openapi/config-values.yaml#L80

zuzzas (dismissed):

ee/fe/modules/500-openvpn/openapi/config-values.yaml
nabokihms, line 21:
```suggestion
    default: {}
```

ee/fe/modules/500-openvpn/openapi/doc-ru-config-values.yaml
```diff
@@ -11,6 +11,7 @@ properties:
   loadBalancer:
     description: |
       Секция опциональных настроек для inlet'а `LoadBalancer`.
+    default: { }
```
nabokihms, line 14:
```suggestion
```

nabokihms (approved):

PR #1358, opened 2022-04-13
[linstor] Update FAQ: add error case

kvaps:
## Description

Describe one error case from @duckhawk, when nodes were eventually evicted from the cluster due to long downtime.

Better UX

## Changelog entries

```changes
section: linstor
type: chore
summary: Add an error case to FAQ, when nodes were eventually evicted from the cluster
impact_level: low
```


z9r5 (approved):

z9r5:

PR #1361, opened 2022-04-13
[cloud-provider-yandex] Change default platform to standard-v3 for new instances created by machine-controller-manager

vitaliy-sn:
## Description

* New instances created by `machine-controller-manager` will use the `standard-v3` platform by default.
    * If you need to change the platform for existing instances, you have to roll out update them or just delete machines in `d8-cloud-instance-manager` namespace.
    * To change the default platform, use the `.spec.platformID` parameter in `YandexInstanceClass`.
* The master and static nodes are set to default platform `standard-v2` to be automatically added to `YandexClusterConfiguration`, which will allow us to change the default value later without any problems.
    * To change the platform, use the `.masterNodeGroup.instanceClass.platform` parameter in `YandexClusterConfiguration` and run `dhctl converge`.

When setting a default value in a spec, it will be set in all YandexInstanceClass resources. This will allow us to change the default value of the platformID in the future without having to rebootstrap the cluster nodes.

## Changelog entries

```changes
section: cloud-provider-yandex
type: feature
summary: Changed default platform to standard-v3 for new instances created by machine-controller-manager.
```

### DIFF
--- a/candi/cloud-providers/yandex/openapi/cluster_configuration.yaml
+++ b/candi/cloud-providers/yandex/openapi/cluster_configuration.yaml
@@ -50,6 +50,7 @@ apiVersions:
                 description: |
                   The type of virtual machine to create.
                 type: string
+                default: standard-v2
               cores:
                 description: |
                   Amount of CPU cores to provision on a Yandex Compute Instance.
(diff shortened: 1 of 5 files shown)


konstantin-axenov:
@vitaliy-sn Please, update description. There is no information that new machines will have platform standard-v3.

konstantin-axenov (approved):

PR #1364, opened 2022-04-13
[go_lib::storage_class_change] fix for storageClass=false

vitaliy-sn:
## Description

Fixed a bug for the case when the storage class is set to "false"

Using false resulted in an error and didn't allow forcing emptyDir on modules.

## Changelog entries

```changes
section: global
type: fix
summary: Fixed a bug for the case when the storage class is set to "false"
```


nabokihms (approved):

z9r5:

PR #1379, opened 2022-04-15
add Andrey as an owner of CCM

EvgenySamoylov:
## Description

Add @zuzzas as a codeowner of the CCM.


nabokihms:
Ok, I'm going to merge it, but let's hold further changes to the code owner file for a while. We need to figure out the duties of a code owner, whether it is comfortable to have code owners or not, etc.

PR #1398, opened 2022-04-18
[cloud-providers-*] Fix cloud providers centos bundle

RomanenkoDenys:
## Description

Change bashible bundle centos-7 to centos in cloud-providers.

We renamed main bundle from centos-7 to centos to support many versions of centos. But we forgot also rename bundles in cloud providers.

## Changelog entries

```changes
section: candi
type: fix
summary: renamed bunde from centos-7 to centos
impact_level: low
```


konstantin-axenov (approved):

z9r5:

z9r5:

PR #1399, opened 2022-04-18
[cloud-provider-aws] ssh-accessible terraform fix

vitaliy-sn:
## Description

The aws_security_group.ssh-accessible resource has been restored in the terraform module security-group, because its removal is not possible, due to the dependence of the master node on it. It will be removed later when the master nodes are associated with the new security-group.

## Changelog entries

```changes
section: cloud-provider-aws
type: fix
summary: ssh-accessible terraform fix
```


RomanenkoDenys:
Description ?

PR #1217, opened 2022-03-30
[cloud-provider-aws] allow ssh access to master nodes only from bastion

vitaliy-sn:
## Description

Allow ssh access to master nodes only from bastion (created by dhctl) for with-nat layout.

## Why do we need it, and what problem does it solve?

This will improve security.

## Changelog entries

```changes
section: cloud-provider-aws
type: fix
summary: Allow ssh access to master nodes only from bastion for with-nat layout
impact_level: low
```

PR #1400, opened 2022-04-18
Revert "[cloud-provider-aws] allow ssh access to master nodes only from bastion"

konstantin-axenov:
## Description

Reverts deckhouse/deckhouse#1217.

[PR #1217](deckhouse/deckhouse#1217) breaks terraform state and manual operations required, so auto converge doesn't work. Further work will be done in https://github.com/deckhouse/deckhouse/pull/1301.

## Changelog entries

```changes
section: cloud-provider-aws
type: fix
summary: Allow ssh access to master nodes only from bastion for with-nat layout
impact_level: low
```


z9r5:

PR #1413, opened 2022-04-19
[node-manager] Remove migration which added and removed labels for control-plane nodes

name212:
## Description
Remove migration which added and removed labels for control-plane nodes

They are no longer needed.

## Changelog entries

```changes
section: node-manager
type: chore
summary: Remove migration which added and removed labels for control-plane nodes.
impact_level: low
```


nabokihms:
I changed the level of impact to low for this because it is ok for Deckhouse to delete migrations between minor releases, and it is stated somewhere on a site.

PR #1415, opened 2022-04-19
[prometheus] deckhouse web interfaces list

yalosev:

## Description
Generate list of deckhouse public web pages (dashboard, docs, prometheus, istio/kiali, etc) and create a table with links on a grafana home page
image

It would be comfortable for a new users to see all resources they might need in one place

## Changelog entries

```changes
section: prometheus
type: feature
summary: Create table with enabled Deckhouse web interfaces on the Grafana home page
```


nabokihms:
image
1. It looks like this new panel does not fit the screen. 
2. Do we need a header for the table? It seems clear where the name is and where the URL is to me.
3. Other tables have transparent backgrounds. It would be best to align this new table with others.

modules/300-prometheus/hooks/metrics_web_interfaces.go
```diff
@@ -0,0 +1,120 @@
...
+
+	rule := ing.Spec.Rules[0]
+
+	host := rule.Host
+	urlPath := ""
+	if len(rule.HTTP.Paths) > 0 {
+		urlPath = rule.HTTP.Paths[0].Path
+	}
+
+	if urlPath != "" {
```
nabokihms, line 91: Seems like a hack to me. Maybe it is better to pass the path explicitly somehow (via annotations?).

modules/500-upmeter/templates/status/ingress.yaml
```diff
@@ -5,7 +5,7 @@ kind: Ingress
 metadata:
   name: status
   namespace: d8-{{ .Chart.Name }}
-  {{- include "helm_lib_module_labels" (list . (dict "app" "status")) | nindent 2 }}
+  {{- include "helm_lib_module_labels" (list . (dict "app" "status" "deckhouse.io/export-domain" "status")) | nindent 2 }}
```
nabokihms, line 8: Let's do the same for the upmeter domain.

shvgn: I expected it to be an annotation. Why is it a label?

yalosev: never mind. It became an annotation :)

modules/300-prometheus/hooks/metrics_web_interfaces.go
yalosev: Added annotation. If annotation is not set - get it from rules

yalosev:
image

ee/fe/modules/500-openvpn/templates/openvpn/ingress.yaml
```diff
@@ -7,6 +7,8 @@ metadata:
   namespace: d8-{{ .Chart.Name }}
   {{- include "helm_lib_module_labels" (list . (dict "app" .Chart.Name)) | nindent 2 }}
   annotations:
+    web.deckhouse.io/export-name: "openvpn"
+    web.deckhouse.io/export-icon: "https://openvpn.net/wp-content/uploads/cropped-openvpn-32x32.png"
```
nabokihms, line 11: @konstantin-axenov, we need a second opinion here. We decided to add icons for all interfaces, and it looks marvelous.


However, we have a problem with icons providers.There are two options and corresponding perks:
1. We can use external links for icons. It is easy to add new interfaces, but we cannot control these icons. If someone decides to delete an icon, we will have to release a new version of Deckhouse to fix it. There is also a problem: users need access to websites with these icons (which will not work for closed environments). 

2. On the other hand, we can add all icons to the Grafana pod and serve them as static images by the Grafana server itself. This approach is solid, yet it makes adding a new interface a little more complicated.

WDYT?

konstantin-axenov: I think that the second option is the right way to do it.

konstantin-axenov: BTW, it looks really nice!

shvgn: Upmeter has its icon tho [modules/500-upmeter/images/webui/src/assets/upmeter.png](https://github.com/deckhouse/deckhouse/blob/main/modules/500-upmeter/images/webui/src/assets/upmeter.png)

yalosev: Done

nabokihms (approved):

nabokihms:
I realized that anyone could add their web interfaces to the front page by annotation desired ingress resources. Good customization feature!

PR #1422, opened 2022-04-19
[node-manager][dhctl] Move creating master ng to hook from dhctl

name212:
## Description
Move creating master ng to hook from dhctl.
In hook, if master node group exists, hook should not affect any changes in master node group.

Right way to creating master node group in node-manager immediately after deploy node group crd.
And we have race between creating master node group and control-plane manager. Control plane manager can redeploy kubelet config  faster than creating master node group from dhctl and we can get negative effects (can not get logs, because cert regenerated without node ip, etc...)

## Changelog entries

```changes
section: node-manager
type: feature
summary: Add hook for creating master node group
impact_level: low
---
section: dhctl
type: feature
summary: Remove creating master node group
impact_level: low
```


deckhouse-BOaTswain:

nabokihms (approved):

PR #1430, opened 2022-04-20
[log-shipper] Fix reloading signal

### LINKED ISSUE #1427: [log-shipper] Config reloading may fail
### Preflight Checklist

### Version

v1.32.0

### Expected Behavior

After changes, the vector configuration should be reloaded.

### Actual Behavior

Log-shipper agent uses the `pidof` function to find the PID of a running vector process. However, pidof returns many pids if more than one instance of vector is running.  In this case, the kill execution fails, and reloader does not retry reloading again.

### Steps To Reproduce

1. Exec into one of the log-shipper agent pods
2. Run `vector vrl` command
3. Change a clusterloggingsource resources in the cluster
4. Watch the `vector-reloader` container logs (it is a log-shipper-agent sidecar container)

### Additional Information

_No response_

### Logs

```shell
Wed Apr 20 07:54:30 UTC 2022 | Reloading vector
/usr/bin/reloader: line 86: kill: 159 149 6: arguments must be process or job IDs
```

nabokihms:

## Description
Use pkill instead of pidof.

By executing pkill reloader sends a reloading signal to every vector process, which is technically correct.

Closes #1427

## Changelog entries

```changes
section: log-shipper
type: fix
summary: Send reloading signal to all vector processes in a container on config change.
impact_level: default
```

### DIFF
--- a/modules/460-log-shipper/images/vector/Dockerfile
+++ b/modules/460-log-shipper/images/vector/Dockerfile
@@ -29,7 +29,7 @@ RUN cargo build \
 FROM $BASE_DEBIAN_BULLSEYE
 RUN mkdir -p /etc/vector \
     && apt-get update \
-    && apt-get install -yq ca-certificates tzdata inotify-tools gettext \
+    && apt-get install -yq ca-certificates tzdata inotify-tools gettext procps \
     && rm -rf /var/cache/apt/archives/*
 COPY --from=build /vector/vector/target/release/vector /usr/bin/vector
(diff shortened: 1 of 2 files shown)


yalosev (approved):

yalosev:
When do we have a case of multiply vector processes running at once?

nabokihms:
If you run `vector top` or `vector vrl` in the log-shipper pod, you will have two processes of vector running simultaneously.

PR #1437, opened 2022-04-21
[prometheus] fix calculating pvc size for local storage

vitaliy-sn:
## Description

Fixed PersistentVolumeClaim size calculation for local storage.

Incorrect calculation led to the need to perform manual actions when rolling out a release.

## Changelog entries

```changes
section: prometheus
type: fix
summary: Fixed PersistentVolumeClaim size calculation for local storage.
```


nabokihms (approved):

z9r5:

PR #1439, opened 2022-04-21
[namespace-configurator] Ignore upmeter probe namespaces

nabokihms:

## Description
Exclude namespaces by heritage labels.

Upmeter deletes namespaces faster than deckhouse can put labels or annotations on them.

## Changelog entries

```changes
section: namespace-configurator
type: fix
summary: Exclude upmeter probe namespaces from namespace-configurator snapshots.
impact_level: default
```


z9r5:

PR #1440, opened 2022-04-21
[candi] added file sync on bashible.sh change

RomanenkoDenys:
## Description

Added filesystem sync on copy bashible-new.sh to bashible.sh

In centos releases with xfs file system, when node is cold-rebooted, bashible.sh truncated to zero size. This is xfs bug.
https://stackoverflow.com/questions/47077828/xfs-rhel7-3-cold-reboot-file-truncate

## Changelog entries

```changes
section: candi
type: fix
summary: added file sync on bashible.sh change
impact_level: low
```


konstantin-axenov (approved):

z9r5:

PR #1443, opened 2022-04-21
[node-manager] Nodes batch bootstrap doc

konstantin-axenov:
## Description

- Add `How do I add a batch of static nodes to a cluster?` to FAQ.
- Create ServiceAccount with access to manual-bootstrap secrets

Sometimes users ask information if bootstrap process of static nodes can be automated. It seems that we have to provide information how it can be done.

## Changelog entries

```changes
section: node-manager
type: feature
summary: Nodes batch bootstrap docs
impact_level: low
```


testing/matrix/linter/rules/roles/placement.go
```diff
@@ -157,6 +159,16 @@ func objectRBACPlacementServiceAccount(m utils.Module, object storage.StoreObjec
 			"Name of ServiceAccount should be equal to %q or %q",
 			serviceAccountName, expectedServiceAccountName,
 		)
+	case strings.HasSuffix(shortPath, "service-account.yaml"):
```
nabokihms, line 162: Let's not add new files to existing rules. It seems fine to me to stick to the `rbac-for-us/rbac-to-us` rule. More files = more entropy. Why do you need to put service accounts in a separate file?

testing/matrix/linter/rules/roles/placement.go
```diff
@@ -157,6 +159,16 @@ func objectRBACPlacementServiceAccount(m utils.Module, object storage.StoreObjec
...
+		if namespace == ServiceAccountDefaultNs {
```
nabokihms, line 163: It is supposed to be a namespace for user service accounts. To prevent overlapping and add our service account to this namespace, Deckhouse should always use a prefix, and it should be enforced.

Yet I still did not get why we needed to use this NS.

modules/040-node-manager/templates/node-group/rbac-to-us.yaml
```diff
@@ -0,0 +1,35 @@
...
+  {{- if and (ne $ng.nodeType "CloudEphemeral") (ne $ng.nodeType "Cloud") }}
+    {{- $manual_ngs = append $manual_ngs (printf "manual-bootstrap-for-%s" $ng.name) }}
+  {{- end }}
+{{- end }}
+{{- if gt (len $manual_ngs) 0 }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: access-to-manual-bootstrap-secrets
```
nabokihms, line 12: Looks like RBAC for us to me.

modules/040-node-manager/templates/node-group/service-account.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: v1
+kind: ServiceAccount
```
nabokihms, line 3: * Rename this file to the `rbac-for-us.yaml`
* Change destination namespaces to the `d8-cloud-instance-manager`
* Change its name

This will fix the problem with the linter.

nabokihms (changes requested):

modules/040-node-manager/templates/node-group/service-account.yaml
konstantin-axenov: Namespace changed to d8-system because of other lint rules.

modules/040-node-manager/templates/node-group/rbac-to-us.yaml
konstantin-axenov: Done.

testing/matrix/linter/rules/roles/placement.go
konstantin-axenov: It seemed to me that this is more like SA for end user and not for us. So this namespace looked like a proper place. Anyway, I fixed it.

testing/matrix/linter/rules/roles/placement.go
konstantin-axenov: I tough that it was a good idea because this was the first case when Deckhouse generates SA that will be used by end user. But I agree that it creates more entropy. Fixed.

modules/040-node-manager/templates/node-group/service-account.yaml
nabokihms: Lint rules allow you to deploy it to the `d8-cloud-instance-manager` namespace. You only need to give the sa a different name.

modules/040-node-manager/docs/FAQ.md
```diff
@@ -15,6 +15,70 @@ To add a new static node (e.g., VM or bare-metal server) to the cluster, you nee
   - Configured the network in the cluster;
 4. Connect to the new node over SSH and run the following command using the data from the secret: `echo <base64> | base64 -d | bash`
 
+## How do I add a batch of static nodes to a cluster?
+If you don't have `NodeGroup` in your cluster, then you can find information how to do it [here](#how-do-i-add-a-static-node-to-a-cluster).
+If you already have `NodeGroup`, you can automate bootstrap process with any automation platform that you prefer. We will use Ansible as example.
```
z9r5, line 20:
```suggestion
If you already have `NodeGroup`, you can automate the bootstrap process with any automation platform that you prefer. We will use Ansible as an example.
```

modules/040-node-manager/docs/FAQ_RU.md
```diff
@@ -15,6 +15,70 @@ search: добавить ноду в кластер, добавить узел 
   - Между узлами кластера настроена сетевая связанность.
 4. Зайти на новый узел по SSH и выполнить команду из Secret'а: `echo <base64> | base64 -d | bash`.
 
+## Как добавить несколько статических узлов в кластер?
+Если у вас в кластере не созданы `NodeGroup`, то вы можете ознакомиться с информацией, как это сделать в [этом разделе](#как-добавить-статичный-узел-в-кластер).
+Если у вас уже созданы `NodeGroup`, то вы можете автоматизировать процесс добавления узлов с помощью любой предпочитаемой платформы автоматизации. Для примера мы будем использовать Ansible.
+
+1. Получите один из адресов Kubernetes API сервера. Обратите внимание, что IP адрес должен быть доступен с узлов, которые добавляются в кластер.
```
z9r5, line 22:
```suggestion
1. Получите один из адресов Kubernetes API-сервера. Обратите внимание, что IP-адрес должен быть доступен с узлов, которые добавляются в кластер.
```

modules/040-node-manager/docs/FAQ_RU.md
```diff
@@ -15,6 +15,70 @@ search: добавить ноду в кластер, добавить узел 
...
+   ```shell
+   kubectl get ep kubernetes -o json | jq '.subsets[0].addresses[0].ip + ":" + (.subsets[0].ports[0].port | tostring)' -r
+   ```
+2. Получить Kubernetes API токен для специального `ServiceAccount`, которым управляет Deckhouse.
```
z9r5, line 26:
```suggestion
2. Получите Kubernetes API-токен для специального `ServiceAccount`, которым управляет Deckhouse.
```

z9r5 (approved):

modules/040-node-manager/templates/node-group/service-account.yaml
konstantin-axenov: Done

nabokihms (approved):

PR #1444, opened 2022-04-21
[prometheus] Add app-root annotation for example `Connecting Prometheus to an external Grafana instance`

name212:
## Description

Add app-root annotation for example `Connecting Prometheus to an external Grafana instance`

By default our prometheus redirect from `/` to `/prometheus/graph`.
Who use our example, will get 404 error for `https://domain/` url.


nabokihms (approved):

nabokihms:
@name212 good catch! Thank you.

z9r5:

Issue #1433, opened 2022-04-21
[log-shipper] Refactor configuration hook


nabokihms:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

There are some dark and repetitive moments in the code in this file:
https://github.com/deckhouse/deckhouse/blob/main/modules/460-log-shipper/hooks/generate_config.go#L108

We need to have one place to control the ordering logic of pipelines.

### Proposed Solution

Refactor!

nabokihms:
@EvgenySamoylov I would like to pick this one because otherwise, it will slow down resolving issues and debugging. I left TODO entries in the code near parts I would like to rewrite.

PR #1475, opened 2022-04-26
[log-shipper] Refactor configuration hook

nabokihms:
## Description
* Separate packages for sources and destinations
* Move logic of the transforms ordering to a single package
* It also fixes the problem when a source is renamed on adding a new destination, which leads to loose checkpoints

Close https://github.com/deckhouse/deckhouse/issues/1433

## Changelog entries

```changes
section: log-shipper
type: fix
summary: Refactor configuration hook
impact_level: low
```


modules/460-log-shipper/hooks/internal/handler/generator.go
```diff
@@ -0,0 +1,183 @@
...
+		input.LogEntry.Warn(err)
+	}
+
+	logConfigGenerator := config.NewLogConfigGenerator()
+
+	for _, source := range clusterSources {
+		pipeline, err := config.NewPipelineCluster(logConfigGenerator, g.logDestinations, &source)
+		if err != nil {
+			input.LogEntry.Warn(err)
+			continue
```
yalosev, line 62: If we get an error we will skip the pipeline but will keep the rest of the config. Does it valid? We will have some cutted config which can lead to unexpected behavior

nabokihms: The pipeline is a valid part of the config (the logical one). As I see it, it is not a big deal because we duplicate sinks, sources, and transforms.

Btw, I think this code is in [master](https://github.com/deckhouse/deckhouse/blob/ecd86c64814c5b7a4bdd13cd2fe3d2ec3c49d1f9/modules/460-log-shipper/hooks/generate_config.go#L239-L257) already.
I just moved it to the dedicated method. And I believe that this code is working like this from the [very beginning (v1.28)](https://github.com/deckhouse/deckhouse/blob/3b69c70aa47dcfe1ed3bc1f360f557aa2273e320/modules/460-log-shipper/hooks/generate_config.go#L199-L219).

modules/460-log-shipper/hooks/internal/handler/generator.go
```diff
@@ -0,0 +1,183 @@
...
+	// load all destinations into the destination map
+	for _, d := range snap {
+		dest := d.(v1alpha1.ClusterLogDestination)
+		destNameToObject[dest.Name] = dest
+	}
+
+	return &Generator{logDestinations: destNameToObject}
+}
+
+func (g *Generator) Do(input *go_hook.HookInput) ([]byte, error) {
```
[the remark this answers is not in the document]
nabokihms: Yeah, I thought about something like this. The main reason is that in my mind, `FomInput` is one way to create a new Generator instance. Theoretically, there can be others like `FromYAML` or `FromMap`.

It looks like this because I try to keep as much logic as possible from the original hook. Then we can write more granular unit tests and change the structure.

yalosev (approved):

PR #1480, opened 2022-04-27
[node-manager] Truncate event message to allowed maximums in update_node_group_status hook

zuzzas:
## Description

Events with messages longer than 1024 bytes fail apiserver validation.

Avoid hook failures:
```
Module hook failed, requeue task to retry after delay. Failed count is 1. Error: 1 error occurred:
        * Create object: events.k8s.io/v1/Event/default/: Event "ng-node-qgvpq" is invalid: message: Invalid value: "": can have at most 1024 characters
```

## Changelog entries

```changes
section: node-manager
type: fix
summary: Truncate event message to allowed maximums in update_node_group_status hook
impact_level: low
```

### DIFF
--- a/modules/040-node-manager/hooks/update_node_group_status.go
+++ b/modules/040-node-manager/hooks/update_node_group_status.go
@@ -351,6 +351,9 @@ func handleUpdateNGStatus(input *go_hook.HookInput) error {
 		}
 		statusMsg := fmt.Sprintf("%s %s", nodeGroup.Error, failureReason)
 		statusMsg = strings.TrimSpace(statusMsg)
+		// truncate to maximum allowed message limit
+		// https://github.com/kubernetes/kubernetes/blob/3e442b74f717ab2f43897d7af50de6114486e459/pkg/apis/core/validation/events.go#L38
+		statusMsg = statusMsg[:1024]
 		if len(statusMsg) > 0 {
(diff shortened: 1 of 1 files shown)


modules/040-node-manager/hooks/update_node_group_status.go
```diff
@@ -351,6 +351,9 @@ func handleUpdateNGStatus(input *go_hook.HookInput) error {
 		}
 		statusMsg := fmt.Sprintf("%s %s", nodeGroup.Error, failureReason)
 		statusMsg = strings.TrimSpace(statusMsg)
+		// truncate to maximum allowed message limit
+		// https://github.com/kubernetes/kubernetes/blob/3e442b74f717ab2f43897d7af50de6114486e459/pkg/apis/core/validation/events.go#L38
+		statusMsg = statusMsg[:1024]
```
yalosev, line 356: if statusMsg is empty (and it could be) you will have a panic here. You should move it 1 line down, under `if len(statusMsg) > 0`

zuzzas: Still won't be enough if it's shorter, but longer than zero. Fixed it!

yalosev (approved):

nabokihms (approved):

PR #1482, opened 2022-04-27
[candi] fixed race condition between old kubernetes-api-proxy-configurator and bashible step

RomanenkoDenys:
## Description

Fixed race condition between old kubernetes-api-proxy-configrator and new kubernetes-api-proxy bashible step.

When we migrate from old kubernetes-api-proxy-configurator to kubernetes-api-proxy static pod, we stop kubernetes-api-proxy-configurator.timer.
But in some cases old kubernetes-api-proxy-configurator runs in parallel with step 052_configure_kubernetes_api_proxy.sh and damages /etc/kubernetes/kubernetes-api-proxy/nginx_new.conf file.
To prevent this, we also need to stop kubernetes-api-proxy-configurator.service.
## Changelog entries

```changes
section: candi
type: fix
summary: fixed race condition between old kubernetes-api-proxy-configurator and bashible step
```


konstantin-axenov (approved):

z9r5:

PR #1493, opened 2022-04-28
[openvpn] remove openvpn.storageClass from deckhouse configmap

vitaliy-sn:
## Description

* Temporarily returned the storageClass parameter in the spec.
* Added removal of openvpn.storageClass parameter from deckhouse configmap to migration hook.

The storageClass parameter is no longer needed.

## Changelog entries

```changes
section: openvpn
type: fix
summary: Removing of openvpn.storageClass parameter from deckhouse configmap
```


ee/fe/modules/500-openvpn/openapi/config-values.yaml
```diff
@@ -205,3 +205,5 @@ properties:
           format: int64
         value:
           type: string
+  storageClass:
+    type: string
```
nabokihms, line 209: The `description` field is missing. This field will be shown in the docs, so we need to provide a description for it and mention that this field is deprecated.

ee/fe/modules/500-openvpn/hooks/easyrsa_migration.go
```diff
@@ -95,6 +95,8 @@ func applyMigrationSecretFilter(obj *unstructured.Unstructured) (go_hook.FilterR
 }
 
 func migration(input *go_hook.HookInput) error {
+	input.ConfigValues.Remove("openvpn.storageClass")
```
nabokihms, line 98: Could you please add a comment on why should we delete this field in the easyrsa migration?

nabokihms (changes requested):

ee/fe/modules/500-openvpn/openapi/config-values.yaml
vitaliy-sn: Wanted to do without it. Added a description.

ee/fe/modules/500-openvpn/hooks/easyrsa_migration.go
vitaliy-sn: Done

nabokihms (approved):

z9r5:

PR #1495, opened 2022-04-28
[docs] Suggest gp3 as disk type in bastion instance

shvgn:
## Description

Suggest gp3

Easy mindless copying

## Changelog entries

```changes
section: docs
type: chore
summary: "Suggest gp3 for bastion instance in AWS-based 'Getting started'"
```


docs/site/_includes/getting_started/aws/partials/config.yml.with_nat.ee.inc
```diff
@@ -101,6 +101,9 @@ withNAT:
       # [<en>] you might consider changing this
       # [<ru>] возможно, захотите изменить
       ami: ami-09a4a23815cdb5e06
+      # [<en>] Disk type
+      # [<ru>] Тип диска
+      diskType: gp3
```
shvgn, line 106: I think these comments are useless, but, well, they are everywhere

PR #1507, opened 2022-05-03
[log-shipper] parse_json max depth

nabokihms:
## Description
TBA

TBA

## Changelog entries

```changes
section: log-shipper
type: feature
summary: Add a max depth parameter to the parse_json VRL function to efficiently parse nested JSON-formatted messages.
impact_level: default
```


nabokihms:
The experiment was successful, so closing it.

PR #1515, opened 2022-05-04
[dhctl][node-manager] Old dockerCfg format support (construct auth from username and password)

### LINKED ISSUE #194: [bashible] install registry packages fails, if registryDockerCFG in old format
### Preflight Checklist

### Version

v1.24

### Expected Behavior

if registryDockerCfg don't contain "auth" field, bashible-apiserver must use username and password fields to  construct registry.auth parameter for bashible.

### Actual Behavior

if registryDockerCfg don't contain "auth" field, bashible-apiserver cannot parse this record properly, and bashible don't get proper formed registry.auth parameter.

### Steps To Reproduce

Run cluster with registryDockerCfg without auth field.

### Additional Information

_No response_

### Logs

_No response_

name212:
## Description
Close #194

Add support in dhctl and bashible-apiserver.

if registryDockerCfg does not contain "auth" field, use username and password fields to construct registry.auth parameter for bashible.

We should support old docker config format.
We can not bootstrap cluster and bashible does not work, if we change deckhouse-registry secret in old format. 

## Changelog entries

```changes
section: dhctl
type: fix
summary: Old dockerCfg format support (construct auth from username and password)
impact_level: low
---
section: node-manager
type: fix
summary: Old dockerCfg format support in bashible-apiserver (construct auth from username and password)
impact_level: low
```

### DIFF
--- a/dhctl/pkg/config/meta.go
+++ b/dhctl/pkg/config/meta.go
@@ -497,7 +497,9 @@ func (m *MetaConfig) LoadVersionMap(filename string) error {
 func (m *MetaConfig) ParseRegistryData() (map[string]interface{}, error) {
 	type dockerCfg struct {
 		Auths map[string]struct {
-			Auth string `json:"auth"`
+			Auth     *string `json:"auth"`
+			Username *string `json:"username"`
+			Password *string `json:"password"`
 		} `json:"auths"`
 	}
 
@@ -514,14 +516,21 @@ func (m *MetaConfig) ParseRegistryData() (map[string]interface{}, error) {
 			return nil, fmt.Errorf("cannot base64 decode docker cfg: %v", err)
 		}
 
-		log.DebugF("parse registry data: dockerCfg after base64 decode = %s\n", bytes)
+		log.DebugF("parse registry data: dockerCfg after base64 decode = %s\n", string(bytes))
 		err = json.Unmarshal(bytes, &dc)
 		if err != nil {
 			return nil, fmt.Errorf("cannot unmarshal docker cfg: %v", err)
 		}
 
 		if registry, ok := dc.Auths[m.Registry.Address]; ok {
-			registryAuth = registry.Auth
+			if registry.Auth != nil {
+				registryAuth = *registry.Auth
+			} else if registry.Username != nil && registry.Password != nil {
+				auth := fmt.Sprintf("%s:%s", *registry.Username, *registry.Password)
+				registryAuth = base64.StdEncoding.EncodeToString([]byte(auth))
+			} else {
+				log.DebugF("auth or username with password not found in dockerCfg %s for %s. Use empty string", string(bytes), m.Registry.Address)
+			}
 		}
 	}
 
--- a/dhctl/pkg/config/meta.go
+++ b/dhctl/pkg/config/meta.go
@@ -516,7 +516,7 @@ func (m *MetaConfig) ParseRegistryData() (map[string]interface{}, error) {
 			return nil, fmt.Errorf("cannot base64 decode docker cfg: %v", err)
 		}
 
-		log.DebugF("parse registry data: dockerCfg after base64 decode = %s\n", string(bytes))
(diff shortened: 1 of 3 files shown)


deckhouse-BOaTswain:

modules/040-node-manager/images/bashible-apiserver/pkg/template/context_builder.go
```diff
@@ -495,7 +502,9 @@ type registryInputData struct {
 
 type dockerCfg struct {
 	Auths map[string]struct {
-		Auth string `json:"auth"`
+		Auth     *string `json:"auth"`
+		Username *string `json:"username"`
+		Password *string `json:"password"`
```
yalosev, line 507: Honestly, I don't see the need of pointers here. You don't need 3 states, only fallback. You can always compare with an empty string and fallback if true. It's not critical, so you can hold this but this gives us extra gc workload without a reason.

We can wait for @RomanenkoDenys opinion

dhctl/pkg/config/meta.go
```diff
@@ -514,14 +516,22 @@ func (m *MetaConfig) ParseRegistryData() (map[string]interface{}, error) {
 			return nil, fmt.Errorf("cannot base64 decode docker cfg: %v", err)
 		}
 
-		log.DebugF("parse registry data: dockerCfg after base64 decode = %s\n", bytes)
+		log.DebugF("parse registry data: dockerCfg after base64 decode = %s\n", string(bytes))
```
RomanenkoDenys, line 519: in the internal logging functions uses fmt.Sprintf, so we don't need to do conversion.
```
func simpleFormat(format string, a ...interface{}) string {
	var msg string
	if len(a) > 0 {
		msg = fmt.Sprintf(format, a...)
	} else {
		msg = fmt.Sprintf("%s", format)
	}

	return msg
}
```

dhctl/pkg/config/meta_test.go
```diff
@@ -131,21 +138,59 @@ provider:
...
+	return base64.StdEncoding.EncodeToString([]byte(auth))
 }
 
-func TestPrepare(t *testing.T) {
-	t.Run("Registry", func(t *testing.T) {
-		dataToRender := map[string]interface{}{
-			"dockerCfg":  generateDockerCfg(),
-			"imagesRepo": "r.example.com/deckhouse/ce/",
+func generateDockerCfg(host, username, password string) string {
+	return fmt.Sprintf(`{ "auths": { "%s": { "auth": "%s" } } }`, host, dockerCfgAuth(username, password))
```
RomanenkoDenys, line 147:
```suggestion
	return fmt.Sprintf(`{"auths":{"%s":{"auth":"%s"}}}`, host, dockerCfgAuth(username, password))
```

modules/040-node-manager/images/bashible-apiserver/pkg/template/context_builder.go
RomanenkoDenys: I think this is not important, but this struct is copy of struct from [dhctl/pkg/config/meta.go](https://github.com/deckhouse/deckhouse/pull/1515/files#diff-4b6226a078c0950de166cd15009acfea9b56a97a8264f43eca39cc72b2703c91), so we can leave this as is or change it in two places.

RomanenkoDenys (changes requested):

modules/040-node-manager/images/bashible-apiserver/pkg/template/context_builder.go
name212: Ok. I checked that empty string is normal case for checking existing auth string. I removed pointers.

dhctl/pkg/config/meta.go
name212: Oh, Thank you. I removed it.

RomanenkoDenys (approved):

yalosev (approved):

PR #1521, opened 2022-05-05
[openvpn] improved migration to secrets

vitaliy-sn:
## Description

Regardless of the status of the certificate, if the file was not found, we try to read it from the pki/revoked directory.

In some older installations, certificates were left in the pki/revoked directory after unrevoke. This bug has already been fixed a long time ago.

## Changelog entries

```changes
section: openvpn
type: fix
summary: improved migration to secrets
```


z9r5:

PR #1522, opened 2022-05-05
[ceph-csi] Fix nodeSelector for csi-node pods in helm_lib

vitaliy-sn:
## Description

Fixed nodeSelector for csi-node DaemonSet in helm_lib.

This is necessary for csi-node pods to run on all types of nodes.

## Changelog entries

```changes
section: ceph-csi
type: fix
summary: Fixed nodeSelector for csi-node pods in helm_lib
```


nabokihms (approved):

z9r5:

PR #1523, opened 2022-05-05
avoid hook failure on absent/problematic Prometheus

yalosev:

## Description
Suppress errors on hook failure

Prometheus instance could be absent or not workable, hook will fail in that case. It's not a problem and shouldn't affect the Deckhouse queue

## Changelog entries

```changes
section: helm
type: fix 
summary: Avoid hook failure on errors
```

### DIFF
--- a/modules/013-helm/hooks/startup_metrics.go
+++ b/modules/013-helm/hooks/startup_metrics.go
@@ -42,29 +42,11 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 func handleDeprecatedAPIStartup(input *go_hook.HookInput, dc dependency.Container) error {
 	input.MetricsCollector.Expire("helm_deprecated_apiversions")
 
-	cl := dc.GetHTTPClient(d8http.WithInsecureSkipVerify())
-
 	// curl -s --connect-timeout 10 --max-time 10 -k -XGET -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" "https://prometheus.d8-monitoring:9090/api/v1/query?query=resource_versions_compatibility"
-	promURL := "https://prometheus.d8-monitoring:9090/api/v1/query?query=resource_versions_compatibility"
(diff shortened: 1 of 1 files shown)


modules/013-helm/hooks/startup_metrics.go
```diff
@@ -42,29 +42,11 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
...
-		return err
-	}
-	defer res.Body.Close()
-
-	var response promMetrics
-	err = json.NewDecoder(res.Body).Decode(&response)
+	response, err := makePromRequest(dc)
 	if err != nil {
-		return err
+		input.LogEntry.Warnf("Prometheus request for previous metrics failed: %s", err)
```
nabokihms, line 48: An error is returned if caused by a failure to speak HTTP (such as a network connectivity problem). It would be best to differentiate somehow whether Prometheus is not working or it is not in the cluster.

modules/013-helm/hooks/startup_metrics.go
```diff
@@ -91,6 +73,34 @@ func handleDeprecatedAPIStartup(input *go_hook.HookInput, dc dependency.Containe
 	return nil
 }
 
+func makePromRequest(dc dependency.Container) (*promMetrics, error) {
+	cl := dc.GetHTTPClient(d8http.WithInsecureSkipVerify())
+
+	promURL := "https://prometheus.d8-monitoring:9090/api/v1/query?query=resource_versions_compatibility"
```
nabokihms, line 79: NIT: Let's move the comment here.

modules/013-helm/hooks/startup_metrics.go
yalosev: Don't think we need this case. With any problems with prometheus we have to skip this hook (cause it's not life-important) and continue the execution of Deckhouse

modules/013-helm/hooks/startup_metrics.go
yalosev: which one? about curl?

PR #1524, opened 2022-05-05
[upmeter] Fix missing init container with migrations for emptyDir

shvgn:
## Description

Upmeter server does not work with emptydirs because migrations init container is skipped on emptyDir

Upmeter does not work on emptydirs

## Changelog entries

```changes
section: upmeter
type: fix
summary: Fix non-working upmeter server on emptyDirs
```


konstantin-axenov (approved):

nabokihms (approved):

z9r5:

PR #1528, opened 2022-05-05
Add APIToken support for Cloudflare

yalosev:

## Changelog entries

```changes
type: fix | feature | chore
```

### DIFF
--- a/modules/101-cert-manager/templates/cert-manager/clusterissuer-cloudflare.yaml
+++ b/modules/101-cert-manager/templates/cert-manager/clusterissuer-cloudflare.yaml
@@ -1,4 +1,4 @@
-{{- if or (hasKey .Values.certManager "cloudflareGlobalAPIKey") (hasKey .Values.certManager "cloudflareEmail") }}
+{{- if or (hasKey .Values.certManager "cloudflareGlobalAPIKey") (hasKey .Values.certManager "cloudflareAPIToken") }}
 ---
 apiVersion: cert-manager.io/v1
 kind: ClusterIssuer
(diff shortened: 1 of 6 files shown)


modules/101-cert-manager/templates/cert-manager/clusterissuer-cloudflare.yaml
```diff
@@ -20,7 +20,14 @@ spec:
     - dns01:
         cloudflare:
           email: "{{ .Values.certManager.cloudflareEmail }}"
+          {{- if and (hasKey .Values.certManager "cloudflareGlobalAPIKey") (.Values.certManager.cloudflareGlobalAPIKey) }}
           apiKeySecretRef:
             name: cloudflare
             key: global-api-key
+          {{- end }}
+          {{- if and (hasKey .Values.certManager "cloudflareAPIToken") (.Values.certManager.cloudflareAPIToken) }}
+          apiTokenSecretRef:
```
nabokihms, line 29: How does it work if a user provides both keys? I assume cert-manager uses api-token first and then falls back to the global API key, but it is just an assumption.

nabokihms: Is it possible to validate it using openapi module spec? (in case someone wants to specify both keys in config)

yalosev: I don't see the way. Because openapi spec for cert-manager contains a lot of parameters for other providers. We can't use anyOf/oneOf here. Maybe I'm missing something

nabokihms (approved):

PR #1539, opened 2022-05-06
[candi] fix aws terraform lookup bastion_instance

RomanenkoDenys:
## Description

Fixed use of lookup function in aws terraform for bastion_instance.
## Changelog entries

When bastion_instance is not set, we set it to `null` instead of `{}`, so upcoming lookup function fails with message `Invalid value for "inputMap" parameter: argument must not be null.`
```changes
section: candi
type: fix
summary:  fix aws terraform lookup bastion_instance
impact_level: low
```


z9r5:

PR #1547, opened 2022-05-11
[docs] Revert dont-use-public-control-plane-images parameter for closed envs getting started and add /install for dhctl image for closed envs getting started

name212:
## Description
Revert dont-use-public-control-plane-images parameter for closed envs getting started and add /install for dhctl image for closed envs getting started

Getting started is not versioned and we removed deprecate `dont-use-public-control-plane-images` in next version.
For closed envs we have incorrect dhctl image.


deckhouse-BOaTswain:

PR #1572, opened 2022-05-13
[dhctl] Update node template on converge

name212:
## Changelog entries

```changes
type: fix | feature | chore
```

### DIFF
--- a/dhctl/pkg/kubernetes/actions/converge/converge.go
+++ b/dhctl/pkg/kubernetes/actions/converge/converge.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/google/go-cmp/cmp"
+
 	"github.com/hashicorp/go-multierror"
 
 	"github.com/deckhouse/deckhouse/dhctl/pkg/config"
@@ -166,7 +168,10 @@ func (r *Runner) converge() error {
 	}
 
 	var nodeGroupsWithStateInCluster []string
+	nodeGroupTemplates := make(map[string]map[string]interface{})
+
 	for _, group := range terraNodeGroups {
+		nodeGroupTemplates[group.Name] = group.NodeTemplate
 		// Skip if node group terraform state exists, we will update node group state below
 		if _, ok := nodesState[group.Name]; ok {
 			nodeGroupsWithStateInCluster = append(nodeGroupsWithStateInCluster, group.Name)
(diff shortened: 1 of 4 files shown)


dhctl/pkg/kubernetes/actions/converge/node.go
```diff
@@ -76,6 +76,56 @@ func GetCloudConfig(kubeCl *client.KubernetesClient, nodeGroupName string) (stri
...
+	if err != nil {
+		return nil, err
+	}
+
+	return res, nil
+}
+
+func UpdateNodeTemplate(kubeCl *client.KubernetesClient, nodeGroupName string, template map[string]interface{}) error {
+	return retry.NewLoop(fmt.Sprintf("Update node template in NodeGroup %q", nodeGroupName), 45, 15*time.Second).Run(func() error {
+		doc, err := kubeCl.Dynamic().
```
nabokihms, line 108: You get node group manifest twice. The first is to get the node template section, and the second is for the update operation.

I believe there is a logical bug.

1. dhctl gets node template from the cluster
2. dhctl shows a diff to the user and asks for confirmation
3. Someone changes the node template in the cluster
4. User confirms node template updating and wipes the previously added date

If you get a node group from the cluster only once, you will not face this bug due to resource version inconsistency, which is the correct behavior from my point of view.

dhctl/pkg/kubernetes/actions/deckhouse/pod.go
```diff
@@ -31,7 +31,7 @@ func GetPod(kubeCl *client.KubernetesClient) (*v1.Pod, error) {
 	}
 
 	if len(pods.Items) != 1 {
-		return nil, ErrListPods
+		return nil, ErrMultiplePodsFound
```
nabokihms, line 34: This is not true. The error will also be shown if there is no Deckhouse pod in the cluster (according to the clause).

nabokihms (changes requested):

dhctl/pkg/kubernetes/actions/converge/node.go
name212: Good point! Thank you. I fixed it.

dhctl/pkg/kubernetes/actions/converge/converge.go
```diff
@@ -610,23 +639,20 @@ func (c *NodeGroupController) tryDeleteNodeGroup(nodeGroup *NodeGroupGroupOption
 		return nil
 	}
 
-	groupInConfig := false
+	return log.Process("converge", fmt.Sprintf("Delete NodeGroup %s", c.name), func() error {
```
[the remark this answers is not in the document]
name212: It is ok. In previous version we can delete node group only. Now, we can update node group if it exists and delete if not. I moved it [here](https://github.com/deckhouse/deckhouse/pull/1572/files#diff-5bf74186d54de5b23e4ee7b3915fb9708050c068098b9e8c81cc47d5d21f36f5R408)

PR #592, opened 2022-01-14
[cni-cilium] New module

zuzzas:
## Why do we need it, and what problem does it solve?

Fixes #206.

## Changelog entries

```changes
module: cni-cilium
type: feature
description: Cilium CNI
note: Without a way to migrate from existing CNIs at this moment.
```

PR #1575, opened 2022-05-16
[cni-cilium] Fixed value modifying hook

zuzzas:
## Description

Fixed a hook so it won't ever return nil.

We rely on this value in Helm templates: `Error: helm upgrade failed: error validating \"\": error validating data: unknown object type \"nil\" in ConfigMap.data.auto-direct-node-routes\n"`.

The error was introduced via a last-minute change in the original Cilium PR that [removed](https://github.com/deckhouse/deckhouse/pull/592#discussion_r869625527) defaulting in OpenAPI.

## Changelog entries

```changes
section: cni-cilium
type: fix
summary: Fixed value modifying hook so that it won't ever leave a value as nil
impact_level: low
```

### DIFF
--- a/modules/021-cni-cilium/hooks/enable_node_routes.go
+++ b/modules/021-cni-cilium/hooks/enable_node_routes.go
@@ -31,23 +31,27 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 }, enableNodeRoutes)
 
 func enableNodeRoutes(input *go_hook.HookInput) error {
+	input.Values.Set("cniCilium.createNodeRoutes", shouldEnableNodeRoutes(input))
+
+	return nil
+}
+
+func shouldEnableNodeRoutes(input *go_hook.HookInput) bool {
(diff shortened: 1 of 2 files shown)


modules/021-cni-cilium/hooks/enable_node_routes.go
```diff
@@ -31,23 +31,31 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 }, enableNodeRoutes)
 
 func enableNodeRoutes(input *go_hook.HookInput) error {
+	if shouldEnableNodeRoutes(input) {
```
nabokihms, line 34:
```go
input.Values.Set("cniCilium.createNodeRoutes", shouldEnableNodeRoutes(input))
```

zuzzas: :heavy_check_mark:

nabokihms (approved):

PR #1579, opened 2022-05-16
[upmeter] Remote write exporter: don't retry sending metrics on 4xx errors from storage

shvgn:
## Description

Fix potential error looop in remote write exporter. Add minor enhancements.

Fixes potential bug in telemetry collection.

## Changelog entries

```changes
section: upmeter
type: chore
summary: Switched the logging of upmeter metrics on info level while exporting.
---
section: upmeter
type: fix
summary: Fixed potential error loops in remote write exporter
impact: If a storage responds with 4xx error, the unaccepted metrics will not be re-sent.
---
section: upmeter
type: fix
summary: 'Added missing User-Agent header to remote write exporter, defined as `Upmeter/1.0 (Deckhouse <edition> <version>)`'
```

### DIFF
--- a/modules/500-upmeter/templates/upmeter/statefulset.yaml
+++ b/modules/500-upmeter/templates/upmeter/statefulset.yaml
@@ -101,6 +101,7 @@ spec:
           {{- range $probeRef := .Values.upmeter.internal.disabledProbes }}
           - --disable-probe={{ $probeRef }}
           {{- end }}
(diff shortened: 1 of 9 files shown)


modules/500-upmeter/templates/upmeter/statefulset.yaml
```diff
@@ -109,7 +110,7 @@ spec:
           - name: UPMETER_LISTEN_PORT
             value: "8091"
           - name: LOG_LEVEL
-            value: "info"
+            value: "debug"
```
shvgn, line 113:
```suggestion
            value: "info"
```

PR #1586, opened 2022-05-19
[prometheus] Add token auth for remote write

nabokihms:

## Description
Add two options - bearerToken (for everyone) and customAuthToken (for okmeter)

Only basic auth is allowed right now, which is not flexible enough. 

## Changelog entries

```changes
section: prometheus
type: feature
summary: Add token auth for Prometheus remote write
```

### DIFF
--- a/modules/300-prometheus/crds/prometheusremotewrite.yaml
+++ b/modules/300-prometheus/crds/prometheusremotewrite.yaml
@@ -49,6 +49,12 @@ spec:
                   - password
                   - username
                   type: object
+                bearerToken:
+                  type: string
+                  description: Bearer token.
+                customAuthToken:
+                  type: string
+                  description: Custom token sent as a value of the X-Auth-Token header.
                 writeRelabelConfigs:
                   description: |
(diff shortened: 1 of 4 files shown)


deckhouse-BOaTswain:

nabokihms:
@konstantin-axenov, according to job logs, there is no problem with the Prometheus module, so I consider e2e tests passed.

modules/300-prometheus/crds/prometheusremotewrite.yaml
```diff
@@ -49,6 +49,12 @@ spec:
                   - password
                   - username
                   type: object
+                bearerToken:
+                  type: string
+                  description: Bearer token.
+                customAuthToken:
```
konstantin-axenov, line 55: Why not xAuthToken?

nabokihms: Thought that we may change the exact header in the future. Using header name is kind of an exposing  implementation details, which I want to avoid.

PR #1598, opened 2022-05-22
[cloud-provider-azure] type -> skuName

zuzzas:
## Description

Fix disk type parameter name.

In the latest (which is an all releases now) azuredisk-csi driver these fields are validated and no new PV can get provisioned due to incorrect value being present. It previously defaulted to `StandardSSD_LRS` and did not throw errors.

## Changelog entries

```changes
section: cloud-provider-azure
type: fix
summary: Fix parameter name `type` -> `skuName`
impact_level: low
```


konstantin-axenov (approved):

z9r5:

PR #1599, opened 2022-05-22
[testing] Add linter for markdown files

z9r5:
## Description
Added markdown linter (https://github.com/markdownlint/markdownlint).

Linter will help to avoid common markup problems and will reduce cases with unexpected results in the process of rendering documentation to HTML.

## Changelog entries
```changes
section: testing
type: feature
summary: Add linter for markdown files.
```


nabokihms (commented):
Hello Artem. There are a couple of concerns regarding this issue.
1. Why do we need a skip label? Are there any problems with the linter? As I see it, the result of this linter should be undeniable as for golangci-lint.
2. How can I run this linter locally? It has a convenient `--fix` flag to automatically deal with some style problems, and it would be best to give this ability to developers.
3. What prevents us from running linter for all `.md` files? Why do we limit it to the PR diff files only?

z9r5:
> Hello Artem. There are a couple of concerns regarding this issue.
> 
> 1. Why do we need a skip label? Are there any problems with the linter? As I see it, the result of this linter should be undeniable as for golangci-lint.

I think that ability to skip the test is required for a while until we are sure that the linter does not give false positive warnings (I saw some issues).

> 2. How can I run this linter locally? It has a convenient `--fix` flag to automatically deal with some style problems, and it would be best to give this ability to developers.
Nice idea. I will add such instruction.

> 3. What prevents us from running linter for all `.md` files? Why do we limit it to the PR diff files only?
Good question... There are MD-files in the repo, which are not rendered - internal docs, files included in modules images. We can fix such files, but it can lead to rebuilding some images.
But, we can check all MD-files e.g. only in docs/folder (and maybe in some other paths). Should I implement it?

z9r5:
- [Failed test example](https://github.com/deckhouse/deckhouse-test-1/runs/6904770038?check_suite_focus=true)
- [Successful test example](https://github.com/deckhouse/deckhouse-test-1/runs/6904798723?check_suite_focus=true)

Changes:
- Added instructions about how to run the linter locally.
- Linter works with all markdown files except those defined in the `testing/.markdownlintignore` file.
- Tune rules.

z9r5:
Changes:
- Added into the /Makefile the `lint-markdown` and the `lint-markdown-fix` commands.
- CI runs make lint-markdown

[Example of a report](https://github.com/deckhouse/deckhouse-test-1/runs/6933273193).

Makefile
```diff
@@ -52,6 +54,16 @@ lint: bin/golangci-lint ## Run linter.
 lint-fix: bin/golangci-lint ## Fix lint violations.
 	bin/golangci-lint run --fix
 
+lint-markdown: ## Run markdown linter.
+	docker pull -q ${MDLINTER_IMAGE}
+	docker run --rm -v ${PWD}:/workdir ${MDLINTER_IMAGE} \
+		--config testing/markdownlint.yaml -p testing/.markdownlintignore "**/*.md"
+
+lint-markdown-fix: ## Run markdown linter and fix problems automatically.
+	docker pull -q ${MDLINTER_IMAGE}
```
nabokihms, line 63: `--pull-always` flag for the `run` command should be enough

.github/workflows/validation.yml
```diff
@@ -330,3 +332,32 @@ jobs:
...
+        uses: actions/download-artifact@v2
+        with:
+          name: pr_diff
+
+      - name: Run check
+        env:
+          DIFF_PATH: ./pr.diff
+          SKIP_LABEL_NAME: ${{ needs.discover.outputs.label_markdown }}
+        run: |
+          ./.github/scripts/validation_run.sh ./testing/validate_markdown.sh
```
nabokihms, line 363: 1. I think having a dedicated script for CI is excessive.
2. Why do we need to use the `validation_run.sh` here? I thought we agreed to run a markdown linter for every file every time. How long does it take to run the validation for all files?

Makefile
z9r5: A separate `docker pull` is here to hide the output of the pulling image process (to avoid printing it in the report). 

Maybe another solution exists...

.github/workflows/validation.yml
z9r5: Changes:
- Removed `validation_run.sh` script
- Added workflow template for linters

Linter passes about 3-4 seconds over a whole repo.

Examples:
- [Successful job example](https://github.com/deckhouse/deckhouse-test-1/runs/6971723727?check_suite_focus=true)
- [Failed job example](https://github.com/deckhouse/deckhouse-test-1/runs/6971690079?check_suite_focus=true)

@nabokihms

nabokihms: I think yes.

PR #1609, opened 2022-05-23
Revert "[monitoring-kubernetes] Enable systemd collector in the node-exporter (#768)"

zuzzas:
## Description

Completely disabled systemd collector.

node-exporter's systemd-collector does not work correctly without adding a new mount inside node-exporter container. It also generates tons of high cardinality metrics that should be narrowed down by providing a regex of monitored systemd Units to the node-exporter. All this work has been done in https://github.com/deckhouse/deckhouse/pull/1587, but since there is no push from customers to use these feature, we'll disable it.

## Changelog entries

```changes
type: fix
summary: Disabled node-exporter's systemd collector. It was not working correctly, so no one is dependent on it.
impact_level: low
```


nabokihms (approved):

z9r5:

PR #1610, opened 2022-05-23
[docs] Fix rendering CRD file with a dot in the name

z9r5:
## Description
Jekyll sanitizes the data filename, so we must delete dots from the file name to access data.

If the CRD file contains a dot in the name, it doesn't render

## Changelog entries
```changes
section: docs
type: fix
summary: Fixed rendering CRD file with a dot in the name.
impact_level: low
```


z9r5:

PR #1627, opened 2022-05-25
[prometheus] fix: rename errors related metrics

diafour:
## Description

Use new metrics names in alert rules.

Errors related metrics was renamed in https://github.com/flant/addon-operator/pull/133

## Changelog entries

```changes
section: prometheus
type: fix
summary: Use new metrics names in alert rules.
impact: Re-enable alerts on hook errors.
impact_level: default
```


nabokihms (approved):

nabokihms:
@diafour, how can we prevent these errors in the future?

Issue #1425, opened 2022-04-19
[ingress-nginx] 0.49 exports metrics with weird vhost label for wildcard ingresses


apolovov:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.31

### Expected Behavior

The vhost label should be `.example.com` for Ingress vhost `*.example.com`.

### Actual Behavior

The vhost label looks like `~^(?<subdomain>[\w-]+)\.example\.com$`.

### Steps To Reproduce

Create Ingress with wildcard vhost, make some requests and fetch the `ingress_nginx_detail_requests_total` metric.

PR #1630, opened 2022-05-26
[ingress-nginx] Fix wildcard vhost metrics

apolovov:
## Description
Fix wildcard vhost label in ingress-controller metrics

Fixes https://github.com/deckhouse/deckhouse/issues/1425.

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: Fix wildcard vhost label in ingress-controller metrics
impact: Ingress controller pods will be restarted
impact_level: high
```

### DIFF
--- a/modules/402-ingress-nginx/images/controller-0-46/rootfs/etc/nginx/lua/pbmetrics.lua
+++ b/modules/402-ingress-nginx/images/controller-0-46/rootfs/etc/nginx/lua/pbmetrics.lua
@@ -207,7 +207,7 @@ local function fill_buffer()
   ngx.var.total_upstream_response_time = 0
   ngx.var.upstream_retries = 0
 
-  local var_server_name = ngx.var.server_name:gsub("^*", "")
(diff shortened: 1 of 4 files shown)


modules/402-ingress-nginx/images/controller-0-46/rootfs/etc/nginx/lua/pbmetrics.lua
```diff
@@ -207,7 +207,7 @@ local function fill_buffer()
   ngx.var.total_upstream_response_time = 0
   ngx.var.upstream_retries = 0
 
-  local var_server_name = ngx.var.server_name:gsub("^*", "")
+  local var_server_name = ngx.var.server_name:gsub("~^%(%?<subdomain>%[\\w%-]%+%)", ""):gsub("^*", ""):gsub("$$", ""):gsub("\\", "")
```
nabokihms, line 210: Let's leave a comment why we added such gsub.

nabokihms (changes requested):
One little thing left. I also think that impact is not low, because ingress controller pods will be restarted.

PR #1648, opened 2022-05-27
[Upmeter] Add alerts for probe garbage

shvgn:
## Description

Add alerts for garbage left by upmeter probes.

We should clean the garbage when it cannot be deleted. Moreover, upmeter agent does not delete
garbage by labels, it only tracks objects by name within single run.

## Changelog entries

```changes
section: upmeter
type: chore
summary: Added alerts for garbage objects left by probes
```

### DIFF
--- a/modules/500-upmeter/monitoring/prometheus-rules/alerting.yaml
+++ b/modules/500-upmeter/monitoring/prometheus-rules/alerting.yaml
@@ -166,3 +166,269 @@
 
           If you have no disk provisioning system in the cluster,
           you can disable ordering volumes for the some-mini through the module settings.
+
+- name: d8.upmeter.resources
+  rules:
+    - alert: D8UpmeterProbeGarbageConfigmap
+      expr: |
+        (
+          count (kube_configmap_info{namespace="d8-upmeter", configmap=~"upmeter-basic-.*"})
+          /
+          count (kube_pod_labels{namespace="d8-upmeter", label_app="upmeter-agent"})
+        ) >= 2
+      for: 10m
+      labels:
+        severity_level: "9"
+        tier: cluster
+        d8_module: upmeter
+        d8_component: upmeter-agent
+      annotations:
+        plk_protocol_version: "1"
+        plk_markup_format: "markdown"
+        plk_create_group_if_not_exists__d8_upmeter_resources_garbage: "D8UpmeterProbeGarbage,tier=cluster,prometheus=deckhouse"
+        plk_grouped_by__d8_upmeter_resources_garbage: "D8UpmeterProbeGarbage,tier=cluster,prometheus=deckhouse"
+        plk_labels_as_annotations: "configmap"
+        summary: Garbage produced by basic probe is not being cleaned.
+        description: |
+          Probe configmaps found.
+
+          Upmeter agents should clean ConfigMaps produced by control-plane/basic probe. There should not be more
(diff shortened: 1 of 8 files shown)


modules/500-upmeter/monitoring/prometheus-rules/alerting.yaml
```diff
@@ -215,3 +215,258 @@
...
+          2. Check that control plane is functional.
+
+          3. Delete configmaps manually:
+
+          `kubectl -n d8-upmeter delete cm -l heritage=upmeter`
+
+    - alert: D8UpmeterProbeGarbageDeployment
+      expr: |
+        (
+          min_over_time  (count (kube_deployment_labels{namespace="d8-upmeter",
```
nabokihms, line 262: Why do you use `_over_time` queries? `min` and `avg` should be enough for this case, as far as I can tell. Over time functions are more resource-consuming.

nabokihms, line 262: Why do you use `_over_time` queries?  `min` and `avg` should be enough for this case, as far as I can tell.

modules/500-upmeter/monitoring/prometheus-rules/alerting.yaml
```diff
@@ -215,3 +215,258 @@
...
+
+- name: d8.upmeter.resources
+  rules:
+    - alert: D8UpmeterProbeGarbageConfigmap
+      expr: |
+        (
+          min_over_time  (count (kube_configmap_info{namespace="d8-upmeter", configmap=~"upmeter-basic-.*"})[5m:])
+          /
+          avg_over_time  (count (kube_pod_labels{namespace="d8-upmeter", label_app="upmeter-agent"})[5m:])
+        ) >= 1
```
nabokihms, line 227: I would simplify the rule to make it less complicated and avoid false positives. Let's say 15 configmaps in the namespaces is a problem.  Just sharing my thoughts, you do not need to change the current logic.

nabokihms, line 227: I would simplify the rule to make it less complicated and avoid false positives. Let's say 15 configmaps in the namespaces is a problem.  Just sharing my thoughts, you do not need to change the current logic.

modules/500-upmeter/monitoring/prometheus-rules/alerting.yaml
```diff
@@ -215,3 +215,258 @@
         summary: Smoke-mini is unavailable
         description: |
           Smoke-mini is unavailable. Please check the linked alerts to find the real problem.
+
+- name: d8.upmeter.resources
+  rules:
+    - alert: D8UpmeterProbeGarbageConfigmap
+      expr: |
```
nabokihms, line 222: Let's set the `for` at least for 5 minutes or more. We do not need to react if configmaps number is back to normal.

nabokihms, line 222: Let's set the `for` value for at least 5 minutes or more. We do not need to react if configmaps number is back to normal on its own.

nabokihms (changes requested):
Overall, everything looks good. I left a couple of suggestions. Please look through them when you have time.

shvgn:
Simplified, added `for` and groups auto-creation

nabokihms (approved):

PR #1661, opened 2022-05-30
[deckhouse-controller] fix: restore hook metrics

diafour:
## Description

Bring fix from shell-operator to restore separate metrics storage for hooks.
- https://github.com/flant/shell-operator/pull/388
- https://github.com/flant/addon-operator/pull/323

No metrics from hooks when separate port for hook metrics is in use after applying https://github.com/flant/shell-operator/pull/388 and https://github.com/flant/addon-operator/pull/315.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary:  Restore separate metrics storage for hooks.
impact_level: low
```


shvgn (approved):
LFGTM

PR #1666, opened 2022-05-30
[ci] GolangCI Linter upgrade to 1.46.2

apolovov:
## Description
GolangCI Linter upgrade to 1.46.2

The 1.42.0 version is no longer available, so we need to upgrade.

## Changelog entries

```changes
section: ci
type: chore
summary: GolangCI Linter upgrade to 1.46.2
impact_level: low
```

### DIFF
--- a/ee/modules/600-flant-integration/hooks/license_test.go
+++ b/ee/modules/600-flant-integration/hooks/license_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 Flant JSC
+Copyright 2021 Flant JSC
 Licensed under the Deckhouse Platform Enterprise Edition (EE) license. See https://github.com/deckhouse/deckhouse/blob/main/ee/LICENSE
 */
 
--- a/ee/modules/600-flant-integration/hooks/license_test.go
+++ b/ee/modules/600-flant-integration/hooks/license_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 Flant JSC
+Copyright 2022 Flant JSC
 Licensed under the Deckhouse Platform Enterprise Edition (EE) license. See https://github.com/deckhouse/deckhouse/blob/main/ee/LICENSE
 */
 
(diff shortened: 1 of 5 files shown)


testing/library/sandbox_runner/runner.go
```diff
@@ -57,15 +56,13 @@ func Run(cmd *exec.Cmd, opts ...SandboxOption) *gexec.Session {
 
 func WithFile(path string, contents []byte, envOpts ...EnvOption) SandboxOption {
 	return func(conf sandboxConfig) error {
-		filePath := filepath.Join(path)
-
-		err := ioutil.WriteFile(filePath, contents, os.FileMode(0644))
+		err := ioutil.WriteFile(path, contents, os.FileMode(0644))
```
yalosev, line 59: I guess `filepath.Join(path)` was made to remove trailing slash if it exists. (https://go.dev/play/p/YsAOH3XYOAO)
Maybe we should keep this behavior in a way like

```
path = strings.TrimSpace(path)
if strings.HasSuffix(path, "/") {
    path = strings.TrimSuffix(path, "/")
}
```

yalosev (approved):

shvgn (approved):

shvgn (dismissed):

yalosev (dismissed):

ee/modules/600-flant-integration/hooks/license_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 Flant JSC
+Copyright 2022 Flant JSC
```
konstantin-axenov, line 2: We don't need to change it. This is the date of creation of this file. [Example](https://github.com/kubernetes/kubernetes/blob/aae07c6f7bb715750a2d0cc77cf81c4e31a4904b/pkg/apis/core/register.go#L2) from Kubernetes repo.

modules/040-node-manager/hooks/discover_standby_ng_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 Flant JSC
+Copyright 2022 Flant JSC
```
apolovov, line 2:
```suggestion
Copyright 2021 Flant JSC
```

testing/library/sandbox_runner/runner.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 Flant JSC
+Copyright 2022 Flant JSC
```
apolovov, line 2:
```suggestion
Copyright 2021 Flant JSC
```

yalosev (approved):

PR #1678, opened 2022-06-01
[cilium] Alerts and policy map limit

zuzzas:
## Description

Alerts and raising policy map limit.

Basic alerts that notify user when cilium agent is not working properly.

Policy map is the largest contributor to eBPF map load. It seems reasonable to increase it preemptively.

## Changelog entries

```changes
section: cni-cilium
type: feature
summary: Raised policy map limit to four times the default. Also added alert for Cilium agent. 
impact_level: low
```

### DIFF
--- a/modules/021-cni-cilium/monitoring/prometheus-rules/agent.yaml
+++ b/modules/021-cni-cilium/monitoring/prometheus-rules/agent.yaml
@@ -1,0 +1,95 @@
+- name: cni-cilium
+  rules:
+  - alert: CiliumAgentControllersRunsFailing
+    expr: max by (pod,namespace) (cilium_controllers_failing) > 0
+    for: 15m
+    labels:
+      d8_module: cni-cilium
+      d8_component: agent
+      tier: cluster
+    annotations:
+      plk_protocol_version: "1"
+      plk_markup_format: "markdown"
+      plk_grouped_by__main: CiliumAgentMalfunctioning,prometheus=deckhouse,tier=~tier
+      plk_create_group_if_not_exists__main: CiliumAgentMalfunctioning,prometheus=deckhouse,tier=~tier
+      summary: Some controllers (internal logical loops) are failing in agent `{{`{{ $labels.namespace }}`}}`/`{{`{{ $labels.pod }}`}}`
+      description: |
+        Check what's going on: `kubectl -n `{{`{{ $labels.namespace }}`}}` logs `{{`{{ $labels.pod }}`}}``
+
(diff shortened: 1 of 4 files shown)


modules/021-cni-cilium/templates/configmap.yaml
```diff
@@ -78,6 +80,7 @@ data:
   enable-remote-node-identity: "true"
 
   bpf-map-dynamic-size-ratio: "0.005"
+  bpf-policy-map-max: "32768"
```
RomanenkoDenys, line 83: We reach the CiliumAgentMapPressureCritical warning - so what do we do? Should this option be configurable?

RomanenkoDenys (approved):

modules/021-cni-cilium/monitoring/prometheus-rules/agent.yaml
```diff
@@ -0,0 +1,89 @@
...
+    for: 5m
+    labels:
+      d8_module: cni-cilium
+      d8_component: agent
+    annotations:
+      plk_protocol_version: "1"
+      plk_markup_format: "markdown"
+      plk_grouped_by__main: CiliumAgentMalfunctioning,prometheus=deckhouse,tier=~tier
+      plk_create_group_if_not_exists__main: CiliumAgentMalfunctioning,prometheus=deckhouse,tier=~tier
+      summary: Some nodes' health endpoints are not reachable by agent `{{`{{ $labels.namespace }}`}}`/`{{`{{ $labels.pod }}`}}`.
```
konstantin-axenov, line 44:
```suggestion
      summary: Some nodes health endpoints are not reachable by agent `{{`{{ $labels.namespace }}`}}`/`{{`{{ $labels.pod }}`}}`.
```

zuzzas: That is not a mistake. It's plural.

image

modules/021-cni-cilium/monitoring/prometheus-rules/agent.yaml
```diff
@@ -0,0 +1,89 @@
+- name: cni-cilium
+  rules:
+  - alert: CiliumAgentControllersRunsNotSucceeding
+    expr: max by (pod,namespace) (cilium_controllers_failing) > 0
+    for: 15m
+    labels:
+      d8_module: cni-cilium
+      d8_component: agent
```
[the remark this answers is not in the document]
zuzzas: Fixed.

PR #1697, opened 2022-06-02
[ingress-nginx] remove requirements

yalosev:
## Description
Remove requirement for ingress nginx

This param should be come in the release 1.33.0. It's a mistake

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: Updated requirements for Ingress Nginx controller versions. If conditions are not met, then further Deckhouse upgrade is blocked.
```


z9r5:

PR #1721, opened 2022-06-06
[prometheus] rollback disk retention size to 80%

vitaliy-sn:

## Description

Set disk retention size to 80%.

When small disks are used in clusters with a large number of metrics, disk cleanup does not have time to complete and space runs out.

## Changelog entries

```changes
section: prometheus
type: fix
summary: Set disk retention size to 80%
```


z9r5:

PR #1742, opened 2022-06-09
[linstor] Add failed cases when drbd module cannot be loaded

kvaps:
## Description

Describe some failed cases when linstor-node can't load drbd module.

Sometimes users could face with problems of loading DRBD module.
This doc trying to cover some of those cases

## Changelog entries

```changes
section: linstor
type: fix
summary: Add failed cases when drbd module cannot be loaded
impact_level: low
```


deckhouse-BOaTswain:

PR #1754, opened 2022-06-09
[helm] log broken release instead of failing

yalosev:

## Description
Don't fail hook on broken releases

Users can have helm releases with incorrect manifests. We shouldn't fail on them, just log it

## Changelog entries

```changes
section: helm
type: fix 
summary: Avoid failing on incorrect helm releases
```


z9r5:
Cherry-picked into release-1.33.

PR #1757, opened 2022-06-09
[ingress-nginx] add patch apply to dockerfile

yalosev:

## Description
Forgot to add patch to Dockerfile

It doesn' work :)

## Changelog entries

```changes
section: ingress-nginx
type: fix 
summary: Fix build of the ingress-nginx 0.33 controller.
```


nabokihms (approved):

z9r5:
Cherry-picked into release-1.33.

PR #1764, opened 2022-06-10
[openvpn] fix set_search_domains hook

vitaliy-sn:

## Description

Added check for the existence of the pushToClientSearchDomains parameter in the config.

This fixes a bug that does not allow you to set additional domains.

## Changelog entries

```changes
section: openvpn
type: fix
summary: Fixes a bug that does not allow you to set additional domains.
```


nabokihms (approved):

z9r5:

PR #1789, opened 2022-06-15
[testing] Add revision history limit lint rule

nabokihms:

Reduce the amount of data stored in the etcd by limiting the number of stored replicasets for each deployment to 2 (default limit is 10)

## Changelog entries

```changes
section: testing
type: feature
summary: Force to use revisionHistoryLimit<=2 for all deployments controlled by Deckhouse
impact_level: low
```


z9r5 (approved):

zuzzas (dismissed):

yalosev (approved):
I would like 3 revisions but never mind

kvaps (approved):

ee/fe/modules/500-basic-auth/templates/deployment.yaml
```diff
@@ -23,6 +23,7 @@ metadata:
   {{- include "helm_lib_module_labels" (list . (dict "app" .Chart.Name)) | nindent 2 }}
 spec:
   {{- include "helm_lib_deployment_strategy_and_replicas_for_ha" . | nindent 2 }}
+  revisionHistoryLimit: 2
```
shvgn, line 26: I see we have 2 as default. I would write a helm line for that, so "2" could be defined only once. And when it changes again, it could produce less diff. With an optional argument, 2 could be overwritten where needed.

testing/matrix/linter/rules/rules.go
```diff
@@ -339,6 +340,46 @@ func newConvertError(object storage.StoreObject, err error) errors.LintRuleError
...
+			return newConvertError(object, err)
+		}
+
+		// https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#revision-history-limit
+		// Revision history limit controls the number of replicasets stored in the cluster for each deployment.
+		// Higher number means higher resource consumption, lower means inability to rollback.
+		//
+		// Since Deckhouse does not use rollback, we can set it to 2 to be able to manually check the previous version.
+		// It is more important to reduce the control plane pressure.
+		idealHistoryLimit := int32(2)
```
RomanenkoDenys, line 359: desiredHistoryLimit ?

ee/fe/modules/500-basic-auth/templates/deployment.yaml
RomanenkoDenys: good point, i think we need helm lib helper for this

RomanenkoDenys (changes requested):

testing/matrix/linter/rules/rules.go
nabokihms: I picked this term because I had seen this word in the doc
>  By default, 10 old ReplicaSets will be kept, however its ideal value depends on the frequency and stability of new Deployments.

It is ok to change the name, but it is not a `desired` value. It is the maximum allowed value, thus I will use the term `max` there.

ee/fe/modules/500-basic-auth/templates/deployment.yaml
nabokihms: Ok, I can rewrite the code using a helper, but I will explain my thoughts first. 

Key points:
1. If we define the helper, it should be used everywhere in the code to keep the same style. It is terrible when we use helpers somewhere and somewhere not. For example:
```
{{ include "helm_lib_revision_history_limit" . }} # revisionHistoryLimit: 2
```
```
revisionHistoryLimit: 0
```

2. To write the helper with the ability to override the default, we still need to use a value. For example:
```
{{ include "helm_lib_revision_history_limit" . }} # revisionHistoryLimit: 2
```
```
{{ include "helm_lib_revision_history_limit" 0 }} # revisionHistoryLimit: 0
```
3. Splitting into two different helpers is also possible but seems like an overkill
```
{{ include "helm_lib_default_revision_history_limit" . }} # revisionHistoryLimit: 2
```
```
{{ include "helm_lib_revision_history_limit" 0 }} # revisionHistoryLimit: 0
```
4. Current variant.
```
revisionHistoryLimit: 2
```
```
revisionHistoryLimit: 0
```
I picked the last one because, for me, it looks less wordy and explicit enough.

nabokihms: Which one should I pick?

kvaps (approved):

ee/fe/modules/500-basic-auth/templates/deployment.yaml
shvgn: How likely this "2" is gonna change to e.g. 1 or 3?

shvgn: Probably my intention to save the diff is just not that important in case of 'find-and-replace one line' scenario. Templating one line looks like an overcomplication for no real value. Thanks for describing all cases. 

Do we want linter for this field to ensure nobody forgets to set 2 or less?

RomanenkoDenys (approved):

PR #1472, opened 2022-04-26
[deckhouse-controller] feat: fast response to ConfigMap changes

diafour:
## TODO

- [x] Update go.mod after merging https://github.com/flant/addon-operator/pull/315 and https://github.com/flant/shell-operator/pull/388

## Description

- Update addon-operator and shell-operator:
  - Faster response to ConfigMap changes.
  - Summary in queue list.
  - Purge and delete before run modules.
- Chore: remove unused registry checker code. It is a Go hook now.

## Why do we need it, and what problem does it solve?

Deckhouse should disable modules ASAP on ConfigMap changes.
 

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: disable modules ASAP on ConfigMap changes.
impact_level: low
```

PR #1790, opened 2022-06-15
[deckhouse-controller] fix: restore exponential backoff for failed hooks

diafour:
## Description

Restore exponential backoff for delays between failed hooks restarts.

Failed hook restarts every second after merging #1472.

See also: https://github.com/flant/shell-operator/pull/407 and https://github.com/flant/addon-operator/pull/326.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Restore exponential backoff for delays between failed hooks restarts.
impact_level: low
```


z9r5:

PR #1792, opened 2022-06-15
[cni-cilium] Fixed enable_node_routes hook

zuzzas:
## Description
The hook now bails if a value is present in a config.

Previously it erroneously set the value to `false` no matter what was set in config.

## Changelog entries
```changes
section: cni-cilium
type: fix
summary: The `enable_node_routes` hook now bails if a value is present in config.
```


z9r5:

PR #1795, opened 2022-06-15
[prometheus] Enable legacy alerting system for Grafana

nabokihms:

Grafana introduced their new alerting system, which is enabled by default for our Grafana version
https://grafana.com/docs/grafana/v8.5/alerting/

With the new alerting system, it is impossible to provision alert managers and add alerts to dashboards using Deckhouse custom resources. The only opportunity, for now, is to return back legacy alerts.

With this

## Changelog entries

```changes
section: prometheus
type: fix
summary: Enable legacy alerting system for Grafana
impact_level: default
```


z9r5 (approved):

z9r5:

PR #1559, opened 2022-05-12
Bump linstor components version

kvaps:
## Description
This PR updates LINSTOR and all components version. 

## Why do we need it, and what problem does it solve?

It includes many fixes to k8s storage backend driver, updated drbd module and scheduler-extender.

- https://github.com/LINBIT/linstor-server/releases/tag/v1.18.1
- https://github.com/piraeusdatastore/linstor-csi/releases/tag/v0.19.0
- https://github.com/LINBIT/drbd/releases/tag/drbd-9.1.7
- https://github.com/piraeusdatastore/linstor-scheduler-extender/pull/3

## Changelog entries

```changes
section: linstor
type: fix
summary: LINSTOR updated to 1.18.1, DRBD module to 9.1.7, linstor-csi to 0.19.0, linstor-scheduler to v0.3.0
impact_level: default
```

PR #1818, opened 2022-06-19
[linstor] fix monitoring of piraeus-operator

kvaps:
## Description

Remove `/metrics-cr` endpoint

There was a regression after updating piraeus-operator to 1.8 https://github.com/deckhouse/deckhouse/pull/1559
Piraeus-operator does not export `/metrics-cr` endpoint anymore.

## Changelog entries

```changes
section: linstor
type: fix
summary: fix monitoring of piraeus-operator
impact_level: low
```

### DIFF
--- a/modules/031-linstor/templates/piraeus-operator/deployment.yaml
+++ b/modules/031-linstor/templates/piraeus-operator/deployment.yaml
@@ -62,6 +62,16 @@ spec:
         imagePullPolicy: IfNotPresent
         args:
         - --create-monitoring=false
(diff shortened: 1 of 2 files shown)


nabokihms (changes requested):
I don't think that this PR fixes the problem. Speaking about the issue, we receive alerts that the target is down, so there should also be a change in the pod monitor for the piraeus-operator.

nabokihms (approved):

z9r5:

PR #1819, opened 2022-06-19
[candi] Automatic patchversions update makefile target

RomanenkoDenys:
## Description

Automatic patchversions update makefile target.

This PR adds Makefile targets to update versions_map file to latest versions of kubernetes control-plane components patchversions. Later we should add GA, to periodically run auto-update and generate PRs for updated versions.
## Changelog entries

```changes
section: candi
type: chore
summary: Automatic patchversions update makefile target.
impact_level: low
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,16 +13,15 @@ MDLINTER_IMAGE = ghcr.io/igorshubovych/markdownlint-cli@sha256:2e22b4979347f70e0
 # besides linux/amd64 (e.g. relevant for mac m1).
 PLATFORM_NAME := $(shell uname -p)
 OS_NAME := $(shell uname)
-LBITS := $(shell getconf LONG_BIT)
 ifneq ($(filter arm%,$(PLATFORM_NAME)),)
 	export WERF_PLATFORM=linux/amd64
 endif
 
 # Set platform for jq
 ifeq ($(OS_NAME), Linux)
-	JQ_PLATFORM = linux
+	JQ_PLATFORM = linux64
 else ifeq ($(OS_NAME), Darwin)
-	JQ_PLATFORM = osx-amd
+	JQ_PLATFORM = osx-amd64
 endif
 
 # Set platform for yq
@@ -174,10 +173,10 @@ docs-down: ## Stop all the documentation containers.
 ##@ Update kubernetes control-plane patchversions
 
 bin/jq: ## Install jq deps for update-patchversion script.
-	curl -sSfL https://github.com/stedolan/jq/releases/download/jq-1.6/jq-$(JQ_PLATFORM)$(LBITS) -o $(PWD)/bin/jq && chmod +x $(PWD)/bin/jq
+	curl -sSfL https://github.com/stedolan/jq/releases/download/jq-1.6/jq-$(JQ_PLATFORM) -o $(PWD)/bin/jq && chmod +x $(PWD)/bin/jq
 
 bin/yq: ## Install yq deps for update-patchversion script.
-	curl -sSfL https://github.com/mikefarah/yq/releases/download/v4.25.3/yq_$(YQ_PLATFORM)_$(YQ_ARCH) -o $(PWD)/bin/yq
+	curl -sSfL https://github.com/mikefarah/yq/releases/download/v4.25.3/yq_$(YQ_PLATFORM)_$(YQ_ARCH) -o $(PWD)/bin/yq && chmod +x $(PWD)/bin/yq
 
 bin/crane: ## Install crane deps for update-patchversion script.
 	curl -sSfL https://github.com/google/go-containerregistry/releases/download/v0.10.0/go-containerregistry_$(OS_NAME)_$(CRANE_ARCH).tar.gz | tar -xzf - crane && mv crane $(PWD)/bin/crane && chmod +x $(PWD)/bin/crane
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ MDLINTER_IMAGE = ghcr.io/igorshubovych/markdownlint-cli@sha256:2e22b4979347f70e0
 # Explicitly set architecture on arm, since werf currently does not support building of images for any other platform
 # besides linux/amd64 (e.g. relevant for mac m1).
 PLATFORM_NAME := $(shell uname -p)
+OS_NAME := $(shell uname)
+LBITS := $(shell getconf LONG_BIT)
 ifneq ($(filter arm%,$(PLATFORM_NAME)),)
 	export WERF_PLATFORM=linux/amd64
 endif
@@ -43,7 +45,7 @@ TESTS_TIMEOUT="15m"
 
 ##@ General
 
-deps: bin/golangci-lint bin/trivy bin/regcopy ## Install dev dependencies.
+deps: bin/golangci-lint bin/trivy bin/regcopy bin/jq bin/yq bin/crane ## Install dev dependencies.
 
 ##@ Tests
 
@@ -144,20 +146,18 @@ docs-down: ## Stop all the documentation containers.
 
 ##@ Update kubernetes control-plane patchversions
 
-.PHONY: update-k8s-patch-versions-deps
-update-k8s-patch-versions-deps: ## Install jq,yq,crane deps for update-patchversion script.
-	make install-jq && make install-yq && make install-crane
+bin/jq: ## Install jq deps for update-patchversion script.
+  ifeq ($(OS_NAME), Linux)
+		JQ_PLATFORM = linux
+  else ifeq ($(OS_NAME), Darwin)
+		JQ_PLATFORM = osx-amd
+  endif
+	curl -sSfL https://github.com/stedolan/jq/releases/download/jq-1.6/jq-$(JQ_PLATFORM)$(LBITS) -o $(PWD)/bin/jq
 
-.PHONY: install-jq
-install-jq: ## Install jq deps for update-patchversion script.
-	cd candi/tools; bash update_kubernetes_patchversions.sh install-jq
-
-.PHONY: install-yq
-install-yq: ## Install yq deps for update-patchversion script.
+bin/yq: ## Install yq deps for update-patchversion script.
 	cd candi/tools; bash update_kubernetes_patchversions.sh install-yq
 
-.PHONY: install-crane
-install-crane: ## Install crane deps for update-patchversion script.
+bin/crane: ## Install crane deps for update-patchversion script.
 	cd candi/tools; bash update_kubernetes_patchversions.sh install-crane
 
 .PHONY: update-k8s-patch-versions
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,33 @@ ifneq ($(filter arm%,$(PLATFORM_NAME)),)
 	export WERF_PLATFORM=linux/amd64
 endif
 
+# Set platform for jq
+ifeq ($(OS_NAME), Linux)
+	JQ_PLATFORM = linux
+else ifeq ($(OS_NAME), Darwin)
+	JQ_PLATFORM = osx-amd
+endif
+
+# Set platform for yq
+ifeq ($(OS_NAME), Linux)
+	YQ_PLATFORM = linux
+else ifeq ($(OS_NAME), Darwin)
+	YQ_PLATFORM = darwin
+endif
+# Set arch for yq
+ifeq ($(PLATFORM_NAME), x86_64)
+	YQ_ARCH = amd64
+else ifeq ($(PLATFORM_NAME), arm)
+	YQ_ARCH = arm64
+endif
+
+# Set arch for crane
+ifeq ($(PLATFORM_NAME), x86_64)
+	CRANE_ARCH = x86_64
+else ifeq ($(PLATFORM_NAME), arm)
+	CRANE_ARCH = arm64
+endif
+
 help:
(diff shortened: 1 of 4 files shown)


candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
+## Changelog entries
+\`\`\`changes
+section: candi
+type: feature
+summary: New Kubernetes control-plane components patchversions.
+impact: Restart Kubernetes control-plane components.
+impact_level: default
+\`\`\`
+EOF
+fi
```
shvgn, line 84: We could use Github Actions with cronjob schedule that whould create the PR for us e.g. every week. 

Along with that, with github actions we can keep the same PR for version updates if versions update twice before the PR merged. It is an idempotent way. 

We can use makefile and generate PR body with changes accordingly. What do you think?

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

```yaml
on:
  # Triggers the workflow every 5 minutes
  schedule:
    - cron: "*/5 * * * *"
```

RomanenkoDenys: i think this is out of scope. We should add this later. But adding automation in our plans.)

shvgn (approved):

candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
+New Kubernetes control-plane components patchversions.
+${PR_DESCRIPTION}
+## Why do we need it, and what problem does it solve?
+Kubernetes control-plane components should be up to date.
+## Changelog entries
+\`\`\`changes
+section: candi
+type: feature
+summary: New Kubernetes control-plane components patchversions.
+impact: Restart Kubernetes control-plane components.
```
shvgn, line 80:
```suggestion
impact: Kubernetes control-plane components will restart.
```

Makefile
```diff
@@ -57,3 +57,9 @@ lint-fix: bin/golangci-lint ## Fix lint violations.
 .PHONY: generate
 generate: ## Run all generate-* jobs in bulk.
 	cd tools; go generate
+
+##@ New kubernetes control-plane patchversions
+
+.PHONY: update-patchversions
+update-patchversions: ## Run update-patchversion script to generate new PR.
```
nabokihms, line 64: The name should be more specific, e.g., update-k8s-patch-versions.

Makefile
```diff
@@ -57,3 +57,9 @@ lint-fix: bin/golangci-lint ## Fix lint violations.
 .PHONY: generate
 generate: ## Run all generate-* jobs in bulk.
 	cd tools; go generate
+
+##@ New kubernetes control-plane patchversions
```
nabokihms, line 61: I think this name is too specific. It sounds like a group for a single target to me. I would not say this blocks us from merging the PR, but any movements will be good.

candi/tools/functions.sh
```diff
@@ -18,21 +18,28 @@ set -Eeo pipefail
...
+    if ! gh --version &>/dev/null; then
+      >&2 echo "ERROR: gh is not installed. Please install it from https://cli.github.com"
       return 1
     fi
 }
 
 function check_crane() {
     if ! crane version &>/dev/null; then
-      >&2 echo "ERROR: crane is not installed"
+      >&2 echo "ERROR: crane is not installed. Please install it from https://github.com/google/go-containerregistry/tree/main/cmd/crane"
```
nabokihms, line 35: I'd like to see targets to install crane and yq binaries in the make file. With this, we can easily do
```
make deps
make update-k8s-patch-versions
```

candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
     fi
   done
 done
 
-cd -
+if [[ "${CREATE_PR}" == "true" ]]; then
+  cd ../../
+  BRANCH="kubernetes-patchversions-$(date +"%y-%m-%d-%H-%M")"
+  git checkout -b "${BRANCH}"
+  git add .
```
nabokihms, line 66: It seems like `git add .` can accidentally add more files than needed. If we know the exact files changed, it is better to add only these files
```suggestion
  git add ../version_map.yml
```

candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
+${PR_DESCRIPTION}
+## Why do we need it, and what problem does it solve?
+Kubernetes control-plane components should be up to date.
+## Changelog entries
+\`\`\`changes
+section: candi
+type: feature
+summary: New Kubernetes control-plane components patchversions.
+impact: Restart Kubernetes control-plane components.
+impact_level: default
```
nabokihms, line 81: I thought that impact should be `high` if control plane components are restarted. Otherwise, there will be no note about the impact in the generated changelog file.

candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
+## Description
+New Kubernetes control-plane components patchversions.
+${PR_DESCRIPTION}
+## Why do we need it, and what problem does it solve?
+Kubernetes control-plane components should be up to date.
+## Changelog entries
+\`\`\`changes
+section: candi
+type: feature
+summary: New Kubernetes control-plane components patchversions.
```
nabokihms, line 79: I'd like to see the list of updated components here, or in the impact field. Restarts of kube-apiserver affect more than restarts of kube-scheduler.

candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
-cd -
+if [[ "${CREATE_PR}" == "true" ]]; then
+  cd ../../
+  BRANCH="kubernetes-patchversions-$(date +"%y-%m-%d-%H-%M")"
+  git checkout -b "${BRANCH}"
+  git add .
+  git commit -m "[candi] New kubernetes control-plane components patchversions"
+  git push --set-upstream origin "${BRANCH}"
+  cat <<EOF | gh pr create -d -B main -F - -t "New kubernetes patchversions"
+## Description
```
nabokihms, line 70: Let's remove the `Description` and the `Why do we need it, and what problem does it solve?` sections because the template may be changed in the future.

I think it would be best to note that this PR was autogenerated with a script instead.

candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
 done
 
-cd -
+if [[ "${CREATE_PR}" == "true" ]]; then
+  cd ../../
+  BRANCH="kubernetes-patchversions-$(date +"%y-%m-%d-%H-%M")"
+  git checkout -b "${BRANCH}"
+  git add .
+  git commit -m "[candi] New kubernetes control-plane components patchversions"
+  git push --set-upstream origin "${BRANCH}"
```
nabokihms, line 68: Using `origin` may be tricky.
For example, my git uses `upstrem` for the github repo.

candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
+  cat <<EOF | gh pr create -d -B main -F - -t "New kubernetes patchversions"
+## Description
+New Kubernetes control-plane components patchversions.
+${PR_DESCRIPTION}
+## Why do we need it, and what problem does it solve?
+Kubernetes control-plane components should be up to date.
+## Changelog entries
+\`\`\`changes
+section: candi
+type: feature
```
nabokihms, line 78: Looks like a `chore` to me.

candi/tools/update_kubernetes_patchversions.sh
```diff
@@ -28,36 +30,55 @@ function check_requirements() {
...
 
-cd -
+if [[ "${CREATE_PR}" == "true" ]]; then
+  cd ../../
+  BRANCH="kubernetes-patchversions-$(date +"%y-%m-%d-%H-%M")"
+  git checkout -b "${BRANCH}"
+  git add .
+  git commit -m "[candi] New kubernetes control-plane components patchversions"
+  git push --set-upstream origin "${BRANCH}"
+  cat <<EOF | gh pr create -d -B main -F - -t "New kubernetes patchversions"
```
nabokihms, line 69: NIT: Let's add a date to it.

candi/tools/functions.sh
```diff
@@ -18,21 +18,28 @@ set -Eeo pipefail
...
 function check_jq() {
     if ! jq --version &>/dev/null; then
-      >&2 echo "ERROR: jq is not installed"
+      >&2 echo "ERROR: jq is not installed. Please install it from https://stedolan.github.io/jq/download"
+      return 1
+    fi
+}
+
+function check_gh() {
+    if ! gh --version &>/dev/null; then
```
nabokihms, line 27: Should I be authorized to use the gh utility? Where should I get the token? I think it is obvious for some engineers, but for others, it may not.

candi/tools/update_kubernetes_patchversions.sh
shvgn: Is control plane restarts important to business-oriented users?

candi/tools/functions.sh
shvgn: We can avoid manual dependencies management if we use GA schedule right away. We could run it every day. 

See 
https://github.com/deckhouse/deckhouse/pull/1819#discussion_r901856171

candi/tools/update_kubernetes_patchversions.sh
nabokihms: Yes, and it is always mentioned in the release notes. Not only control plane, but all components.

candi/tools/functions.sh
nabokihms: I have a strong opinion that every script executed on CI can be executed locally as well (and should produce the same result).

If we only expect this script to run on CI, why is gh tool used for API interactions to open prs instead? Why do we check dependencies before the start?

shvgn: Okay, I'm sure @RomanenkoDenys definitely achieved that locally. 

IMO, it's enough to automate everything except PR creation to say that we have the same tool for both CI and developer machine. Running `make` and getting the diff will do.

GA lets us not bother with such a dependency as `gh`. Some developers don't use `gh` at all (I do). Some developers would prefer to open web interface and type there.

I would prefer not to bother developers at all with such a routine. And make it fully on daily basis acutomatically. The workflow is as simple as running `make` and creating a PR like we do it for changelogs. This is easy to achieve.

nabokihms: Definitely agree! Had the same though.

Makefile
RomanenkoDenys: good name, i used it

candi/tools/update_kubernetes_patchversions.sh
RomanenkoDenys: removed create of PR

candi/tools/update_kubernetes_patchversions.sh
RomanenkoDenys: removed create of PR

candi/tools/update_kubernetes_patchversions.sh
RomanenkoDenys: removed create of PR

candi/tools/update_kubernetes_patchversions.sh
RomanenkoDenys: removed create of PR

candi/tools/functions.sh
RomanenkoDenys: i removed pr creation, we will add GA later

candi/tools/functions.sh
RomanenkoDenys: yes, good point

nabokihms: There is a conflict with the makefile from the main branch because there is the deps target already.
Could you take a look and rearrange your code in line with it (e,g use targets to download dependencies)?

nabokihms (commented):
I left a comment. Despite it, everything looks good.

candi/tools/functions.sh
RomanenkoDenys: resolved

Makefile
```diff
@@ -110,3 +110,25 @@ cve-report: ## Generate CVE report for a Deckhouse release.
 cve-base-images: ## Check CVE in our base images.
   ##~ Options: SEVERITY=CRITICAL,HIGH
 	./tools/cve/base-images.sh
+
+##@ Update kubernetes control-plane patchversions
+
+.PHONY: update-k8s-patch-versions-deps
+update-k8s-patch-versions-deps: ## Install jq,yq,crane deps for update-patchversion script.
+	make install-jq && make install-yq && make install-crane
```
shvgn, line 118: Reusing targets can be one-liner. Great target names, by the way. They are self-explanatory.

```make
.PHONY: update-k8s-patch-versions-deps
update-k8s-patch-versions-deps: install-jq install-yq install-crane
```

shvgn (dismissed):
LGTM

Makefile
shvgn: You are welcome to resolve this thread if you won't feel like changing it.

nabokihms (approved):

PR #1839, opened 2022-06-22
[control-plane-manager] Bolt-on healthCheckNodePort of the control-plane Service and trafficPolicy: Local

zuzzas:
## Description

We'll allocate from the end of the [default](https://github.com/kubernetes/kubernetes/blob/0520d75838d28007896c9af9eff2485ae8b84e86/pkg/kubeapiserver/options/options.go#L26) NodePort range.

Kubernetes' port allocation algorithm starts from the beginning. By choosing the port from the end, we are safe from port collisions.

## Changelog entries

```changes
type: feature
summary: Bolt-on healthCheckNodePort of the control-plane Service and trafficPolicy: Local
impact_level: default
```

### DIFF
--- a/modules/040-control-plane-manager/templates/service.yaml
+++ b/modules/040-control-plane-manager/templates/service.yaml
@@ -14,6 +14,9 @@ metadata:
       {{- end }}
     {{- end }}
 spec:
+  externalTrafficPolicy: Local
+  # must be changed if NodePort range is changed via "--service-node-port-range"
+  healthCheckNodePort: 32767
   type: LoadBalancer
     {{- if .Values.controlPlaneManager.apiserver.loadBalancer.sourceRanges }}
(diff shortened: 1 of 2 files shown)


z9r5 (approved):

modules/040-control-plane-manager/templates/service.yaml
```diff
@@ -14,6 +14,9 @@ metadata:
       {{- end }}
     {{- end }}
 spec:
+  externalTrafficPolicy: Local
+  # must be changed if NodePort range is changed via "--service-node-port-range"
+  healthCheckNodePort: 32767
```
konstantin-axenov, line 19: Isn't this field immutable?

konstantin-axenov (approved):

z9r5:

PR #1851, opened 2022-06-22
[deckhouse] improve deckhouse update process

yalosev:
## Description
Suspend only pending releases. Apply first release automatically on deckhouse bootstrap

Improve update process. Suspending deployed release leads to implicit behavior. Also as first pending release

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Automatically apply first release on deckhouse bootstrap
```

### DIFF
--- a/modules/020-deckhouse/hooks/update_deckhouse_image.go
+++ b/modules/020-deckhouse/hooks/update_deckhouse_image.go
@@ -108,6 +108,9 @@ type deckhousePodInfo struct {
 	Ready     bool   `json:"ready"`
 }
 
+// while cluster bootstrapping we have the tag for deckhouse image like: alpha, beta, early-access, stable, rock-solid
+// it is set via dhctl, which does not know anything about releases and tags
+// We can use this bootstrap image for applying first release without any requirements (like update windows, canary, etc)
 func (dpi deckhousePodInfo) isBootstrapImage() bool {
 	colonIndex := strings.LastIndex(dpi.Image, ":")
 	if colonIndex == -1 {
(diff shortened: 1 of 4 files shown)


modules/020-deckhouse/hooks/update_deckhouse_image.go
```diff
@@ -108,6 +108,21 @@ type deckhousePodInfo struct {
 	Ready     bool   `json:"ready"`
 }
 
+func (dpi deckhousePodInfo) isBootstrapImage() bool {
+	arr := strings.Split(dpi.Image, ":")
```
nabokihms, line 112: A plain split is not ok here because it does not allow registries with a port, e.g., `registry.test.com:5000/deckhouse:alpha` (valid private registry address).

There are two options to parse the tag correctly:
1. `ParseReference` method from the `go-containerregistry` library.
2. A method copied from Docker https://github.com/deckhouse/deckhouse/blob/a7535a49dcee6ac56ff8ba9fa7ea83106e01cabb/global-hooks/discovery/deckhouse_registry.go#L169

And we also need a test for it.

yalosev: Makes sense. Thank you.
But as we need only tag here we can use LastIndex for figuring it out

nabokihms: The last index can also be a problem. Example:
```
registry.deckhouse.io:alpha@sha256:abcdefg
```

nabokihms (approved):

nabokihms (changes requested):
Oh, sorry. The one comment is still left.

z9r5:

PR #1853, opened 2022-06-23
[cni-cilium] createNodeRoutes everywhere

zuzzas:
## Description

Set `createNodeRoutes` to `true` everywhere except clouds other than openstack and vsphere.

We'll create routes everywhere except cloud providers that already has CCMs which create routes between Nodes.

## Changelog entries

```changes
section: cni-cilium
type: feature
summary: Set `createNodeRoutes` to `true` everywhere except clouds other than openstack and vsphere.
impact_level: low
```


z9r5 (dismissed):

modules/031-linstor/docs/FAQ.md
```diff
@@ -5,6 +5,7 @@ title: "The linstor module: FAQ"
 ## What is difference between LVM and LVMThin?
 
 Briefly:
+
```
nabokihms, line 8: Rebase is required.

modules/021-cni-cilium/hooks/enable_node_routes.go
```diff
@@ -44,14 +46,14 @@ func shouldEnableNodeRoutes(input *go_hook.HookInput) bool {
 	}
 
 	providerRaw, ok := input.Values.GetOk("global.clusterConfiguration.cloud.provider")
-	if !ok {
-		return false
-	}
+	if ok {
```
nabokihms, line 49: I'd like to see `default: true` here with the ability to override it through the cni configuration secret.
https://github.com/deckhouse/deckhouse/blob/8f7710e5baaabab2624a201e8e9887b69b0b6771/modules/021-cni-cilium/openapi/config-values.yaml#L15

Is there any problem with enabling it by default?

zuzzas: We can do this, the cloud routes will be set regardless of the cloud we are in. Until the CNI configuration secret is deployed in the cloud-provider-* module. Which is later in the initialization order than cni-cilium. At which point we'll (perhaps) revert to `enableNodeRoutes: false`. But the routes will already be created in the OS, which may adversely impact networking in the cluster.
How do you envision this?

nabokihms: I think cilium will be automatically enabled right after its config is deployed.
Order will not be
```
020-deckhouse -> 021-cni-cilium (with wrong settings) -> 030-cloud-provider-yandex -> 021-cni-cilium (with ok settings)
```
It will be
```
020-deckhouse -> 030-cloud-provider-yandex -> 021-cni-cilium (with ok settings)
```

nabokihms: The problem occurs only if cni-cilium is manually enabled in clouds.

nabokihms: We've discussed it in Google Meet and agreed that there is no better solution for this issue.

deckhouse-BOaTswain:
Aye, aye, @nabokihms. I've started the workflow for label 'e2e/run/azure'...

nabokihms (approved):

z9r5:

PR #1859, opened 2022-06-23
[Doc] Fix sample-limit comments

shvgn:
## Description

Fix YAML comments in docs.

## Changelog entries

```changes
section: docs
type: fix
summary: Fixed typos in sample-limit clarifications.
impact_level: low
```

### DIFF
--- a/modules/340-monitoring-custom/docs/README.md
+++ b/modules/340-monitoring-custom/docs/README.md
@@ -57,7 +57,7 @@ metadata:
     prometheus.deckhouse.io/path: "/my_app/metrics"           # /metrics by default
     prometheus.deckhouse.io/query-param-format: "prometheus"  # '' by default
     prometheus.deckhouse.io/allow-unready-pod: "true"         # by default, NON-ready Pods are ignored
-    prometheus.deckhouse.io/sample-limit: "5000"              # by default, the sample is limited to 1000 metrics for a single Pod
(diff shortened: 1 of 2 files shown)


modules/340-monitoring-custom/docs/README.md
```diff
@@ -57,7 +57,7 @@ metadata:
     prometheus.deckhouse.io/path: "/my_app/metrics"           # /metrics by default
     prometheus.deckhouse.io/query-param-format: "prometheus"  # '' by default
     prometheus.deckhouse.io/allow-unready-pod: "true"         # by default, NON-ready Pods are ignored
-    prometheus.deckhouse.io/sample-limit: "5000"              # by default, the sample is limited to 1000 metrics for a single Pod
```
nabokihms, line 60: This comment meant that by default, the limit is 1000, but it is overridden to 5000 with the annotation. The same logic is applied to previous comments, e.g.:
 ```
 prometheus.deckhouse.io/allow-unready-pod: "true" # by default, NON-ready Pods are ignored
 ```
The allow-unready-pod annotation value is true, but by default, NON-ready pods are ignored.

z9r5 (approved):

modules/340-monitoring-custom/docs/README.md
shvgn: Okay, so I misunderstood it out of context.  We should close the PR then. Sorry for bothering.

PR #1861, opened 2022-06-23
[deckhouse-controller] fix: display help when deckhouse-controller is invoked without arguments

diafour:
## Description

Remove default for start command to prevent starting second instance of operator.

Accidental invoking deckhouse-controller without arguments starts an operator. This behaviour complicates diagnostics.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Display help when deckhouse-controller is invoked without arguments.
impact_level: low
```


deckhouse-BOaTswain:

PR #1864, opened 2022-06-24
[snapshot-controller] Reduce amount of replicas for snapshot-controller to a maximum of 2

kvaps:
## Description

Reduce amount of replicas for snapshot-controller to a maximum of 2

If amount of system-nodes less than master-nodes, then some replicas of snapshot-controller always stuck in `Pending`

## Changelog entries

```changes
section: snapshot-controller
type: fix
summary: Reduce amount of replicas for snapshot-controller to a maximum of 2
impact: low
```


nabokihms (approved):

z9r5:

PR #1866, opened 2022-06-24
[monitoring-kubernetes] node-exporter: parse containerd config

yalosev:

## Description
Managed clusters (EKS, Yandex, GKE) with deckhouse as module couldn't work with kubelet-eviction-threshold-exporter because they don't have crictl and also socket can have different passes. But all of them have /etc/containerd/config.toml, so we can parse it and extract root direction

kubelet-eviction-threshold-exporter doesn't work in platform managed kubernetes.
```
kubelet-eviction-threshold-exporter /bin/loop: line 33: crictl: command not found      
```

## Changelog entries

```changes
type: fix
summary: Make `kubelet-eviction-threshold-exporter` workable in managed kubernetes platforms (AWS, GKE, Yandex, etc)
```

### DIFF
--- a/modules/340-monitoring-kubernetes/images/kubelet-eviction-thresholds-exporter/loop
+++ b/modules/340-monitoring-kubernetes/images/kubelet-eviction-thresholds-exporter/loop
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2021 Flant JSC
+# Copyright 2022 Flant JSC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/modules/340-monitoring-kubernetes/images/kubelet-eviction-thresholds-exporter/loop
+++ b/modules/340-monitoring-kubernetes/images/kubelet-eviction-thresholds-exporter/loop
@@ -48,7 +48,11 @@ runtime_root_dir () {
     curl -sS --unix-sock /var/run/docker.sock http://system/info | jq -re '.DockerRootDir'
     ;;
   containerd)
-    dasel -f /var/run/containerd-config.toml '.root' 2>/dev/null || echo "/var/lib/containerd"
+    root_dir="$(grep "^root" /etc/containerd/config.toml | sed "s/ //g"  | sed "s/\"//g" |cut -d"=" -f2)"
+    if [ -z "${root_dir}" ]; then
+      root_dir="/var/lib/containerd"
+    fi
+    echo ${root_dir}
     ;;
   *)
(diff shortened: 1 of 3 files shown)


modules/340-monitoring-kubernetes/images/kubelet-eviction-thresholds-exporter/loop
```diff
@@ -1,7 +1,25 @@
 #!/bin/bash
 
+# Copyright 2021 Flant JSC
```
nabokihms, line 3:
```suggestion
# Copyright 2022 Flant JSC
```

nabokihms (dismissed):
I'd like Denys to also have a look on this PR, yet it lgtm.

modules/340-monitoring-kubernetes/images/kubelet-eviction-thresholds-exporter/loop
```diff
@@ -30,7 +48,7 @@ runtime_root_dir () {
     curl -sS --unix-sock /var/run/docker.sock http://system/info | jq -re '.DockerRootDir'
     ;;
   containerd)
-    crictl --runtime-endpoint unix:/var/run/containerd/containerd.sock info | jq -r '.config.containerdRootDir'
+    dasel -f /var/run/containerd-config.toml '.root' 2>/dev/null || echo "/var/lib/containerd"
```
RomanenkoDenys, line 51:
```
root_dir="$(grep "^root" /etc/containerd/config.toml | sed "s/ //g"  | sed "s/\"//g" |cut -d"=" -f2)"
if [-z "${root_dir}" ]; then
  root_dir="/var/lib/containerd"
fi
echo ${root_dir}
```
and we cannot install dasel

RomanenkoDenys (changes requested):

modules/340-monitoring-kubernetes/images/kubelet-eviction-thresholds-exporter/loop
yalosev: It could be:
`# root = "/var/lib/containerd"`
`#root = "/var/lib/containerd"`

and many other cases. I guess dasel is cheaper here

yalosev: maybe use something like:
`grep "^root\|^#root\|^# root"`

yalosev: nevermind, this case is closed

nabokihms (approved):

deckhouse-BOaTswain:
Aye, aye, @nabokihms. I've started the workflow for label 'e2e/run/yandex-cloud'...

z9r5:

PR #1867, opened 2022-06-24
[monitoring-kubernetes-control-plane] Deprecated APIs dashboard

nabokihms:

## Description
Add GKE-like dashboard to Grafana to analyze deprecated APIs

image

## Changelog entries

```changes
section: monitoring-kubernetes-control-plane
type: feat
summary: Deprecated APIs dashboard
```


yalosev:
2 point: we can exclude Deckhouse releases from check. Currently we check all releases `LabelSelector: "owner=helm,status=deployed"` but we can exclude d8 ones.
4 we don't check outdated releases, only the last one. So metric does't contain this situation

PR #1872, opened 2022-06-25
[cloud-provider-openstack] Support for volume type with spaces in name

konstantin-axenov:

## Description

If volume type name has spaces, replace spaces with dashes for StorageClass name

Some OpenStack clouds have volumes with spaces in names
```
+--------------------------------------+---------+-----------+
| ID                                   | Name    | Is Public |
+--------------------------------------+---------+-----------+
| bb49fb86-7279-44ed-ae1f-7a55e1771754 | SSD R1  | True      |
```
For such volume type name Deckhouse tries to create StorageClass with name `ssd r1` but this is forbidden.

## Changelog entries

```changes
type: fix
summary: Support for volume type with spaces in name
```

### DIFF
--- a/ee/modules/030-cloud-provider-openstack/hooks/discover_volume_types.go
+++ b/ee/modules/030-cloud-provider-openstack/hooks/discover_volume_types.go
@@ -38,7 +38,7 @@ func handleDiscoverVolumeTypes(input *go_hook.HookInput) error {
 
 	var openstackVolumeTypes []string
 	if os.Getenv("D8_IS_TESTS_ENVIRONMENT") != "" {
-		openstackVolumeTypes = []string{"__DEFAULT__", "some-foo", "bar", "other-bar", "SSD R1", "-Xx$&? -foo", " YY fast SSD -foo"}
+		openstackVolumeTypes = []string{"__DEFAULT__", "some-foo", "bar", "other-bar", "SSD R1", "-Xx$&? -foo", " YY fast SSD -foo."}
 	} else {
 		openstackVolumeTypes, err = getVolumeTypesArray()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ee/modules/030-cloud-provider-openstack/hooks/discover_volume_types.go
```diff
@@ -49,7 +49,7 @@ func handleDiscoverVolumeTypes(input *go_hook.HookInput) error {
 	storageClassesMap := make(map[string]string, len(openstackVolumeTypes))
 
 	for _, vt := range openstackVolumeTypes {
-		storageClassesMap[strings.ToLower(strings.ReplaceAll(vt, "_", ""))] = vt
+		storageClassesMap[strings.ToLower(strings.ReplaceAll(strings.ReplaceAll(vt, "_", ""), " ", "-"))] = vt
```
nabokihms, line 52: This is an ok solution, but I'd like to fix it once and forever.
There are functions that converts strings to valid DNS1123 domain labels, e.g., 

```go
const (
	DNS1123NameMaximumLength         = 63
	DNS1123NotAllowedCharacters      = "[^-a-z0-9]"
	DNS1123NotAllowedStartCharacters = "^[^a-z0-9]+"
)

func makeDNS1123Compatible(name string) string {
	name = strings.ToLower(name)

	nameNotAllowedChars := regexp.MustCompile(DNS1123NotAllowedCharacters)
	name = nameNotAllowedChars.ReplaceAllString(name, "")

	nameNotAllowedStartChars := regexp.MustCompile(DNS1123NotAllowedStartCharacters)
	name = nameNotAllowedStartChars.ReplaceAllString(name, "")

	if len(name) > DNS1123NameMaximumLength {
		name = name[0:DNS1123NameMaximumLength]
	}

	return name
}

```

What do you think?

konstantin-axenov: Good point. I had a thought about making a generic solution. Then I thought that we want to change spaces to dashes because it gives more proper names for StorageClasses. But as I think now, it's better just to remove all invalid characters as it was done for underscores previously. I will apply your suggestion.

konstantin-axenov: Done

PR #1888, opened 2022-06-28
[ingress-nginx] Refactor startup hook

yalosev:

## Description
Refactor startup hook, use kubeclient instead of snapshots

Snopshots don't work at startup phase. We have an extra run of the hook without any usefulness

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: Refactor startup hook
impact_level:  low
```

### DIFF
--- a/modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
+++ b/modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// this hook figure out minimal ingress controller version at the beginning and on IngressNginxController creation
+// Deckhouse would not update minor version before pod is ready, so this hook will execute at least once (on sync)
+
 package hooks
 
 import (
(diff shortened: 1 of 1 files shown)


go_lib/dependency/dependency.go
```diff
@@ -257,6 +263,27 @@ func (mdc *mockedDependencyContainer) SetK8sVersion(ver k8s.FakeClusterVersion)
 	mdc.K8sClient = cli
 }
 
+// CreateDynamicFromYAML create dynamic object from yaml string
+// you can create multiply objects of the same kind by separating yaml with '---'
+func (mdc *mockedDependencyContainer) CreateDynamicFromYAML(gvr schema.GroupVersionResource, data string) error {
+	arr := strings.Split(data, "---")
```
diafour, line 269: This split pattern is error-prone. There is a 'ListFromYamlDocs' in [kube-client](https://github.com/flant/kube-client/blob/master/manifest/manifest_list.go#L5) that is used for monitoring Helm resources in addon-operator.

diafour:
> Snopshots don't work at startup phase. We have an extra run of the hook without any usefulness

You can also use "Synchronization only OnKubernetes": remove onStartup and use OnKubernetes with `executeHookOnEvent: false` to get snapshots once at module startup.

go_lib/dependency/dependency.go
yalosev: Well. It was for tests only but you are right. I have fixed it

yalosev:
seems like we don't need onStartup binding here at all

nabokihms (approved):

deckhouse-BOaTswain:

PR #1889, opened 2022-06-28
[ingress-nginx] hpa: measure average cpu usage for DaemonSet

yalosev:

## Description
We have to calculate average CPU load for all pods in a DaemonSet but we are calculating the summary of them at the moment

HPA works wrong, ordering too many new replicas

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: fix hpa calculating for IngressNginxController. Calculate average CPU load instead of summary
```


nabokihms (approved):

z9r5:

PR #1892, opened 2022-06-28
[ingress-nginx] Delete ingressClass objects on ingress migration to > 1.0

yalosev:

## Description
Delete IngressClass object when we migrate from controller < 1.0 to controller > 1.0 version

IngressClassesfor controller > 1.0 and < 1.0 have different controller spec field. This field is immutable, so we have to delete all this object and recreate it. Without this hook D8 will stuck if you just edit the version of a IngressNginxController

## Changelog entries

```changes
section: ingress-nginx
type: fix 
summary: Handle IngressClass resources when migrating from controller < 1.0 version to > 1.0 version
```

### DIFF
--- a/modules/402-ingress-nginx/hooks/migrate_ingress_class.go
+++ b/modules/402-ingress-nginx/hooks/migrate_ingress_class.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2022 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package hooks
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
+	"github.com/flant/addon-operator/sdk"
+	v1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/utils/pointer"
+)
+
+// migration from 0.33 to 1.1 have to change ingress class but it's controller field is immutable
+// we have to handle this situation manually - delete old IngressClass resources
+
+var _ = sdk.RegisterFunc(&go_hook.HookConfig{
+	OnBeforeHelm: &go_hook.OrderedConfig{Order: 30},
+	Kubernetes: []go_hook.KubernetesConfig{
+		{
+			Name:                         "ingressClasses",
+			ApiVersion:                   "networking.k8s.io/v1",
+			Kind:                         "IngressClass",
+			ExecuteHookOnSynchronization: pointer.Bool(false),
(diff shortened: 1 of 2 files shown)


modules/402-ingress-nginx/hooks/migrate_ingress_class.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// migration from 0.33 to 1.1 have to change ingress class but it's controller field is immutable
+// we have to handle this situation manually - delete old IngressClass resources
+
+var _ = sdk.RegisterFunc(&go_hook.HookConfig{
+	OnBeforeHelm: &go_hook.OrderedConfig{Order: 30},
+	Kubernetes: []go_hook.KubernetesConfig{
+		{
+			Name:                         "ingressClasses",
```
nabokihms, line 39: NIT: we usually use snake_case for binding names.

modules/402-ingress-nginx/hooks/migrate_ingress_class.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			return err
+		}
+
+		version, ok, err := unstructured.NestedString(controller.Spec, "controllerVersion")
+		if err != nil {
+			return err
+		}
+		if !ok {
+			version = input.Values.Get("ingressNginx.defaultControllerVersion").String()
```
nabokihms, line 103: Is this part required? As far as I remember, Deckhouse always sets the `controllerVersion` on setting internal values. See this code
https://github.com/deckhouse/deckhouse/blob/5bbc1347c94030821a82a492e78850fd1c238d5c/modules/402-ingress-nginx/hooks/get_ingress_controllers.go#L138-L150

modules/402-ingress-nginx/hooks/migrate_ingress_class.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	existingClasses := make(map[string]string, len(snap))
+	for _, s := range snap {
+		class := s.(d8IngressClass)
+		existingClasses[class.Name] = class.Controller
+	}
+
+	conArray := input.Values.Get("ingressNginx.internal.ingressControllers").Array()
+	edgeVersion, _ := semver.NewVersion("v1.0.0")
```
nabokihms, line 81: Nit: To be in line with the other part of your code
```
edgeVersion := semver.MustParse("v1.0.0")
```

modules/402-ingress-nginx/hooks/migrate_ingress_class.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		if !ok {
+			version = input.Values.Get("ingressNginx.defaultControllerVersion").String()
+		}
+		semV := semver.MustParse(version)
+		if semV.GreaterThan(edgeVersion) {
+			controllerName = fmt.Sprintf("ingress-nginx.deckhouse.io/%s", ingressClass)
+		}
+
+		existingController, ok := existingClasses[ingressClass]
```
nabokihms, line 110: It should be done earlier in the code when the hook gets the `ingressClass` property (L96).

nabokihms (commented):
Looks good. I left some comments, but this PR is fine enough already.

modules/402-ingress-nginx/hooks/migrate_ingress_class.go
yalosev: This variable (from your code) is used only for metrics, we don't change value in `controller` object
And in templates we use:
```
{{- $controllerVersion := $crd.spec.controllerVersion | default $context.Values.ingressNginx.defaultControllerVersion }}
```

modules/402-ingress-nginx/hooks/migrate_ingress_class.go
yalosev: Moved to var block

nabokihms (approved):

z9r5:

PR #1894, opened 2022-06-28
[monitoring-kubernetes] handle absent containerd directory

yalosev:

## Description
Create containerd directory if it does not exist.

Pod could run if host does not contain containerd directory (docker runtime for example). FileOrCreate does not create parent directory

## Changelog entries

```changes
type: fix 
summary: Fixed handling absent containerd directory
```

### DIFF
--- a/modules/340-monitoring-kubernetes/templates/node-exporter/daemonset.yaml
+++ b/modules/340-monitoring-kubernetes/templates/node-exporter/daemonset.yaml
@@ -87,6 +87,8 @@ spec:
           mountPath: /var/run/node-exporter-textfile
         - name: dockersock
           mountPath: /var/run/docker.sock
+        - name: containerddir
+          mountPath: /etc/containerd
         - name: containerdconfig
           mountPath: /etc/containerd/config.toml
         - name: kube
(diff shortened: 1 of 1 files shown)


modules/340-monitoring-kubernetes/templates/node-exporter/daemonset.yaml
```diff
@@ -87,6 +87,8 @@ spec:
           mountPath: /var/run/node-exporter-textfile
         - name: dockersock
           mountPath: /var/run/docker.sock
+        - name: containerddir
```
nabokihms, line 90: If the whole containerd directory is mounted, I think the following contained config mount is not required.

modules/340-monitoring-kubernetes/templates/node-exporter/daemonset.yaml
```diff
@@ -138,6 +140,10 @@ spec:
       - name: dockersock
         hostPath:
           path: /var/run/docker.sock
+      - name: containerddir
```
nabokihms, line 143: I assume it would be best to set the `subPath` and abstain from two mounts, or just mount the whole directory.
```yaml
volumes:
- name: containerddir
   hostPath:
     path: /etc/containerd
     type: DirectoryOrCreate
 ...
 volumeMounts:
- name: containerddir
   mountPath: /etc/containerd
   subPath: config.toml
```

yalosev: The whole directory is ok. Fixed

z9r5:

PR #1897, opened 2022-06-29
[ci] Move variables to werf

nabokihms:

## Description
* Create YAML-formatted files with variables used for building by werf (two more version maps)
* Remove ci_include files for terraform variables and base images variables
* Load all version maps on top of werf.yaml file (images, terraform, kubernetes)
* Replace `{{ env "BASE_IMAGE" }}` with `{{ .Images.BASE_IMAGE }}`

1. Our CI system does not support changing any env variable in the PR. Thus if we want to change base images or terraform versions, developers should commit their changes to the main branch. Kubernetes version map works perfectly for this case because werf utilizes variables from it directly, bypassing CI. This PR inherits Kubernetes versions map logic for other version maps.

2. Because files are now in YAML format, it is possible to use them in other scripts (like we did for automatic Kubernetes versions updater).

3. To build an image, developers had to provide args from CI. Now, all required dependencies are loaded directly to the werf.yaml, so it is not needed to export env variables in local environments.

## Changelog entries

```changes
section: ci
type: feature
summary: Move variables to werf
impact_level: low
```


deckhouse-BOaTswain:

RomanenkoDenys (changes requested):
Everything is fine. But in my opinion, you should use yaml-notation, for example .Images.BaseGolangBuster instead of .Images.BASE_GOLANG_BUSTER. But that doesn't block this PR.

werf.yaml
```diff
@@ -1,21 +1,25 @@
...
-      limit:
-        in: 720h
 gitWorktree:
   forceShallowClone: true
 ---
+# Base Images
+{{ $_ := set . "Images" (.Files.Get "candi/image_versions.yml" | fromYaml) }}
+  {{- range $k, $v := .Images }}
+    {{ $baseImagePath := (printf "%s%s" $.Images.REGISTRY_PATH (trimSuffix "/" $v)) }}
+    {{- if ne $k "REGISTRY_PATH" }}
```
RomanenkoDenys, line 11: Why we remove .Images.REGISTRY_PATH ? It may useful in future.

nabokihms: REGISTRY_PATH value is concatenated with other images from the file, except REGISTRY_PATH. It is still in the dictionary but without changes.

RomanenkoDenys: sorry, I was inattentive. All is ok.

RomanenkoDenys (approved):

PR #1900, opened 2022-06-29
[candi] Fix disk size for static nodes in YandexClusterConfiguration

shvgn:
## Description

- Fix inconsistency between the docs, config schema, and terraform variables
- Make examples more realistic

- Default size is 20G, not 50G as stated in docs
- Nodegroup disk size is not passed to terraform due to typo
- Examples in docs are misleading

## Changelog entries

```changes
section: candi
type: fix
summary: Fixed disk size config for CloudPermanent nodes in YandexClusterConfiguration
```


candi/cloud-providers/yandex/terraform-modules/master-node/variables.tf
```diff
@@ -49,7 +49,7 @@ locals {
   platform              = lookup(local.master_instance_class, "platform", "standard-v2")
   cores                 = local.master_instance_class.cores
   memory                = local.master_instance_class.memory / 1024
-  disk_size_gb          = lookup(local.master_instance_class, "diskSizeGB", 20)
+  disk_size_gb          = lookup(local.master_instance_class, "diskSizeGb", 20)
```
shvgn, line 52:
```suggestion
  disk_size_gb          = lookup(local.master_instance_class, "diskSizeGb", 50)
```

candi/cloud-providers/yandex/terraform-modules/master-node/variables.tf
```diff
@@ -49,7 +49,7 @@ locals {
   platform              = lookup(local.master_instance_class, "platform", "standard-v2")
   cores                 = local.master_instance_class.cores
   memory                = local.master_instance_class.memory / 1024
-  disk_size_gb          = lookup(local.master_instance_class, "diskSizeGB", 20)
+  disk_size_gb          = lookup(local.master_instance_class, "diskSizeGB", 50)
```
shvgn, line 52: Should we change this?

RomanenkoDenys: I cannot understand, how we should converge state of existing clusters ? Manually ? Maybe we need a migration to set diskSizeGB to 20 in existing installations ?

RomanenkoDenys (commented):
I cannot understand, how we should converge state of existing clusters ? Manually ? Maybe we need a migration to set diskSizeGB to 20 in existing installations ?

candi/cloud-providers/yandex/terraform-modules/master-node/variables.tf
shvgn: Indeed. Let it be a separate decision

name212 (dismissed):

nabokihms (approved):

PR #1903, opened 2022-06-29
[chrony] Refactor alerts

konstantin-axenov:

## Description

- Replace metric `node_ntp_sanity metrics` with `node_timex_sync_status`
- Calculate node's clock drift using metric `node_ntp_offset_seconds`

Alert `NTPDaemonOnNodeDoesNotSynchronizeTime` was triggered if difference between node time and chrony server time was bigger than 1ms. But chrony doesn't act as ntpd that just instantly corrects time.

> In normal operation, chronyd slews the time when it needs to adjust the system clock. For example, to correct a system clock which is 1 second slow, chronyd slightly increases the amount by which the system clock is advanced on each clock interrupt, until the error is removed. Note that at no time does time run backwards with this method.
> On most Unix systems it is not desirable to step the system clock, because many programs rely on time advancing monotonically forwards.
> When the chronyd daemon is initially started, it is possible that the system clock is considerably in error. Attempting to correct such an error by slewing might not be sensible, since it might take several hours to correct the error by this means.

Instead of it, we just check that chrony is working using metric `node_timex_sync_status`. And to check the results of chrony work, we check `node_ntp_offset_seconds` in NodeTimeOutOfSync alert.

## Checklist
- [ ] e2e tests are passed.

## Changelog entries

```changes
section: chrony
type: fix
summary: Refactored alerts.
```

### DIFF
--- a/modules/470-chrony/monitoring/prometheus-rules/chrony.yaml
+++ b/modules/470-chrony/monitoring/prometheus-rules/chrony.yaml
@@ -4,8 +4,8 @@
     expr: abs(node_ntp_offset_seconds)
 
   - alert: NTPDaemonOnNodeDoesNotSynchronizeTime
-    expr: (min by (node) (node_ntp_sanity)) == 0
-    for: 2h
+    expr: (min by (node) (node_timex_sync_status)) == 0
+    for: 30m
     labels:
       severity_level: "5"
       tier: cluster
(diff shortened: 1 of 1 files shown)


RomanenkoDenys (commented):
I don't think that's the best solution. We should discuss this.

RomanenkoDenys (approved):

z9r5:

PR #1907, opened 2022-06-29
[istio] PodMonitor fix for fresh istio version

apolovov:
## Description
PodMonitor fix for data-plane with fresh istio version.

There was PodMonitor's labelSelector which used the `istio.io/rev` label to detect istio data-plane pods. Now there isn't such label. Now we use `service.istio.io/canonical-name`.

## Checklist
- [ ] e2e tests are passed.

## Changelog entries

```changes
section: istio
type: fix
summary: Data-plane metrics scrape fix for fresh istio versions.
```

### DIFF
--- a/ee/modules/110-istio/templates/podmonitor.yaml
+++ b/ee/modules/110-istio/templates/podmonitor.yaml
@@ -35,7 +35,7 @@ spec:
       targetLabel: namespace
   selector:
     matchExpressions:
-    - key: istio.io/rev
+    - key: service.istio.io/canonical-name
       operator: Exists
   namespaceSelector:
     matchNames:


ee/modules/110-istio/templates/podmonitor.yaml
```diff
@@ -35,7 +35,7 @@ spec:
       targetLabel: namespace
   selector:
     matchExpressions:
-    - key: istio.io/rev
+    - key: service.istio.io/canonical-name
```
nabokihms, line 38: What happens if the label will be updated right now before updating the istio?
Do both istio versions set the `service.istio.io/canonical-name` label?

nabokihms (approved):

z9r5:

PR #1935, opened 2022-07-02
[user-authn] Refactor dex probes

nabokihms:

## Description
* Refactor liveness and readiness probes align with the [official Dex helm chart](https://github.com/dexidp/helm-charts/blob/da8d7fe3369b2c9cf89a59ae7f86a6f1cf449cf7/charts/dex/templates/deployment.yaml#L95-L102). The liveness probe endpoint returns an ok message without checking database connectivity.
* Collect metrics from Dex (there are approximately 40 metrics). 
* Add a simple alert that Prometheus can scrape one of Dex targets.

1. Previous Dex liveness probes led to container restarts, which was a cause of pods lying in the backoff loop status
2. At least simple monitoring is required to detect problems with Dex

Dex pod will be restarted.

## Checklist
- [x] e2e tests are passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Refactor Dex probes, and collect metrics from Dex.
impact_level: default
```

### DIFF
--- a/modules/150-user-authn/monitoring/prometheus-rules/dex.yaml
+++ b/modules/150-user-authn/monitoring/prometheus-rules/dex.yaml
@@ -1,7 +1,7 @@
 - name: d8.dex.availability
   rules:
-  - alert: D8DexTargetDown
-    expr: max by (job) (up{job="dex", namespace="d8-user-authn"} == 0)
+  - alert: D8DexAllTargetsDown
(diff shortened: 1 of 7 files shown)


deckhouse-BOaTswain:
Aye, aye, @nabokihms. I've started the workflow for label 'e2e/run/azure'...

name212 (approved):

PR #2025, opened 2022-07-14
[deckhouse]: disruption updates

yalosev:

## Description
This patch allows to manually confirm disruptive updates by setting `deckhouse.update.disruptionApprovalMode` value to `Manual`. If release is marked as disruptive, it will wait in a following state: 
```
  status:
    approved: false
    message: 'Release requires disruption approval (`kubectl annotate DeckhouseRelease
      v1-36-0 release.deckhouse.io/disruption-approved=true`): Default IngressNginxController
      version 0.33 will be automatically changed to 1.1. This action will restart
      all controllers with non-specified version'
    phase: Pending
```

until user will annotate it and approve disruption

Give users the ability to control potentially dangerous releases.

## Checklist
- [ ] e2e tests are passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Ability to manually control disruptive releases
```


modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,23 @@
+# DeckhouseRelease details
+
+## Available annotations
+
+| annotation                               | description                                                                        |
+|------------------------------------------|------------------------------------------------------------------------------------|
+| release.deckhouse.io/force               | Apply specified release without any checks. Force deploy.                          |
+| release.deckhouse.io/disruption-approved | Approve release with disruptive changes. Works if update.disruptionMode is Manual. |
+| release.deckhouse.io/approved            | Approve release for deployment. Works if update.mode is Manual.                    |
```
z9r5, line 9:
```suggestion
| `release.deckhouse.io/force`               | Apply specified release without any checks. Force deploy.                          |
| `release.deckhouse.io/disruption-approved` | Approve release with disruptive changes. Works if `update.disruptionMode` is Manual. |
| `release.deckhouse.io/approved`            | Approve release for deployment. Works if `update.mode` is Manual.                    |
```

modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Disruptive release
+
+It's a release with some potentially dangerous changes (change some default value / behavior / docker -> containerd, etc)
+To handle this release, you should add disruption check logic in a release `X-1`, for example - register requirements.DisruptionFunc in init() [example](modules/402-ingress-nginx/hooks/requirements.go)
+And then in a release `X` you should add record `"disruption:$functionName": "true"` to the [requirements.json](requirements.json) file.
+For example:
```
z9r5, line 16:
```suggestion

For example:
```

modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,23 @@
...
+In the release 1.35.0 set disruption check logic into ingressNginx
+In the release 1.36.0 set `"disruption:ingressNginx": "true"` to the requirements.json
+
+## Release requirements
+
+For checking some precondition/requirement - register CheckFunc like [here](modules/402-ingress-nginx/hooks/requirements.go)
+and then add `"$functionName": "$version"` to the [requirements.json](requirements.json) file. (Example: `"ingressNginx": "0.33"`)
```
z9r5, line 23:
```suggestion
For checking some precondition/requirement — register CheckFunc like [here](modules/402-ingress-nginx/hooks/requirements.go)
and then add `"$functionName": "$version"` to the [requirements.json](requirements.json) file. (e.g., `"ingressNginx": "0.33"`)
```

z9r5 (dismissed):

modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
```diff
@@ -80,6 +81,9 @@ func discoverMinimalNginxVersion(input *go_hook.HookInput) error {
 		ctrl := s.(ingressNginxController)
 		if ctrl.Version == "" {
 			ctrl.Version = input.Values.Get("ingressNginx.defaultControllerVersion").String()
+			if ctrl.Version == "0.33" {
+				isDisruptionUpdate = true
+			}
```
shvgn, line 86: We don't we migrate other versions? Like 0.40+.

modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
```diff
@@ -80,6 +81,9 @@ func discoverMinimalNginxVersion(input *go_hook.HookInput) error {
 		ctrl := s.(ingressNginxController)
 		if ctrl.Version == "" {
```
shvgn, line 82: If the controller has explicitly specified "0.33", `isDisruptionUpdate` will not become `true`, and hence `hasDisruptionVersionUpdate` will not become `true`. 

If it is desired, please leave a comment, that we migrate only defaults, not *all* outdated controllers. 

If it is not desired, I suggest not to nest the second `if`

```go
if ctrl.Version == "" {
    ctrl.Version = input.Values.Get("ingressNginx.defaultControllerVersion").String()
}

if ctrl.Version == "0.33" {
    isDisruptionUpdate = true
}
```

modules/402-ingress-nginx/hooks/requirements.go
```diff
@@ -52,5 +56,14 @@ func init() {
...
+	disruptionCheckFunc := func() (bool, string) {
+		reason := ""
+		if hasDisruptionVersionUpdate {
+			reason = "Default IngressNginxController version 0.33 will be automatically changed to 1.1. This action will restart all controllers with non-specified version"
+		}
+		return hasDisruptionVersionUpdate, reason
+	}
+
+	requirements.Register("ingressNginx", checkRequirementFunc)
+	requirements.RegisterDisruptionFunc("disruption:ingressNginx", disruptionCheckFunc)
```
shvgn, line 68: Why do we need to specify `disruption` prefix here explicitly?

modules/020-deckhouse/docs/internal/RELEASE.md
shvgn, line 23:
```suggestion
and then add `"$functionName": "$version"` to the [requirements.json](requirements.json) file. (Example: `"disruption:ingressNginx": "0.33"`)
```

modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,23 @@
...
+| release.deckhouse.io/approved            | Approve release for deployment. Works if update.mode is Manual.                    |
+
+## Disruptive release
+
+It's a release with some potentially dangerous changes (change some default value / behavior / docker -> containerd, etc)
+To handle this release, you should add disruption check logic in a release `X-1`, for example - register requirements.DisruptionFunc in init() [example](modules/402-ingress-nginx/hooks/requirements.go)
+And then in a release `X` you should add record `"disruption:$functionName": "true"` to the [requirements.json](requirements.json) file.
+For example:
+In the release 1.35.0 set disruption check logic into ingressNginx
+In the release 1.36.0 set `"disruption:ingressNginx": "true"` to the requirements.json
```
shvgn, line 18: Imagine stable release channel. Can we add a disruption check in 1.35.5 (not 1.35.0), and update to 1.36.10 for this machinery to work?

shvgn, line 18: Can I add both check and disruption to 1.36.0? What's the point in splitting check and disruption between different versions?

shvgn (commented):
I'd like to have some clarifications.

go_lib/dependency/requirements/requirements.go
```diff
@@ -38,10 +44,31 @@ func Register(key string, f CheckFunc) {
...
 		return true, nil
 	}
+
+	if strings.HasPrefix(key, DisruptionPrefix) {
+		return true, nil
+	}
+
 	f, err := defaultRegistry.GetByKey(key)
 	if err != nil {
 		panic(err)
```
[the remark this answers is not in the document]
yalosev: it's just a protection. if you are iterating over all requirements map without filtering disruption changes shouldn't trigger requirements check here

go_lib/dependency/requirements/requirements.go
```diff
@@ -50,31 +77,59 @@ func CheckRequirement(key, value string, getter ValueGetter) (bool, error) {
...
+// CheckFunc check come precondition, comparing desired value (requirementValue) with current value (getter)
 type CheckFunc func(requirementValue string, getter ValueGetter) (bool, error)
+type DisruptionFunc func() (bool, string)
 
 type ValueGetter interface {
 	Get(path string) gjson.Result
 }
 
 type requirementsResolver interface {
 	Register(key string, f CheckFunc)
```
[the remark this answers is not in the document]
yalosev: makes sense. It was only `Register` method here, I'm just extended this interface. Maybe we should also refactor previous methods

modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
yalosev: Because it's not about deprecating all old controllers. It's a preparation to change only one value [here](https://github.com/deckhouse/deckhouse/blob/main/modules/402-ingress-nginx/openapi/config-values.yaml#L4)

modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
yalosev: If controller is explicitly specified - it's out of this scope. User will have 5 more releases to migrate manually/controllable to the new one. We have an alert about deprecated releases. Picture looks like that:
1. In the release 1.35 we will add this check
2. In the release 1.36 we will change default controller. This will trigger disruption update
3. In the release 1.40 we will remove controllers <1.1. This action will be protected with requirement `"ingressNginx": "1.1"`

modules/402-ingress-nginx/hooks/requirements.go
yalosev: We don't. It would be added automatically if not set. But I decided to spread names here just for eye-simplicity

modules/020-deckhouse/docs/internal/RELEASE.md
yalosev: Well, we are putting all this values into openapi values. They will look and process much more complicated, that why I decided to highlight it with the prefix

modules/020-deckhouse/docs/internal/RELEASE.md
yalosev: We could and this will work but we want to avoid this behavior. Because in a some airgap installation client can jump from 1.35.3 to 1.36.10. And 1.35.3 wouldn't have this machinery, so we will meet a problem. It's just an agreement, cause as I said, machinery will work here

yalosev: You could have
```
"ingressNginx": "1.1",
"disruption:ingressNginx": "true"
```
in the same release. It would work, cause these checks pass in a row

In my mind they are splitted like that:
 - RequirementCheck (like `"ingressNginx": "1.1"`, or `"k8s": "1.19"`) - it's some strict check of outer dependencies which deny a release deploy until a requirement is met. (**hard block**)
 - DisruptionCheck - it's a check, that warn user about some meaningful changes, which are controlled by our code and logic (simple: we are changing some behavior) (**soft block**)

modules/020-deckhouse/docs/internal/RELEASE.md
shvgn: Okay. Nesting is inconvenient. 

What I still miss is the asymmetry in notations. I got it why I am wrong here. It would be more clear to me, if we had more explicit notations. E.g.:

```json
{
  "check:ingressNginx": "1.1",
  "disruption:ingressNginx": "1.1"
}
```

What do you think?

modules/402-ingress-nginx/hooks/requirements.go
shvgn: When I see "ingressNginx" I cannot understand if there is an implicit prefix or no prefix at all. I suggest adding prefixes explicitly to make it less confusing. Probably, we should introduce "check" prefix for it.

modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
shvgn: Got it.

go_lib/dependency/requirements/requirements.go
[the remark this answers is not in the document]
shvgn: As you can see based on my questions, all these check-disruption relationships are not very clear to me. So I suggest make explicit notations for them.

go_lib/dependency/requirements/requirements.go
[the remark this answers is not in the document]
shvgn: The same can be achieved with an error, isn't it? Panic stack trace in logs is less convenient than an error IMO.

modules/402-ingress-nginx/hooks/requirements.go
shvgn: Oh, I don't mean to add them here, only in requirements.json. Method names `RegisterCheck` and `RegisterDisruption` are clean enough.

modules/020-deckhouse/docs/internal/RELEASE.md
shvgn: Ok. The development flow is clear to me:

To have a disruptive update in `1.X.*`, one should do:

1. Make a PR to `1.(X-1).0` with **both** precondition check and update requirement. In `requirements.json' the check should be turned on, but the update requirement is not.
2. Make PR  to `1.X.0` which turns the update requirement on in `requirements.json`.

modules/020-deckhouse/docs/internal/RELEASE.md
shvgn: Let's stick to the domain and terminology a little bit. 

I guess, what we fill in `requirements.json` is not a disruption in any sense. It is a precondition check and an update requirement. We describe only the condition data and the fact of the requirement should be met:

1. precondition **check** is "1.1" for key "ingressNginx"
2. update **requirement** must be met (being "true") for "ingressNginx" to update Deckhouse

And there can be no disruption! 

- If the cluster has already had no default controller version, the check condition has already passed, and the requirement is already met. So there will be no disruption. 
- If a client notices the disruptive alert/notification and satisfies the **requirement**, there will be no disruption. 

So, I would pick more concise wording for this file.

```go
type ... interface {
    SetCoditionCheckFunc(string, func)
    Check(string)

    SetRequirementFunc(string, func)
    IsRequired(string)
}
```

```json
{
    "check:ingressNginx": "1.1",
    "require:ingressNginx": "true" // BTW why not boolean here?
}
```

What do you think?

modules/020-deckhouse/docs/internal/RELEASE.md
shvgn: So, in order to have Deckhouse in working state, we need to precaution clients in `X-1` release, because if the requirement is not met, release `X` will not work. The client will be notified, but Deckhouse will not be stuck because of the inevitable migration. I see.

yalosev:
refactored, with respect of RT conversation

modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,32 @@
...
+- RequirementCheck (like `"ingressNginx": "1.1"`, or `"k8s": "1.19"` in the section `requirements`) - it's some strict check of outer dependencies which deny a release deploy until a requirement is met. (hard block)
+- DisruptionCheck - (like `"ingressNginx"` in the section `disruptions`) - it's a check, that warn user about some meaningful changes, which are controlled by our code and logic (simple: we are changing some behavior / default value) (soft block)
+
+### Disruptive release
+
+It's a release with some potentially dangerous changes (change some default value / behavior / docker -> containerd, etc)
+To handle this release, you should add disruption check logic in a release at least `X-1` (previous release), for example — register DisruptionFunc in init() [example](modules/402-ingress-nginx/hooks/requirements.go).
+And add record for a specified release, where this logic will be checked. (e.g.: `"1.36": ["ingressNginx"]` to the [release.yaml](release.yaml) file, section `disruptions`.
+
+Example:
```
shvgn, line 24:
```suggestion
How to add a disruptive change:
```

shvgn (dismissed):
Basically LGTM, although I'd to make naming more concise in metrics and config, and probably docs as well.

go_lib/dependency/requirements/requirements.go
```diff
@@ -20,65 +20,117 @@ import (
...
-func (r *requirementsRegistry) GetByKey(key string) (CheckFunc, error) {
+func (r *requirementsRegistry) RegisterDisruption(key string, f DisruptionFunc) {
+	r.disruptions[key] = f
+}
+
+func (r *requirementsRegistry) GetCheckByKey(key string) (CheckFunc, error) {
 	f, ok := r.checkers[key]
 	if !ok {
-		return nil, fmt.Errorf("check function for %q requirement is not registred", key)
+		return nil, errors.Wrap(ErrNotRegistered, fmt.Sprintf("requirement with a key: %s", key))
```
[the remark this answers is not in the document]
yalosev: hm, then errors.Is() wouldn't work, would it?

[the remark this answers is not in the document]
yalosev: It works, but I will keep Wrap, as more obvious one

shvgn (approved):
LGTM

werf.yaml
```diff
@@ -589,24 +589,40 @@ shell:
...
+      EOF
+      {{ end }}
+git:
+  - add: /
+    to: /deckhouse
+    includePaths:
+      - release.yaml
+---
+image: release-channel-version
+from: spotify/scratch
```
nabokihms, line 617:
```suggestion
from: {{ .Images.BASE_SCRATCH }}
```
It seems like a bug. This image must be built from the [`BASE_SCRATCH`](https://github.com/deckhouse/deckhouse/blob/51a1b3a4b158a05ef2e7042451d2ada9cf52e5b5/candi/image_versions.yml#L27) image.

modules/020-deckhouse/docs/USAGE_RU.md
```diff
@@ -76,6 +76,24 @@ deckhouse: |
 kubectl patch DeckhouseRelease v1-26-0 --type=merge -p='{"approved": true}'
 ```
 
+### Ручное подтверждение потенциально опасных(disruptive) обновлений
```
nabokihms, line 79:
```suggestion
### Ручное подтверждение потенциально опасных (disruptive) обновлений
```

modules/020-deckhouse/docs/USAGE_RU.md
```diff
@@ -76,6 +76,24 @@ deckhouse: |
...
+
+При необходимости возможно включить ручное подтверждение потенциально опасных(disruptive) обновлений (которые меняют значения по-умолчанию или поведение некоторых модулей). Сделать это можно следующим образом:
```
nabokihms, line 81:
```suggestion
При необходимости возможно включить ручное подтверждение потенциально опасных (disruptive) обновлений (которые меняют значения по-умолчанию или поведение некоторых модулей). Сделать это можно следующим образом:
```

modules/402-ingress-nginx/hooks/requirements.go
```diff
@@ -52,5 +56,14 @@ func init() {
 		return true, nil
 	}
 
-	requirements.Register("ingressNginx", f)
+	disruptionCheckFunc := func() (bool, string) {
+		reason := ""
+		if hasDisruptionVersionUpdate {
+			reason = "Default IngressNginxController version 0.33 will be automatically changed to 1.1. This action will restart all controllers with non-specified version"
```
nabokihms, line 62:
```suggestion
			reason = "Default IngressNginxController version 0.33 will be automatically changed to 1.1. This action will restart all controllers with non-specified version."
```

modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
```diff
@@ -71,6 +71,7 @@ func discoverMinimalNginxVersion(input *go_hook.HookInput) error {
 
 	var minVersion *semver.Version
 	classVersionMap := make(map[string]*semver.Version)
+	var isDisruptionUpdate = false
```
nabokihms, line 74: NIT
```suggestion
	var isDisruptionUpdate bool
```

nabokihms: Or
```
isDisruptionUpdate := false
```

modules/020-deckhouse/docs/USAGE.md
```diff
@@ -76,6 +76,24 @@ Manual confirmation of the update to the version `v1.26.0`:
...
+  ...
+  releaseChannel: Stable
+  update:
+    disruptionApproval: Manual
+```
+
+In this mode, it will be necessary to confirm each minor disruptive update with an annotation:
+
+```shell
+kubectl annotate DeckhouseRelease v1-36-0 release.deckhouse.io/disruption-approved=true
```
nabokihms, line 94: How does it work if I have both `update.disruptionApproval` and `update.approvalMode` set to `Manual`? Do I need to approve release twice? Maybe there is something we can optimize in this case.

modules/020-deckhouse/docs/USAGE.md
```diff
@@ -76,6 +76,24 @@ Manual confirmation of the update to the version `v1.26.0`:
...
+### Manual disruption update confirmation
+
+If necessary, it is possible to enable manual confirmation of disruptive updates (updates that change the default values or behavior). This can be done as follows:
+
+```yaml
+deckhouse: |
+  ...
+  releaseChannel: Stable
+  update:
+    disruptionApproval: Manual
```
nabokihms, line 88:
```suggestion
    disruptionApprovalMode: Manual
```
wdyt?

nabokihms (changes requested):

modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,32 @@
...
+
+| annotation                               | description                                                                        |
+|------------------------------------------|------------------------------------------------------------------------------------|
+| `release.deckhouse.io/force`               | Apply specified release without any checks. Force deploy.                          |
+| `release.deckhouse.io/disruption-approved` | Approve release with disruptive changes. Works if `update.disruptionApproval` is Manual. |
+| `release.deckhouse.io/approved`            | Approve release for deployment. Works if `update.mode` is Manual.                    |
+
+## Difference between disruption check and requirement check
+
+All release settings, requirements and disruptions are stored in the file [release.yaml](release.yaml)
```
z9r5, line 13:
```suggestion
All release settings, requirements, and disruptions are stored in the [release.yaml](release.yaml) file.
```

modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,32 @@
...
+
+- RequirementCheck (like `"ingressNginx": "1.1"`, or `"k8s": "1.19"` in the section `requirements`) - it's some strict check of outer dependencies which deny a release deploy until a requirement is met. (hard block)
+- DisruptionCheck - (like `"ingressNginx"` in the section `disruptions`) - it's a check, that warn user about some meaningful changes, which are controlled by our code and logic (simple: we are changing some behavior / default value) (soft block)
```
z9r5, line 16:
```suggestion
- RequirementCheck (like `"ingressNginx": "1.1"`, or `"k8s": "1.19"` in the section `requirements`) — it's some strict check of outer dependencies which deny a release deploy until a requirement is met (hard block).
- DisruptionCheck - (like `"ingressNginx"` in the section `disruptions`) — it's a check that warn a user about some meaningful changes, which are controlled by our code and logic (simple: we are changing some behavior or default value) (soft block).
```

modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Disruptive release
+
+It's a release with some potentially dangerous changes (change some default value / behavior / docker -> containerd, etc)
```
z9r5, line 20:
```suggestion
It's a release with some potentially dangerous changes (e.g. which changes some default value, behavior, or changes docker to containerd, etc.).
```

modules/020-deckhouse/docs/internal/RELEASE.md
```diff
@@ -0,0 +1,32 @@
...
+To handle this release, you should add disruption check logic in a release at least `X-1` (previous release), for example — register DisruptionFunc in init() [example](modules/402-ingress-nginx/hooks/requirements.go).
+And add record for a specified release, where this logic will be checked. (e.g.: `"1.36": ["ingressNginx"]` to the [release.yaml](release.yaml) file, section `disruptions`.
+
+How to add a disruptive change:
+- In current release (say, 1.35.0) set disruption check logic into the `ingressNginx` function;
```
z9r5, line 25:
```suggestion
- In current release (say, 1.35.0) set disruption check logic into the `ingressNginx` function.
```

z9r5 (dismissed):

modules/402-ingress-nginx/hooks/requirements.go
yalosev: `,` after 1.1

yalosev: changed to comma

modules/020-deckhouse/docs/USAGE.md
yalosev: Yes, we will approve twice. It's different flows we don't need to mix them.
First: we approve a release
Second: if the release has disruption changes - we approve our confirmation (after one more brain storm process ;)

modules/020-deckhouse/docs/USAGE.md
yalosev: Done

nabokihms (approved):

PR #2067, opened 2022-07-20
[monitoring-ping] consider source host packet loss when monitoring ping

marriva:
## Description

Change expression for `NodePingPacketLoss` alert to consider the packet loss of source host when monitoring ping

The existing expression aggregates ping packet loss across the entire cluster and does not consider losses from a particular node.

We should receive an alert when some node in cluster cannot not reach another node.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: monitoring-ping
type: fix
summary: consider source host packet loss when monitoring ping
```

### DIFF
--- a/modules/340-monitoring-ping/monitoring/prometheus-rules/node-ping.yaml
+++ b/modules/340-monitoring-ping/monitoring/prometheus-rules/node-ping.yaml
@@ -3,17 +3,17 @@
   - alert: NodePingPacketLoss
     expr: >-
       (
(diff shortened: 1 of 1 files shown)


marriva:
@nabokihms take a look, please

nabokihms:
Could you please elaborate more on why do we need this change?

The only problem I see is that for clusters with, e.g., 150 nodes users will have 150 * 150 possible alerts, and if there is a problem with a single node network - users will also receive 150 alerts instead of one. Maybe we can solve your problem the other way?

PR #2111, opened 2022-07-26
[node-local-dns] Revert service account for prevent pod stuck in Terminating state

name212:
Signed-off-by: Nikolay Mitrofanov <nikolay.mitrofanov@flant.com>

## Description
Kubelet has issue when service-account mounted to pod is deleted before pod is removed. And pod hangs in Terminating state.
We moved node-local-dns to kube-system namespace in 1.34 release. Helm removes `node-local-dns` service account before some pods were terminated, and a lot of pods can be stuck in Terminating state. To fix it, we should restart kubelet on every node with stuck pod.

Exclude manual actions during moving node-local-dns to kube-system namespace.

Node local dns pods should not stuck in terminating status in `d8-system` namespace

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-local-dns
type: fix
summary: Revert service account to prevent Pod from getting stuck in a Terminating state.
impact_level: default
```

### DIFF
--- a/ee/modules/350-node-local-dns/templates/rbac-for-us.yaml
+++ b/ee/modules/350-node-local-dns/templates/rbac-for-us.yaml
@@ -1,4 +1,15 @@
 ---
+# Needed for migration of node-local-dns from d8-system namespace to kube-system.
+# Kubelet has issue when config map mounted to pod is deleted before pod is removed.
+# And pod hangs in `Terminating` state.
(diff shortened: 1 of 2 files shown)


yalosev (approved):

nabokihms (approved):

z9r5:
Cherry-picked into release-1.34

PR #2139, opened 2022-07-29
[candi] fix discover node ip bashible step

RomanenkoDenys:
Signed-off-by: Denis Romanenko <denis.romanenko@flant.com>

## Description

Exclude virtual interfaces like flannel, cni0 etc. from node ip autodiscover.

When staticClusterConfiguration is not set, we try to autodiscover node ip address.
If the node has only one interface discover is ok, but when flannel is deployed in cluster, the second interface appears and autodiscover step fails.
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: fix discover node ip bashible step
impact_level: low
```

### DIFF
--- a/candi/bashible/common-steps/node-group/003_discover_node_ip.sh.tpl
+++ b/candi/bashible/common-steps/node-group/003_discover_node_ip.sh.tpl
@@ -34,15 +34,14 @@ fi
 {{- end }}
 
 {{- if eq .nodeGroup.nodeType "Static" }}
-ip_route_get_cmd="ip route show scope link proto kernel"
   {{- if and (hasKey .nodeGroup "static") (hasKey .nodeGroup.static "internalNetworkCIDRs")}}
 internal_network_cidrs={{ .nodeGroup.static.internalNetworkCIDRs | join " " | quote }}
   {{- end }}
-
 if [[ -z "$internal_network_cidrs" ]]; then
   # if internal network cidrs is not set, and the node has one interface, use its network as internal_network_cidr
-  if [[ "$($ip_route_get_cmd | wc -l)" -eq 1 ]]; then
-    internal_network_cidrs="$($ip_route_get_cmd | awk '{print $1}')"
+  ip_route_get_cmd="$(ip route show scope link proto kernel | grep -vE 'cni|docker|flannel|cilium|nodelocaldns')"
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:
Aye, aye, @RomanenkoDenys. I've started the workflow for label 'e2e/run/static'...

candi/bashible/common-steps/node-group/003_discover_node_ip.sh.tpl
```diff
@@ -34,15 +34,14 @@ fi
...
-ip_route_get_cmd="ip route show scope link proto kernel"
   {{- if and (hasKey .nodeGroup "static") (hasKey .nodeGroup.static "internalNetworkCIDRs")}}
 internal_network_cidrs={{ .nodeGroup.static.internalNetworkCIDRs | join " " | quote }}
   {{- end }}
-
 if [[ -z "$internal_network_cidrs" ]]; then
   # if internal network cidrs is not set, and the node has one interface, use its network as internal_network_cidr
-  if [[ "$($ip_route_get_cmd | wc -l)" -eq 1 ]]; then
-    internal_network_cidrs="$($ip_route_get_cmd | awk '{print $1}')"
+  ip_route_get_cmd="$(ip route show scope link proto kernel | grep -vE 'cni|docker|flannel|cilium|nodelocaldns')"
```
konstantin-axenov, line 42: I don't get why we can't use default route to discover a single interface. Also, I don't think specifying exact virtual interfaces is a good idea. Why not to use?
```
ls -la /sys/class/net | grep "\->" | grep -v virtual
lrwxrwxrwx  1 root root 0 Aug  3 12:40 ens3 -> ../../devices/pci0000:00/0000:00:03.0/virtio0/net/ens3
```

RomanenkoDenys: Using the default route for iface detection is not a good idea, because in closed contours the default route may not be available due to security concerns.
But using /sys/class/net is, in my opinion, a good solution. I will use it.

deckhouse-BOaTswain:
Aye, aye, @konstantin-axenov. I've started the workflow for label 'e2e/run/static'...

konstantin-axenov (approved):

PR #2146, opened 2022-08-01
[deckhouse] Fix webhook handler certificate

yalosev:

## Description
Recreate TLS certificate for webhook handler.

Certificate for webhook handler has 1 year expire time when other certificates have 10 years. We should recreate TLS certificate for webhook handler. Also generic hook for creating certificates was changed to handle more complex solutions and check certificate expire time

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Fix webhook handler TLS certificate expire time
impact: Webhook handler will restart. During the handler restart deckhouse controller could generate a few error messages when it will not be able to access the webhook. It should be resolved in the next 15 seconds ish.
impact_level: high
```


modules/020-deckhouse/hooks/generate_selfsigned_ca.go
```diff
@@ -20,97 +20,38 @@ import (
...
-		var err error
-		sefSignedCA, err := certificate.GenerateCA(input.LogEntry, webhookServiceHost)
-		if err != nil {
-			return fmt.Errorf("cannot generate selfsigned ca: %v", err)
-		}
+const (
+	webhookServiceHost = "webhook-handler.d8-system.svc"
+)
 
+var _ = hooks.RegisterInternalTLSHook(hooks.GenSelfSignedTLSHookConf{
```
name212, line 31: Looks good! But i should remind that old hook generated certificate for 1 year, `go_lib/hooks/internal_tls/hook.go` generates certificate for 10 years. I think it is not a problem but it changes hook semantic.

go_lib/hooks/internal_tls/hook.go
```diff
@@ -55,6 +58,23 @@ type GenSelfSignedTLSHookConf struct {
 	// Data in values store as plain text
 	// In helm templates you need use `b64enc` function to encode
 	FullValuesPathPrefix string
+
+	// You can set Paths for internal values explicitly or use FullValuesPathPrefix
+	CAValuesPath   string
```
name212, line 63: When I did this hook, we discussed with @nabokihms about this paths. I as you added internal paths in the configuration and we decided that is overkill. Better way use `FullValuesPathPrefix` and change paths in module.

go_lib/hooks/internal_tls/hook.go
```diff
@@ -31,6 +31,9 @@ import (
 type GenSelfSignedTLSHookConf struct {
 	// SANs - list of domains to include into certificate
 	SANs []string
+	// SANsGenerator dynamic SANs generator based in hook input (e.x.: values like cluster domain)
+	SANsGenerator func(input *go_hook.HookInput) []string
```
name212, line 35: I look to it and i have one idea.

We can pass  `GenSelfSignedTLSHookConf` generator (`func (*go_hook.HookInput) GenSelfSignedTLSHookConf` ) to `RegisterInternalTLSHook` function. In `RegisterInternalTLSHook` call generator and pass config to `genSelfSignedTLS`

In this case we can use only `SANs` field. It a much cleaner way than use two fields to getting SAN's.

Yep, we will change func signature, but it easy to fix in another places, because we have only 2 places to use this hook: `000-common` (it is integrate tests for internal hooks) and `301-prometheus-metrics-adapter` and in there places you only wrap configuration to function.

modules/020-deckhouse/hooks/generate_selfsigned_ca.go
```diff
@@ -20,97 +20,38 @@ import (
 	"fmt"
```
name212, line 20: Because we work here, we can rename hook name, for example, to `generate_webhook_tls_certificate`. When i look at `generate_selfsigned_ca` name i do not understand what hook will do.

modules/020-deckhouse/hooks/generate_selfsigned_ca.go
name212: @nabokihms I do not know, is it problem for us?

yalosev: It was a main goal of this patch :) self-signed certificate for webhook handler was created for 1 year when it should be created for 10 years.
All other changes - it's a try to make generic solution for working with self-signed certificates, with an expiration check also

name212:
As I remember while webhook is restarting we will get error about convert and validation our resources. Our release manager often was triggered for this errors. I think we need to add to impact message next information that while webhook is restarting we can get errors in deckhouse-controller.

name212 (changes requested):
I've added some requests to code improvements.

go_lib/hooks/internal_tls/hook.go
yalosev: Well. This is not comfortable because:
a) We are losing a flexibility
b) `FullValuesPathPrefix` gives us a plain structure while the vast majority of our modules have a different structure, like:
```
      customCertificateData:
        type: object
        properties:
          tls.crt:
            type: string
            x-examples:
              - plainstring
          tls.key:
            type: string
            x-examples:
              - plainstring
          ca.crt:
            type: string
            x-examples:
              - plainstring
```
or 
```
          serverCerts:
            type: object
            default: {}
            properties:
              ca:
                type: string
                x-examples: ["---ROOT PEM---"]
              key:
                type: string
                x-examples: ["---KEY PEM---"]
              cert:
                type: string
                x-examples: ["---CERT PEM---"]
```

Simplify them to a plain struct will create more chaos imho, look at cilium-hubble module for example, it has a lot of certs.


That's why I kept backward compatibility but I want to have an ability to set values path explicitly

modules/020-deckhouse/hooks/generate_selfsigned_ca.go
name212: oh, ok. i've resolved this

go_lib/hooks/internal_tls/hook.go
yalosev: It's overcomplicated. We will pass the Factory to `RegisterInternalTLSHook` and then to the handlerFunc of that hook. Then we will return a function from the returning function there. Seems like callback hell :)

My first idea was to make SANs generator like an variadic args `Option` but then I analyzed the config structure.
`CN` - is static
`Namespace` - is static
`TLSSecretName` - is static
`FullValuesPathPrefix` or `ValuesPaths` - are static

and only `SANs` could be dependable on a values input. That's why I decided to put generator near the SANS field. It's easy to observe it when you are researching the conf struct in IDE

modules/020-deckhouse/hooks/generate_selfsigned_ca.go
yalosev: Makes sense

go_lib/hooks/internal_tls/hook.go
name212: We are waiting @nabokihms  here.

go_lib/hooks/internal_tls/hook.go
name212: > Seems like callback hell 
Ok, I've accept your conclusion.

> and only SANs could be dependable on a values input. That's why I decided to put generator near the SAN's field. It's easy to observe it when you are researching the conf struct in IDE
Ok, then I suggest to remove SAN's field and stay only generator. I am confused about two way to pass SAN's and i think we need to stay only one.

deckhouse-BOaTswain:

deckhouse-BOaTswain:
Aye, aye, @z9r5. I've started the workflow for label 'e2e/run/azure'...

go_lib/hooks/internal_tls/hook.go
nabokihms, line 35: For certificates, signed by Kubernetes, we used a diffirent logic for the same purpose:

https://github.com/deckhouse/deckhouse/blob/c59c3dd101035b233ad35f28bc3f26eabcc11168/modules/150-user-authn/hooks/order_dex_tls_certificate.go#L31-L36

This logic should be either applied to this hook or changed according to new standards. I personally do not like dynamic san creation from any value, but using cluster and public domains is ok.

go_lib/hooks/internal_tls/hook.go
```diff
@@ -121,17 +140,33 @@ func genSelfSignedTLS(conf GenSelfSignedTLSHookConf) func(input *go_hook.HookInp
 		} else {
 			// Certificate is in the snapshot => load it.
 			cert = input.Snapshots["secret"][0].(certificate.Certificate)
+			// update certificate if less than 6 month left. We create certificate for 10 years, so it looks acceptable
+			// and we don't need to create Crontab schedule
+			expiring, err := certificate.IsCertificateExpiringSoon([]byte(cert.Cert), 4380*time.Hour) // 6 month
```
nabokihms, line 145:
```suggestion
			expiring, err := certificate.IsCertificateExpiringSoon([]byte(cert.Cert), 4380*time.Hour) // 6 month
```
Let's make the IsCertificateExpiringSoon function accept a certificate as a string if it is more convenient.

go_lib/hooks/internal_tls/hook.go
nabokihms: I think we even want to go further and use the same structure in all modules. It sounds like an idea for significant refactoring.

go_lib/hooks/internal_tls/hook.go
yalosev: But now it's realized in [another way](https://github.com/deckhouse/deckhouse/blob/main/go_lib/hooks/internal_tls/hook.go#L127) and is used like [this](https://github.com/deckhouse/deckhouse/blob/main/modules/301-prometheus-metrics-adapter/openapi/values.yaml#L24).

Your case will look like:
```
a:
  myCertPEM: string
  myCertCA: string
  myCertKey: string
b:
  myCertPEM: string
  myCertCA: string
  myCertKey: string
  anotherCertPEM: string
  anotherCertCA: string
  anotherCertKey: string
```

yalosev: We can change it like this:
```
ca = strings.Join([]string{FullValuesPathPrefix, "ca"}, ".")
key = strings.Join([]string{FullValuesPathPrefix, "key"}, ".")
cert = strings.Join([]string{FullValuesPathPrefix, "cert"}, ".")
```

then it would work like a charm

nabokihms: Let's go with your proposal then. Looks great.

nabokihms: Maybe we also should trim dots at the end of `FullValuesPathPrefix`.

go_lib/hooks/internal_tls/hook.go
nabokihms: So, it was my mistake. This function is already used in several places.

deckhouse-BOaTswain:

deckhouse-BOaTswain:

yalosev:
Reffactor all! %)

go_lib/hooks/tls_certificate/internal_tls.go
```diff
@@ -0,0 +1,267 @@
...
+
+		// Note that []byte values will be encoded in base64. Use strings here!
+		input.Values.Set(caPath, cert.CA)
+		input.Values.Set(certPath, cert.Cert)
+		input.Values.Set(keyPath, cert.Key)
+		return nil
+	}
+}
+
+func isOutdatedCert(certData string, desiredSANSs []string) (bool, error) {
```
nabokihms, line 193: From the function's name, devs can guess that it only checks for the expiration of a certificate, which is not valid. Let's consider changing the name to something more suitable.

modules/301-prometheus-metrics-adapter/hooks/pma_server_tls.go
```diff
@@ -16,18 +16,23 @@ limitations under the License.
 
 package hooks
 
-import hooks "github.com/deckhouse/deckhouse/go_lib/hooks/internal_tls"
+import (
+	hooks "github.com/deckhouse/deckhouse/go_lib/hooks/tls_certificate"
```
nabokihms, line 20: NIT: It seems that the alias does not help to clarify the situation. I'd like to have the tls_certificate here.

go_lib/hooks/tls_certificate/internal_tls.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/flant/shell-operator/pkg/kube_events_manager/types"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/utils/net"
+
+	"github.com/deckhouse/deckhouse/go_lib/certificate"
+)
+
+const (
+	certOutdateDuration = 4380 * time.Hour // 6 month
```
nabokihms, line 34: Despite this constant being internal, I'd like to clarify that this is the value for self-signed certs.

go_lib/hooks/tls_certificate/internal_tls.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	if time.Until(cert.NotAfter) < certOutdateDuration {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func generateNewTLS(input *go_hook.HookInput, cn string, sans []string) (certificate.Certificate, error) {
```
nabokihms, line 241:
```suggestion
func generateNewSelfSignedTLS(input *go_hook.HookInput, cn string, sans []string) (certificate.Certificate, error) {
```
Because it now lies in the generic tls certificate library (but there is already a function with this name :thinking:).

nabokihms (approved):
Looks good. I left some comments, but they are all about variables and function names.

modules/301-prometheus-metrics-adapter/hooks/pma_server_tls.go
yalosev: autorefactor

nabokihms (approved):

go_lib/hooks/tls_certificate/internal_tls.go
```diff
@@ -0,0 +1,277 @@
...
+		// Note that []byte values will be encoded in base64. Use strings here!
+		input.Values.Set(caPath, cert.CA)
+		input.Values.Set(certPath, cert.Cert)
+		input.Values.Set(keyPath, cert.Key)
+		return nil
+	}
+}
+
+// check certificate duration and SANs list
+func isIrrelevantCert(certData string, desiredSANSs []string) (bool, error) {
```
shvgn, line 203: It's better to keep single-side connotation between booleans and meanings. Now positive side 'true' is on negative side 'irrelevance'. We have to flip booleans in head.

I propose to keep it direct:

```go
certOk, err := checkCARelevance(...)
caOk, err := checkCertRelevance(...)
```

go_lib/hooks/tls_certificate/order_certificate.go
```diff
@@ -335,7 +320,7 @@ func shouldGenerateNewCert(cert []byte, request OrderCertificateRequest, duratio
 	return false, nil
 }
 
-func compareArrays(a, b []string) bool {
+func arrayAreEqual(a, b []string) bool {
```
shvgn, line 323: One of

```suggestion
func arraysAreEqual(a, b []string) bool {
```

```suggestion
func slicesAreEqual(a, b []string) bool {
```

modules/020-deckhouse/hooks/generate_webhook_tls_certificate.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright 2021 Flant JSC
```
shvgn, line 2: Probably,

```suggestion
Copyright 2022 Flant JSC
```

modules/020-deckhouse/hooks/generate_webhook_tls_certificate_test.go
```diff
@@ -109,9 +105,108 @@ var _ = Describe("Deckhouse hooks :: generate_selfsigned_ca ::", func() {
...
 })
+
+var (
+	actualCert = ""
+	actualCA   = ""
+	actualKey  = ""
+
+	expiredCert = ""
+	expiredCA   = ""
+	expiredKey  = ""
```
shvgn, line 137: Just... why? Switching global values from generators seems so indirect. You could use returned values instead and their fields.

shvgn (changes requested):
Nitpicking, but still want to have those little changes

modules/020-deckhouse/hooks/generate_webhook_tls_certificate_test.go
yalosev: Well, I have changed it but in my opinion it looks more dirty, too many return parameters. But I don't care too much about tests structure, so, it's up to you

go_lib/hooks/internal_tls/hook.go
yalosev: Done

go_lib/hooks/internal_tls/hook.go
yalosev: another PR I guess

go_lib/hooks/internal_tls/hook.go
```diff
@@ -31,6 +31,9 @@ import (
 type GenSelfSignedTLSHookConf struct {
 	// SANs - list of domains to include into certificate
 	SANs []string
```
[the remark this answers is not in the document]
yalosev: Done

go_lib/hooks/tls_certificate/order_certificate.go
yalosev: Done

modules/020-deckhouse/hooks/generate_webhook_tls_certificate.go
yalosev: Done. It was an old hook so usually I don't change the copyright there but nvm

go_lib/hooks/tls_certificate/internal_tls.go
yalosev: It depends on workflow. In this case we are not interested in valid certs, only in invalid.
`CheckXXX` functions are too wide-spreaded and if they return anything except error, it makes a code blurred. Check is successful (has no error), why we need one more variable for it. So, it can be used but I try to avoid such kind of functions.
It was `IsOutdatedCert` before but then we have added the SANs check to it and have decided to set the name like this. I assume we can keep it and rename it later on some function reuse action, if needed.

shvgn: I would not argue. Just saying that `notOk, err := f()` is less convenient than `ok, err := f()`

name212 (approved):

modules/020-deckhouse/hooks/generate_webhook_tls_certificate_test.go
shvgn: Thank you!

PR #2159, opened 2022-08-04
[upmeter] Tweak agent HTTP timeout

shvgn:
## Description

Set more sane timeouts to upmeter-agent sender.

Reduces dummy logs about reached timeout and request cancellations.

Reduced or diminished logging or cancelled requests in upmeter-agent

## Checklist

## Changelog entries
```changes
section: upmeter
type: fix
summary: Tweak HTTP sending timeouts to reduce cancellation logs
impact_level: low
```

### DIFF
--- a/modules/500-upmeter/images/upmeter/cmd/upmeter/parse_args.go
+++ b/modules/500-upmeter/images/upmeter/cmd/upmeter/parse_args.go
@@ -94,13 +94,13 @@ func parseAgentArgs(cmd *kingpin.CmdClause, config *agent.Config) {
 		Default("false").
 		BoolVar(&config.ClientConfig.TLS)
 
-	cmd.Flag("send-interval", "The interval for sending episodes to the server.").
-		Envar("UPMETER_SEND_INTERVAL").
+	cmd.Flag("export-interval", "Exporting interval when sending from WAL.").
+		Envar("UPMETER_EXPORT_INTERVAL").
 		Default("1s").
 		DurationVar(&config.Interval)
 
-	cmd.Flag("send-timeout", "Sending response timeout before retry.").
-		Envar("UPMETER_SEND_TIMEOUT").
+	cmd.Flag("export-timeout", "Exporting response timeout before retry.").
+		Envar("UPMETER_EXPORT_TIMEOUT").
 		Default("5s").
 		DurationVar(&config.ClientConfig.Timeout)
 
--- a/modules/500-upmeter/images/upmeter/cmd/upmeter/parse_args.go
+++ b/modules/500-upmeter/images/upmeter/cmd/upmeter/parse_args.go
@@ -96,8 +96,13 @@ func parseAgentArgs(cmd *kingpin.CmdClause, config *agent.Config) {
 
 	cmd.Flag("period", "The period of episodes sending to server, and at the same the client timeout.").
 		Envar("UPMETER_SEND_INTERVAL").
-		Default("5s").
+		Default("1s").
 		DurationVar(&config.Interval)
+
+	cmd.Flag("period", "The period of episodes sending to server, and at the same the client timeout.").
+		Envar("UPMETER_SEND_TIMEOUT").
+		Default("5s").
+		DurationVar(&config.ClientConfig.Timeout)
 
 	// Database
 	cmd.Flag("db-path", "SQLite file path.").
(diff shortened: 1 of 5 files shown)


modules/500-upmeter/images/upmeter/pkg/agent/sender/client.go
```diff
@@ -142,15 +144,17 @@ func createHttpTransport(caPath string) (*http.Transport, error) {
...
-		TLSClientConfig: &tls.Config{
-			RootCAs: caCertPool,
-		},
-	}
+	tr.TLSClientConfig = &tls.Config{RootCAs: caCertPool}
 
 	return tr, nil
 }
 
+func newTransport(timeout time.Duration) *http.Transport {
```
nabokihms, line 152: It is unclear that the timeout argument is actually a ResponseHeaderTimeout.

modules/500-upmeter/images/upmeter/pkg/agent/sender/client.go
```diff
@@ -89,11 +90,14 @@ func (c *Client) Send(reqBody []byte) error {
...
-func NewHttpClient(config *ClientConfig, timeout time.Duration) *http.Client {
-	client, err := createSecureHttpClient(config.TLS, config.CAPath, timeout)
+func NewHttpClient(config *ClientConfig) *http.Client {
+	client, err := createSecureHttpClient(config.TLS, config.CAPath, config.Timeout)
 	if err != nil {
 		log.Errorf("falling back to default HTTP client: %v", err)
-		return &http.Client{Timeout: timeout}
+		return &http.Client{
+			Timeout:   config.Timeout,
+			Transport: newTransport(config.Timeout / 2),
```
nabokihms, line 99: 1. Why do we set timeout / 2 to read headers? For example, if there are only headers in the response, it's ok to spend the whole time reading headers.

2. Shouldn't other timeouts be set, e.g., connection timeout, handshake timeout?

3. There is the same division (timeout / 2) in two places. Maybe it is better to put this logic into the `newTransport` function? It makes more sense to me cos we will accept some generic timeout and then describe the logic of why we use `timeout / 2` for headers inside the function.

modules/500-upmeter/images/upmeter/cmd/upmeter/parse_args.go
```diff
@@ -94,10 +94,15 @@ func parseAgentArgs(cmd *kingpin.CmdClause, config *agent.Config) {
 		Default("false").
 		BoolVar(&config.ClientConfig.TLS)
 
-	cmd.Flag("period", "The period of episodes sending to server, and at the same the client timeout.").
-		Envar("UPMETER_PERIOD").
+	cmd.Flag("send-interval", "The period of episodes sending to server, and at the same the client timeout.").
```
nabokihms, line 97:
```suggestion
	cmd.Flag("send-interval", "The period of sending episodes to the upmeter server.").
```

modules/500-upmeter/images/upmeter/cmd/upmeter/parse_args.go
```diff
@@ -94,10 +94,15 @@ func parseAgentArgs(cmd *kingpin.CmdClause, config *agent.Config) {
...
+		Envar("UPMETER_SEND_INTERVAL").
 		Default("1s").
-		DurationVar(&config.Period)
+		DurationVar(&config.Interval)
+
+	cmd.Flag("send-timeout", "The period of episodes sending to server, and at the same the client timeout.").
```
nabokihms, line 102:
```suggestion
	cmd.Flag("send-timeout", "The timeout of sending episodes to the upmeter server.").
```

modules/500-upmeter/images/upmeter/pkg/agent/sender/client.go
shvgn: Let it be just "a transport timeout"! it is reused in multiple places.

modules/500-upmeter/images/upmeter/pkg/agent/sender/client.go
shvgn: 1. Agreed. I removed the division.
2. I agree. That's why there was a copy of http.DefaultClient, where all the timeouts are set. And I changed only one that matters. It is done differently now, anyway.
3. I'd rather keep it explicit in two places than changing passed value implicitly under the hood. You pass a value, but it is halved! Anyway, I removed the division due to the first point.

modules/500-upmeter/images/upmeter/cmd/upmeter/parse_args.go
shvgn: The text has changed

modules/500-upmeter/images/upmeter/cmd/upmeter/parse_args.go
shvgn: The text has changed

nabokihms (approved):

PR #2165, opened 2022-08-05
[testing] support focusing for matrix tests linter

diafour:
## Description

Calculate focused modules before running module linter in matrix tests.

Ignore errors in other modules while testing one module.

Running 

```
FOCUS=openvpn go test ./testing/matrix/ -v
```

should not complain about errors in other modules.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: fix
summary: support focusing for matrix tests linter
impact_level: low
```

### DIFF
--- a/testing/matrix/linter/rules/modules/modules.go
+++ b/testing/matrix/linter/rules/modules/modules.go
@@ -142,7 +142,7 @@ func helmignoreModuleRule(name, path string) errors.LintRuleError {
 	return errors.EmptyRuleError
 }
 
-func GetDeckhouseModulesWithValuesMatrixTests() ([]utils.Module, error) {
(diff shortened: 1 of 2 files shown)


testing/matrix/linter/rules/modules/modules.go
```diff
@@ -167,8 +167,18 @@ func GetDeckhouseModulesWithValuesMatrixTests() ([]utils.Module, error) {
 		modulesPaths = append(modulesPaths, result...)
 	}
 
+	hasFocusNames := len(focusNames) > 0
 	var lintRuleErrorsList errors.LintRuleErrorsList
 	for _, modulePath := range modulesPaths {
+		if hasFocusNames {
```
nabokihms, line 173: This check is not required. Even if the map is empty, it works fine, so there will be fewer clauses in the code without this line.

nabokihms (approved):

Issue #2049, opened 2022-07-18
[istio] D8IstioDataPlanePatchVersionMismatch refactoring — use build_tag metric


apolovov:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

There is revisions_monitoring hook which compares desired and actual data-plane version. It is better to use image_build metric from istio-proxy.

### Proposed Solution

It is better to use image_build metric from istio-proxy.

PR #2181, opened 2022-08-09
[istio] Data plane versions monitoring refactoring

apolovov:
## Description
Data plane versions monitoring refactoring. 

Fixes https://github.com/deckhouse/deckhouse/issues/2049.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: feature
summary: Data plane versions monitoring refactoring. 
```

### DIFF
--- a/ee/modules/110-istio/hooks/revisions_monitoring.go
+++ b/ee/modules/110-istio/hooks/revisions_monitoring.go
@@ -169,6 +169,15 @@ func revisionsMonitoring(input *go_hook.HookInput, dc dependency.Container) erro
 			}
 		}
 
+		// TODO: get rid of everything but this only metric https://github.com/deckhouse/deckhouse/issues/2182
+		labels := map[string]string{
+			"namespace":        istioPod.GetNamespace(),
+			"dataplane_pod":    istioPod.GetName(),
+			"desired_revision": desiredRevision,
+			"revision":         istioPod.getIstioRevision(),
+		}
+		input.MetricsCollector.Set("d8_istio_pod_revision", 1, labels, metrics.WithGroup(revisionsMonitoringMetricsGroup))
+
 		if !internal.Contains(revisionsToInstall, desiredRevision) {
 			// ALARM! Desired revision isn't configured to install
 			labels := map[string]string{
--- a/ee/modules/110-istio/hooks/revisions_monitoring_test.go
+++ b/ee/modules/110-istio/hooks/revisions_monitoring_test.go
@@ -315,7 +315,7 @@ spec:
 			Expect(string(f.LogrusOutput.Contents())).To(HaveLen(0))
 
 			m := f.MetricsCollector.CollectedMetrics()
-			Expect(m).To(HaveLen(7))
+			Expect(m).To(HaveLen(16))
 			Expect(m[0]).To(BeEquivalentTo(operation.MetricOperation{
 				Group:  revisionsMonitoringMetricsGroup,
 				Action: "expire",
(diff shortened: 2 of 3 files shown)


ee/modules/110-istio/hooks/revisions_monitoring.go
```diff
@@ -169,6 +169,15 @@ func revisionsMonitoring(input *go_hook.HookInput, dc dependency.Container) erro
 			}
 		}
 
+		// TODO: get rid of everything but this only metric
```
apolovov, line 172:
```suggestion
		// TODO: get rid of everything but this only metric https://github.com/deckhouse/deckhouse/issues/2182
```

ee/modules/110-istio/hooks/revisions_monitoring.go
```diff
@@ -169,6 +169,15 @@ func revisionsMonitoring(input *go_hook.HookInput, dc dependency.Container) erro
 			}
 		}
 
+		// TODO: get rid of everything but this only metric https://github.com/deckhouse/deckhouse/issues/2182
+		labels := map[string]string{
+			"namespace":        istioPod.GetNamespace(),
+			"dataplane_pod":    istioPod.GetName(),
+			"desired_revision": desiredRevision,
```
shvgn, line 176: Probably "pod" will do? Why do wee need this "dataplane" detail?

ee/modules/110-istio/hooks/revisions_monitoring_test.go
```diff
@@ -315,7 +315,7 @@ spec:
 			Expect(string(f.LogrusOutput.Contents())).To(HaveLen(0))
 
 			m := f.MetricsCollector.CollectedMetrics()
-			Expect(m).To(HaveLen(7))
+			Expect(m).To(HaveLen(16))
```
shvgn, line 318: Completely unobvious what happened here and how I easily see the right number from the tests code

shvgn: I suggest splitting this big test case with 11 test cases, each for one pod case. In every case for a pod, you can check that it either adds to the metric or, importantly, that it is not added. 

For now, it is just hard to track what pods are not added. It took some time for me to see the "ignored" one which is the first, but is should be easily be read and not make people read through this. I still didn't figure out which else pod is missing here. I'd like not to do this kind of hand job. We can structure tests better and reduce boilerplate to track narrow cases instead of all-in-one assertion.

shvgn (changes requested):
The test is hard to understand

ee/modules/110-istio/hooks/revisions_monitoring.go
apolovov: Deckhouse will rewrite it to exported_pod. The "pod" will be smth like "deckhouse-xxx-yyy".

ee/modules/110-istio/hooks/revisions_monitoring_test.go
apolovov: There will be refactoring in https://github.com/deckhouse/deckhouse/issues/2182. So the tests will be refactored too.

ee/modules/110-istio/hooks/revisions_monitoring.go
shvgn: Oh, I see. Thanks

ee/modules/110-istio/hooks/revisions_monitoring_test.go
shvgn: I'll take your word for it!

shvgn (dismissed):
Just because you promised to refactor the test 😄

ee/modules/110-istio/monitoring/prometheus-rules/revisions.yaml
```diff
@@ -69,8 +69,34 @@
         ### pod-wide configuration
         kubectl -n {{$labels.namespace}} get pods -l istio.io/rev={{$labels.desired_revision}}
         ```
-  - alert: D8IstioDataPlanePatchVersionMismatch
-    expr: max by (namespace, revision, actual_sidecar_image_tag) (d8_istio_data_plane_patch_version_mismatch == 1)
+  - alert: D8IstioDataPlaneVersionMismatch
+    expr: |
+      (
```
nabokihms, line 74: For complicated promql, I'd like to see an exaplanation about its parts. If a new developer needs to add something to this query or alter it, it will take some time to figure out what is happening.

apolovov: Done.

nabokihms (approved):

PR #2183, opened 2022-08-10
[ingress-nginx] add validation for controller version

yalosev:

## Description
Add validation webhook

Ingress controllers < 1.1 are forbidden for k8s >= 1.22. We have to notify users about it and deny creation of such controllers

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: feature
summary: Validate ingress controllers compatibility with k8s version
```

### DIFF
--- a/modules/402-ingress-nginx/webhooks/validating/ingressnginxcontroller
+++ b/modules/402-ingress-nginx/webhooks/validating/ingressnginxcontroller
@@ -91,12 +91,14 @@ function __main__() {
   # // End check controller versions
 
   # Only ingress controller >= 1.1 is allowed for kubernetes >= 1.22
-  if semver::lt "$ctrlVersion" "1.0.0"; then
-      k8sVersion=$(kubectl version --short -o json | jq -r '.serverVersion.gitVersion' | sed 's/v//')
-      if semver::ge "$k8sVersion" "1.22.0"; then
-        forbid "only IngressNginxController 1.1 is allowed for k8s >= 1.22"
-        exit 0
-      fi
+  if context::jq -e -r '.review.request.operation == "CREATE"' >/dev/null 2>&1; then
+    if semver::lt "$ctrlVersion" "1.0.0"; then
+        k8sVersion=$(kubectl version --short -o json | jq -r '.serverVersion.gitVersion' | sed 's/v//')
+        if semver::ge "$k8sVersion" "1.22.0"; then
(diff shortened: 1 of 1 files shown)


modules/402-ingress-nginx/webhooks/validating/ingressnginxcontroller
```diff
@@ -90,6 +90,15 @@ function __main__() {
   done < <(kubectl get ingressnginxcontrollers.deckhouse.io -o json | jq --arg class $ctrlClass --arg name $ctrlName -cr '.items[] | select(.spec.ingressClass == $class and .metadata.name != $name) | .spec.controllerVersion // 0.33')
   # // End check controller versions
 
+  # Only ingress controller >= 1.1 is allowed for kubernetes >= 1.22
+  if semver::lt "$ctrlVersion" "1.0.0"; then
+      k8sVersion=$(kubectl version --short -o json | jq -r '.serverVersion.gitVersion' | sed 's/v//')
+      if semver::ge "$k8sVersion" "1.22.0"; then
+        forbid "only IngressNginxController 1.1 is allowed for k8s >= 1.22"
```
nabokihms, line 97: I think this is not good to validate version for UPDATE requests. For any reason, if a user has an outdated version of ingress-nginx controller in the cluster, there is no way to change anything in this controller spec until the version is changed.

yalosev: Lets take the situation:
1. User has updated k8s to 1.22 and have ingress controller 0.49

at this moment his controller wouldn't work
2. User is trying to update node-selector for controller or add more headers - get forbidden error, cause of this validation

Yes, he couldn't update addition spec but his controller is not working anyway and user has one more notification about it (maybe he didn't check ingress pods and don't know anything about the crash). 
Do we have any reason to left a user with a broken controller when he want to change anything else?

nabokihms: All you've described is a side effect.

If notifying users about controllers with deprecated versions is required, it would be best to let them now before they want to update the spec, e.g., by alerting.

If a user already knows that there are outdated controllers in the cluster, it is okay to update anything in these controllers spec.

yalosev: *sigh*

Done

nabokihms (approved):

Issue #2309, opened 2022-08-23
[istio] STRICT mtls mode hinders metrics scraping


apolovov:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.34

### Expected Behavior

Prometheus must be able to scrape metrics from workloads with `PeerAuthentication` `mtls.mode` = STRICT.

### Actual Behavior

It is impossible to scrape metrics without proper client cert issued by istio ca.

### Additional Information

Proposal:
* Create a hook:
  * Subscribe to d8-istio/cacert secret
  * If there is:
    * generate csr: 
      * for a year
      * X509v3 Extended Key Usage — TLS Web Client Authentication
      * X509v3 Basic Constraints — CA:FALSE
      * X509v3 Subject Alternative Name  — URI:spiffe://{ clusterDomain }/ns/d8-monitoring/sa/prometheus
    *  sign in with ca-key.pem and ca-cert.pem
    * and then rotate it every month
* Put the cert to a secret `d8-monitoring/prometheus-scraper-istio-mtls`
* Introduce a new annotation — `prometheus.deckhouse.io/istio-mtls: "true"`. It will force Prometheus using the certificate above for scraping. To make it work in will need to add a selector [here](https://github.com/deckhouse/deckhouse/blob/c69ec246137b09465ae7ff5e3338d1ed94ba4d5d/modules/340-monitoring-custom/templates/service-monitor.yaml#L19).

PR #2332, opened 2022-08-24
[prometheus] add prometheus-scraper-istio-mtls secret

fl64:
## Description

Added a feature to scrape metrics from istio-strict-mtls-ed exporters.

To scrape metrics from workloads with PeerAuthentication mtls.mode = STRICT. Issue (#2309)

The new secret `prometheus-scraper-istio-mtls` is generated for istio-enabled clusters and is mounted to Prometheus pod for scraping metrics.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: feature
summary: Create mTLS secret to scrape metrics from workloads with PeerAuthentication mtls.mode = STRICT.
impact_level: default
```


apolovov (changes requested):

modules/300-prometheus/hooks/generate_istio_mtls_cert_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	Context("Empty cluster and minimal settings", func() {
+		BeforeEach(func() {
+			f.BindingContexts.Set(f.KubeStateSet(``))
+			f.RunHook()
+		})
+
+		It("Hook must execute successfully", func() {
+			Expect(f).To(ExecuteSuccessfully())
+		})
```
apolovov, line 123: Check if the certificate wasn't generated.

modules/300-prometheus/hooks/generate_istio_mtls_cert_test.go
```diff
@@ -0,0 +1,196 @@
...
+			f.RunHook()
+
+		})
+		It("Hook must execute successfully", func() {
+			Expect(f).To(ExecuteSuccessfully())
+			Expect(f.ValuesGet(mTLSCrtPath).Exists()).To(BeTrue())
+			Expect(f.ValuesGet(mTLSKeyPath).Exists()).To(BeTrue())
+			cert, err := base64.StdEncoding.DecodeString(f.KubernetesResource("Secret", "d8-monitoring", "prometheus-scraper-istio-mtls").Field(`data.tls\.crt`).String())
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(f.ValuesGet(mTLSCrtPath).String()).To(BeEquivalentTo(cert))
```
apolovov, line 151: the same for key pls.

modules/300-prometheus/hooks/generate_istio_mtls_cert_test.go
```diff
@@ -0,0 +1,202 @@
...
+		Expect(err).ShouldNot(HaveOccurred())
+
+		certPool := x509.NewCertPool()
+		ok := certPool.AppendCertsFromPEM(istioCA)
+		Expect(ok).To(BeTrue())
+
+		opts := x509.VerifyOptions{
+			Roots:       certPool,
+			KeyUsages:   []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+			CurrentTime: time.Now().Add(time.Hour * 8030), // ~ 11 month
```
apolovov, line 191: add an Except for checking if the KeyUsages doesn't contain ServerAuth.

apolovov (changes requested):

modules/340-monitoring-custom/docs/README.md
```diff
@@ -32,6 +32,7 @@ To enable the `monitoring-custom` module to collect application metrics, you mus
       prometheus.deckhouse.io/tls: "true"  # you don't need to specify this annotation if metrics are sent over http
     ```
 
+- When using service mesh [Istio](../../ee/modules/110-istio) in mTLS STRICT mode, add the following Service annotation to force collecting metrics with proper mTLS certificate: `prometheus.deckhouse.io/istio-mtls: "true"`. Note that the application metrics must be exported via pure http without TLS.
```
z9r5, line 35:
```suggestion
- When using service mesh [Istio](../110-istio/) in mTLS STRICT mode, add the following Service annotation to force collecting metrics with proper mTLS certificate: `prometheus.deckhouse.io/istio-mtls: "true"`. Note that the application metrics must be exported via pure HTTP without TLS.
```

modules/340-monitoring-custom/docs/README_RU.md
```diff
@@ -32,6 +32,8 @@ search: prometheus
       prometheus.deckhouse.io/tls: "true"  # если метрики отдаются по http, эту аннотацию указывать не нужно
     ```
 
+- При использовании service mesh [Istio](../../ee/modules/110-istio) в режиме STRICT mTLS, для подобающего сбора метрик необходимо указать следующую аннотацию для Service: `prometheus.deckhouse.io/istio-mtls: "true"`. Важно, что метрики приложения должны экспортироваться по протоколу http без TLS.
```
z9r5, line 35:
```suggestion
- При использовании service mesh [Istio](../110-istio/) в режиме STRICT mTLS, для сбора метрик необходимо указать следующую аннотацию у Service: `prometheus.deckhouse.io/istio-mtls: "true"`. Важно, что метрики приложения должны экспортироваться по протоколу HTTP без TLS.
```

z9r5: @fl64 Not sure about my fixes. Check, please.

z9r5 (approved):

modules/300-prometheus/templates/prometheus-scraper-istio-mtls.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.prometheus.internal.prometheusScraperIstioMTLS }}
+  {{- if hasKey .Values.prometheus.internal.prometheusScraperIstioMTLS "certificate" }}
```
apolovov, line 2:
```suggestion
{{- if hasKey .Values.prometheus.internal.prometheusScraperIstioMTLS "certificate" }}
```

modules/300-prometheus/templates/prometheus-scraper-istio-mtls.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: prometheus-scraper-istio-mtls
+  namespace: d8-monitoring
+  {{- include "helm_lib_module_labels" (list .) | nindent 2 }}
+type: kubernetes.io/tls
+data:
+  tls.crt: {{ .Values.prometheus.internal.prometheusScraperIstioMTLS.certificate | b64enc }}
+  tls.key: {{ .Values.prometheus.internal.prometheusScraperIstioMTLS.key | b64enc }}
+  {{- end }}
+{{- end }}
```
apolovov, line 15:
```suggestion
{{- end }}
```

apolovov (changes requested):

apolovov (dismissed):

z9r5 (approved):

modules/300-prometheus/hooks/generate_istio_mtls_cert.go
```diff
@@ -0,0 +1,214 @@
...
+					MatchNames: []string{"d8-monitoring"},
+				},
+			},
+			NameSelector: &types.NameSelector{
+				MatchNames: []string{"prometheus-scraper-istio-mtls"},
+			},
+		},
+	},
+	Schedule: []go_hook.ScheduleConfig{
+		{Name: "cron", Crontab: "0 0 * * *"}, // every day at 00:00 am
```
nabokihms, line 72: We usually use magical 42 4 for it.
https://github.com/deckhouse/deckhouse/blob/7e8fe14a81468cce4ddb43ad22b4ca1aa0efa7ae/go_lib/hooks/tls_certificate/order_certificate.go#L143

modules/300-prometheus/hooks/generate_istio_mtls_cert.go
```diff
@@ -0,0 +1,214 @@
...
+
+	if keyBytes, ok = secret.Data["tls.key"]; ok {
+		cert.Key = string(keyBytes)
+	} else {
+		return nil, fmt.Errorf("can't get key from secert %v", secret.Name)
+	}
+	return cert, nil
+}
+
+func applySecertIstioCAFilter(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
```
nabokihms, line 101: As I see it, the code is the same for both filters. Only certificate and key keys are different. 

Let's rewrite the code to reuse the generic logic using decorators.
```go
func filterCertificate(key, cert string, obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
  ...
}

func filterCa(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
  return filterCertificate("ca-cert.pem", "ca-key.pem", obj)
}

func filterCert(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
  return filterCertificate("tls.crt", "tls.key", obj)
}
```

modules/300-prometheus/templates/prometheus/prometheus.yaml
```diff
@@ -144,6 +144,9 @@ spec:
       threshold.extended-monitoring.flant.com/disk-bytes-warning: "97"
       threshold.extended-monitoring.flant.com/disk-bytes-critical: "99"
       checksum/scraper: {{ include (print $.Template.BasePath "/prometheus-scraper-tls.yaml") . | sha256sum }}
+{{- if hasKey .Values.prometheus.internal.prometheusScraperIstioMTLS "certificate" }}
+      checksum/scraper-istio-mtls: {{ include (print $.Template.BasePath "/prometheus-scraper-istio-mtls.yaml") . | sha256sum }}
```
nabokihms, line 148: Does it mean that Prometheus will be restarted every month?

modules/340-monitoring-custom/templates/_helpers.tpl
```diff
@@ -61,11 +61,21 @@
 {{- end }}
 
 {{- define "tls_config" }}
+{{- $tls_secret_name := . }}
```
nabokihms, line 64: :+1:

modules/340-monitoring-custom/templates/service-monitor.yaml
```diff
@@ -17,13 +17,24 @@ spec:
...
   - scheme: https
-    {{- include "tls_config" . | nindent 4 }}
+    {{- include "tls_config" "" | nindent 4 }}
     relabelings:
     {{- include "service_label_selector" . | nindent 4 }}
     {{- include "keep_targets_for_schema" (list "service" "https") | nindent 4 }}
     {{- include "endpoint_by_service_port_name" "https" | nindent 4 }}
     {{- include "base_relabeling" "service" | nindent 4 }}
 
+    {{- if .Values.monitoringCustom.internal.prometheusScraperIstioMTLSEnabled }}
```
nabokihms, line 27: I'd like to avoid additional the secret subscription hook and this if clause. In my opinion it is ok to always have this monitor part if istio is enabled in the cluster.

modules/340-monitoring-custom/templates/service-monitor.yaml
```diff
@@ -17,13 +17,24 @@ spec:
...
+  # Istio mTLS target with samples limited to 1000
+  - scheme: https
+    {{- include "tls_config" "prometheus-scraper-istio-mtls" | nindent 4 }}
+    relabelings:
+    {{- include "service_label_selector" . | nindent 4 }}
+    {{- include "keep_targets_for_schema" (list "service" "istio-mtls") | nindent 4 }}
+    {{- include "endpoint_by_service_port_name" "http" | nindent 4 }}
```
nabokihms, line 34: Seems like a problem. In the monitoring-custom module we use port names to distinguish the protocol - `http-metrics` for http and `https-metrics` for https. 

However, this line of the code filters ports by the http-metrics name. Is this the expected behavior? What will happen if users sets the prometheus.deckhouse.io/tls=true annotation?

nabokihms: The same question goes to the pod monitor.

modules/300-prometheus/hooks/generate_istio_mtls_cert.go
apolovov: They also return different types.
	var cert certificate.Certificate
	var cert certificate.Authority
it is a problem for decorator.

modules/300-prometheus/templates/prometheus/prometheus.yaml
apolovov: It does.

modules/340-monitoring-custom/templates/service-monitor.yaml
apolovov: No, if there istn't a mtls-client-secret, prometheus swears dirty.

modules/300-prometheus/hooks/generate_istio_mtls_cert.go
nabokihms: Yeah, overlooked it. Thanks.

modules/340-monitoring-custom/templates/service-monitor.yaml
nabokihms: Yeah, but if mtls is enabled it will be restarted monthly?

modules/300-prometheus/templates/prometheus/prometheus.yaml
nabokihms: Prometheus can hot reload certificates. The config reloader sidecar detects changes and triggers reloading.

The only problem is if a tcp session lasts for more than a client certificate lifetime,
but I thought the whole point to rotate certificate eleven month ahead was in avoiding excessive Prometheus restarts.

modules/340-monitoring-custom/templates/service-monitor.yaml
apolovov: It is a well-thought approach.
1. Why "http-metrics" — application exports metric in plain http, after enabling istio you don't need to rename ports.
2. We don't support applications that export metric in https. In this case, we need to use two certificates. Probably it worth mentioning in docs.
3. What will happen if users sets the prometheus.deckhouse.io/tls=true annotation? (and istio-mtls=true simultanuesly) — the metrics wont be scraped.

modules/340-monitoring-custom/templates/service-monitor.yaml
nabokihms: Dirty in which way?

fl64: I have an idea to remove the annotation with mtls secret checksum from the Prometheus resource and add an annotation checksum for PodMonitor and ServiceMonitor, so when the certificate changes, only these two resources will be recreated, and Prometheus will not restart.

modules/300-prometheus/templates/prometheus/prometheus.yaml
fl64: I have an idea to remove the annotation with mtls secret checksum from the Prometheus resource and add an annotation checksum for PodMonitor and ServiceMonitor, so when the certificate changes, only these two resources will be recreated, and Prometheus will not restart.

apolovov: Let's check the Prometheus behaviour. @fl64 pls, change the mtls cert lifetime to 5min and check if prometheus watch the cert.

modules/340-monitoring-custom/templates/service-monitor.yaml
apolovov: Prometheus skips ServiceMonitor (or PodMonitor) if there isn't a secret.

modules/300-prometheus/templates/prometheus/prometheus.yaml
fl64: @apolovov 
The test was completed successfully. Prometheus successfully reloads certificates if they are changed.

modules/300-prometheus/hooks/generate_istio_mtls_cert.go
fl64: fixed

apolovov (approved):

z9r5 (approved):

nabokihms (approved):

PR #2350, opened 2022-08-26
[ingress-nginx] pbmetrics improvement

yalosev:

## Description
- Premature removal
`Premature timer expiration happens when the Nginx worker process is trying to shut down, as in an Nginx configuration reload triggered by the HUP signal or in an Nginx server shutdown.`
It seems like a good idea to send collected metrics on a worker shutdown. We are not using recursive timers, so can skip premature handler. Also we can close the tcp socket on premature

- Remove table copy
`Lua is single-threaded by design, one and only one thread can access the data at same moment.`
We can avoid an extra copy of a metrics table. Like an node event-loop nginx stops main thread execution when calls a timer's callback function, so we can avoid extra time and memory consumption

- We can backport it to 1.35, cause it already has an ingress controller reboot
- Bump protobuf to avoid memory leak

Speed up metrics collection. Low memory consumption

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: Improve metrics collection script
impact_level: default
```

### DIFF
--- a/modules/402-ingress-nginx/images/controller-0-33/Dockerfile
+++ b/modules/402-ingress-nginx/images/controller-0-33/Dockerfile
@@ -25,7 +25,7 @@ FROM quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.26.1@sha25
 USER root
 RUN apt-get update \
   && apt-get install -y --no-install-recommends patch gcc build-essential \
-  && luarocks install lua-protobuf 0.3.2-0 \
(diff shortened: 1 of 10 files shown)


yalosev:
Probably, we can change buffer size. For 1 request we send 18 metrics. Buffer with preallocated 200000 slots is enough for 10000 requests per second per worker. I guess we can lower it at least twice

nabokihms (approved):

PR #2357, opened 2022-08-29
[testing][ci] Add a check if the repository contains the actual CI workflow

z9r5:
## Description
Adds a check if the repository contains the actual GitHub workflow (that no one forgot to regenerate it).

Sometimes, after changing CI templates, authors forget to regenerate workflows.

- CI has to fail completely if workflow templates have been changed, but workflow has not been regenerated.
- CI should not fail if workflow templates have been changed and workflow has been regenerated.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: testing
type: chore
summary: Added a check if the repository contains the actual GitHub workflow (that no one forgot to regenerate it).
impact_level: low
```

### DIFF
--- a/.github/ci_templates/build.yml
+++ b/.github/ci_templates/build.yml
@@ -18,6 +18,25 @@ steps:
 # </template: go_generate_template>
 {!{ end }!}
 
+{!{ define "workflow_render_template" }!}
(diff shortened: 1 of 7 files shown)


diafour (approved):
LGTM

PR #2421, opened 2022-09-08
[candi][dhctl] Render `detect_bundle.sh` as template before execute

name212:
## Description

Render detect_bundle.sh as template before execute

`detect_bundle.sh` pass to new cloud node via `user-data`. In AWS `user-data` has content size limit to 16kb.
For decrease size we remove comments via helm-template.

E2e should pass

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix 
summary: Render detect_bundle.sh as template before execute.
impact_level: low
```

### DIFF
--- a/dhctl/pkg/template/bundle.go
+++ b/dhctl/pkg/template/bundle.go
@@ -34,7 +34,7 @@ const (
 	bashibleDir      = "/var/lib/bashible"
 	candiBashibleDir = candiDir + "/bashible"
 	stepsDir         = bashibleDir + "/bundle_steps"
-	detectBundlePath = "/deckhouse/candi/bashible/detect_bundle.sh"
+	detectBundlePath = candiBashibleDir + "/detect_bundle.sh"
 )
 
 type saveFromTo struct {
--- a/dhctl/pkg/template/bundle.go
+++ b/dhctl/pkg/template/bundle.go
@@ -15,8 +15,12 @@
 package template
 
 import (
+	"bytes"
 	"fmt"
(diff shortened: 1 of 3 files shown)


RomanenkoDenys (dismissed):
LGTM

dhctl/pkg/template/bundle.go
```diff
@@ -30,6 +34,7 @@ const (
 	bashibleDir      = "/var/lib/bashible"
 	candiBashibleDir = candiDir + "/bashible"
 	stepsDir         = bashibleDir + "/bundle_steps"
+	detectBundlePath = "/deckhouse/candi/bashible/detect_bundle.sh"
```
nabokihms, line 37:
```suggestion
	detectBundlePath = candiBashibleDir + "/detect_bundle.sh"
```

dhctl/pkg/template/bundle.go
```diff
@@ -184,3 +189,57 @@ func withoutNodeGroup(data map[string]interface{}) map[string]interface{} {
...
+
+	if _, err = outFile.WriteString(content); err != nil {
+		return "", err
+	}
+
+	if err = outFile.Sync(); err != nil {
+		return "", err
+	}
+
+	if err = outFile.Chmod(0777); err != nil {
```
nabokihms, line 240:
```
	if err = outFile.Chmod(0777); err != nil {
```
Do we use 777 for scripts? I think for other files we use different permissions.

PR #2426, opened 2022-09-08
[linter][monitoring] check prometheus rules

### LINKED ISSUE #163: [linter][monitoring] check prometheus rules
### Preflight Checklist

For now, we are not testing Prometheus rules. And if we merge misformatted rules to the main branch it will break the Deckhouse queue because Helm won't be able to deploy rules to the cluster.

### Proposed Solution

Consider implementing unit tests of Prometheus rules using the [promtool](https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/).

### Additional Information

_No response_

gmaiorov:
## Description

Implement Prometheus rule checking using [Promtool](https://github.com/prometheus/prometheus/tree/v2.37.0/cmd/promtool).

Fixes #163.

* `bin/promtool` Makefile target added;
* `promtool check rules <rule-file>` executed during matrix tests for every module containing Prometheus rules

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: testing
type: feature
summary: Implement Prometheus rule checking
impact_level: low
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,10 @@ else ifeq ($(PLATFORM_NAME), arm)
 	CRANE_ARCH = arm64
 endif
 
+# Set host arch & OS for golang-based programs, e.g. Prometheus
+GOHOSTARCH := $(shell go env GOHOSTARCH)
+GOHOSTOS := $(shell go env GOHOSTOS)
+
 help:
 	@printf -- "${FORMATTING_BEGIN_BLUE}%s${FORMATTING_END}\n" \
 	"" \
@@ -67,19 +71,29 @@ help:
 
 GOLANGCI_VERSION = 1.46.2
 TRIVY_VERSION= 0.28.1
+PROMTOOL_VERSION = 2.37.0
 TESTS_TIMEOUT="15m"
 
 ##@ General
 
-deps: bin/golangci-lint bin/trivy bin/regcopy bin/jq bin/yq bin/crane ## Install dev dependencies.
+deps: bin/golangci-lint bin/trivy bin/regcopy bin/jq bin/yq bin/crane bin/promtool ## Install dev dependencies.
 
 ##@ Tests
 
-.PHONY: tests-modules tests-matrix tests-openapi
+bin/promtool-${PROMTOOL_VERSION}/promtool:
+	mkdir -p bin/promtool-${PROMTOOL_VERSION}
+	curl -sSfL https://github.com/prometheus/prometheus/releases/download/v${PROMTOOL_VERSION}/prometheus-${PROMTOOL_VERSION}.${GOHOSTOS}-${GOHOSTARCH}.tar.gz | tar zxf - -C bin/promtool-${PROMTOOL_VERSION} --strip=1 prometheus-${PROMTOOL_VERSION}.${GOHOSTOS}-${GOHOSTARCH}/promtool
+
+.PHONY: bin/promtool
+bin/promtool: bin/promtool-${PROMTOOL_VERSION}/promtool
+	rm -f bin/promtool
+	ln -s /deckhouse/bin/promtool-${PROMTOOL_VERSION}/promtool bin/promtool
+
+.PHONY: tests-modules tests-matrix tests-openapi tests-prometheus
 tests-modules: ## Run unit tests for modules hooks and templates.
 	go test -timeout=${TESTS_TIMEOUT} -vet=off ./modules/... ./global-hooks/... ./ee/modules/... ./ee/fe/modules/...
 
-tests-matrix: ## Test how helm templates are rendered with different input values generated from values examples.
+tests-matrix: bin/promtool ## Test how helm templates are rendered with different input values generated from values examples.
   ##~ Options: FOCUS=module-name
 	go test ./testing/matrix/ -v
 
--- a/testing/matrix/linter/rules/modules/prometheus_rules.go
+++ b/testing/matrix/linter/rules/modules/prometheus_rules.go
@@ -65,7 +65,10 @@ func writeTempRuleFileFromObject(m utils.Module, marshalledYaml []byte) (path st
 		return "", err
 	}
 	defer func(renderedFile *os.File) {
-		err = renderedFile.Close()
+		errClose := renderedFile.Close()
+		if err == nil {
+			err = errClose
+		}
 	}(renderedFile)
 
 	_, err = renderedFile.Write(marshalledYaml)
--- a/testing/matrix/linter/rules/modules/prometheus_rules.go
+++ b/testing/matrix/linter/rules/modules/prometheus_rules.go
@@ -34,25 +34,29 @@ type checkResult struct {
 	errMsg  string
 }
 
+type rulesCacheStruct struct {
+	cache map[string]checkResult
+	mu    sync.RWMutex
+}
+
 const promtoolPath = "/deckhouse/bin/promtool"
 
-var (
-	rulesCache   = map[string]checkResult{}
-	rulesCacheMu = sync.RWMutex{}
-)
+var rulesCache = rulesCacheStruct{
+	cache: map[string]checkResult{},
+	mu:    sync.RWMutex{},
+}
 
 func promtoolAvailable() bool {
 	info, _ := os.Stat(promtoolPath)
 	return info != nil && (info.Mode().Perm()&0111 != 0)
 }
 
-func marshalChartYaml(object storage.StoreObject) (marshal []byte, hash string, err error) {
-	marshal, err = yaml.Marshal(object.Unstructured.Object["spec"])
+func marshalChartYaml(object storage.StoreObject) ([]byte, string, error) {
+	marshal, err := yaml.Marshal(object.Unstructured.Object["spec"])
 	if err != nil {
-		return
+		return nil, "", err
 	}
-	hash = newSHA256(marshal)
-	return
+	return marshal, newSHA256(marshal), nil
 }
 
 func writeTempRuleFileFromObject(m utils.Module, marshalledYaml []byte) (path string, err error) {
@@ -80,7 +84,8 @@ func checkRuleFile(path string) error {
 }
 
 func newSHA256(data []byte) string {
-	return string(sha256.Sum256(data)[:])
+	hash := sha256.Sum256(data)
+	return string(hash[:])
 }
 
 func createPromtoolError(m utils.Module, errMsg string) errors.LintRuleError {
@@ -114,9 +119,9 @@ func PromtoolRuleCheck(m utils.Module, object storage.StoreObject) errors.LintRu
 			)
 		}
 
-		rulesCacheMu.RLock()
-		res, ok := rulesCache[hash]
-		rulesCacheMu.RUnlock()
+		rulesCache.mu.RLock()
+		res, ok := rulesCache.cache[hash]
+		rulesCache.mu.RUnlock()
 		if ok {
 			if !res.success {
 				return createPromtoolError(m, res.errMsg)
@@ -141,17 +146,17 @@ func PromtoolRuleCheck(m utils.Module, object storage.StoreObject) errors.LintRu
 		err = checkRuleFile(path)
 		if err != nil {
 			errorMessage := string(err.(*exec.ExitError).Stderr)
-			rulesCacheMu.Lock()
-			rulesCache[hash] = checkResult{
+			rulesCache.mu.Lock()
+			rulesCache.cache[hash] = checkResult{
 				success: false,
 				errMsg:  errorMessage,
 			}
-			rulesCacheMu.Unlock()
+			rulesCache.mu.Unlock()
 			return createPromtoolError(m, errorMessage)
 		}
-		rulesCacheMu.Lock()
-		rulesCache[hash] = checkResult{success: true}
-		rulesCacheMu.Unlock()
+		rulesCache.mu.Lock()
+		rulesCache.cache[hash] = checkResult{success: true}
+		rulesCache.mu.Unlock()
 	}
 	return errors.EmptyRuleError
 }
--- a/testing/matrix/linter/rules/modules/prometheus_rules.go
+++ b/testing/matrix/linter/rules/modules/prometheus_rules.go
@@ -47,8 +47,8 @@ var rulesCache = rulesCacheStruct{
 }
 
 func promtoolAvailable() bool {
-	info, _ := os.Stat(promtoolPath)
-	return info != nil && (info.Mode().Perm()&0111 != 0)
+	info, err := os.Stat(promtoolPath)
+	return err == nil && (info.Mode().Perm()&0111 != 0)
 }
 
 func marshalChartYaml(object storage.StoreObject) ([]byte, string, error) {
--- a/testing/matrix/linter/rules/modules/prometheus_rules.go
+++ b/testing/matrix/linter/rules/modules/prometheus_rules.go
@@ -1,0 +1,162 @@
+/*
+Copyright 2022 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modules
+
+import (
+	"crypto/sha256"
+	"os"
+	"os/exec"
+	"sync"
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/deckhouse/deckhouse/testing/matrix/linter/rules/errors"
+	"github.com/deckhouse/deckhouse/testing/matrix/linter/storage"
+	"github.com/deckhouse/deckhouse/testing/matrix/linter/utils"
+)
+
+type checkResult struct {
+	success bool
+	errMsg  string
+}
+
+const promtoolPath = "/deckhouse/bin/promtool"
+
+var (
+	rulesCache   = map[string]checkResult{}
+	rulesCacheMu = sync.RWMutex{}
+)
+
+func promtoolAvailable() bool {
+	info, _ := os.Stat(promtoolPath)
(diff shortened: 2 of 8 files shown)


gmaiorov:
Example runs available here:
* [Pipeline for broken Prometheus rules](https://github.com/deckhouse/deckhouse/actions/runs/3069247078/jobs/4957672804#step:9:81)
* [Pipeline for fixed Prometheus rules](https://github.com/deckhouse/deckhouse/actions/runs/3069327235/jobs/4957845982#step:9:1)

Makefile
```diff
@@ -67,19 +71,29 @@ help:
...
+bin/promtool-${PROMTOOL_VERSION}/promtool:
+	mkdir -p bin/promtool-${PROMTOOL_VERSION}
+	curl -sSfL https://github.com/prometheus/prometheus/releases/download/v${PROMTOOL_VERSION}/prometheus-${PROMTOOL_VERSION}.${GOHOSTOS}-${GOHOSTARCH}.tar.gz | tar zxf - -C bin/promtool-${PROMTOOL_VERSION} --strip=1 prometheus-${PROMTOOL_VERSION}.${GOHOSTOS}-${GOHOSTARCH}/promtool
+
+.PHONY: bin/promtool
+bin/promtool: bin/promtool-${PROMTOOL_VERSION}/promtool
+	rm -f bin/promtool
+	ln -s /deckhouse/bin/promtool-${PROMTOOL_VERSION}/promtool bin/promtool
+
+.PHONY: tests-modules tests-matrix tests-openapi tests-prometheus
```
RomanenkoDenys, line 92: Why we need different versions of promtool ? I think we need only one version, so
```
bin/promtool:
  curl -sSfL https://github.com/prometheus/prometheus/releases/download/v${PROMTOOL_VERSION}/prometheus-${PROMTOOL_VERSION}.${GOHOSTOS}-${GOHOSTARCH}.tar.gz | tar zxf - -C bin --strip=1 prometheus-${PROMTOOL_VERSION}.${GOHOSTOS}-${GOHOSTARCH}/promtool
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+	success bool
+	errMsg  string
+}
+
+const promtoolPath = "/deckhouse/bin/promtool"
+
+var (
+	rulesCache   = map[string]checkResult{}
+	rulesCacheMu = sync.RWMutex{}
+)
```
RomanenkoDenys, line 42: why we create module mutex ?
I think we do something like that:
```go
type rulesCacheStruct {
  rulesCache map[string]checkResult{}
  mu sync.Mutex
}
```
https://gobyexample.com/mutexes

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+
+func promtoolAvailable() bool {
+	info, _ := os.Stat(promtoolPath)
```
RomanenkoDenys, line 45: Why we suppress error there ?
from doc - Stat returns a FileInfo describing the named file.
If there is an error, it will be of type *PathError.

RomanenkoDenys, line 45: Why we suppress error there ?
from doc - Stat returns a FileInfo describing the named file.
If there is an error, it will be of type *PathError.

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func marshalChartYaml(object storage.StoreObject) (marshal []byte, hash string, err error) {
+	marshal, err = yaml.Marshal(object.Unstructured.Object["spec"])
+	if err != nil {
+		return
+	}
+	hash = string(newSHA256(marshal))
+	return
+}
```
RomanenkoDenys, line 56:
```suggestion
func marshalChartYaml(object storage.StoreObject) ([]byte, string, error) {
	marshal, err := yaml.Marshal(object.Unstructured.Object["spec"])
	if err != nil {
		return nil, "", err
	}
	return marshal, string(newSHA256(marshal)), nil
}
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+		return "", err
+	}
+
+	err = renderedFile.Close()
+	if err != nil {
+		return renderedFile.Name(), err
+	}
+
+	return renderedFile.Name(), nil
+}
```
RomanenkoDenys, line 75:
```suggestion
	return renderedFile.Name(), err
}
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+	info, _ := os.Stat(promtoolPath)
+	return info != nil && (info.Mode().Perm()&0111 != 0)
+}
+
+func marshalChartYaml(object storage.StoreObject) (marshal []byte, hash string, err error) {
+	marshal, err = yaml.Marshal(object.Unstructured.Object["spec"])
+	if err != nil {
+		return
+	}
+	hash = string(newSHA256(marshal))
```
RomanenkoDenys, line 54:
```suggestion
	hash = newSHA256(marshal)
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func newSHA256(data []byte) []byte {
+	hash := sha256.Sum256(data)
+	return hash[:]
+}
```
RomanenkoDenys, line 89:
```suggestion
func newSHA256(data []byte) string {
	return string(sha256.Sum256(data)[:])
}
```

RomanenkoDenys (changes requested):

Makefile
gmaiorov: I believe we should use the same version of Promtool as the Prometheus version used in Deckhouse to address potential issues with modified rule syntax or new Prometheus features.
This leads to the need of supporting all of the versions of Promtool found across the repository.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: Thanks, I'll account that & update PR.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: I'm relying on `info` being nil in case of errors, checking it in following comparison in the return statement.
Is it considered a bad practice?

gmaiorov: Perhaps this change request is a duplicated previous one.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: This suggestion conflicts with previous one due to changed `newSHA256()` signature, so I cannot apply it directly.
I'll add corresponding change manually.
BTW, does named return parameters considered bad practice?

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: It seems that golang doesn't allow to create slice from function's return due to slice needing array to be allocated on heap, when return is allocated on stack. And slice needed to convert [32]byte to []byte.
Perhaps I need to revert that.

gmaiorov: I've addresed this request in 4b92715708a0d0b324c42edfc3952b24edb332a0.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: I've addresed this request in 4b92715708a0d0b324c42edfc3952b24edb332a0.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: I've addresed this request in 4b92715708a0d0b324c42edfc3952b24edb332a0.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: Fixed error handling in df881a0d01d16cfceb2c26acbc0efe0e9ff955a3.

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+func marshalChartYaml(object storage.StoreObject) ([]byte, string, error) {
+	marshal, err := yaml.Marshal(object.Unstructured.Object["spec"])
+	if err != nil {
+		return nil, "", err
+	}
+	return marshal, newSHA256(marshal), nil
+}
+
+func writeTempRuleFileFromObject(m utils.Module, marshalledYaml []byte) (path string, err error) {
+	renderedFile, err := os.CreateTemp("", m.Name+".*.yml")
```
RomanenkoDenys, line 63:
```suggestion
	renderedFile, err := os.CreateTemp("", filepath.Join(m.Name,"*.yml"))
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	_, err = renderedFile.Write(marshalledYaml)
+	if err != nil {
+		return "", err
+	}
+
+	err = renderedFile.Close()
```
RomanenkoDenys, line 73: I think better to use defer to close file.

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,162 @@
...
+	return renderedFile.Name(), err
+}
+
+func checkRuleFile(path string) error {
+	promtoolComand := exec.Command(promtoolPath, "check", "rules", path)
+	_, err := promtoolComand.Output()
+	if err != nil {
+		return err
+	}
+	return nil
```
RomanenkoDenys, line 83:
```suggestion
  return err
```

RomanenkoDenys (changes requested):
small suggestions

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: Joining name & extension provides output like `<name>/*.yml`, which we doesn't need here.
The goal is to create filename pattern for `os.CreateTemp()` looking like `<name>.*.yml`, so concatenating is platform-independent in this case because there is no separator added.

RomanenkoDenys: yes, it's my mistake.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: Thanks, it's much more logical, though I'll have to add this suggestion manually because of indentation issues (2 whitespaces instead of tab).

deckhouse-BOaTswain:
Aye, aye, @gmaiorov. I've started the workflow for label 'e2e/run/aws'...

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: I've added suggested change in https://github.com/deckhouse/deckhouse/pull/2426/commits/1b7b38b1735ba6747c17d0320059408dbb527205.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: I've added suggested change in https://github.com/deckhouse/deckhouse/commit/1b7b38b1735ba6747c17d0320059408dbb527205.

deckhouse-BOaTswain:

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func writeTempRuleFileFromObject(m utils.Module, marshalledYaml []byte) (path string, err error) {
+	renderedFile, err := os.CreateTemp("", m.Name+".*.yml")
+	if err != nil {
+		return "", err
+	}
+	defer func(renderedFile *os.File) {
+		err = renderedFile.Close()
+	}(renderedFile)
```
RomanenkoDenys, line 69: You cannot handle error in this place. Standard example - https://gobyexample.com/defer. Or use panic.

deckhouse-BOaTswain:
Aye, aye, @gmaiorov. I've started the workflow for label 'e2e/run/gcp'...

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: I'm rewriting named return here, so I think it's possible to handle error in deferred function in such way, though I've found possible error overwrite if such happens during temp file write. I've fixed this in 835bd5ab5b1886832a70f0ec4f3e7f518946a7e2.

RomanenkoDenys (approved):

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,164 @@
...
+
+type rulesCacheStruct struct {
+	cache map[string]checkResult
+	mu    sync.RWMutex
+}
+
+const promtoolPath = "/deckhouse/bin/promtool"
+
+var rulesCache = rulesCacheStruct{
+	cache: map[string]checkResult{},
```
yalosev, line 45:
```suggestion
	cache: make(map[string]checkResult),
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,164 @@
...
+	return marshal, newSHA256(marshal), nil
+}
+
+func writeTempRuleFileFromObject(m utils.Module, marshalledYaml []byte) (path string, err error) {
+	renderedFile, err := os.CreateTemp("", m.Name+".*.yml")
+	if err != nil {
+		return "", err
+	}
+	defer func(renderedFile *os.File) {
+		errClose := renderedFile.Close()
```
yalosev, line 68:
```suggestion
		_ = renderedFile.Close()
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,164 @@
...
+		if err == nil {
+			err = errClose
+		}
```
yalosev, line 71:
```suggestion
		_ = renderedFile.Close()
```

temporary file will be removed, we can ignore an error here

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,164 @@
...
+				"MODULE060",
+				m.Name,
+				m.Path,
+				"Error creating temporary rule file from Helm chart:\n%s",
+				err.Error(),
+			)
+		}
+		defer func(name string) {
+			_ = os.Remove(name)
+		}(path)
```
yalosev, line 146:
```suggestion
		path, err := writeTempRuleFileFromObject(m, marshal)
		defer func(name string) {
			_ = os.Remove(name)
		}(path)
		if err != nil {
			return errors.NewLintRuleError(
				"MODULE060",
				m.Name,
				m.Path,
				"Error creating temporary rule file from Helm chart:\n%s",
				err.Error(),
			)
		}
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,164 @@
...
+		if err == nil {
+			err = errClose
+		}
+	}(renderedFile)
+
+	_, err = renderedFile.Write(marshalledYaml)
+	if err != nil {
+		return "", err
+	}
+
```
yalosev, line 78:
```suggestion
_ = renderedFile.Sync()
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,164 @@
...
+		"MODULE060",
+		moduleLabel(m.Name),
+		m.Path,
+		"Promtool check failed for Helm chart:\n%s",
+		errMsg,
+	)
+}
+
+func PromtoolRuleCheck(m utils.Module, object storage.StoreObject) errors.LintRuleError {
+	if object.Unstructured.GetKind() == "PrometheusRule" {
```
yalosev, line 104:
```
if object.Unstructured.GetKind() != "PrometheusRule" {
    return errors.EmptyRuleError
}
```

testing/matrix/linter/rules/modules/prometheus_rules.go
```diff
@@ -0,0 +1,164 @@
...
+		if !promtoolAvailable() {
```
yalosev, line 105: Let's move it to the upper level of the linter to check it just once

yalosev: I'd like `Run(...)` global function or `controller.Run(...)` function

yalosev (changes requested):
Left comments

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: I've moved check for promtool to linter.Run() function in b4d61b24525f0e153b094f800f0ee299044b2937, now this check is performed once for every module with existing Prometheus rules.

It seems to be possible to move this check to matrix.TestMatrix() to execute it only once, but would make whole matrix tests to depend on promtool, and not just modules with Prometheus rules. And I think it could break tests abstraction level.

testing/matrix/linter/rules/modules/prometheus_rules.go
gmaiorov: Thanks! I've addressed this in 3b4989d60493609feba0b9c4d94e879c67d40829.

yalosev (approved):

RomanenkoDenys (approved):
LGTM

nabokihms (approved):

PR #2451, opened 2022-09-12
[node-manager] Fail early-oom gracefully

zuzzas:
## Description

Fail with a helpful log message and generate an alert if PSI subsystem is unavailable.

There are OS versions that have PSI disabled by default.

Pod is not in CrashloopBackOff and an alert is generated.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Fail early-oom gracefully with a helpful log message and a low-severity alert if PSI subsystem is unavailable.
impact_level: default
```


deckhouse-BOaTswain:
Aye, aye, @zuzzas. I've started the workflow for label 'e2e/run/yandex-cloud'...

modules/040-node-manager/images/early-oom/src/psi-monitor.go
```diff
@@ -35,16 +40,42 @@ const (
 	sysrqOOMCharacter = "f"
 )
 
+var psiAvailable = prometheus.NewGauge(prometheus.GaugeOpts{
```
yalosev, line 43: `psiUnavailable` ?

modules/040-node-manager/images/early-oom/src/psi-monitor.go
```diff
@@ -55,6 +86,25 @@ func main() {
 	}
 }
 
+func probePSISupport() error {
+	fs, err := procfs.NewDefaultFS()
+	if err != nil {
+		log.Fatal(err)
```
yalosev, line 92: you have error handling but fail with log.Fatal. Is it necessarily?

modules/040-node-manager/images/early-oom/src/psi-monitor.go
```diff
@@ -35,16 +40,42 @@ const (
...
+	go func() {
+		if err := server.ListenAndServe(); err != nil {
+			log.Fatal(err)
+		}
+	}()
+
+	psiSupportErr := probePSISupport()
+	if psiSupportErr != nil {
+		log.Println(psiSupportErr)
+		psiAvailable.Set(1)
```
yalosev, line 68: Because this one is reading in an opposite meaning

zuzzas: Fixed.

modules/040-node-manager/images/early-oom/src/psi-monitor.go
zuzzas: Not necessary. Bad copypasta while refactoring.

modules/040-node-manager/images/early-oom/src/psi-monitor.go
zuzzas: Fixed.

yalosev (approved):

modules/040-node-manager/images/early-oom/src/psi-monitor.go
```diff
@@ -35,16 +40,42 @@ const (
...
+		}
+	}()
+
+	psiSupportErr := probePSISupport()
+	if psiSupportErr != nil {
+		log.Println(psiSupportErr)
+		psiUnavailable.Set(1)
+
+		sig := <-c
+		shutdown(sig, server)
```
RomanenkoDenys, line 71: One question. When psiSupportErr is not nil, do we set metric and wait SIGTERM from kubelet due to liveness probe fails ?  Is this expected behavior ?

RomanenkoDenys: Or we should set metric and do nothing in main cycle ? If our pod is never be ready we will get unwanted alerts about it (KubernetesDaemonSetReplicasUnavailable, KubernetesDaemonSetNotUpToDate).

RomanenkoDenys (commented):
added one little question

modules/040-node-manager/images/early-oom/src/psi-monitor.go
yalosev: I guess we can remove `sig := <-c` and everything would be fine

yalosev: Or maybe we should check PSISupport before starting the http server?

RomanenkoDenys: No, all is ok. Liveness probe checked from kube-rbac-proxy so this code works as expected. If we remove 70 line, we go to the cycle which tries to do oom work.

RomanenkoDenys (approved):
LGTM

deckhouse-BOaTswain:
Cherry pick PR [2516](https://github.com/deckhouse/deckhouse/pull/2516) to the branch [release-1.36](https://github.com/deckhouse/deckhouse/tree/release-1.36) successful!

Issue #2409, opened 2022-09-06
[kube-dns] Self-signed certificate in d8-kube-dns-sts-pods-hosts-appender-webhook is not reissued after expiration


shvgn:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.35

### Expected Behavior

The issued TLS certificate is valid only for 1 year. It should be re-issued before it expires.

### Actual Behavior

The certificate is set to secrets and is reused as is despite the expiration. It prevents from the creation of statefulset pods.

https://github.com/deckhouse/deckhouse/blob/main/modules/042-kube-dns/hooks/generate_selfsigned_ca.go#L84

An example is Upmeter server pod and `contoller-manager` probe. The probe fails because it is based on statefulset pod lifecycle.

### Steps To Reproduce

1. Add domain alias.
2. Wait 1 year until self-signed certificate expires.
3. Try to re-create statefulset pods.

### Additional Information

```
# kubectl -n d8-system exec -t deploy/deckhouse -- deckhouse-controller module values kube-dns -ojson | jq -r .kubeDns.internal.stsPodsHostsAppenderWebhook.crt | openssl x509 -dates -nocert

notBefore=Sep  1 13:58:00 2021 GMT
notAfter=Sep  1 13:58:00 2022 GMT
```

### Logs

```shell
Warning  FailedCreate      2m43s              statefulset-controller  create Pod upmeter-0 in StatefulSet upmeter failed error: Internal error occurred: failed calling webhook "sts-pods-hosts-appender-conversion.flant.com": Post "https://d8-kube-dns-sts-pods-hosts-appender-webhook.kube-system.svc:443/mutate?timeout=3s": x509: certificate has expired or is not yet valid: current time 2022-09-06T12:41:33Z is after 2022-09-01T13:58:00Z
```

PR #2539, opened 2022-09-19
[kube-dns] fix certificate generation

yalosev:

## Description
Use generic hook for tls certificate generation

Certificate could be expired and would not be recreated (see #2409)

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: kube-dns
type: fix
summary: Fix tls certificate generation
impact_level: default
```


shvgn (approved):
Yo dawg, we've added a hook within a hook so you can check a certificate before you check a certificate.

LGTM

nabokihms (approved):

PR #2544, opened 2022-09-19
[candi][cloud-provider-yandex] Fix routes for multi-zonal clusters when using WithNATInstance layout

konstantin-axenov:

## Description

Add routes to netplan config in user-data for NAT instance.

If we have `nodeNetworkCIDR` set to 192.168.199.0/24 then we have 3 subnets. NAT instance knows only routes to the subnet where it's located, but has no routes to other subnets in VPC. So network on nodes in other subnets doesn't work properly. This pr creates correct routes for the whole VPC:
```
192.168.199.0/24 dev eth1 proto static scope link
```
Network is already configured similarly for master nodes using bashible:
```
root@p-k-aksenov-tmp-ks-master-0:~# cat /etc/netplan/999-cim-eth1.yaml 
network:
  version: 2
  ethernets:
    eth1:
      dhcp4: true
      dhcp4-overrides:
        use-hostname: false
        use-routes: false
      match:
        macaddress: d0:1d:10:f8:95:5d
      routes:
      - to: 192.168.199.0/24
        scope: link
```

There must be no new alerts D8TerraformStateExporterClusterStateChanged in Yandex Cloud clusters using layout `WithNatInstance`

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: Fix routes for multi-zonal clusters when using WithNATInstance layout
```

### DIFF
--- a/candi/cloud-providers/yandex/terraform-modules/vpc-components/nat_instance.tf
+++ b/candi/cloud-providers/yandex/terraform-modules/vpc-components/nat_instance.tf
@@ -82,6 +82,9 @@ locals {
             use-hostname: false
             use-routes: false
             use-domains: false
+          routes:
+          - to: "${var.node_network_cidr}"
+            link
     EOF
 
     netplan apply


deckhouse-BOaTswain:
Aye, aye, @konstantin-axenov. I've started the workflow for label 'e2e/run/yandex-cloud'...

nabokihms (approved):

PR #2579, opened 2022-09-23
[istio] remove cpu limist for istio-proxy

fl64:
## Description
Cpu limits for istio-proxy sidecars has been removed.

In some cases the isiot-proxy gets throttled, we want to get rid of the CPU limits set for istio-proxy.

All istio-proxy sidecars will be creater without cpu limits.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: chore
summary: Cpu limits for istio-proxy sidecars has been removed.
impact_level: default
```

### DIFF
--- a/ee/modules/110-istio/hooks/patch_injector_configmap.go
+++ b/ee/modules/110-istio/hooks/patch_injector_configmap.go
@@ -9,14 +9,15 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/deckhouse/deckhouse/ee/modules/110-istio/hooks/internal"
-	_ "github.com/evanphx/json-patch"
+
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
 	"github.com/flant/addon-operator/sdk"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


apolovov:
Try to sed the limit in /var/lib/istio/manifests/charts/istio-control/istio-discovery/values.yaml during the build (https://github.com/deckhouse/deckhouse/blob/main/ee/modules/110-istio/images/operator-v1x13/Dockerfile).

fl64:
> Try to sed the limit in /var/lib/istio/manifests/charts/istio-control/istio-discovery/values.yaml during the build (https://github.com/deckhouse/deckhouse/blob/main/ee/modules/110-istio/images/operator-v1x13/Dockerfile).

Alas, that didn't help either.

ee/modules/110-istio/hooks/patch_injector_configmap.go
```diff
@@ -0,0 +1,81 @@
+/*
+Copyright 2022 Flant JSC
+Licensed under the Deckhouse Platform Enterprise Edition (EE) license. See https://github.com/deckhouse/deckhouse/blob/main/ee/LICENSE
+*/
+
```
apolovov, line 5:
```suggestion

// The hook will lose relevance after solving the issue — https://github.com/istio/istio/issues/40078.

```

apolovov (approved):

PR #2619, opened 2022-10-01
[log-shipper] Fix many source to one destination pipelines

nabokihms:

## Description
Two bugs were spotted and fixed:
* Parse JSON rule is added even if there are no log filter rules.
* Only the last works for many sources that point to a single destination.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: log-shipper
type: fix
summary: Stop generating pointless 'parse_json' transform, which improves performance.
impact_level: default
---
section: log-shipper
type: fix
summary: Fix the bug when the many sources point to the same input and only the last is working.
impact_level: default
```

### DIFF
--- a/go_lib/set/set.go
+++ b/go_lib/set/set.go
@@ -17,6 +17,7 @@ limitations under the License.
 package set
 
 import (
+	"encoding/json"
 	"sort"
 
 	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
@@ -103,4 +104,8 @@ func (s Set) Slice() []string {
 
 func (s Set) Size() int {
 	return len(s)
+}
+
+func (s Set) MarshalJSON() ([]byte, error) {
+	return json.Marshal(s.Slice())
 }
--- a/modules/460-log-shipper/hooks/internal/vector/source/file.go
+++ b/modules/460-log-shipper/hooks/internal/vector/source/file.go
@@ -21,6 +21,8 @@ import (
 	"github.com/deckhouse/deckhouse/modules/460-log-shipper/apis/v1alpha1"
 )
 
+var _ apis.LogSource = (*File)(nil)
+
 // File represents `file` vector source
 // https://vector.dev/docs/reference/configuration/sources/file/
 type File struct {
--- a/modules/460-log-shipper/hooks/internal/vector/source/kubernetes.go
+++ b/modules/460-log-shipper/hooks/internal/vector/source/kubernetes.go
@@ -27,6 +27,8 @@ import (
 
 const defaultGlobCooldownMs = 1000
 
+var _ apis.LogSource = (*Kubernetes)(nil)
+
 // Kubernetes represents a source for collecting Kubernetes logs.
 //
 // Because of how selectors work in Kubernetes, it is not possible to declare OR selector.
--- a/modules/460-log-shipper/hooks/internal/vector/transform/testdata/extra-labels.json
+++ b/modules/460-log-shipper/hooks/internal/vector/transform/testdata/extra-labels.json
@@ -1,9 +1,7 @@
 [
 	{
 		"drop_on_abort": false,
-		"inputs": [
-			"testit"
-		],
+		"inputs": [],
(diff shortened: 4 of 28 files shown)


deckhouse-BOaTswain:

go_lib/set/set.go
```diff
@@ -104,3 +105,7 @@ func (s Set) Slice() []string {
 func (s Set) Size() int {
 	return len(s)
 }
+
+func (s Set) MarshalJSON() ([]byte, error) {
+	return json.Marshal(s.Slice())
```
yalosev, line 110: you are changing old behavior of the Set marshalling. Wouldn't it be broken in other places?

modules/460-log-shipper/hooks/internal/vector/source/file.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/deckhouse/deckhouse/modules/460-log-shipper/apis/v1alpha1"
 )
 
+var _ apis.LogSource = (*File)(nil)
```
yalosev, line 24: what is this line for?
In the composer.go file we have `newLogSource` which calls `source.NewFile`. if interface `apis.LogSource ` is not implemented - it wouldn't compile at that place. We don't need to check it one more time

modules/460-log-shipper/hooks/internal/vector/source/kubernetes.go
```diff
@@ -27,6 +27,8 @@ import (
 
 const defaultGlobCooldownMs = 1000
 
+var _ apis.LogSource = (*Kubernetes)(nil)
```
yalosev, line 30: same here ^

modules/460-log-shipper/hooks/internal/vector/transform/testdata/extra-labels.json
```diff
@@ -1,9 +1,7 @@
 [
 	{
 		"drop_on_abort": false,
-		"inputs": [
-			"testit"
-		],
+		"inputs": [],
```
yalosev, line 4: ? How we can have a transform without an input? Something have changed in the vector?

yalosev: I got that you don't want to have an transform if no rules set. So in that case we have a global transform without inputs and it's just not working, right?

go_lib/set/set.go
nabokihms: Nah, everything's fine. I checked that sets aren't used anywhere. In addition, it is strange if marshalling set to something like `{"test":{},"test-2":{}}` is expected behaviour.

I'd like to leave this be.

modules/460-log-shipper/hooks/internal/vector/source/file.go
nabokihms: > if interface apis.LogSource is not implemented - it wouldn't compile at that place.
Yeah, this is kind of a safety measurement. The behavior of the `newLogSource` can change in the future, but this line's only purpose is to check the interface. I did it just to be hundred percent sure everything is fine with the code.

modules/460-log-shipper/hooks/internal/vector/transform/testdata/extra-labels.json
nabokihms: Eww, this testdata shows only the part of the config. 

The actual change that caused the inputs array to be empty here is this piece of the code
https://github.com/deckhouse/deckhouse/blob/1d85d1f9e117d9a444265008f9221c71b379bdae/modules/460-log-shipper/hooks/internal/vector/transform/transforms.go#L28-L44

The `prevInput` variable was equal to the `inputName` and always added as the input for all transform maps. This is an error because `inputName` is never a valid input name. Inputs for transforms were overridden further in the composer's code.

nabokihms:
Closes https://github.com/deckhouse/deckhouse/issues/2658

yalosev (approved):

RomanenkoDenys (approved):

deckhouse-BOaTswain:
Cherry pick PR [2661](https://github.com/deckhouse/deckhouse/pull/2661) to the branch [release-1.36](https://github.com/deckhouse/deckhouse/tree/release-1.36) successful!

deckhouse-BOaTswain:
Cherry pick PR [2662](https://github.com/deckhouse/deckhouse/pull/2662) to the branch [release-1.37](https://github.com/deckhouse/deckhouse/tree/release-1.37) successful!

PR #2628, opened 2022-10-03
[cloud-provider-*] fire an alert if ccm reconcile failed

### LINKED ISSUE #237: [cloud-provider-*] fire an alert if ccm reconcile failed
### Preflight Checklist

The situation may occur when the state of cluster objects (e.g. Services) does not converge with cloud resources (e.g. LoadBalancers). It may happen due to some errors, but we will never know about that in advance.

### Proposed Solution

In our maintained [yandex-ccm](https://github.com/deckhouse/yandex-cloud-controller-manager) we can fail readiness probe on reconciling errors. 

But would be better to find a general way suitable for any cloud provider.

### Additional Information

_No response_

gmaiorov:
Signed-off-by: Gleb Maiorov <gleb.maiorov@flant.com>

## Description

This PR exposes CCM metrics for multiple cloud providers and adds Prometheus alerts for failed reconciliation.

Fixes #237 
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cloud-provider-*
type: feature
summary: Add alert for CCM reconciliation fail 
impact: May restart CCM.
impact_level: default
```

### DIFF
--- a/ee/modules/030-cloud-provider-openstack/templates/cloud-controller-manager/deployment.yaml
+++ b/ee/modules/030-cloud-provider-openstack/templates/cloud-controller-manager/deployment.yaml
@@ -86,6 +86,7 @@ spec:
           - --bind-address=$(HOST_IP)
           - --secure-port=10471
           - --v=2
+          - --authorization-always-allow-paths="/metrics"
           env:
           - name: HOST_IP
             valueFrom:


gmaiorov:
Probably stale and should be redone.

PR #2767, opened 2022-10-18
[testing] Bunch of matrix tests optimizations

nabokihms:

### **Cache templates validation result.** 
Matrix tests can generate different values file sets from the examples of the OpenAPI spec. It is possible to have global values combinations that influence one module template and make no difference for other modules. Values are different, but the result of the render is the same. 

To improve performance for this case, a hash map can help cache the render result and avoid excessive parsing/linting.

### Reduce values allocations
Values are generated as `map[string]interface{}` objects. Both Helm engine and validating function work with the same type of objects. However, there wore unnecessary yaml.Marshal <-> yaml.Unmarshal conversions in matrix tests, which affected the CPU time consumption and GC run frequency.

### Fix values_matrix_tests.yaml

Tests are fixes for two modules - `keepalived` and `monitoring-kubernetes-control-plane`.

### Prometheus rules marshaling

We can calculate the hash of the bytes representation of PrometheusRules objects on parsing Helm templates and avoid marshaling, reducing memory allocations.

This is an attempt to improve matrix test performance.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: feature
summary: Bunch of matrix tests optimizations
impact_level: low
```


deckhouse-BOaTswain:

RomanenkoDenys (approved):
Looks good !

PR #2927, opened 2022-10-31
[ingress-nginx] Remove deprecated versions (< 1.1)

yalosev:

## Description
Remove old versions of the ingress nginx

We don't use them anymore

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: feature
summary: Remove Ingress Nginx controllers of versions below 1.1.
```


nabokihms (approved):

modules/402-ingress-nginx/crds/ingress-nginx.yaml
```diff
@@ -58,7 +58,7 @@ spec:
                     One of the supported NGINX Ingress Controller versions.
 
                     **By default**: the version in the module settings is used.
-                  enum: ['0.33', '0.46', '0.48', '0.49', '1.1']
```
nabokihms, line 61: It can be a problem. You cannot remove fields from enum without changing the apiversion.

yalosev: Why do you think so? I remove not field but it's value. Requirements check would prevent updating to the 1.40 release without updating ingress controller to the 1.1 version.
We did the same with 0.28 and 0.29 versions, it didn't cause any problems

PR #2933, opened 2022-10-31
[dhctl] Add version in dhctl image

gyrter:
## Description

Add version file to install image and dhctl command.

New users want's to have easy method to determine installer version.

File with version in image and built-in version string for dhctl

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: feature
summary: Add version in dhctl image
impact_level: default
```

### DIFF
--- a/dhctl/cmd/dhctl/main.go
+++ b/dhctl/cmd/dhctl/main.go
@@ -126,7 +126,7 @@ func main() {
 		return nil
 	})
 
-	kpApp.Version("v0.0.0").Author("Flant")
+	kpApp.Version(app.AppVersion).Author("Flant")
 
 	go func() {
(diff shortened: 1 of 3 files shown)


name212 (commented):
I don't like variant with injection version as ld flag.
We have file `/deckhouse/version`. May be we can read this file (if file not found, get 'dev' version) and get version from it?

gyrter:
No problem. We can read file.

name212:
please add DCO https://github.com/apps/dco for your commits

name212 (changes requested):
DCO for commits

gyrter:
DCO added

name212 (approved):

Issue #2233, opened 2022-08-17
[openapi] Incorrect regex for resources


konstantin-axenov:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.34

### Expected Behavior

This is valid config
```
  prometheus: |
    vpa:
      maxCPU: 1.5
      maxMemory: 5.0Gi
```

### Actual Behavior

This regex
```
^[0-9]+(E|P|T|G|M|k|Ei|Pi|Ti|Gi|Mi|Ki)?$
```
Leads to error
```
"ConvergeModules failed in phase 'StandBy', requeue task to retry after delay. Failed count is 162. Error: config not valid: 3 errors occurred:\n\t* 'prometheus' module section in ConfigMap/deckhouse is not valid\n\t* \"prometheus.vpa.maxMemory\" must validate one and only one schema (oneOf). Found none valid\n\t* prometheus.vpa.maxMemory should match '^[0-9]+(E|P|T|G|M|k|Ei|Pi|Ti|Gi|Mi|Ki)?$'\n\n"
```

z9r5:
Fixed in #2940

PR #2940, opened 2022-11-01
[global] memory validation fix — float support for dimensioned digits

duckhawk:
## Description
memory validation fix — float support for dimensioned digits like 3.5Gi.

We couldn't use the dot in values with units.
Closes issue #2233

Memory validation allows using notations like '5.0Gi'

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: openapi
type: chore
summary: Memory validation allows to use notation like '5.0Gi'
impact: no impact expected
impact_level: low
```

### DIFF
--- a/candi/openapi/node_group.yaml
+++ b/candi/openapi/node_group.yaml
@@ -287,7 +287,7 @@ spec:
                                 Describes the amount of memory that will not be held by standby holder on Nodes from this NodeGroup.
 
                                 The value can be an absolute number of bytes (for example, 128974848) as well as a fixed-point number using one of memory suffixes: G, Gi, M, Mi.
-                              pattern: '^[0-9]+(E|P|T|G|M|K|Ei|Pi|Ti|Gi|Mi|Ki)?$'
+                              pattern: '^[0-9]+(\.[0-9]+)?(E|P|T|G|M|K|Ei|Pi|Ti|Gi|Mi|Ki)?$'
                               x-kubernetes-int-or-string: true
                     classReference:
                       description: |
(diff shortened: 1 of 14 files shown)


yalosev (approved):
LGTM
but fix tests ;)

candi/openapi/node_group.yaml
```diff
@@ -1518,7 +1518,7 @@ spec:
                                 Describes the amount of memory that will not be held by standby holder on Nodes from this NodeGroup.
 
                                 The value can be an absolute number of bytes (for example, 128974848) as well as a fixed-point number using one of memory suffixes: G, Gi, M, Mi.
-                              pattern: '^[0-9]+(E|P|T|G|M|K|Ei|Pi|Ti|Gi|Mi|Ki)?$'
```
apolovov, line 1521: Why K is different?

ee/modules/110-istio/openapi/config-values.yaml
```diff
@@ -392,15 +392,15 @@ properties:
...
+                      pattern: "^[0-9]+(\.[0-9]+)?(E|P|T|G|M|k|Ei|Pi|Ti|Gi|Mi|Ki)?$"
                     - type: number
                   min:
                     description: |
                       Minimum allowed memory requests.
                     default: '256Mi'
                     oneOf:
                     - type: string
-                      pattern: "^[0-9]+(E|P|T|G|M|k|Ei|Pi|Ti|Gi|Mi|Ki)?$"
+                      pattern: "^[0-9]+(\.[0-9]+)?(E|P|T|G|M|k|Ei|Pi|Ti|Gi|Mi|Ki)?$"
```
yalosev, line 403: You have to use single quote, not double. Otherwise `\.` will lead to error

candi/openapi/node_group.yaml
apolovov: Don't change it, add a little k to the pattern. And be sure if it is supported.

apolovov: Get off the K's. Focus on dots.

apolovov: Don't change em.

apolovov (changes requested):
Also, add openapi-case-tests.yaml to related openapi's.

ee/modules/110-istio/openapi/openapi-case-tests.yaml
```diff
@@ -1,5 +1,14 @@
 positive:
   configValues:
-    - {}
+    - controlPlane:
+        resourcesManagement:
+          vpa:
+            memory:
+              max: "2G"
+              min: "1G"
```
apolovov, line 8:
```suggestion
              max: "2.000G"
              min: "1.123G"
```

PR #2959, opened 2022-11-03
[node-manager] refactor standby nodes overprovisioning

yalosev:

## Description
Refactor standby node overprovisioning

Now standby-holder resource requests are miscalculated. We decided to order standby-holder requests as a percentage of an existing node. In this way we will always have an appropriate pod which can hold a node.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Calculate resource requests for a stanby-holder Pod as a percentage of a node's capacity.
```


modules/040-node-manager/hooks/discover_standby_ng.go
```diff
@@ -63,6 +63,8 @@ func standbyNodeGroupFilter(obj *unstructured.Unstructured) (go_hook.FilterResul
 
 	needStandby := false
 	maxPerZone := 0
+	overprovisioningRate := int64(50) // default: 50%
```
RomanenkoDenys, line 66: You have default in openapi spec, why you set default in this place ?

RomanenkoDenys (changes requested):
little suggestions

modules/040-node-manager/hooks/discover_standby_ng.go
yalosev: standbyHolder section could not be set. User can set only `standby: 5` for example. Sure we can put `default: {}` there but it will generate extra values, which are usually unneeded.
We should have struct like this:
```
standby:
  type: object
  properties:
    count:
      type: integer
    overprovisioningRate:
      type: integer
```
honestly but not now.
Maybe I have missed... and I have to remove default value from an openapi spec

z9r5 (approved):

deckhouse-BOaTswain:

RomanenkoDenys (approved):
LGTM

deckhouse-BOaTswain:
Aye, aye, @konstantin-axenov. I've started the workflow for label 'e2e/run/aws'...

konstantin-axenov (approved):

PR #2988, opened 2022-11-08
[extended-monitoring] Restart if metrics were last collected more than 15 minutes

nabokihms:

Sometimes extended monitoring exporter can stop sending metrics because of Kubernetes API server connection loss of something.  This PR adds the check that new metrics are collected. Otherwise, kubelet will restart the pod because of the failed liveness probe.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: extended-monitroing
type: fix
summary: Restart if metrics were last collected more than 15 minutes
impact_level: default
```

### DIFF
--- a/modules/340-extended-monitoring/images/extended-monitoring-exporter/src/extended-monitoring.py
+++ b/modules/340-extended-monitoring/images/extended-monitoring-exporter/src/extended-monitoring.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 from concurrent.futures.thread import ThreadPoolExecutor
+from datetime import datetime, timedelta
 from itertools import chain
 from threading import Thread
 from time import sleep
@@ -272,12 +273,14 @@ def _get_metrics():
 class GetHandler(BaseHTTPRequestHandler):
     _response = ""
     _populated = False
+    _last_observe = datetime.now()
 
     @classmethod
     def get_metrics(cls):
         # setting string variable is atomic in Python
         cls._response, quantity = _get_metrics()
         logging.info('Metrics are collected successfully. Batches quantity: {}'.format(quantity))
+        cls._last_observe = datetime.now()
 
     @classmethod
     def loop_get_metrics(cls):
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:

modules/340-extended-monitoring/images/extended-monitoring-exporter/src/extended-monitoring.py
```diff
@@ -301,8 +304,10 @@ def do_GET(self):
             return
 
         if self.path == "/healthz":
-            self.send_response(200)
+            # Fail if metrics were last collected more than 15 minutes ago
+            self.send_response(200 if self.__class__._last_observe > (datetime.now() - timedelta(minutes=15)) else 400)
```
zuzzas, line 308:
```suggestion
            self.send_response(200 if self.__class__._last_observe > (datetime.now() - timedelta(minutes=15)) else 500)
```

That is not a Bad Request, there is something wrong with the server.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

zuzzas (changes requested):

modules/340-extended-monitoring/images/extended-monitoring-exporter/src/extended-monitoring.py
nabokihms: Ok, it makes sense.

modules/340-extended-monitoring/images/extended-monitoring-exporter/src/extended-monitoring.py
```diff
@@ -301,8 +304,10 @@ def do_GET(self):
             return
 
         if self.path == "/healthz":
-            self.send_response(200)
+            # Fail if metrics were last collected more than 15 minutes ago
+            self.send_response(200 if self.__class__._last_observe > (datetime.now() - timedelta(minutes=15)) else 500)
```
shvgn, line 308: For the sake of readability of the decision making

```suggestion
            fresh_enough = self.__class__._last_observe > (datetime.now() - timedelta(minutes=15))
            self.send_response(200 if fresh_enough else 500)
```

shvgn (dismissed):
Despite a small nitpick, LGTM

PR #3012, opened 2022-11-10
[tools] fix in registry migration script

duckhawk:
## Description
Added additional checks to the registry migration script. The script will exit with an error, in case of:
- There is no path in registry URL.
- Domain doesn't match RFC 1123 regex.
- It's impossible to get a Bearer token from the new registry.

In some cases, there can be problems after Deckhouse migration to other registry.

Script will exit with error, in case of:
- There is no path in registry URL.
- Domain doesn't match RFC 1123 regex.
- It's impossible to get a Bearer token from the new registry.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: tools
type: fix
summary: Added additional checks to registry migration script.
impact_level: low
```

### DIFF
--- a/tools/change-registry.sh
+++ b/tools/change-registry.sh
@@ -103,7 +103,7 @@ function parse_args() {
 
   domain_validator="^[a-z0-9][-a-z0-9\.]*[a-z]$"
   if ! [[ $REGISTRY_ADDRESS =~ $domain_validator ]]; then
-    >&2 echo "Domain must contains only symbols from [a-z0-9-.], first symbol must be letter or number, and last symbol must be letter: $REGISTRY_ADDRESS."
+    >&2 echo "Registry domain doesn't fit the regex "^[a-z0-9][-a-z0-9\.]*[a-z]$": $REGISTRY_ADDRESS."
     exit 1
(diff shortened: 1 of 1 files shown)


apolovov (changes requested):
There must be validation with error message like "registry url doesn't fit pattern", not auto-fix.

tools/change-registry.sh
```diff
@@ -81,10 +101,23 @@ function parse_args() {
...
+  fi
+
   if [[ "$REGISTRY_CAFILE" != "" ]] && [[ ! -f "$REGISTRY_CAFILE" ]]; then
     >&2 echo "Cannot find ca file: $REGISTRY_CAFILE."
     exit 1
   fi
+
+  TOKEN="$(bb-rp-get-token)"
+  if [[ "$TOKEN" == "" ]]; then
+    >&2 echo "Cannot get Bearer token from registry"
```
apolovov, line 117:
```suggestion
    >&2 echo "Cannot get Bearer token from registry $REGISTRY_URL"
```

apolovov (approved):

PR #3040, opened 2022-11-14
[DX] Introduce formatting option in common Makefile

shvgn:
## Description

Introduces an opinionated and common go formatting tool, gofumpt.

It makes the formatting more strict, and, more important, automatic and independent from a
particular developer environment settings.

Code formatting should be more consistent including comments due to the corresponding gofumpt 0.4
feature.

## Checklist
- [ ] e2e tests passed.
- [ ] The code if formatted
- [ ] CI check for the formatting is introduced

## Changelog entries

```changes
section: tools
type: feature
summary: Added repo-wise formatting for go code and CI verification.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,14 +83,15 @@ help:
 
 
 GOLANGCI_VERSION = 1.46.2
-TRIVY_VERSION= 0.28.1
+GOFUMPT_VERSION  = 0.4.0
+TRIVY_VERSION    = 0.28.1
 PROMTOOL_VERSION = 2.37.0
-GATOR_VERSION = 3.9.0
-TESTS_TIMEOUT="15m"
+GATOR_VERSION    = 3.9.0
+TESTS_TIMEOUT    = 15m
 
 ##@ General
 
-deps: bin/golangci-lint bin/trivy bin/regcopy bin/jq bin/yq bin/crane bin/promtool bin/gator bin/werf ## Install dev dependencies.
+deps: bin/golangci-lint bin/gofumpt bin/trivy bin/regcopy bin/jq bin/yq bin/crane bin/promtool bin/gator bin/werf ## Install dev dependencies.
 
 ##@ Tests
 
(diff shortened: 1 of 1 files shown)


apolovov:
```
$ make bin/gofumpt
mkdir -p bin
test -f bin/gofumpt-0.4.0 || curl -sLo bin/gofumpt-0.4.0 https://github.com/mvdan/gofumpt/releases/download/v0.4.0/gofumpt_v0.4.0_Darwin_x86_64
chmod u+x bin/gofumpt-0.4.0
ln -s gofumpt-0.4.0 bin/gofumpt
```
```
$ cat ./bin/gofumpt-0.4.0
Not Found
```

shvgn:
@apolovov I cannot reproduce this.

apolovov:
@shvgn 
curl  https://github.com/mvdan/gofumpt/releases/download/v0.4.0/gofumpt_v0.4.0_Darwin_x86_64

404

Makefile
```diff
@@ -132,12 +133,21 @@ bin/golangci-lint:
...
+.PHONY: lint lint-fix format
 lint: ## Run linter.
-	golangci-lint run
+	bin/golangci-lint run
 
 lint-fix: ## Fix lint violations.
-	golangci-lint run --fix
+	bin/golangci-lint run --fix
+
+format:
```
nabokihms, line 149: How it works if a dev forgets to call the format command? Will the CI pipeline fail? If we want to use gofumpt as the main formatter, we need to enforce it by providing additional CI checks (if it doesn't already work this way).

Makefile
```diff
@@ -132,12 +133,21 @@ bin/golangci-lint:
...
+bin/gofumpt:
+	mkdir -p bin
+	test -f bin/gofumpt-$(GOFUMPT_VERSION) || curl -sLo bin/gofumpt-$(GOFUMPT_VERSION) https://github.com/mvdan/gofumpt/releases/download/v$(GOFUMPT_VERSION)/gofumpt_v$(GOFUMPT_VERSION)_$(OS_NAME)_$(PLATFORM_NAME)
+	chmod u+x bin/gofumpt-$(GOFUMPT_VERSION)
+	ln -s gofumpt-$(GOFUMPT_VERSION) bin/gofumpt
+
+.PHONY: lint lint-fix format
 lint: ## Run linter.
-	golangci-lint run
+	bin/golangci-lint run
```
nabokihms, line 144: NIT: I assume that $PWD/bin is in the $PATH, so you don't need to add it here, yet I could be mistaken.

Makefile
```diff
@@ -175,7 +185,7 @@ bin:
 	mkdir -p bin
 
 bin/regcopy: bin ## App to copy docker images to the Deckhouse registry
-	cd tools/regcopy; go build -o bin/regcopy
+	cd tools/regcopy; go get -a; go build -o bin/regcopy
```
nabokihms, line 188: What does `go get -a` in this context?

nabokihms (changes requested):

Makefile
shvgn: I don't have $PWD/bin in my PATH. I don't expect it to work. I noticed that this Makefile called golangci-lint installed by Homebrew, and it has different version

Makefile
shvgn: The action verification is yet to be done

Makefile
shvgn: Forces to download missing modules. Without it, regcopy failed to build on my (fresh) machine

Makefile
```diff
@@ -19,10 +19,12 @@ endif
 
 # Set platform for deps
 ifeq ($(OS_NAME), Linux)
+	PLATFORM = linux
 	JQ_PLATFORM = linux64
 	YQ_PLATFORM = linux
 	TRDL_PLATFORM = linux
```
shvgn, line 25: We have common PLATFORM and ARCH values and a couple of special cases: werf and jq. Why should we treat every tool as a special case?

Makefile
nabokihms: https://github.com/deckhouse/deckhouse/blob/6c008d46b45ec6df7a67eedb9ca17f0e13a2f889/Makefile#L1

PR #3055, opened 2022-11-16
[ingress-nginx] less stringent rules for validating Deckhouse service ingresses

### LINKED ISSUE #2950: [ingress-nginx] Deckhouse get stuck if ingress controller exists without available replicas
### Preflight Checklist

### Version

v1.39

### Expected Behavior

We discussed that this is misconfiguration of Deckhouse. But maybe we can do something to prevent such problem.

### Actual Behavior

Deckhouse get stuck on release update

### Steps To Reproduce

1. Create ingress controller with bad node-selector and enabled validation.
2. Wait while validation webhook is updated
3. Ingress resources can't be upgraded

### Additional Information

_No response_

### Logs

_No response_

duckhawk:
## Description
Added different ValidatingWebhookConfiguration webhooks for system and application ingresses, with different failurePolicy, to avoid Deckhouse stuck during update.

Fix #2950 

Deckhouse won't stuck during upgrade if ingress controller exists without available replicas.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: Added different ValidatingWebhookConfiguration webhooks for service and application ingresses, with different failurePolicy, to avoid Deckhouse stuck during update.
impact_level: low
```

### DIFF
--- a/modules/402-ingress-nginx/templates/controller/admission.yaml
+++ b/modules/402-ingress-nginx/templates/controller/admission.yaml
@@ -15,6 +15,11 @@ webhooks:
     {{- if $crd.spec.validationEnabled }}
   - name: {{ $crd.name }}.validate.d8-ingress-nginx
     matchPolicy: Equivalent
+    namespaceSelector:
+      matchExpressions:
+      - key: heritage
+        operator: NotIn
+        values: deckhouse
     rules:
       - apiGroups:
           - networking.k8s.io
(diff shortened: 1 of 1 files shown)


modules/402-ingress-nginx/templates/controller/admission.yaml
```diff
@@ -34,6 +40,39 @@ webhooks:
...
+          - v1
+        operations:
+          - CREATE
+          - UPDATE
+        resources:
+          - ingresses
+        scope: Namespaced
+    failurePolicy: Ignore
+    sideEffects: None
+    timeoutSeconds: 30
```
yalosev, line 70: I think we can decrease the timeout here. 30 seconds timeout is needed for huge client's ingresses but for our ingresses we can use 3-5 seconds, especially with Ignore failure policy

yalosev (changes requested):
Think about timeout for internal ingresses

modules/402-ingress-nginx/templates/controller/admission.yaml
duckhawk: @yalosev  Maybe at least 10 seconds? I afraid, that 3 seconds can cause false positive even when ingress alive

yalosev (approved):

PR #3103, opened 2022-11-22
[bootstrap] Abort node bootstrap when XFS partition with ftype=0 parameter founded

### LINKED ISSUE #2501: [candi] warn if there is xfs with fstype=0 detected
### Preflight Checklist

There are weird errors when you try to set up a node with xfs fstype=0:
```
FATA[0000] getting status of runtime: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService
```
Also cri has error state:
```
# ctr plugin ls
TYPE                            ID                       PLATFORMS      STATUS
io.containerd.content.v1        content                  -              ok
io.containerd.snapshotter.v1    aufs                     linux/amd64    skip
io.containerd.snapshotter.v1    btrfs                    linux/amd64    skip
io.containerd.snapshotter.v1    devmapper                linux/amd64    error
io.containerd.snapshotter.v1    native                   linux/amd64    ok
io.containerd.snapshotter.v1    overlayfs                linux/amd64    error
io.containerd.snapshotter.v1    zfs                      linux/amd64    skip
io.containerd.metadata.v1       bolt                     -              ok
io.containerd.differ.v1         walking                  linux/amd64    ok
io.containerd.gc.v1             scheduler                -              ok
io.containerd.service.v1        introspection-service    -              ok
io.containerd.service.v1        containers-service       -              ok
io.containerd.service.v1        content-service          -              ok
io.containerd.service.v1        diff-service             -              ok
io.containerd.service.v1        images-service           -              ok
io.containerd.service.v1        leases-service           -              ok
io.containerd.service.v1        namespaces-service       -              ok
io.containerd.service.v1        snapshots-service        -              ok
io.containerd.runtime.v1        linux                    linux/amd64    ok
io.containerd.runtime.v2        task                     linux/amd64    ok
io.containerd.monitor.v1        cgroups                  linux/amd64    ok
io.containerd.service.v1        tasks-service            -              ok
io.containerd.internal.v1       restart                  -              ok
io.containerd.grpc.v1           containers               -              ok
io.containerd.grpc.v1           content                  -              ok
io.containerd.grpc.v1           diff                     -              ok
io.containerd.grpc.v1           events                   -              ok
io.containerd.grpc.v1           healthcheck              -              ok
io.containerd.grpc.v1           images                   -              ok
io.containerd.grpc.v1           leases                   -              ok
io.containerd.grpc.v1           namespaces               -              ok
io.containerd.internal.v1       opt                      -              ok
io.containerd.grpc.v1           snapshots                -              ok
io.containerd.grpc.v1           tasks                    -              ok
io.containerd.grpc.v1           version                  -              ok
io.containerd.grpc.v1           cri                      linux/amd64    error
```

### Proposed Solution

```s_info /var/lib/docker | grep ftype=0 || PANIC```

### Additional Information

_No response_

### PR DESCRIPTION
**Pull request by duckhawk**
## Description
Now node bootstrap will fail if node has an XFS partition with ftype=0 parameter. (if such partition exists, there is a risk of an unsuccessful installation of docker/containerd)

Fix #2501 

Now node bootstrap will fail if node has an XFS partition with ftype=0 parameter.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: bootstrap
type: chore
summary: Now node bootstrap will fail if node has an XFS partition with ftype=0 parameter.
impact_level: low
```


deckhouse-BOaTswain:
Aye, aye, @duckhawk. I've started the workflow for label 'e2e/run/aws'...

candi/bashible/bundles/centos/bootstrap.sh.tpl
```diff
@@ -40,6 +40,13 @@ if [ "${VERSION_ID}" == "8" ] ; then
   alternatives --set python /usr/bin/python3
 fi
 
+for FS_NAME in `egrep -v -e '(^[[:space:]]*$|^#)' /etc/fstab | awk '{ print $1 }'`; do
```
apolovov, line 43: It is legal to have weird xfs anywhere but /var/lib/contaierd.

candi/bashible/bundles/centos/bootstrap.sh.tpl
```diff
@@ -40,6 +40,13 @@ if [ "${VERSION_ID}" == "8" ] ; then
...
+  if  [[ `xfs_info $FS_NAME | grep --count ftype=0` -eq 1 ]]; then
```
apolovov, line 44: What if there isn't a xfs_info? and why don't just grep:
```suggestion
  if  xfs_info $FS_NAME | grep -q ftype=0; then
```

candi/bashible/bundles/centos/bootstrap.sh.tpl
```diff
@@ -40,6 +40,13 @@ if [ "${VERSION_ID}" == "8" ] ; then
...
+     echo "XFS file system with fstop=0 was found on the node ($FS_NAME). This may cause problems (https://www.suse.com/support/kb/doc/?id=000020068), please fix it and try again."
```
apolovov, line 45:
```suggestion
     >&2 echo "XFS file system with fstype=0 was found ($FS_NAME). This may cause problems (https://www.suse.com/support/kb/doc/?id=000020068), please fix it and try again."
```

apolovov (changes requested):

candi/bashible/bundles/centos/bootstrap.sh.tpl
duckhawk: I checked on centos 7/8/9 and rocky 9, looks like xfs_info exists in centos-like distros by default.

Fixed to grep.

candi/bashible/bundles/centos/bootstrap.sh.tpl
duckhawk: It's to complicated to check all variants in script, also, it's unusual case, so, it's simplest method.

RomanenkoDenys, line 43: Why we use /etc/fstab instead of mount | grep xfs ?

candi/bashible/bundles/centos/bootstrap.sh.tpl
```diff
@@ -40,6 +40,13 @@ if [ "${VERSION_ID}" == "8" ] ; then
   alternatives --set python /usr/bin/python3
 fi
 
+for FS_NAME in `egrep -v -e '(^[[:space:]]*$|^#)' /etc/fstab | awk '{ print $1 }'`; do
+  if xfs_info $FS_NAME | grep -q ftype=0; then
```
RomanenkoDenys, line 44: What we do if we do not have xfs_info binary installed ?

RomanenkoDenys (changes requested):

candi/bashible/bundles/centos/bootstrap.sh.tpl
duckhawk: I checked on centos 7/8/9 and rocky 9, looks like xfs_info exists in centos-like distros by default.
So, I didn't added installation of xfsprogs because of bootstrap.sh  size limitation (we must keep it smaller then 16384 bytes for cloud-init)

candi/bashible/bundles/centos/bootstrap.sh.tpl
duckhawk: Didn't thought about mount in this case. Fixed.

apolovov (changes requested):

candi/bashible/bundles/centos/bootstrap.sh.tpl
```diff
@@ -40,6 +40,13 @@ if [ "${VERSION_ID}" == "8" ] ; then
   alternatives --set python /usr/bin/python3
 fi
 
+for FS_NAME in `mount | grep 'type xfs' | awk '{ print $1 }'`; do
```
RomanenkoDenys, line 43:
```suggestion
for FS_NAME in $(mount -l -t xfs | awk '{ print $1 }'); do
```

RomanenkoDenys (commented):
This problem exists only in centos ? Do we have similar problems in alteros and redos ?

RomanenkoDenys:
> > This problem exists only in centos ? Do we have similar problems in alteros and redos ?
> 
> AFAIK we use one bundle for all Centos-like OS, isn't it?

Only for CE releases
For EE/FE releases we have different bundles in /ee/candi

RomanenkoDenys (dismissed):

apolovov (approved):

apolovov (approved):

RomanenkoDenys (approved):

Issue #3097, opened 2022-11-22
[deckhouse-config] more improvements


diafour:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

Some improvements after merging #1729.

### Code

- [ ] 1. res.IsConverted seems redundant. Also, looping over allConfigs can be [deduplicated](https://github.com/deckhouse/deckhouse/pull/1729#discussion_r1025156172).
- [ ] 2. Hard-coded increment for versions. This can reduce arguments for conversion functions.
- [ ] 3. Use slice instead of `conversions map[int]*Conversion` and call Sort in the Add method. See this [comment](https://github.com/deckhouse/deckhouse/pull/1729#discussion_r1025091771).

### Corner cases

- [ ] 1. There is a chance that conversion functions may fail to convert a configuration before AddonOperator starts. There is no way to update image to the hotfix version in this situation.
- 2. Possible soultions:
  2.1. Use separate Deployment for deckhouse updates.
  2.2. Passthrough error to the deckhouse-config module.
- [ ] 3. Drop support for obsolete versions. There should be a support in Disruptive Update mechanism to prevent update when there are ModuleConfig resources with dropped versions.
- [ ] 4. Empty spec.settings (`spec: {"version": 2, "settings": {} }`). There are some modules with required fields on the top level. It is a good idea to show error if spec.settings are empty.

### Helpers
- [ ] 1. Create README for developers on conversion concepts and testing. It may be wiki page.
- [ ] 2. README on how to delete all created resources to start migration again. #2977
- [ ] 3. Create snippet to help migrate configurations #2980

### Features

- [ ] 1. Mutating webhook and an option to enable auto-migration. It may be useful for "one-time" or developer clusters to auto migrate ModuleConfig resources.
- [ ] 2. Write latest settings into the "latest-version-settings-applied" annotation (as with the "last-applied-configuration" annotation).
- [ ] 3. Auto-migration "always on". Enable mutating webhook for all clusters. Discover if 3-way-merge patch in Helm will merge spec.settings. "Always on" will not be an option in this case.
[...]

PR #1729, opened 2022-06-07
[deckhouse-config] new module: use ModuleConfig custom resource for configuration

diafour:
## Description

Linked issues: https://github.com/flant/addon-operator/issues/269

Use multiple ModuleConfig custom resources instead of one ConfigMap/deckhouse.

### Migration from cm/deckhouse to ModuleConfig resources

<details><summary>Schema</summary>
<image src="https://user-images.githubusercontent.com/1055474/200574794-9c85ef59-17cf-48ae-b9a7-d736b23463d1.png" alt="Migration to ModuleConfig resources" />
</details>

## Why do we need it, and what problem does it solve?

- Store configuration in separate CRs instead of the single ConfigMap/deckhouse.
- More clear syntax for module settings.
- Add validating webhook to report about invalid configuration early.
- Use Git, Helm or other IaC tools to manage Deckhouse configuration.
- Add explicit version of module settings, add conversion machinery instead of «migration hooks» with OnStartup binding.

## What is the expected result?

- New ConfigMap created (`kubectl -n d8-system get configmap/deckhouse-generated-config-do-not-edit -o yaml`)
- ModuleConfig resources are created (`kubectl get moduleconfigs`)
- New webhook started (`kubectl get validatingwebhookconfiguration/deckhouse-config-webhook`)
- It is not possible to edit or delete cm/deckhouse (`kubectl -n d8-system delete cm/deckhouse`)

## Checklist

- [x] https://github.com/flant/addon-operator/pull/330
- [x] https://github.com/flant/shell-operator/pull/431
- [x] #2133 
- [x] https://github.com/flant/addon-operator/pull/340
- [x] The code is covered by unit tests.
- [ ] e2e tests passed.
- [x] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Use ModuleConfig objects to configure deckhouse modules.
impact: |
  ConfigMap/deckhouse is forbidden to edit, use ModuleConfig object to change module confguration.
impact_level: high
---
section: istio
type: feature
summary: auth.password option is deprecated
impact: |
  auth.password option is deprecated, consider to enable user-authn module.
impact_level: high
---
section: prometheus
type: feature
summary: auth.password option is deprecated
impact: |
  auth.password option is deprecated, consider to enable user-authn module.
impact_level: high
---
section: cilium-hubble
type: feature
summary: auth.password option is deprecated
impact: |
[...]

PR #3125, opened 2022-11-24
[deckhouse-config] fix: ignore ModuleConfig resources with invalid spec.settings

diafour:
## Description

- Split ConfigValidator.Validate to three methods: ValidateCR, ConvertToLatest, Validate.
- Use ConvertToLatest in initial creation of ModuleConfig
- Ignore ModuleConfig with invalid spec.settings

Giving this invalid "openvpn" section in the ConfigMap/deckhouse:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: deckhouse
  namespace: d8-system
data:
  global: |
    ...
  openvpn: |
    pushToClientRoutes:
        - 10.111.0.0/16
        - 10.222.0.0/16
    pushToClientSearchDomains:
        - pew.mad
  openvpnEnabled: "true"
```

ModuleConfig/openvpn will be created with following status:

A continuation of #1729: update to release with #1729 may fail on outdated cm/deckhouse. Recovery from this fail is non-automatic.

It also fix possible problems with conversions at start (see Corner cases 1 and 5 in #3097)

Deckhouse not fail on cm/deckhouse with outdated invalid module sections.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-config
type: fix
summary: Ignore ModuleConfig resources with invalid spec.settings.
impact_level: low
```


deckhouse-BOaTswain:
Aye, aye, @z9r5. I've started the workflow for label 'e2e/run/aws'...

deckhouse-BOaTswain:
Aye, aye, @z9r5. I've started the workflow for label 'e2e/run/aws'...

deckhouse-BOaTswain:
Aye, aye, @z9r5. I've started the workflow for label 'e2e/run/gcp'...

go_lib/deckhouse-config/validator.go
```diff
@@ -46,74 +46,139 @@ type ValuesValidator interface {
...
 	}
 
-	// Settings are present, check version using conversion chain.
+	// Check if there is registered conversion for the version and if the version is the latest.
 	chain := conversion.Registry().Chain(cfg.GetName())
 	latestVer := chain.LatestVersion()
 
+	// Check if version is unknown.
 	if !chain.IsKnownVersion(cfg.Spec.Version) {
 		previousVersions := concatIntList(chain.PreviousVersionsList())
```
shvgn, line 108: To outline the goal of the variable I'd call `previousVersions` to `previousVersionsMessage` or `prevVersionsMessage` .

go_lib/deckhouse-config/validator.go
```diff
@@ -46,74 +46,139 @@ type ValuesValidator interface {
...
+		errs = append(errs, v.ConversionError)
+	}
+	if v.ValidationError != "" {
+		errs = append(errs, v.ValidationError)
+	}
+	return strings.Join(errs, ", ")
+}
+
+// ValidateCR checks if ModuleConfig resource is well-formed.
+func (c *ConfigValidator) ValidateCR(cfg *d8cfg_v1alpha1.ModuleConfig) ValidationResult {
```
shvgn, line 81: Let's call this method `ValidateModuleConfig` ?

shvgn: Nit: Instead of `ValidationResult`, it could return `error` that `ValidationResult` generates. And then Everything else can be deduced without `ValidationResult`, for example non-nil spec.settings.

go_lib/deckhouse-config/validator.go
```diff
@@ -46,74 +46,139 @@ type ValuesValidator interface {
...
+	// Check if version is unknown.
 	if !chain.IsKnownVersion(cfg.Spec.Version) {
 		previousVersions := concatIntList(chain.PreviousVersionsList())
 		if previousVersions != "" {
 			previousVersions = fmt.Sprintf(", or one of previous versions: %s", previousVersions)
 		}
 
-		return result, fmt.Errorf("spec.version=%d is invalid. Use latest version %d%s", cfg.Spec.Version, latestVer, previousVersions)
+		msg := fmt.Sprintf("spec.version=%d is unsupported. Use latest version %d%s", cfg.Spec.Version, latestVer, previousVersions)
+		if result.HasVersionedSettings {
```
shvgn, line 114: On line 99 you just set effectively 
```go
result.HasVersionedSettings = cfg.Spec.Settings != nil
```

And here you check the same fact.

shvgn: In general I'd not distinguish empty settings from any other state. Why would we want to give it special attention?

go_lib/deckhouse-config/validator.go
```diff
@@ -46,74 +46,139 @@ type ValuesValidator interface {
...
+	}
+	result.ValidateCRWarning = versionMsg
+	return result
+}
+
+// ConvertToLatest checks if ModuleConfig resource is well-formed and runs conversions for spec.settings is needed.
+func (c *ConfigValidator) ConvertToLatest(cfg *d8cfg_v1alpha1.ModuleConfig) ValidationResult {
+	result := c.ValidateCR(cfg)
+	if result.HasError() || !result.HasVersionedSettings {
+		return result
```
shvgn, line 137: Nitpicking

This code would be cleaner this way:

```go
if cfg.Settings == nil {
  return nil 
}
err := := c.ValidateCR(cfg)
if err != nil {
  return err
}
```

Seems, the ValidationResult copies the state of MC and you now track two states of the same data (MC + ValidationResult) instead of one. But what you are trying to do is to get an error, and the knowledge about non-empty settings and changed version. The latter two are easily accessible without ValidationResult. 

If we are to throw away ValidationResult, the code will remain responsible to combine multiple errors into one. [multierr](https://pkg.go.dev/github.com/uber-go/multierr#Combine) package must be useful here.

Additionally, we can wrap errors if we need to split the processing of them. But seems is it not the case.

Additionally, I think currently it is not idiomatic go code.

shvgn (changes requested):

global-hooks/deckhouse-config/startup_sync_test.go
```diff
@@ -161,7 +161,7 @@ unknownModule: |
 			})
 		})
 
-		Context("giving invalid ConfigMap", func() {
+		Context("giving invalid generated ConfigMap", func() {
 			BeforeEach(func() {
 				// Emulate migrated Deployment/deckhouse.
 				_ = os.Setenv("ADDON_OPERATOR_CONFIG_MAP", d8config.GeneratedConfigMapName)
```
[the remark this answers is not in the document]
diafour: Done

global-hooks/deckhouse-config/startup_sync.go
```diff
@@ -202,19 +209,26 @@ func modifyDeckhouseDeploymentToUseGeneratedConfigMap(patchCollector *object_pat
 	patchCollector.Filter(modify, "apps/v1", "Deployment", d8config.DeckhouseNS, "deckhouse")
 }
 
+// syncModuleConfigs updates generated ConfigMap using ModuleConfig resources.
 func syncModuleConfigs(input *go_hook.HookInput, generatedCM *v1.ConfigMap, allConfigs []*d8cfg_v1alpha1.ModuleConfig) error {
+	properCfgs := make([]*d8cfg_v1alpha1.ModuleConfig, 0)
```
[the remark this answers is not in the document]
diafour: I've put it to https://github.com/deckhouse/deckhouse/issues/3097

global-hooks/deckhouse-config/startup_sync.go
```diff
@@ -143,15 +143,22 @@ func createInitialModuleConfigs(input *go_hook.HookInput, cmData map[string]stri
...
 		}
+		// Update spec.settings to converted settings.
 		if res.IsConverted {
 			cfg.Spec.Settings = res.Settings
 			cfg.Spec.Version = res.Version
 		}
+		properCfgs = append(properCfgs, cfg)
 	}
 
 	input.LogEntry.Infof("Create %d ModuleConfig objects", len(configs))
```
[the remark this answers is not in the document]
diafour: Done

go_lib/deckhouse-config/validator.go
diafour: Done

go_lib/deckhouse-config/validator.go
diafour: Discussed. Decision is to make refactoring later.
Use error wrapping and use v1alpha1.ModuleConfig object instead of additional type.

go_lib/deckhouse-config/validator.go
diafour: I've removed this public field, it is now a helper func.

go_lib/deckhouse-config/validator.go
diafour: See https://github.com/deckhouse/deckhouse/pull/3125#discussion_r1032373001

go_lib/deckhouse-config/validator.go
```diff
@@ -46,74 +46,139 @@ type ValuesValidator interface {
...
+		// Resource is valid without spec.version and spec.settings.
+		return result
 	}
-	if cfg.Spec.Settings == nil && cfg.Spec.Version > 0 {
-		return result, fmt.Errorf("spec.version is forbidden when settings are not specified")
+
+	// Can run conversions and validations if spec.version and spec.settings are specified.
+	if cfg.Spec.Settings == nil {
+		// Warn about spec.version without spec.settings.
+		result.ValidateCRWarning = "spec.version is redundant without spec.settings"
```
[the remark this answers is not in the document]
diafour: Discussed:
1. The version alone is not a valid case, but the contract is to ignore it and display warning in the status.
This contract may be distractive and we need to make it more clearer in the future.
2. Empty spec.settings and absent spec.settings should behave equaly.

global-hooks/deckhouse-config/startup_sync.go
```diff
@@ -161,8 +161,8 @@ func createInitialModuleConfigs(input *go_hook.HookInput, cmData map[string]stri
 		properCfgs = append(properCfgs, cfg)
 	}
 
-	input.LogEntry.Infof("Create %d ModuleConfig objects", len(configs))
 	for _, cfg := range configs {
+		input.LogEntry.Infof("Create ModuleConfig/%s", cfg.GetName())
```
shvgn, line 165: "Create" is imperative. Maybe 'Creating'? 'Created'?

go_lib/deckhouse-config/validator.go
```diff
@@ -93,30 +77,28 @@ func (c *ConfigValidator) ValidateCR(cfg *d8cfg_v1alpha1.ModuleConfig) Validatio
 	// Can run conversions and validations if spec.version and spec.settings are specified.
 	if cfg.Spec.Settings == nil {
 		// Warn about spec.version without spec.settings.
-		result.ValidateCRWarning = "spec.version is redundant without spec.settings"
-	} else {
-		// Resource has both spec.settings and spec.version — it is ok to convert and validate with OpenAPI schema.
-		result.HasVersionedSettings = true
+		result.Warning = "spec.version is redundant without spec.settings"
```
shvgn, line 80: "spec.version has no effect without spec.settings, defaults from the latest config version will be applied"

shvgn (approved):

global-hooks/deckhouse-config/startup_sync.go
```diff
@@ -143,24 +143,31 @@ func createInitialModuleConfigs(input *go_hook.HookInput, cmData map[string]stri
...
 		if res.IsConverted {
 			cfg.Spec.Settings = res.Settings
 			cfg.Spec.Version = res.Version
 		}
+		properCfgs = append(properCfgs, cfg)
 	}
 
-	input.LogEntry.Infof("Create %d ModuleConfig objects", len(configs))
 	for _, cfg := range configs {
+		input.LogEntry.Infof("Create ModuleConfig/%s", cfg.GetName())
```
diafour, line 165:
```suggestion
		input.LogEntry.Infof("Creating ModuleConfig/%s", cfg.GetName())
```

go_lib/deckhouse-config/validator.go
```diff
@@ -46,74 +46,123 @@ type ValuesValidator interface {
...
+		// Resource is valid without spec.version and spec.settings.
+		return result
 	}
-	if cfg.Spec.Settings == nil && cfg.Spec.Version > 0 {
-		return result, fmt.Errorf("spec.version is forbidden when settings are not specified")
+
+	// Can run conversions and validations if spec.version and spec.settings are specified.
+	if cfg.Spec.Settings == nil {
+		// Warn about spec.version without spec.settings.
+		result.Warning = "spec.version is redundant without spec.settings"
```
diafour, line 80:
```suggestion
		result.Warning = "spec.version has no effect without spec.settings, defaults from the latest version of settings schema will be applied"
```

shvgn (approved):

yalosev (approved):

deckhouse-BOaTswain:

PR #3222, opened 2022-12-05
[node-manager] fix ng capacity handling

yalosev:

## Description
Fix node capacity value generation

In some cases wrong/missing capacity spec can lead to broken ng spec. I think we can set `nodeCapacity` value only if it's valid.  We have a protection inside the templates `{{- if $ng.nodeCapacity }}` so we will have an error and unscalable from zero node group but it would have a change to break a whole node-group spec and remove all nodes (because of `cloudInstances` is not set)

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Fix node-group template generation when `minPerZone==0` and capacity is not set.
```


konstantin-axenov:
@yalosev Can we create test for this case?

yalosev:
> @yalosev Can we create test for this case?

Will try. We have a trash there with `D8TestInstanceClass`. As I remember I have tried to make tests but don't know what to do with an InstanceClass.
Will make one more try

konstantin-axenov (approved):

deckhouse-BOaTswain:
Cherry pick PR [3229](https://github.com/deckhouse/deckhouse/pull/3229) to the branch [release-1.41](https://github.com/deckhouse/deckhouse/tree/release-1.41) successful!

PR #3263, opened 2022-12-08
[prometheus] lock main queue during update

vitaliy-sn:

## Description

Lock main queue during deckhouse update in clusters with HA enabled.

When we are updating deckhouse and restarting a prometheus pod - we can left without any monitoring and alerts.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: feature
summary: Lock main queue during deckhouse update in clusters with HA enabled.
```

### DIFF
--- a/modules/300-prometheus/hooks/lock_main_queue_during_update.go
+++ b/modules/300-prometheus/hooks/lock_main_queue_during_update.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2022 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package hooks
+
+import (
+	"fmt"
+
+	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
+	"github.com/flant/addon-operator/sdk"
+	"github.com/flant/shell-operator/pkg/kube_events_manager/types"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/utils/pointer"
+)
+
+/*
+Description:
+	locks deckhouse main queue while prometheus Pod will be Ready
+*/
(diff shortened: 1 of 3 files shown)


deckhouse-BOaTswain:
Aye, aye, @vitaliy-sn. I've started the workflow for label 'e2e/run/gcp'...

deckhouse-BOaTswain:
Aye, aye, @vitaliy-sn. I've started the workflow for label 'e2e/run/gcp'...

modules/300-prometheus/hooks/lock_main_queue_during_update.go
```diff
@@ -0,0 +1,119 @@
...
+	return cpod, nil
+}
+
+func handleLockMainQueue(input *go_hook.HookInput) error {
+	if !input.Values.Get("global.clusterIsBootstrapped").Bool() {
+		input.LogEntry.Info("Cluster is not yet bootstrapped, not locking main queue")
+		return nil
+	}
+
+	if !input.Values.Get("global.discovery.clusterControlPlaneIsHighlyAvailable").Bool() {
```
name212, line 95: we can disable high availability in the module config. Do we have to consider this case too?

modules/300-prometheus/hooks/lock_main_queue_during_update.go
```diff
@@ -0,0 +1,119 @@
...
+	cpod := prometheusPod{
+		Name:    pod.Name,
+		IsReady: isReady,
+	}
+
+	return cpod, nil
+}
+
+func handleLockMainQueue(input *go_hook.HookInput) error {
+	if !input.Values.Get("global.clusterIsBootstrapped").Bool() {
```
name212, line 90: Let's imagine the situation: we set up a cluster with one static node (frontend) and one system node.
The frontend node is bootstrapped before the system node. The node is created, the cluster is considered bootstrapped. Then this hook executes, sees that the cluster is bootstrapped, but there are no pods and locates the main queue. Question: will the system node be created?

modules/300-prometheus/hooks/lock_main_queue_during_update_test.go
```diff
@@ -0,0 +1,136 @@
...
+			f.BindingContexts.Set(f.KubeStateSet(runningNotReadyPods))
+			f.ValuesSet("global.discovery.clusterControlPlaneIsHighlyAvailable", false)
+			f.RunHook()
+		})
+
+		It("Must be executed successfully", func() {
+			Expect(f).To(ExecuteSuccessfully())
+		})
+
+	})
```
name212, line 134: I think we need some tests:
- when cluster is not bootstrapped
- prometheus is not in HA mode

name212 (commented):
tests and questions

name212 (changes requested):
questions and tests

modules/300-prometheus/hooks/lock_main_queue_during_update.go
vitaliy-sn: I think yes. Added high availability check from module config.

PR #3399, opened 2022-12-22
[deckhouse]: Add auth to update notification

shvgn:
## Description

Add authentication settings to the update notification hook.

For the inner Flant infrastructure, we need to be able to send update notifications to the clients
from their clusters.

The update notification hook can be configured to send  authenticated notifications requests.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Added authentication settings to the update notification hook
```

### DIFF
--- a/modules/002-deckhouse/openapi/config-values.yaml
+++ b/modules/002-deckhouse/openapi/config-values.yaml
@@ -137,7 +137,7 @@ properties:
                 "requirements":  {"k8s": "1.20.0"},
                 "changelogLink": "https://github.com/deckhouse/deckhouse/changelog/1.36.md",
                 "applyTime": "2023-01-01T14:30:00Z00:00",
-                "message": "New Deckhouse Release 1.36 is available. Release will be applied at: Friday, 01-Jan-22 14:30:00 UTC"
+                "message": "New Deckhouse Release 1.36 is available. Release will be applied at: Friday, 01-Jan-23 14:30:00 UTC"
               }
               ```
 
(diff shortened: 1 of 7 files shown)


deckhouse-BOaTswain:
Aye, aye, @shvgn. I've started the workflow for label 'e2e/run/gcp'...

deckhouse-BOaTswain:
Aye, aye, @shvgn. I've started the workflow for label 'e2e/run/yandex-cloud'...

yalosev (approved):

nabokihms (approved):

modules/002-deckhouse/openapi/config-values.yaml
```diff
@@ -161,6 +164,31 @@ properties:
               The update mechanism ensures that Deckhouse will not be updated before the specified time.
 
               When using update windows, the Deckhouse update after the notification will happen at the nearest possible update window, but not before the time specified in `minimalNotificationTime` expires.
+          auth:
+            type: object
```
nabokihms, line 168: 1. There should be validation that either basic auth or token can be provided.
2. The naming can be the same as for PrometheusRemoteWrite (I think we use the same option names in Upmeter too).

https://deckhouse.io/documentation/v1/modules/300-prometheus/cr.html#prometheusremotewrite-v1-spec
```
bearerToken:
basicAuth:
  username:
  password:
```
And with this structure, it is easier to validate that only one field is provided.

3. We usually use the `auth` key to describe module authentication params (that will be used to authenticate and authorize users to get access to this module). Still, maybe it is ok because, in this case, `auth` is a nested field.

nabokihms (changes requested):
Some thoughts about configuration.

modules/002-deckhouse/openapi/config-values.yaml
shvgn: Improved.

nabokihms (approved):

PR #3428, opened 2022-12-28
[node-local-dns] Switch to "verified" mode

zuzzas:
## Description

Switched from "immediate" to "verified" stale_cache mode.

Some programs resolve a Service name only once and then attempt to connect to it without retrying DNS resolution. stale_cache's `immediate` mode results in node-local-dns sending a response (granted, with TTL 0) to the client, and only then attempting to refresh the cache. In `verified` mode, node-local-dns will attempt to first get a fresh response and only send the stale one to the client if the refresh fails.

https://coredns.io/plugins/cache/

https://github.com/coredns/coredns/blob/66f2ac7568ccb0178cc9ce6dbd7320bcd3428d64/plugin/cache/handler.go#L43-L66

`stale_cache` always attempts to send a fresh Resource Record to the client.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: node-local-dns
type: fix
summary: Switched stale cache behavior from `immediate` to `verified`
impact_level: default
```

### DIFF
--- a/ee/modules/350-node-local-dns/templates/configmap.yaml
+++ b/ee/modules/350-node-local-dns/templates/configmap.yaml
@@ -16,7 +16,7 @@ data:
         success 39936
         denial 9984
         prefetch 10 1m 25%
-        serve_stale 1h immediate
+        serve_stale 1h verified
       }
       reload 2s
       loop


deckhouse-BOaTswain:
Aye, aye, @zuzzas. I've started the workflow for label 'e2e/run/yandex-cloud'...

RomanenkoDenys (approved):
LGTM

konstantin-axenov (approved):

PR #3554, opened 2023-01-16
[chrony] Adding a dot to the end of each NTP server address to avoid resolve problems. Allowing correction of large time difference

duckhawk:
## Description

Adding a dot to the end of each NTP server address to avoid resolve problems. Allowing correction of large time difference.

Incorrect DNS resolve is possible in case of wildcard domains in search. In case of a large difference in local time with the world time correction may not work. This PR corrects both situations.

Chrony correct work.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: chrony
type: fix
summary: Adding a dot to the end of each NTP server address to avoid resolve problems. Allowing correction of large time difference.
impact_level: low
```

### DIFF
--- a/modules/470-chrony/images/chrony/start.sh
+++ b/modules/470-chrony/images/chrony/start.sh
@@ -34,11 +34,20 @@ cmdallow 127/8
 allow 127/8
 bindaddress 127.0.0.1
 driftfile /var/run/chrony/chrony.drift
-makestep 1.0 3
+makestep 1.0 -1
 rtcsync
 EOF
 for NTP_SERVER in ${NTP_SERVERS}; do
-  echo "pool ${NTP_SERVER} iburst" >> /var/run/chrony/chrony.conf
+  # Check, if there is an IP address, then pass it to file
+  if [[ $NTP_SERVER =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
+      echo "pool ${NTP_SERVER} iburst" >> /var/run/chrony/chrony.conf
+  # Else, if there is a dot in the end of DNS-address, then pass it to file
+  elif [[ $NTP_SERVER =~ ^.*\.$ ]]; then
+      echo "pool ${NTP_SERVER} iburst" >> /var/run/chrony/chrony.conf
+  # In all other cases, we should add dot, then pass it to file
+  else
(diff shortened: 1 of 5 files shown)


z9r5 (dismissed):

modules/470-chrony/images/chrony/start.sh
```diff
@@ -34,10 +34,13 @@ cmdallow 127/8
...
 bindaddress 127.0.0.1
 driftfile /var/run/chrony/chrony.drift
-makestep 1.0 3
+makestep 1.0 -1
 rtcsync
 EOF
 for NTP_SERVER in ${NTP_SERVERS}; do
+  if [ ${NTP_SERVER:0-1} != "." ]; then
+    NTP_SERVER=$NTP_SERVER"."
+  fi
```
RomanenkoDenys, line 43: Please use trimSuffix func in helm for it.

RomanenkoDenys (changes requested):

modules/470-chrony/images/chrony/start.sh
duckhawk: But we need to add a dot here as a suffix, not to remove it.

duckhawk: This script runs on container startup, so we can't use helm functions here. Also, I'd added check for IP address in ntp servers.

modules/470-chrony/images/chrony/start.sh
```diff
@@ -34,11 +34,20 @@ cmdallow 127/8
...
+  # Check, if there is an IP address, then pass it to file
+  if [[ $NTP_SERVER =~ ^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
+      echo "pool ${NTP_SERVER} iburst" >> /var/run/chrony/chrony.conf
+  # Else, if there is a dot in the end of DNS-address, then pass it to file
+  elif [[ $NTP_SERVER =~ ^.*\.$ ]]; then
+      echo "pool ${NTP_SERVER} iburst" >> /var/run/chrony/chrony.conf
+  # In all other cases, we should add dot, then pass it to file
+  else
+      echo "pool ${NTP_SERVER}. iburst" >> /var/run/chrony/chrony.conf
+  fi
```
RomanenkoDenys, line 50: Why we do not do it in the helm manifest ?

RomanenkoDenys (commented):
question

RomanenkoDenys (approved):
LGTM

PR #3618, opened 2023-01-24
[secret-copier] Add annotations with timestamp

yalosev:
## Description
Add annotation to a copied secret

It's useful for debug purposes to have an annotation when secret was created/update by secret-copier

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: secret-copier
type: chore
summary: Add annotation with create/update timestamp to copied secrets
```

### DIFF
--- a/modules/600-secret-copier/hooks/handler.go
+++ b/modules/600-secret-copier/hooks/handler.go
@@ -73,6 +73,12 @@ func ApplyCopierSecretFilter(obj *unstructured.Unstructured) (go_hook.FilterResu
 		Type:        secret.Type,
 		Data:        secret.Data,
 	}
+
+	if secret.Namespace != v1.NamespaceDefault {
+		// desired secret (target secret in a namespace) must not have annotations to satisfy DeepEqual function
+		secret.Annotations = nil
+	}
+
 	// Secrets with that label lead to D8CertmanagerOrphanSecretsChecksFailed alerts.
 	delete(s.Labels, "certmanager.k8s.io/certificate-name")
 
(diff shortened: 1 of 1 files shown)


modules/600-secret-copier/hooks/handler.go
```diff
@@ -159,9 +159,13 @@ func copierHandler(input *go_hook.HookInput, dc dependency.Container) error {
 	secretsDesired := make(map[string]*Secret)
 	for _, s := range secrets {
 		secret := s.(*Secret)
+
 		// Secrets that are not in namespace `default` are existing Secrets.
 		if secret.Namespace != v1.NamespaceDefault {
 			path := SecretPath(secret)
+			// remove annotations because desired secret doesn't have them
+			// annotations map have to be nil that secrets are equal
+			secret.Annotations = nil
```
apolovov, line 168: It is better to do in ApplyCopierSecretFilter.

apolovov: And i didn't get the comment.

apolovov (changes requested):

modules/600-secret-copier/hooks/handler.go
yalosev: Nope, annotations are needed after apply function, we use some value from them. But below this code we have a DeepEqual function, which is comparing `desired` secret with `existing` secret and `desired` secret does not have annotations, so we have to remove them from `existing` to satisfy equaty

apolovov (approved):

RomanenkoDenys (approved):

PR #3624, opened 2023-01-25
[istio] add CLOUD_PROVIDER env var for istio proxy

fl64:
## Description
Added control of an environment variable for obtaining cloud provider metadata for istio proxy

Automatic cloud provider detection must be disabled to avoid unwanted network interaction

For istio-proxy, getting cloud provider metadata is available for: azure, aws, gcp for other cases - disabled.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: fix
summary: Added control of an environment variable for obtaining cloud provider metadata for istio proxy
impact_level: low
```

### DIFF
--- a/ee/modules/110-istio/templates/control-plane/iop.yaml
+++ b/ee/modules/110-istio/templates/control-plane/iop.yaml
@@ -1,8 +1,10 @@
 {{ $supportedProviders := list "aws" "gcp" "azure" }}
-{{ $currentProvider := .Values.global.clusterConfiguration.cloud.provider | lower }}
 {{ $cloudPlatform := "none" }}
-{{ if has $currentProvider $supportedProviders }}
-  {{ $cloudPlatform = $currentProvider }}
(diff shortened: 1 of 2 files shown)


ee/modules/110-istio/templates/control-plane/iop.yaml
```diff
@@ -1,3 +1,10 @@
+{{ $supportedProviders := list "aws" "gcp" "azure" }}
+{{ $currentProvider := .Values.global.clusterConfiguration.cloud.provider | lower }}
```
apolovov, line 2: It will fail in static cluster. There isn't a `cloud` section.

apolovov (changes requested):

ee/modules/110-istio/templates/control-plane/iop.yaml
fl64: i didn't know about it, ok i wil fix it

apolovov (approved):

PR #3628, opened 2023-01-25
[bashible] fix bb-event-error-create

RomanenkoDenys:
Signed-off-by: Denis Romanenko <denis.romanenko@flant.com>

## Description

Fix error in bb-event-error-create bashible function.

1. When bashible tries to create an event, we get an error because we do not have access from system:nodes group to events.k8s.io api.
2. If cluster in bootstrapping state, we cannot have working API to send events, so we simply suppress event creating errors to prevent bashible fail. 
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: Fix error in bb-event-error-create bashible function.
impact_level: low
```

### DIFF
--- a/candi/bashible/bashible.sh.tpl
+++ b/candi/bashible/bashible.sh.tpl
@@ -36,20 +36,18 @@ function bb-event-error-create() {
         kind: Event
         metadata:
           name: bashible-error-${eventName}
-          annotations:
-            timestamp: '$(date -u +"%Y-%m-%dT%H:%M:%SZ")'
         regarding:
           apiVersion: v1
           kind: Node
           name: '$(hostname -s)'
           uid: "$(kubectl_exec get node $(hostname -s) -o jsonpath='{.metadata.uid}')"
         note: '$(tail -c 500 ${eventLog})'
-        reason: Failed
+        reason: BashibleStepFailed
         type: Warning
         reportingController: bashible
         reportingInstance: '$(hostname -s)'
         eventTime: '$(date -u +"%Y-%m-%dT%H:%M:%S.%6NZ")'
-        action: "Binding"
+        action: "BashibleStepExecution"
 EOF
 }
 
--- a/candi/bashible/bashible.sh.tpl
+++ b/candi/bashible/bashible.sh.tpl
@@ -31,7 +31,7 @@ function bb-event-error-create() {
     step="$1"
     eventName="$(echo -n "$(hostname -s)")-$(echo $step | sed 's#.*/##; s/_/-/g')"
     eventLog="/var/lib/bashible/step.log"
-    kubectl_exec apply -f - || true <<EOF
+    kubectl_exec apply -f - <<EOF || true
         apiVersion: events.k8s.io/v1
         kind: Event
         metadata:
--- a/candi/bashible/bashible.sh.tpl
+++ b/candi/bashible/bashible.sh.tpl
@@ -27,10 +27,11 @@ function bb-event-error-create() {
     # underscore with dash due to regexp.
     # All of stderr outputs are stored in the eventLog file.
     # step is used as argument for function call.
+    # If event creation failed, error from kubectl suppressed.
     step="$1"
     eventName="$(echo -n "$(hostname -s)")-$(echo $step | sed 's#.*/##; s/_/-/g')"
     eventLog="/var/lib/bashible/step.log"
-    kubectl_exec apply -f - <<EOF
+    kubectl_exec apply -f - || true <<EOF
         apiVersion: events.k8s.io/v1
         kind: Event
         metadata:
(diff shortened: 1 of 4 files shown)


deckhouse-BOaTswain:
Aye, aye, @RomanenkoDenys. I've started the workflow for label 'e2e/run/aws'...

deckhouse-BOaTswain:

candi/bashible/bashible.sh.tpl
```diff
@@ -27,10 +27,11 @@ function bb-event-error-create() {
     # underscore with dash due to regexp.
     # All of stderr outputs are stored in the eventLog file.
     # step is used as argument for function call.
+    # If event creation failed, error from kubectl suppressed.
     step="$1"
     eventName="$(echo -n "$(hostname -s)")-$(echo $step | sed 's#.*/##; s/_/-/g')"
     eventLog="/var/lib/bashible/step.log"
-    kubectl apply -f - <<EOF
+    kubectl_exec apply -f - <<EOF || true
```
yalosev, line 34: should not we also move `stderr` to `/dev/null` ?

RomanenkoDenys: no because if kubectl exit code > 0 bashible fails

yalosev (dismissed):

testing/matrix/linter/rules/roles/placement.go
```diff
@@ -316,7 +316,7 @@ func handleNestedRBACForUs(m utils.Module, object storage.StoreObject, shortPath
 
 	switch {
 	case strings.HasPrefix(objectName, localPrefix):
-		if namespace != m.Namespace {
+		if namespace != m.Namespace && namespace != "default" {
```
nabokihms, line 319: Oh no 😞

nabokihms: This condition must not be changed. `default` is considered a system namespace, so the rule's name must start with `d8:` prefix.

modules/040-node-manager/templates/bashible/rbac-for-us.yaml
```diff
@@ -52,6 +52,20 @@ rules:
...
+metadata:
+  name: bashible-events
+  namespace: default
+  {{- include "helm_lib_module_labels" (list . ) | nindent 2 }}
+rules:
+  - apiGroups: ["events.k8s.io"]
+    resources:
+      - events
+    verbs:
+      - "*"
```
nabokihms, line 67: Too broad, consider specifying only required verbs

candi/bashible/bashible.sh.tpl
```diff
@@ -41,7 +42,7 @@ function bb-event-error-create() {
           apiVersion: v1
           kind: Node
           name: '$(hostname -s)'
-          uid: "$(kubectl get node $(hostname -s) -o jsonpath='{.metadata.uid}')"
+          uid: "$(kubectl_exec get node $(hostname -s) -o jsonpath='{.metadata.uid}')"
```
nabokihms, line 45: BTW, uid is not required and can be avoided.

nabokihms (changes requested):
Some questions regarding the event error create function:
1. The reason field is always `Failed`. `Failed` is too abstract. I expected to see an instruction on what should I do next, or at least, what is broken.

2. The action field is always equal to `Binding`. I assume it should be something like `BashibleFailed`.
> action is what action was taken/failed regarding to the regarding object. It is machine-readable. This field cannot be empty for new Events and it can have at most 128 characters.

What is the meaning of the `Binding`?

3. Why is there the `annotations.timestamp` field? It seems like eventTime is enough.

testing/matrix/linter/rules/roles/placement.go
apolovov: btw, it is better to use v1.NamespaceDefault.

candi/bashible/bashible.sh.tpl
yalosev: Nope. Without uid event is not binded to the object. I mean when you make `kubectl describe node X` - you will not fetch an event without id

testing/matrix/linter/rules/roles/placement.go
RomanenkoDenys: fixed by proper way

RomanenkoDenys:
> Some questions regarding the event error create function:
> 
> 1. The reason field is always `Failed`. `Failed` is too abstract. I expected to see an instruction on what should I do next, or at least, what is broken.

changed to BashibleStepFailed

> 2. The action field is always equal to `Binding`. I assume it should be something like `BashibleFailed`.
> 
> > action is what action was taken/failed regarding to the regarding object. It is machine-readable. This field cannot be empty for new Events and it can have at most 128 characters.
> 
> What is the meaning of the `Binding`?

changed to BashibleStepExecution

> 
> 3. Why is there the `annotations.timestamp` field? It seems like eventTime is enough.

removed

modules/040-node-manager/templates/bashible/rbac-for-us.yaml
RomanenkoDenys: changed to get, list, create, update, patch

yalosev (approved):

nabokihms (approved):

deckhouse-BOaTswain:
Aye, aye, @RomanenkoDenys. I've started the workflow for label 'e2e/run/aws'...

apolovov (approved):

PR #3660, opened 2023-01-27
[deckhouse] add bash wrapper for handling USR signals

fl64:
## Description
Added bash wrapper, allowing processing of USR1, USR2 signals, which is used to restart the deckhouse child process.

This supervisor is required for the use of 3rd party modules.

By receiving USR1, USR2 the deckhouse child process will receive a TERM signal and be restarted.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Added bash wrapper for handling USR signals.
impact_level: default
```

### DIFF
--- a/deckhouse-controller/entrypoint.sh
+++ b/deckhouse-controller/entrypoint.sh
@@ -34,4 +34,32 @@ EOF
 
 cat ${MODULES_DIR}/values-${bundles_map[$bundle]}.yaml >> ${MODULES_DIR}/values.yaml
 
-exec /sbin/tini -- /usr/bin/deckhouse-controller start
+PID=0
+EXITCODE=0
+
+for SIG in SIGUSR1 SIGUSR2 SIGINT SIGTERM SIGHUP SIGQUIT; do
+  trap "signal_handler ${SIG}" "${SIG}"
+done
+
+signal_handler() {
+  case "${1}" in
+  "SIGUSR1" | "SIGUSR2")
+    kill "${PID}"
+    wait "${PID}"
+    run_deckhouse
+    ;;
+  *)
+    kill -"${1}" "${PID}"
+    ;;
+  esac
+}
+
+run_deckhouse() {
+  /usr/bin/deckhouse-controller start &
+  PID="${!}"
+  wait "${PID}"
+  EXITCODE="${?}"
+}
+
+run_deckhouse
+exit "${EXITCODE}"
--- a/deckhouse-controller/entrypoint.sh
+++ b/deckhouse-controller/entrypoint.sh
@@ -40,12 +40,8 @@ set +e
 PID=0
 EXITCODE=0
 
-for SIG in SIGUSR1 SIGUSR2 SIGINT SIGTERM SIGHUP SIGQUIT; do
-  trap "signal_handler ${SIG}" "${SIG}"
-done
-
 signal_handler() {
-  echo "{\"msg\": \"Catch signal ${1}\"}"
+  echo "{\"level\":\"warning\", \"msg\": \"Catch signal ${1}\"}"
   case "${1}" in
   "SIGUSR1" | "SIGUSR2")
     kill "${PID}"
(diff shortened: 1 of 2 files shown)


deckhouse-controller/entrypoint.sh
```diff
@@ -34,4 +34,37 @@ EOF
...
+    run_deckhouse
+    ;;
+  *)
+    kill -"${1}" "${PID}"
+    ;;
+  esac
+}
+
+run_deckhouse() {
+  /usr/bin/deckhouse-controller start &
```
nabokihms, line 62: tiny is missing here. Deckhous is using it to cleanup zombies left after running hooks written on python and bash.

deckhouse-controller/entrypoint.sh
```diff
@@ -34,4 +34,37 @@ EOF
...
-exec /sbin/tini -- /usr/bin/deckhouse-controller start
+set +o pipefail
+set +e
+
+PID=0
+EXITCODE=0
+
+for SIG in SIGUSR1 SIGUSR2 SIGINT SIGTERM SIGHUP SIGQUIT; do
+  trap "signal_handler ${SIG}" "${SIG}"
+done
```
kvaps, line 45: Let's move the logic under the functions definitions

deckhouse-controller/entrypoint.sh
```diff
@@ -34,4 +34,37 @@ EOF
...
+    ;;
+  esac
+}
+
+run_deckhouse() {
+  /sbin/tini -s -- /usr/bin/deckhouse-controller start &
+  PID="${!}"
+  echo "{\"level\":\"info\", \"msg\": \"New deckhouse PID ${PID}\"}"
+  wait "${PID}"
+  EXITCODE="${?}"
```
kvaps, line 66: IMHO setting global variables from functions is slightly confusing, I would prefer to use `return $?` instead

deckhouse-controller/entrypoint.sh
fl64: fixed

deckhouse-controller/entrypoint.sh
fl64: Could you be a little more specific?

deckhouse-controller/entrypoint.sh
```diff
@@ -34,4 +34,37 @@ EOF
...
+}
+
+run_deckhouse
+exit "${EXITCODE}"
```
kvaps, line 70: This can be omitted, because run_deckhouse is already latest command.
I don't see the problem of seeing this, just see no reason

deckhouse-controller/entrypoint.sh
kvaps: solved through videocall

deckhouse-controller/entrypoint.sh
kvaps: solved through videocall

deckhouse-controller/entrypoint.sh
```diff
@@ -34,4 +34,37 @@ EOF
...
+    run_deckhouse
+    ;;
+  *)
+    kill -"${1}" "${PID}"
+    ;;
+  esac
+}
+
+run_deckhouse() {
+  /sbin/tini -s -- /usr/bin/deckhouse-controller start &
```
kvaps, line 58: I'd like to use tini as a main supervisor to run entrypoint.sh, but not vise versa

kvaps (approved):

kvaps (approved):
lgtm

werf.yaml
```diff
@@ -255,9 +255,10 @@ ansible:
         {{- .Env | default "unknown" | nindent 8 }}
       dest: /deckhouse/edition
 
-  # Because of https://github.com/flant/werf/issues/1741 just make symlink for the entry point
-  - name: "Make symlink for deckhouse entrypoint in the right location"
-    shell: ln -s /deckhouse/deckhouse-controller/entrypoint.sh /deckhouse/deckhouse
+  - name: "Override deckhouse entrypoint to use tini as supervisor"
+    shell: |
+      echo -e "#!/bin/bash\nexec tini -- /deckhouse/deckhouse-controller/entrypoint.sh" > /deckhouse/deckhouse
```
yalosev, line 260:
```
cat <<"EOF" > /deckhouse/deckhouse
#!/bin/bash
exec tini -- /deckhouse/deckhouse-controller/entrypoint.sh
EOF
```

fl64: Alas, it didn't work as expected, so I had to go for a dirty hack

yalosev (approved):

nabokihms (approved):

PR #4047, opened 2023-03-09
[extended-monitoring] exclude openvpn-pki secrets from CertificateSec…

stratnevpy:
## Description
Exclude openvpn clients secrets from alert CertificateSecretExpiredSoon

At the moment, the certificates that are issued to clients for connecting via openvpn get into the alert.

Need to exclude that secrets from alerting 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix 
summary: Exclude openvpn clients secrets from alert CertificateSecretExpiredSoon
impact_level: low
```

### DIFF
--- a/modules/340-extended-monitoring/monitoring/prometheus-rules/certificates/certificate.yaml
+++ b/modules/340-extended-monitoring/monitoring/prometheus-rules/certificates/certificate.yaml
@@ -3,7 +3,7 @@
   - alert: CertificateSecretExpiredSoon
     expr: |
       max by (name, namespace) (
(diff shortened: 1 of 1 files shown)


zuzzas (approved):

nabokihms:
On hold for now. We discussed it internally:
* certificates monitoring is opt-in
* it does not seem worth filtering ovpn certs by default because there will be no way to enable it

One more thing, some descriptions of alerts contain info about certmanager, which can be misleading.

nabokihms:
Closing this as not planned

PR #4099, opened 2023-03-14
[log-shipper] Add key_field and exclude fileds for rate limit config for CLusterLogDestination

### LINKED ISSUE #3994: [log-shipper] Rate limit improvements (QoS)
### Preflight Checklist

There is a [rate limit feature](https://deckhouse.io/documentation/v1/modules/460-log-shipper/cr.html#clusterlogdestination-v1alpha1-spec-ratelimit) that can help users to throttle requests to the destination.  It can prevent overloads. However, it is not convenient for QoS. If a namespace produces more logs than it is supposed to, other namespaces will be limited alongside.

### Proposed Solution

Add the [`key_field`](https://vector.dev/docs/reference/configuration/transforms/throttle/#key_field) to be able to limit requests by namespace or other metadata info, e.g., by controller.

### Additional Information

_No response_

alex123012:

Fix #3994
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: log-shipper
type: feature
summary: Add key_field and exclude fileds for rate limit config for CLusterLogDestination
impact_level: default
```


alex123012:
**Note**: before merge - check what happens when `keyField` isn't valid log label

modules/460-log-shipper/hooks/testdata/throttle-with-filter/result.json
```diff
@@ -0,0 +1,159 @@
...
+  "transforms": {
+    "transform/destination/test-es-dest/00_elastic_dedot": {
+      "drop_on_abort": false,
+      "inputs": [
+        "transform/source/test-source/01_clean_up"
+      ],
+      "source": "if exists(.pod_labels) {\n    .pod_labels = map_keys(object!(.pod_labels), recursive: true) -\u003e |key| { replace(key, \".\", \"_\") }\n}",
+      "type": "remap"
+    },
+    "transform/destination/test-es-dest/01_ratelimit": {
```
nabokihms, line 30: There is a problem that each rate_limit transform has it's [own state](https://vector.dev/docs/reference/configuration/transforms/throttle/). Entries are limited separately. 

If you have a 500 entries limit and two rules in the exclude section with the following conditions for logs
1. Are not excluded.
2. Excluded only because of the first rate limit rule.
the actual limit of logs will be 1000.

To solve this, we need to find a way to build a single VRL for all excludes.

modules/460-log-shipper/hooks/internal/vector/transform/throttle.go
```diff
@@ -17,24 +17,84 @@ limitations under the License.
...
+
+const (
+	excludeField  = "exclude"
+	keyFieldField = "key_field"
 )
 
 // ThrottleTransform adds throttling to event's flow.
-func ThrottleTransform(rl v1alpha1.RateLimitSpec) *DynamicTransform {
+func ThrottleTransform(rl v1alpha1.RateLimitSpec) ([]apis.LogTransform, error) {
+	if rl.Excludes != nil && rl.KeyField != "" {
```
nabokihms, line 35: Didn't get this. 

i think it is valid to have non empty excludes with an empty key field.

modules/460-log-shipper/hooks/internal/vector/transform/throttle.go
```diff
@@ -17,24 +17,84 @@ limitations under the License.
...
+		if err := setExcludeThrottleTransformDynamicArg(throttleTransform.DynamicArgsMap, &filter); err != nil {
+			return nil, err
+		}
+
+		throttleTransforms[i] = throttleTransform
+	}
+	return throttleTransforms, nil
+}
+
+func setExcludeThrottleTransformDynamicArg(m map[string]interface{}, exclude *v1alpha1.Filter) error {
```
nabokihms, line 56: NIT: can be inlined because this function is used in one place.

modules/460-log-shipper/hooks/testdata/throttle-with-filter/result.json
alex123012: ok, got it, I'll try to fix

modules/460-log-shipper/hooks/internal/vector/transform/throttle.go
```diff
@@ -17,24 +17,90 @@ limitations under the License.
...
+	if err != nil {
+		return nil, fmt.Errorf("error rendering exclude rule for dynamic transform: %w", err)
+	}
+
+	return &condition, nil
+}
+
+func combineThrottleTransformExcludes(excludes []string) string {
+	resultExcludeConds := make([]string, len(excludes)+1)
+	resultExcludeConds[0] = "matchedExcludeCond"
```
nabokihms, line 97: NIT
```suggestion
	resultExcludeConds[0] = "matched_exclude_cond"
```

modules/460-log-shipper/hooks/internal/vector/transform/throttle.go
```diff
@@ -17,24 +17,90 @@ limitations under the License.
...
+	resultCond := fmt.Sprintf("%s = false;\n", resultExcludeConds[0])
+
+	for i, excludeCond := range excludes {
+		resultExcludeConds[i+1] = fmt.Sprintf("matchedExcludeCond%d", i)
```
nabokihms, line 101: I don't think you need this variable.  In VRL, each block is variable, so you can just merge them.

```rust
{ true } || { false }

true
```



Instead of rendering all the rules and saving them to an array, the much cleaner way to do this merge is using a string builder.
```go
var trottleResult strings.Builder
trottleResult.WriteString("false")

for _, filter := range excludes {
  	rule := getRuleOutOfFilter(exclude)
	condition, err := rule.Render(vrl.Args{"filter": filter})
	if err != nil {
		return nil, fmt.Errorf("error rendering exclude rule for dynamic transform: %w", err)
	}
     
    trottleResult.WriteString(" || ")
    trottleResult.WriteString(condition)
}

return trottleResult.String()
```

nabokihms: It seems like an abstraction leak, vrl building in transform package, but it can live like this for a while.

nabokihms (commented):
A couple of things left, but overall it looks good.

alex123012:
@nabokihms there were also typo in buffer settings (from #4095), i've fixed it there, is it ok?

nabokihms (approved):

PR #4216, opened 2023-03-24
fix shell executable

HeyLazySunnyKid:
## Description
Fix shell binary using by ansible in FAQ example. Ansible shell module use /bin/sh by default, but the script is written on bash.

Correct example in documentation without syntax error in shell during execution.

Try to add node, using ansible script in documentation without syntax errors.

## Checklist
- [X] Documentation updated according to the changes.

## Changelog entries
```changes
section: docs
type: fix
summary: Fix the Ansible script example for adding a node.
impact: documentation
impact_level: low
```


z9r5 (approved):

z9r5:
@HeyLazySunnyKid Thanks

PR #4245, opened 2023-03-28
feat: command to enable/disable module via ModuleConfig

diafour:
## Description

Add commands to deckhouse-controller binary:

- `deckhouse-controller module module-name enable`
- `deckhouse-controller module module-name enable`

Commands change `spec.enabled` flag or create new ModuleConfig resource.

Required exposing DebugServer from AddonOperator: https://github.com/flant/addon-operator/pull/369

This PR gives users a way to enable/disable module without YAML editing.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Add commands to enable/disable modules without YAML editing
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/debug/module_enable.go
+++ b/deckhouse-controller/pkg/debug/module_enable.go
@@ -1,0 +1,81 @@
+package debug
+
+import (
+	"fmt"
+	"net/http"
+
+	addon_operator "github.com/flant/addon-operator/pkg/addon-operator"
+	sh_app "github.com/flant/shell-operator/pkg/app"
+	sh_debug "github.com/flant/shell-operator/pkg/debug"
+	"github.com/go-chi/chi/v5"
+	"gopkg.in/alecthomas/kingpin.v2"
+
+	deckhouse_config "github.com/deckhouse/deckhouse/go_lib/deckhouse-config"
+)
+
+func RegisterModuleEnableRoutes(dbgSrv *sh_debug.Server, op *addon_operator.AddonOperator) {
+	dbgSrv.RoutePOST("/module/{name}/{action:(disable|enable)}", func(r *http.Request) (interface{}, error) {
+		modName := chi.URLParam(r, "name")
+		if modName == "" {
+			return nil, fmt.Errorf("'name' parameter is required")
+		}
+
+		action := chi.URLParam(r, "action")
+		switch action {
+		case "enable":
+			err := deckhouse_config.SetModuleConfigEnabledFlag(op.KubeClient, modName, true)
+			if err != nil {
+				return nil, fmt.Errorf("Enable module failed: %v", err)
+			}
+			return "Module enabled", nil
+		case "disable":
+			err := deckhouse_config.SetModuleConfigEnabledFlag(op.KubeClient, modName, true)
+			if err != nil {
+				return nil, fmt.Errorf("Disable module failed: %v", err)
+			}
+			return "Module disabled", nil
+		}
+		return nil, fmt.Errorf("Unknown action '%s' for module '%s'", action, modName)
+	})
+}
+
+func DefineModuleConfigDebugCommands(kpApp *kingpin.Application) {
+	moduleCmd := kpApp.GetCommand("module")
+
+	var moduleName string
+	moduleEnableCmd := moduleCmd.Command("enable", "Enable module via spec.enabled flag in ModuleConfig resource.").
+		Action(func(c *kingpin.ParseContext) error {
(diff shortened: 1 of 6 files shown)


go_lib/deckhouse-config/kube_helpers.go
```diff
@@ -99,3 +103,54 @@ func GetAllConfigs(kubeClient k8s.Client) ([]*d8cfg_v1alpha1.ModuleConfig, error
...
+			return err
+		}
+		return nil
+	}
+
+	// Create new ModuleConfig if absent.
+	newCfg := &d8cfg_v1alpha1.ModuleConfig{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ModuleConfig",
+			APIVersion: "deckhouse.io/v1alpha1",
```
shvgn, line 136: I believe we shouldn't use hardcoded versions in the middle of the function.

go_lib/deckhouse-config/kube_helpers.go
```diff
@@ -99,3 +103,54 @@ func GetAllConfigs(kubeClient k8s.Client) ([]*d8cfg_v1alpha1.ModuleConfig, error
...
+func SetModuleConfigEnabledFlag(kubeClient k8s.Client, name string, enabled bool) error {
+	// This should not happen, because DebugServer is started after Kubernetes client initialization, but check it anyway.
+	if kubeClient == nil {
+		return fmt.Errorf("kubernetes client not initialized, try again later")
+	}
+
+	gvr := d8cfg_v1alpha1.GroupVersionResource()
+	unstructuredObj, err := kubeClient.Dynamic().Resource(gvr).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil && !k8errors.IsNotFound(err) {
+		return err
```
shvgn, line 117: I suggest wrapping errors with a description: 

```go
return fmt.Errorf("getting module config: %w", err)
return fmt.Errorf("updating module config: %w", err)
return fmt.Errorf("creating module config: %w", err)
```

In this function, with kube client we 1) get, 2) update, or 3) create an object. If we have any non-404 issue, we should just get an error from the API. I'd prefer not to loose track of a specific action causing the problem.

deckhouse-controller/pkg/debug/module_enable.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+			return "Module enabled", nil
+		case "disable":
+			err := deckhouse_config.SetModuleConfigEnabledFlag(op.KubeClient, modName, true)
+			if err != nil {
+				return nil, fmt.Errorf("Disable module failed: %v", err)
+			}
+			return "Module disabled", nil
+		}
+		return nil, fmt.Errorf("Unknown action '%s' for module '%s'", action, modName)
```
shvgn, line 38: 1. "For module" sounds like modules support different sets of actions which is not the case. 
2. I'd start with lower case
3. If path-wise action is not supported in HTTP, the handler should return 404

```go
return nil, fmt.Errorf("switching module %s: action '%s' is not supported", modName, action)
```

deckhouse-controller/pkg/debug/module_enable.go
```diff
@@ -0,0 +1,81 @@
...
+			return nil
+		})
+	moduleDisableCmd.Arg("module_name", "").Required().StringVar(&moduleName)
+	// --debug-unix-socket <file>
+	sh_app.DefineDebugUnixSocketFlag(moduleEnableCmd)
+}
+
+func ModuleEnable(client *sh_debug.Client, modName string) ([]byte, error) {
+	url := fmt.Sprintf("http://unix/module/%s/enable", modName)
+	return client.Post(url, nil)
```
shvgn, line 75: Why are we using unix socket instead of calling the kube client directly?

shvgn (changes requested):

deckhouse-controller/pkg/debug/module_enable.go
diafour: Hm... that's my focus on debug commands. Yeah, there is no need to use DebugServer.

deckhouse-controller/pkg/debug/module_enable.go
diafour: There is no 'unknown action' error after update.

go_lib/deckhouse-config/kube_helpers.go
diafour: Done

shvgn (approved):
LGTM

shvgn (approved):

yalosev (approved):

PR #4295, opened 2023-04-04
[candi] fix in bashible.sh.tpl

duckhawk:
## Description

Fix in bashible template.

In bash v5.2 logic changed and $(<file) directive cause error if file not exists. (looks like it was changed [here](https://github.com/bminor/bash/blob/master/CHANGES#L299))

Working bashible even if file with node checksum ($CONFIGURATION_CHECKSUM_FILE) do not exist.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: Fix in bashible template.
impact_level: low
```

### DIFF
--- a/candi/bashible/bashible.sh.tpl
+++ b/candi/bashible/bashible.sh.tpl
@@ -210,7 +210,7 @@ function main() {
   fi
 
 {{ if eq .runType "Normal" }}
-  if [[ "$(<$CONFIGURATION_CHECKSUM_FILE)" == "$CONFIGURATION_CHECKSUM" ]] 2>/dev/null; then
+  if [[ -f $CONFIGURATION_CHECKSUM_FILE ]] && [[ "$(<$CONFIGURATION_CHECKSUM_FILE)" == "$CONFIGURATION_CHECKSUM" ]] 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


candi/bashible/bashible.sh.tpl
```diff
@@ -210,6 +210,7 @@ function main() {
   fi
 
 {{ if eq .runType "Normal" }}
+  if [[ ! -e $CONFIGURATION_CHECKSUM_FILE ]]; then touch $CONFIGURATION_CHECKSUM_FILE; fi
   if [[ "$(<$CONFIGURATION_CHECKSUM_FILE)" == "$CONFIGURATION_CHECKSUM" ]] 2>/dev/null; then
```
RomanenkoDenys, line 214:
```suggestion
  if [[ -f $CONFIGURATION_CHECKSUM_FILE ]] && [[ "$(<$CONFIGURATION_CHECKSUM_FILE)" == "$CONFIGURATION_CHECKSUM" ]] 2>/dev/null; then
```

RomanenkoDenys (changes requested):
some nits

RomanenkoDenys (approved):

PR #4310, opened 2023-04-05
[node-manager] User draining

yalosev:
## Description
Add annotation for user-initiated node drain

We need to separate 2 draining sources:
- bashible
- user

Bashible drain is fully handled by deckhouse, so we don't need any manual operations here.
User drain is initiated via the annotation and then user have to uncordon node manually

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: feature
summary: Add annotation `update.node.deckhouse.io/draining=user` for starting node drain process
```

### DIFF
--- a/modules/040-node-manager/hooks/handle_draining.go
+++ b/modules/040-node-manager/hooks/handle_draining.go
@@ -38,9 +38,9 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 	Kubernetes: []go_hook.KubernetesConfig{
 		{
 			Name:                         "nodes_for_draining",
-			WaitForSynchronization:       pointer.BoolPtr(false),
-			ExecuteHookOnSynchronization: pointer.BoolPtr(false),
-			ExecuteHookOnEvents:          pointer.BoolPtr(false),
+			WaitForSynchronization:       pointer.Bool(false),
+			ExecuteHookOnSynchronization: pointer.Bool(false),
+			ExecuteHookOnEvents:          pointer.Bool(false),
 			ApiVersion:                   "v1",
 			Kind:                         "Node",
 			LabelSelector: &v1.LabelSelector{
@@ -54,6 +54,9 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 			FilterFunc: drainFilter,
 		},
 	},
+	Settings: &go_hook.HookConfigSettings{
+
+	}
 	Schedule: []go_hook.ScheduleConfig{
 		{
 			Name:    "draining_schedule",
@@ -70,15 +73,24 @@ func drainFilter(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
 		return nil, err
 	}
 
-	var isDraining bool
-	if _, ok := node.Annotations["update.node.deckhouse.io/draining"]; ok {
+	var (
+		isDraining     bool
+		drainingSource string
+	)
+	if source, ok := node.Annotations["update.node.deckhouse.io/draining"]; ok {
 		isDraining = true
+		if source == "" {
+			drainingSource = "bashible"
+		} else {
+			drainingSource = source
+		}
 	}
 
 	return drainingNode{
-		Name:          node.Name,
-		IsDraining:    isDraining,
-		Unschedulable: node.Spec.Unschedulable,
+		Name:           node.Name,
+		IsDraining:     isDraining,
+		DrainingSource: drainingSource,
+		Unschedulable:  node.Spec.Unschedulable,
 	}, nil
 }
 
@@ -127,11 +139,11 @@ func handleDraining(input *go_hook.HookInput, dc dependency.Container) error {
 		}
 
 		wg.Add(1)
-		go func(nodeName string) {
+		go func(nodeName, drainingSource string) {
 			defer wg.Done()
 			err = drain.RunNodeDrain(drainHelper, nodeName)
-			drainingNodesC <- drainedNodeRes{NodeName: nodeName, Err: err}
-		}(dNode.Name)
+			drainingNodesC <- drainedNodeRes{NodeName: nodeName, DrainingSource: drainingSource, Err: err}
+		}(dNode.Name, dNode.DrainingSource)
 	}
 
 	go func() {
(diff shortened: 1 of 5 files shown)


deckhouse-BOaTswain:

name212 (approved):

modules/040-node-manager/hooks/handle_draining.go
```diff
@@ -70,15 +68,33 @@ func drainFilter(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
...
 	}
 
-	var isDraining bool
-	if _, ok := node.Annotations["update.node.deckhouse.io/draining"]; ok {
-		isDraining = true
+	var (
+		drainingSource string
+		drainedSource  string
+	)
+	if source, ok := node.Annotations["update.node.deckhouse.io/draining"]; ok {
```
shvgn, line 75: In hooks (not in tests) this annotation key is repeated 4 times. I suggest extracting it to a constant

modules/040-node-manager/hooks/handle_draining.go
```diff
@@ -70,15 +68,33 @@ func drainFilter(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
...
+		// keep backward compatibility
+		if source == "" {
+			drainingSource = "bashible"
+		} else {
+			drainingSource = source
+		}
+	}
+
+	if source, ok := node.Annotations["update.node.deckhouse.io/drained"]; ok {
```
shvgn, line 84: In hooks (not in tests) this annotation key is repeated 3 times. I suggest extracting it to a constant

modules/040-node-manager/hooks/handle_draining.go
```diff
@@ -107,9 +123,19 @@ func handleDraining(input *go_hook.HookInput, dc dependency.Container) error {
 	snap := input.Snapshots["nodes_for_draining"]
 	for _, s := range snap {
 		dNode := s.(drainingNode)
-		if !dNode.IsDraining {
+		if !dNode.isDraining() {
+			// if node has drained annotation (`user` source) and .spec.unscheduled == false - remove the annotation
```
shvgn, line 127:
```suggestion
			// If the node became schedulable, but 'drained' annotation is still on it, remove the obsolete annotation
```

modules/040-node-manager/hooks/handle_draining.go
```diff
@@ -107,9 +123,19 @@ func handleDraining(input *go_hook.HookInput, dc dependency.Container) error {
...
+			if !dNode.Unschedulable && dNode.DrainedSource == "user" {
+				input.PatchCollector.MergePatch(removeDrainedAnnotation, "v1", "Node", "", dNode.Name)
+			}
 			continue
 		}
+
+		// if node has drained annotation (`user` source) and new draining annotation - remove the 'drained' annotation
```
shvgn, line 134:
```suggestion
		// If the node is marked for draining while is has been drained, remove the 'drained' annotation
```

modules/040-node-manager/hooks/handle_draining.go
```diff
@@ -144,30 +170,50 @@ func handleDraining(input *go_hook.HookInput, dc dependency.Container) error {
...
-		input.PatchCollector.MergePatch(drainAnnotationsPatch, "v1", "Node", "", drainedNode.NodeName)
+		input.PatchCollector.MergePatch(newDrainAnnotationPatch(drainedNode.DrainingSource), "v1", "Node", "", drainedNode.NodeName)
 	}
 
 	return nil
 }
 
-var (
-	drainAnnotationsPatch = map[string]interface{}{
+func newDrainAnnotationPatch(source string) map[string]interface{} {
```
shvgn, line 179:
```suggestion
func newDrainedAnnotationPatch(source string) map[string]interface{} {
```

modules/040-node-manager/hooks/handle_draining.go
```diff
@@ -54,11 +55,8 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 			FilterFunc: drainFilter,
 		},
 	},
-	Schedule: []go_hook.ScheduleConfig{
-		{
-			Name:    "draining_schedule",
-			Crontab: "* * * * *",
-		},
+	Settings: &go_hook.HookConfigSettings{
+		ExecutionMinInterval: 30 * time.Second,
```
shvgn, line 59: What if we switch to labels instead of annotations to have more reactive UX?

modules/040-node-manager/hooks/handle_draining.go
```diff
@@ -107,9 +123,19 @@ func handleDraining(input *go_hook.HookInput, dc dependency.Container) error {
 	snap := input.Snapshots["nodes_for_draining"]
 	for _, s := range snap {
 		dNode := s.(drainingNode)
-		if !dNode.IsDraining {
+		if !dNode.isDraining() {
+			// if node has drained annotation (`user` source) and .spec.unscheduled == false - remove the annotation
+			if !dNode.Unschedulable && dNode.DrainedSource == "user" {
```
shvgn, line 128: What about `bashible` source or any other source? Where are they handled?

shvgn (changes requested):

modules/040-node-manager/hooks/handle_draining.go
```diff
@@ -107,9 +123,19 @@ func handleDraining(input *go_hook.HookInput, dc dependency.Container) error {
 	snap := input.Snapshots["nodes_for_draining"]
 	for _, s := range snap {
 		dNode := s.(drainingNode)
```
shvgn, line 125: It would be a little more obvious if at the loop beginning we had additional `if` to skip irrelevant nodes. (not draining and not drained)

yalosev: We had this if one line below

modules/040-node-manager/hooks/handle_draining.go
yalosev: In the update_approval hook

modules/040-node-manager/hooks/handle_draining.go
yalosev: Well, it would be more reactive but these kinds of labels look weird.
Annotations is a kind of status of the object, some realtime information
while label is an attribute, which describe node in some way

modules/040-node-manager/hooks/handle_draining.go
shvgn: I see the `continue` line (L111). It is just not that obvious.

shvgn (approved):

modules/040-node-manager/hooks/handle_draining.go
yalosev: Done

modules/040-node-manager/hooks/handle_draining.go
yalosev: Done

name212 (approved):

PR #4325, opened 2023-04-06
[node-manager] Doc: mark spec.operatingSystem.manageKernel as deprecated

shvgn:
## Description

Marking `NodeGroup#spec.operatingSystem.manageKernel` as deprecated for clarity.

For clarity

## Changelog entries

```changes
section: node-manger
type: chore
summary: Marked NodeGroup field `spec.operatingSystem.manageKernel` as deprecated.
impact_level: low
```


candi/openapi/doc-ru-node_group.yaml
```diff
@@ -150,7 +150,7 @@ spec:
                   properties:
                     manageKernel:
                       description: |
-                        Автоматическое управление ядром операционной системы.
+                        Этот параметр не используется. Раньше он включл автоматическое управление ядром операционной системы.
```
shvgn, line 153:
```suggestion
                        Этот параметр не используется. Раньше он включaл автоматическое управление ядром операционной системы.
```

PR #4352, opened 2023-04-10
[flant-integration] add deckhouse controllers resource consumption metrics hook

### LINKED ISSUE #4253: [flant-integration] Collect CPU and Memory
### Preflight Checklist

Clusters under our control have all the metrics about all the controllers. However, it is not convenient for Flant engineers to analyze.
1. How many CPU and Memory modules utilize in staging / production environments.
2. The impact on resource consumption after rolling a new release.
3. Distinguish exceptional cases when modules consume more resources than usual.
4. Plan VPA defaults for min and max requests.

### Proposed Solution

Deckhouse clusters, most of the time, carry a running Prometheus instance. The idea is to scrape CPU and Memory metrics in Grafana agent from Prometheus (only average values consumed by all controllers in kube-system and d8- namespaces).

### Additional Information

_No response_

alex123012:
## Description

This PR adds new hook for collecting metrics of resource consumption from deckhouse controllers

Fix #4253 
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: flant-integration
type: feature
summary: add d8 controllers resource consumption metrics
impact_level: default
```


ee/modules/600-flant-integration/images/flant-pricing/hooks/controller_metrics.py
```diff
@@ -0,0 +1,200 @@
...
+
+    controller.cpu = get_cpu_prometheus(
+        controller=controller,
+        mock_data=cpu_mock_data,
+        mock_data_file=MOCK_DATA_FILE,
+    )
+    controller.memory = get_memory_prometheus(
+        controller=controller,
+        mock_data=mem_mock_data,
+        mock_data_file=MOCK_DATA_FILE,
```
shvgn, line 128: If we need mocks for tests, the mocks should live in tests and should not pollute business logic.

I suggest making facade class that abstracts how you get the mertics. It is a single class with single method (get/fetch/whatever). Its instance can be passed as a dependency. The implementation of such an interface with fixture (another class) can be used in tests. It would simplify the code and separate concerns better.

ee/modules/600-flant-integration/images/flant-pricing/hooks/controller_metrics.yaml
```diff
@@ -0,0 +1,68 @@
...
+    queue: /controller_metrics_deploy
+    apiVersion: apps/v1
+    kind: Deployment
+    group: main
+    executeHookOnEvent: []
+    executeHookOnSynchronization: false
+    keepFullObjectsInMemory: false
+    jqFilter: |
+      {
+        "controller_name": .metadata.name,
```
shvgn, line 18: Nitpicking. Do we need to repeat `controller_` prefix? I think we know the context well enough to be shorter.

ee/modules/600-flant-integration/images/flant-pricing/hooks/controller_metrics.py
```diff
@@ -0,0 +1,200 @@
...
+    cpu: Optional[float] = None
+    memory: Optional[float] = None
+
+
+def main(ctx: hook.Context):
+    # Generate list of Controllers from snapshots
+    controllers = process_controllers(snapshots=ctx.snapshots)
+
+    # Generate metrics from Controllers list
+    metrics = generate_metrics(controllers=controllers)
```
shvgn, line 54: Nitpicking. I would not use keyword arguments when a function has very few arguments. The name of variables is clean enough to see what's going on (ctx.snapshots, controllers)

shvgn (changes requested):

ee/modules/600-flant-integration/images/flant-pricing/hooks/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    if labels:
+        metric_labels = [f'{k}="{v}"' for k, v in labels.items()]
+    return f"{metric_name}{{{','.join(metric_labels)}}}"
+
+
+def prometheus_function_builder(func: str, metric: str, interval: str = None) -> str:
+    """build prometheus function from func, metric name and internval"""
+
+    interval_str = ""
+    if interval is not None:
```
shvgn, line 41: If interval is expected to be string, it is better to make it `""` by default. 

```python
def ... ( ... interval: str = "")

if not interval:
    # will work 
```

ee/modules/600-flant-integration/images/flant-pricing/hooks/controller_metrics.py
```diff
@@ -0,0 +1,189 @@
...
+        return prometheus_query_value(
+            query=self.resource_consumption_query.format(
+                prometheus_function_builder(
+                    func="rate",
+                    metric=resource_metric(
+                        metric_name="container_cpu_usage_seconds_total",
+                        controller=controller,
+                    ),
+                    interval=self.interval,
+                ),
```
shvgn, line 86: I think this place is too bloated with wrapping functions. It is hard to get what is going on, what is the resulting expression. I'd make it more explicit.

ee/modules/600-flant-integration/images/flant-pricing/hooks/controller_metrics_test.py
```diff
@@ -0,0 +1,174 @@
...
+# See https://github.com/deckhouse/deckhouse/blob/main/ee/LICENSE
+#
+
+from controller_metrics import HookRunner, AbstractMetricCollector, Controller
+from shell_operator import hook
+
+
+class MockMetricCollector(AbstractMetricCollector):
+    __cpu_values = {
+        ("dex", "user-authn", "Deployment", "d8-user-authn"): 1.0,
```
shvgn, line 14: Let's add a comment on this structure, IIRC:

```
# controller, module, kind, namespace
```

ee/modules/600-flant-integration/images/flant-pricing/hooks/controller_metrics.py
shvgn: E.g. we always use interval. Why it is optional in `prometheus_function_builder`? I'd write its content right here instead of extracting a function.

Also, `resource_metric` adds almost nothing as for me.

In `query.format(..., ...)`, we have two arguments, but it is a little too noisy to read in 8 lines.

I don't think it is a blocker for merge, but I'd make this place a more clean.

ee/modules/600-flant-integration/images/flant-pricing/hooks/controller_metrics.py
```diff
@@ -0,0 +1,189 @@
...
+
+    metric_group = "group_d8_controller_metrics"
+    cpu_metric_name = "flant_pricing_controller_average_cpu_usage_seconds"
+    memory_metric_name = "flant_pricing_controller_average_memory_working_set_bytes:without_kmem"
+    interval = "5m"
+
+    def collect(self, ctx: hook.Context, controllers: List[Controller]):
+        '''Export metrics to hook context from Controllers list'''
+
+        ctx.metrics.expire(self.metric_group)
```
shvgn, line 39: Labels `"labels": controller_metric_labels(ctrl)` are the same. I'd write them here once, and would not extract a function. We can reuse variable instead of a function, so all the metric fields would be in one place.

ee/modules/600-flant-integration/images/flant-pricing/hooks/controller_metrics_test.py
```diff
@@ -0,0 +1,174 @@
...
+        return self.__memory_values[(controller.name, controller.module,
+                                     controller.kind, controller.namespace)]
+
+
+def test_controller_metrics():
+    hook_runner = HookRunner(MockMetricCollector())
+    out = hook.testrun(hook_runner.run, binding_context)
+
+    assert out.metrics.data == expected_metrics
+    assert not out.kube_operations.data
```
shvgn, line 41: Looks good!

ee/modules/600-flant-integration/images/flant-pricing/hooks/utils.py
```diff
@@ -0,0 +1,97 @@
...
+def prometheus_query_value(query: str) -> float:
+    """query prometheus from query and get only value result"""
+
+    query_result = prometheus_query(query)
+    if len(query_result) > 0:
+        return float(query_result[0].get("value", [0, 0])[1])
+    return 0.0
+
+
+def make_get_request(
```
shvgn, line 73: You are not 'making' it, you are 'doing' it :)

shvgn (dismissed):
There a space for improvements. Probably not blocking ones.

shvgn (approved):

name212 (approved):

PR #4408, opened 2023-04-17
fix ClusterLogDestination loki endpoint

pavelmironchenko-ws:
## Description
ClusterLogDestination CR uses d8-loki namespace instead of d8-monitoring

At the moment the current configuration with the specified service in d8-loki namespace does not work

This fix may be in the patch release

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: loki
type: fix
summary: Fix ClusterLogDestination loki endpoint
impact_level: default
```

### DIFF
--- a/modules/462-loki/templates/logshipper-log-destination.yaml
+++ b/modules/462-loki/templates/logshipper-log-destination.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   type: Loki
   loki:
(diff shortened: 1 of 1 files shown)


nabokihms:
@pavelmironchenko-ws thank you for your contribution!

nabokihms (approved):

z9r5:
Cherry-picked into release-1.45

PR #4582, opened 2023-05-05
[node-manager] Fix Ready condition for node groups

name212:
## Description
Set `Ready` condition to `True` If node group has not nodes. 

Cannot deploy node group with werf.

Cannot deploy node group with werf.
## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Fix the Ready condition for node groups.
impact_level: default
```

### DIFF
--- a/modules/040-node-manager/hooks/internal/conditions/calculate.go
+++ b/modules/040-node-manager/hooks/internal/conditions/calculate.go
@@ -191,9 +191,7 @@ func CalculateNodeGroupConditions(ng NodeGroup, nodes []*Node, currentConditions
 		curTime, _ = time.Parse(time.RFC3339, timeStr)
 	}
 
-	hasNodes := false
 	for _, node := range nodes {
-		hasNodes = true
 		if !node.Unschedulable {
 			schedulableNodes++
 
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:

modules/040-node-manager/hooks/internal/conditions/calculate.go
```diff
@@ -214,7 +216,11 @@ func CalculateNodeGroupConditions(ng NodeGroup, nodes []*Node, currentConditions
 		}
 	}
 
-	isReady := false
+	isReady := true
+	if hasNodes {
```
yalosev, line 220:
```suggestion
	if len(nodes) > 0 {
```

modules/040-node-manager/hooks/internal/conditions/calculate.go
```diff
@@ -214,7 +216,11 @@ func CalculateNodeGroupConditions(ng NodeGroup, nodes []*Node, currentConditions
 		}
 	}
 
-	isReady := false
+	isReady := true
```
yalosev, line 219: or `isReady := len(nodes) == 0`

name212: yep. tnx

modules/040-node-manager/hooks/internal/conditions/calculate.go
name212: fixed

yalosev (approved):

PR #4651, opened 2023-05-16
[admission-policy-engine] update grafana dashboard definition

### LINKED ISSUE #4415: [admission-policy-engine] deduplicate OPA constrain violations in the Grafana Dashboard
### Preflight Checklist

Currently Grafana Dashboard (Security/Admission Policy Engine) has the table `OPA violations` with a stream of events. They have timestamp for each occurred event. Users have to deduplicate these events somehow to fix them one-by-one.
I don't think that we need a timestamp for each event. We can group it and show deduplicated list of violations for fixing.

image

### Proposed Solution

- Fix [this query](https://github.com/deckhouse/deckhouse/blob/main/modules/015-admission-policy-engine/monitoring/grafana-dashboards/security/admission-policy-engine.json#L445) with removing time from it
- Change grafana table: remove the `time` column and apply instant value
- Check [another query](https://github.com/deckhouse/deckhouse/blob/main/modules/015-admission-policy-engine/monitoring/grafana-dashboards/security/admission-policy-engine.json#L435) in this table. Seems it doesn't work. I don't remember what for it was made. Maybe we can just remove it

As a result we want to see a table with just active deduplicated events, like that:
image

### Additional Information

_No response_

miklezzzz:
## Description
Admission-policy-engine security dashboard definition in Grafana has been updated:
* Previously disabled and incorrect query has been removed;
* The query type has been changed to instant (from vector) and grouping by time has been removed.

OperationPolicy CRD has been updated:
* spec.policies.requiredLabels.watchKinds pattern regex expression has been fixed (wasn't working as it was intended)

Previously, OPA Violations table showed multiple entries for the same violations with different timestamps (range vector), rendering the whole table pretty cumbersome. After the aforementioned changes the table includes only unique records for the violating objects for the selected period of time.
Close #4415 

OPA Violations table of Admission-policy-engine Dashboard in Grafana shows only unique entries.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: chore
summary: Admission-policy-engine security dashboard in Grafana has been updated so that OPA Violations table doesn't show multiple entries for the same violations.
impact_level: default
```

### DIFF
--- a/modules/015-admission-policy-engine/crds/operation-policy.yaml
+++ b/modules/015-admission-policy-engine/crds/operation-policy.yaml
@@ -109,7 +109,7 @@ spec:
                           minItems: 1
                           items:
                             type: string
-                            pattern: '^[a-z]?/[a-zA-Z]+$'
+                            pattern: '^[a-z]*/[a-zA-Z]+$'
                             example: ["apps/Deployment", "/Pod", "networking.k8s.io/Ingress"]
                     requiredProbes:
                       type: array
--- a/modules/015-admission-policy-engine/monitoring/grafana-dashboards/security/admission-policy-engine.json
+++ b/modules/015-admission-policy-engine/monitoring/grafana-dashboards/security/admission-policy-engine.json
@@ -298,7 +298,9 @@
       "type": "row"
     },
     {
-      "datasource": "$ds_prometheus",
+      "datasource": {
+        "uid": "$ds_prometheus"
+      },
       "fieldConfig": {
         "defaults": {
           "color": {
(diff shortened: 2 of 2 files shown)


miklezzzz:
@yalosev now it looks like this 
image

does it make sense to change to something like this?
image
because it isn't clear in which namespace the violating pod(object?) resides.

yalosev:
@miklezzzz I guess you can add only `namespace` column. Like this:
Namespace | Pod Name

It's dashboard with `Pod Security Standards`. It couldn't contain any objects except pods

miklezzzz:
this is what it looks like now
image
@yalosev

yalosev (approved):

PR #4771, opened 2023-05-25
[candi] Old binary deletion fixes

zuzzas:
## Description

Fixed a migration that did not include /opt/deckhouse/bin creation. Also save PATH to a previous jq location to avoid breaking the bashible.sh update process.

bashible step failed mid-process due to a previous issue that was addressed in v1.46.1. Here we protect from losing path to a jq binary even more.

More proofing against bashible exploding due to missing jq binary.

jq is always available to bashible.sh main script.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: Fixed a migration that did not include /opt/deckhouse/bin creation. Also save PATH to a previous jq location to avoid breaking the bashible.sh update process.
impact_level: low
```

### DIFF
--- a/candi/bashible/bashible.sh.tpl
+++ b/candi/bashible/bashible.sh.tpl
@@ -142,7 +142,7 @@ function get_bundle() {
 }
 
 function main() {
-  export PATH="/opt/deckhouse/bin:$PATH"
+  export PATH="/opt/deckhouse/bin:/usr/local/bin:$PATH"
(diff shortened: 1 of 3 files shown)


deckhouse-BOaTswain:
Cherry pick PR [4780](https://github.com/deckhouse/deckhouse/pull/4780) to the branch [release-1.46](https://github.com/deckhouse/deckhouse/tree/release-1.46) successful!

PR #4600, opened 2023-05-09
[linstor] Update LINSTOR and CSI components

kvaps:
## Description

- Update LINSTOR to 1.22.1
- Update all CSI images to latest versions
- Update DRBD 9.2.3

Includes https://github.com/deckhouse/deckhouse/pull/4639

## Why do we need it, and what problem does it solve?

bug fixes

## What is the expected result?

Updated LISNTOR and CSI images

## Checklist
- [ ] The code is covered by unit tests.
- [x] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: linstor
type: chore
summary: update LINSTOR to 1.22.1
impact_level: default
---
section: linstor
type: chore
summary: update all CSI images to latest versions
impact_level: default
---
section: linstor
type: chore
summary: Update DRBD 9.2.3
impact_level: default
---
section: ceph-csi
type: chore
summary: update all CSI images to latest versions
impact_level: default
---
section: cloud-provider-aws
type: chore
summary: update all CSI images to latest versions
impact_level: default
---
section: cloud-provider-azure
type: chore
summary: update all CSI images to latest versions
impact_level: default
---
section: cloud-provider-gcp
type: chore
summary: update all CSI images to latest versions
impact_level: default
---
section: cloud-provider-yandex
type: chore
summary: update all CSI images to latest versions
impact_level: default
---
section: cloud-provider-openstack
type: chore
summary: update all CSI images to latest versions
impact_level: default
---
section: cloud-provider-vsphere
type: chore
summary: update all CSI images to latest versions
impact_level: default
```

PR #4788, opened 2023-05-25
[linstor] Improve DRBD alerts

kvaps:
## Description

- Fix spaas alert grouping
- Add DRBD version alert
- Update `hack/update.sh` helper script to keep alert version up-to-date

- Fix stucking group in monitoring
- Alert about DRBD version which requests node reboot

We just updated DRBD version, it's nice to get informed about that https://github.com/deckhouse/deckhouse/pull/4600

- Fix SPAAS alert
- New alert about DRBD version

изображение

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: linstor
type: chore
summary: Improve DRBD alerts
impact: |
  Users might receive new alert about DRBD version change.
  It's not important to perform update immediately thus the severity is low, but it's better to put it in plan because new drbd versions often include many bug fixes
impact_level: default
```

### DIFF
--- a/modules/041-linstor/monitoring/prometheus-rules/drbd-version.yaml
+++ b/modules/041-linstor/monitoring/prometheus-rules/drbd-version.yaml
@@ -1,7 +1,7 @@
 - name: kubernetes.drbd.device_state
   rules:
     - alert: D8DRBDVersionIsOutdated
-      expr: drbd_version{kmod!="9.2.3"}
+      expr: drbd_version{kmod!="__DRBD_VERSION__"}
       for: 5m
       labels:
         severity_level: "9"
@@ -16,7 +16,7 @@
           LINSTOR node {{ $labels.node }} has outdated DRBD version
 
           in use: {{ $labels.kmod }}
-          expected: 9.2.3
+          expected: __DRBD_VERSION__
 
           The recommended course of action:
           - reboot the node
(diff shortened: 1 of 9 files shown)


RomanenkoDenys (approved):

modules/041-linstor/monitoring/prometheus-rules/drbd-version.yaml
```diff
@@ -0,0 +1,22 @@
+- name: kubernetes.drbd.device_state
+  rules:
+    - alert: D8DRBDVersionIsOutdated
+      expr: drbd_version{kmod!="9.2.3"}
```
apolovov, line 4: it is better to store actual drbd version in values. Describe a default value for argument in openapi/values.yaml virtualization.drbdVersion. The sed in update.sh will be much more reliable.

apolovov (changes requested):

modules/041-linstor/monitoring/prometheus-rules/drbd-version.yaml
kvaps: I think openapi spec is not a best place for storing default value because:

1) it will affect whole module values, which is not good for patch release
2) it's difficult to update with sed

I added additional helper instead
https://github.com/deckhouse/deckhouse/blob/017985f137220eecbcc4927e0feecc5a3984182e/modules/041-linstor/templates/_drbd_version.tpl#L1

Please take a look. If you don't agree I modify openapi/values of course

apolovov: It is much better, but now we reinvented templates.  We use default values for storing istio version, so it is the best of the worst options.

kvaps: Ok, done

AleksZimin (approved):

apolovov (approved):

apolovov:
Closed in favor of #4882

PR #4843, opened 2023-06-02
Set discard_unpacked_layers to true

asviel:
## Description

This change allows to not keep compressed image layers in containerd's content store once they have been unpacked.

I also suggest to clean containerd's content store in order to delete those layers which were pulled before this change (this change does not lead to deletion of previously pulled image layers)

Containerd, unlike Docker, stores pulled images twice: as compressed layers (as in registry) and as unpacked snaphots, which are used to run containers. This leads to increased disk space consumption (up to 35% percent). Read more here https://github.com/containerd/containerd/blob/main/docs/content-flow.md

We will save up to 35% of disk space.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: feature
summary: Not keep compressed image layers in containerd's content store once they have been unpacked.
impact: 'All `containerd` daemons will restart.'
impact_level: default
```


candi/bashible/common-steps/node-group/050_configure_and_start_containerd.sh.tpl
```diff
@@ -194,6 +194,25 @@ fi
...
+
+  ctr -n k8s.io content prune references
+  if [ $? -eq 0 ]; then
+    touch /var/lib/bashible/containerd_content_store_is_cleared
+  else
+    bb-log-warning "Failed to clean up containerd's content store. Skip this for now"
+  fi
+fi
+
+
```
RomanenkoDenys, line 215: I think this should be NodeGroupConfiguration step, not bashible

RomanenkoDenys (changes requested):
some nits

candi/bashible/common-steps/node-group/050_configure_and_start_containerd.sh.tpl
asviel: Moved to the node-manager module

modules/040-node-manager/templates/nodegroupconfiguration-prune-containerd-content.yaml
```diff
@@ -0,0 +1,45 @@
...
+    bb-log-info "Removing compressed image layers from containerd's content store"
+
+    ctr -n k8s.io content prune references
+    if [ $? -eq 0 ]; then
+        touch /var/lib/bashible/containerd_content_store_is_cleared
+    else
+        bb-log-warning "Failed to clean up containerd's content store. Skip this for now"
+    fi
+    fi
+    {{- end }}
```
RomanenkoDenys, line 45:
```suggestion
    {{- if eq .cri "Containerd" }}
    if [ -f /var/lib/bashible/containerd_content_store_is_cleared ] ; then
      exit 0
    fi
    bb-log-info "Removing compressed image layers from containerd's content store"
    if ctr -n k8s.io content prune references; then 
        touch /var/lib/bashible/containerd_content_store_is_cleared
    else
        bb-log-warning "Failed to clean up containerd's content store. Skip this for now"
    fi
    {{- end }}
```

RomanenkoDenys (approved):
LGTM

deckhouse-BOaTswain:

PR #4927, opened 2023-06-13
[nginx-ingress] set default ingress class to nginx

### LINKED ISSUE #4799: [ingress-nginx] Make "nginx" ingress class the default
### Preflight Checklist

Not all users know what is ingress class and why they should care. Also, users tend to create single controller. But everybody need to write something in the corresponding field in ingress controller manifest.  

Here where all ingress classes are caught and created
https://github.com/deckhouse/deckhouse/blob/main/modules/402-ingress-nginx/templates/controller/ingress-class.yaml#L10-L12 

### Proposed Solution

Make `spec.ingressClass` "nginx" by default

### Additional Information

_No response_

miklezzzz:
## Description
Having required to specify the .spec.ingressClass when defining an ingress nginx controller might seem daunting for some users. We can provide default value "nginx" if the field is left blank.

It makes it a bit easier to define an ingress nginx controller.
Close #4799 

In case the .ingressClass isn't provided on create/update, we set it to 'nginx' via crd default field.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: feature
summary: Set ingressClass to 'nginx' if not explicitly set.
impact_level: default
```

### DIFF
--- a/modules/402-ingress-nginx/crds/ingress-nginx.yaml
+++ b/modules/402-ingress-nginx/crds/ingress-nginx.yaml
@@ -36,7 +36,7 @@ spec:
                     **Caution!** If you set it to "nginx", then Ingress resources lacking the `kubernetes.io/ingress.class` annotation or `spec.ingressClassName` field will also be handled.
                   example: 'nginx'
                   default: 'nginx'
-                  pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
+                  pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
                 inlet:
                   type: string
                   description: |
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:

miklezzzz:
image

modules/402-ingress-nginx/crds/ingress-nginx.yaml
```diff
@@ -35,6 +35,7 @@ spec:
 
                     **Caution!** If you set it to "nginx", then Ingress resources lacking the `kubernetes.io/ingress.class` annotation or `spec.ingressClassName` field will also be handled.
                   example: 'nginx'
+                  default: 'nginx'
                   pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
```
yalosev, line 39: I guess we also have to change pattern here, according to https://github.com/deckhouse/deckhouse/pull/4932/

yalosev (changes requested):
pattern change required

modules/402-ingress-nginx/crds/ingress-nginx.yaml
miklezzzz: done, also updated secret name pattern with the same value

yalosev (approved):

nabokihms (approved):

PR #4955, opened 2023-06-16
[loki] Added kube-rbac-proxy sidecar in loki statefullset

### LINKED ISSUE #4315: [loki] Protect with kube-rbac-proxy
### Preflight Checklist

Anyone can write and read logs now.

### Proposed Solution

Protect Loki with kube-rbac-proxy like Prometheus and other software.

### Additional Information

_No response_

NaNameUz3r:
## Description

Added kube-rbac-proxy sidecar container in loki statefulset, to protect it from unauthorized access.

Fix #4315

I guess it need to be back-ported into the patch release, because it is security related staff  ¯\_(ツ)_/¯

Loki 3100 port protected with kube-rbac-proxy.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: loki
type: fix
summary: "kube-rbac-proxy sidecar has beed added to the loki statefulset"
impact_level: default
```

### DIFF
--- a/modules/462-loki/templates/configmap.yaml
+++ b/modules/462-loki/templates/configmap.yaml
@@ -10,7 +10,7 @@ data:
     auth_enabled: false
 
     server:
-      http_listen_port: 3100
+      http_listen_port: 3099
 
     limits_config:
       retention_period: {{ .Values.loki.retentionPeriodHours | default 168 }}h
--- a/modules/462-loki/templates/statefulset.yaml
+++ b/modules/462-loki/templates/statefulset.yaml
@@ -59,7 +59,7 @@ spec:
           - name: KUBE_RBAC_PROXY_CONFIG
             value: |
               upstreams:
-              - upstream: http://127.0.0.1:3100/
+              - upstream: http://127.0.0.1:3099/
                 path: /
                 authorization:
                   resourceAttributes:
@@ -90,7 +90,7 @@ spec:
             mountPath: /loki
         ports:
           - name: http-metrics
-            containerPort: 3100
+            containerPort: 3099
             protocol: TCP
         livenessProbe:
           httpGet:
--- a/modules/462-loki/templates/statefulset.yaml
+++ b/modules/462-loki/templates/statefulset.yaml
@@ -7,7 +7,10 @@ metadata:
   namespace: d8-monitoring
   {{- include "helm_lib_module_labels" (list . (dict "app" .Chart.Name)) | nindent 2 }}
 spec:
-  {{- include "helm_lib_resources_management_vpa_spec"  (list "apps/v1" "StatefulSet" .Chart.Name "loki" .Values.loki.resourcesManagement ) | nindent 2}}
+  {{- include "helm_lib_resources_management_vpa_spec"  (list "apps/v1" "StatefulSet" .Chart.Name "loki" .Values.loki.resourcesManagement ) | nindent 2 }}
+  {{- if .Values.loki.resourcesManagement.vpa }}
+    {{- include "helm_lib_vpa_kube_rbac_proxy_resources" . | nindent 4 }}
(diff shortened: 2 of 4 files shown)


modules/462-loki/templates/configmap.yaml
```diff
@@ -10,7 +10,7 @@ data:
     auth_enabled: false
 
     server:
-      http_listen_port: 3100
+      http_listen_port: 3099
```
nabokihms, line 13: `http_listen_address: 127.0.0.1` is also required

modules/462-loki/templates/statefulset.yaml
```diff
@@ -49,7 +90,7 @@ spec:
             mountPath: /loki
         ports:
           - name: http-metrics
-            containerPort: 3100
```
nabokihms, line 52: This port should not be accessible because the listen address should be 127.0.0.1.
 All communications with the loki should be provided only through kube-rbac-proxy.

modules/462-loki/templates/statefulset.yaml
```diff
@@ -37,6 +40,44 @@ spec:
...
+                    apiVersion: v1
+                    resource: statefulsets
+                    subresource: logs
+                    name: loki
+        resources:
+          requests:
+        {{- include "helm_lib_module_ephemeral_storage_only_logs" . | nindent 12 }}
+        {{- if not ( .Values.global.enabledModules | has "vertical-pod-autoscaler-crd") }}
+        {{- include "helm_lib_container_kube_rbac_proxy_resources" . | nindent 12 }}
+        volumeMounts:
```
nabokihms, line 77: This volume mount is under the `{{- if not ( .Values.global.enabledModules | has "vertical-pod-autoscaler-crd") }}`. It doesn't seem right.

modules/462-loki/templates/statefulset.yaml
```diff
@@ -37,6 +40,44 @@ spec:
...
+              - upstream: http://127.0.0.1:3099/
+                path: /
+                authorization:
+                  resourceAttributes:
+                    namespace: d8-monitoring
+                    apiGroup: apps
+                    apiVersion: v1
+                    resource: statefulsets
+                    subresource: logs
+                    name: loki
```
nabokihms, line 71: You also need access rights to Prometheus to scrape this endpoint, like here:
https://github.com/deckhouse/deckhouse/blob/main/modules/200-operator-prometheus/templates/rbac-to-us.yaml

Besides, it is also required to give access to grafana and log-shipper to loki:
1. A token should be added here https://github.com/deckhouse/deckhouse/blob/c1ff4bf5f1b241af666f1dfe2d27947e4e694f93/modules/462-loki/templates/logshipper-log-destination.yaml#L7-L10
2. And here https://github.com/deckhouse/deckhouse/blob/c1ff4bf5f1b241af666f1dfe2d27947e4e694f93/modules/462-loki/templates/grafana-datasource.yaml#L8-L15

nabokihms (changes requested):

modules/462-loki/templates/statefulset.yaml
```diff
@@ -37,6 +40,44 @@ spec:
...
+        {{- include "helm_lib_module_container_security_context_read_only_root_filesystem" . | nindent 8 }}
+        image: {{ include "helm_lib_module_common_image" (list . "kubeRbacProxy") }}
+        args:
+          - "--secure-listen-address=$(KUBE_RBAC_PROXY_LISTEN_ADDRESS):3100"
+          - "--v=2"
+          - "--logtostderr=true"
+          - "--stale-cache-interval=1h30m"
+        ports:
+          - containerPort: 3100
+            name: loki
```
nabokihms, line 53:
```suggestion
            name: https-metrics

```

PR #4968, opened 2023-06-19
[prometheus] add external labels knob

### LINKED ISSUE #4673: [prometheus] Add externalLabels option
### Preflight Checklist

For multicluster setups, it is required to distinguish which cluster sends which metrics and alerts.

### Proposed Solution

Add the `externalLabels` to the module config.

### Additional Information

Provide an example for HA setups with the pods name as an extrnalLabel using environment variables.
```
externalLabels:
  prometheus_replica: $(POD_NAME)
```

miklezzzz:
## Description
Adds the way to set custom external labels for the main Prometheus. According to the description: "The labels to add to any time series or alerts when communicating with external systems (federation, remote storage, Alertmanager)". The value of `prometheus` label is set to `deckhouse` by default and can't be changed. It's possible to set label values through resolving the environment variables of the `config-relloader` container( some meaningful examples: `SHARD`, `POD_NAME`, `HOSTNAME`). This env substitution functionality is provided by the `prometheus-config-reloader` binary, so no restart is required.

It allows a user to set custom external labels depending on their needs. 
Close #4673

It is possible to set external labels through the Prometheus module configuration.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: feature
summary: Add module external labels setting.
impact_level: default
```

### DIFF
--- a/modules/300-prometheus/openapi/config-values.yaml
+++ b/modules/300-prometheus/openapi/config-values.yaml
@@ -121,7 +121,10 @@ properties:
     description: |
       The set of external labels to add to the metrics.
       
-      It is possible to expand environment variables (for example: `HOSTNAME`, `POD_NAME`, `SHARD`) in external labels.
+      It's possible to expand the environment variables of the `config-reloader` container in external labels
+      such as:
+      * `HOSTNAME`/`POD_NAME` - contains the name of the pod (for example `prometheus-main-0`, `prometheus-main-1`, etc.).
+      * `SHARD` - contains the [shard](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/shards-and-replicas.md) number.
     additionalProperties:
       type: string
   https:
--- a/modules/300-prometheus/openapi/config-values.yaml
+++ b/modules/300-prometheus/openapi/config-values.yaml
@@ -115,16 +115,15 @@ properties:
   externalLabels:
     type: object
     x-examples:
-      - prometheus: dev
-        prometheus_replica: ${HOSTNAME}
-        env: $MY_ENV
+      - prometheus_replica: $(POD_NAME)
+        shard: $(SHARD)
+        hostname: $(HOSTNAME)
     description: |
       The set of external labels to add to the metrics.
       
-      It is possible to expand [environment variables](https://prometheus.io/docs/prometheus/latest/feature_flags/#expand-environment-variables-in-external-labels) in external labels.
+      It is possible to expand environment variables (for example: `HOSTNAME`, `POD_NAME`, `SHARD`) in external labels.
     additionalProperties:
       type: string
-      maxLength: 63
   https:
     type: object
     x-examples:
--- a/modules/300-prometheus/openapi/config-values.yaml
+++ b/modules/300-prometheus/openapi/config-values.yaml
@@ -112,6 +112,19 @@ properties:
       The class of the Ingress controller used for Grafana/Prometheus.
 
       An optional parameter. By default, the `modules.ingressClass` global value is used.
+  externalLabels:
+    type: object
+    x-examples:
+      - prometheus: dev
+        prometheus_replica: ${HOSTNAME}
+        env: $MY_ENV
+    description: |
+      The set of external labels to add to the metrics.
+      
+      It is possible to expand [environment variables](https://prometheus.io/docs/prometheus/latest/feature_flags/#expand-environment-variables-in-external-labels) in external labels.
+    additionalProperties:
+      type: string
+      maxLength: 63
   https:
     type: object
     x-examples:
--- a/modules/300-prometheus/templates/prometheus/prometheus.yaml
+++ b/modules/300-prometheus/templates/prometheus/prometheus.yaml
@@ -55,9 +55,6 @@ spec:
   enableRemoteWriteReceiver: true
   enableFeatures:
   - memory-snapshot-on-shutdown # https://ganeshvernekar.com/blog/prometheus-tsdb-snapshot-on-shutdown/
-{{- if .Values.prometheus.externalLabels }}
-  - expand-external-labels # https://prometheus.io/docs/prometheus/latest/feature_flags/#expand-environment-variables-in-external-labels
-{{- end }}
   imagePullSecrets:
   - name: deckhouse-registry
   listenLocal: true
@@ -137,10 +134,9 @@ spec:
   scrapeInterval: {{ .Values.prometheus.scrapeInterval | default "30s"}}
   evaluationInterval: {{ .Values.prometheus.scrapeInterval | default "30s" }}
   externalLabels:
+    prometheus: deckhouse
 {{- with .Values.prometheus.externalLabels }}
     {{- . | toYaml | nindent 4 }}
-{{- else }}
-    prometheus: deckhouse
 {{- end }}
   prometheusExternalLabelName: ""
   replicaExternalLabelName: ""
--- a/modules/300-prometheus/templates/prometheus/prometheus.yaml
+++ b/modules/300-prometheus/templates/prometheus/prometheus.yaml
@@ -134,10 +134,10 @@ spec:
   scrapeInterval: {{ .Values.prometheus.scrapeInterval | default "30s"}}
   evaluationInterval: {{ .Values.prometheus.scrapeInterval | default "30s" }}
   externalLabels:
-    prometheus: deckhouse
 {{- with .Values.prometheus.externalLabels }}
     {{- . | toYaml | nindent 4 }}
 {{- end }}
+    prometheus: deckhouse
   prometheusExternalLabelName: ""
   replicaExternalLabelName: ""
   serviceAccountName: prometheus
--- a/modules/300-prometheus/templates/prometheus/prometheus.yaml
+++ b/modules/300-prometheus/templates/prometheus/prometheus.yaml
@@ -55,6 +55,9 @@ spec:
   enableRemoteWriteReceiver: true
   enableFeatures:
   - memory-snapshot-on-shutdown # https://ganeshvernekar.com/blog/prometheus-tsdb-snapshot-on-shutdown/
+{{- if .Values.prometheus.externalLabels }}
+  - expand-external-labels # https://prometheus.io/docs/prometheus/latest/feature_flags/#expand-environment-variables-in-external-labels
+{{- end }}
   imagePullSecrets:
   - name: deckhouse-registry
   listenLocal: true
@@ -134,7 +137,11 @@ spec:
   scrapeInterval: {{ .Values.prometheus.scrapeInterval | default "30s"}}
   evaluationInterval: {{ .Values.prometheus.scrapeInterval | default "30s" }}
   externalLabels:
+{{- if .Values.prometheus.externalLabels }}
+    {{- .Values.prometheus.externalLabels | toYaml | nindent 4 }}
+{{- else }}
     prometheus: deckhouse
+{{- end }}
   prometheusExternalLabelName: ""
   replicaExternalLabelName: ""
   serviceAccountName: prometheus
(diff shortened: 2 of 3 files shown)


modules/300-prometheus/templates/prometheus/prometheus.yaml
```diff
@@ -55,6 +55,9 @@ spec:
   enableRemoteWriteReceiver: true
   enableFeatures:
   - memory-snapshot-on-shutdown # https://ganeshvernekar.com/blog/prometheus-tsdb-snapshot-on-shutdown/
+{{- if .Values.prometheus.externalLabels }}
+  - expand-external-labels # https://prometheus.io/docs/prometheus/latest/feature_flags/#expand-environment-variables-in-external-labels
+{{- end }}
```
nabokihms, line 60: 1. This should not be required because the Prometheus pod has a config reloader sidecar that does the job. Probably, the flag is not required.
2. If a flag is still required, it should always be enabled. Otherwise, if an external label is added to the module config, Prometheus will be restarted.

modules/300-prometheus/templates/prometheus/prometheus.yaml
```diff
@@ -134,7 +137,11 @@ spec:
   scrapeInterval: {{ .Values.prometheus.scrapeInterval | default "30s"}}
   evaluationInterval: {{ .Values.prometheus.scrapeInterval | default "30s" }}
   externalLabels:
+{{- if .Values.prometheus.externalLabels }}
+    {{- .Values.prometheus.externalLabels | toYaml | nindent 4 }}
+{{- else }}
```
nabokihms, line 142: NIT
```suggestion
{{- with .Values.prometheus.externalLabels }}
    {{- . | toYaml | nindent 4 }}
{{- else }}
```

modules/300-prometheus/templates/prometheus/prometheus.yaml
```diff
@@ -134,7 +137,11 @@ spec:
...
     prometheus: deckhouse
```
nabokihms, line 143: This label should always be present. Let's move it out of the condition. The only question is whether users want to override the value or not. 
🤔

modules/300-prometheus/openapi/config-values.yaml
```diff
@@ -112,6 +112,19 @@ properties:
       The class of the Ingress controller used for Grafana/Prometheus.
 
       An optional parameter. By default, the `modules.ingressClass` global value is used.
+  externalLabels:
+    type: object
+    x-examples:
+      - prometheus: dev
+        prometheus_replica: ${HOSTNAME}
+        env: $MY_ENV
```
nabokihms, line 120: This does not seem right.

Users cannot specify any env variables, only a limited subset of envs of the reloader container.

The example os envs from my cluster:
```
KUBERNETES_PORT=tcp://10.222.0.1:443
KUBERNETES_SERVICE_PORT=443
GRAFANA_DEX_AUTHENTICATOR_PORT_443_TCP_ADDR=10.222.133.254
CONTROL_PLANE_PROXY_SERVICE_HOST=10.222.191.230
CONTROL_PLANE_PROXY_PORT_8080_TCP_ADDR=10.222.191.230
HOSTNAME=prometheus-main-0
GRAFANA_DEX_AUTHENTICATOR_PORT_443_TCP_PORT=443
PROMETHEUS_METRICS_ADAPTER_PORT_443_TCP=tcp://10.222.119.79:443
GRAFANA_DEX_AUTHENTICATOR_PORT_443_TCP_PROTO=tcp
SHLVL=1
PROMETHEUS_METRICS_ADAPTER_SERVICE_HOST=10.222.119.79
CONTROL_PLANE_PROXY_PORT_8080_TCP_PORT=8080
TRICKSTER_PORT=tcp://10.222.85.133:443
TRICKSTER_SERVICE_PORT=443
HOME=/
CONTROL_PLANE_PROXY_SERVICE_PORT_HTTPS_METRICS=8080
CONTROL_PLANE_PROXY_PORT_8080_TCP_PROTO=tcp
CONTROL_PLANE_PROXY_SERVICE_PORT=8080
CONTROL_PLANE_PROXY_PORT=tcp://10.222.191.230:8080
GRAFANA_DEX_AUTHENTICATOR_SERVICE_PORT_HTTPS=443
PROMETHEUS_METRICS_ADAPTER_SERVICE_PORT=443
GRAFANA_DEX_AUTHENTICATOR_PORT_443_TCP=tcp://10.222.133.254:443
PROMETHEUS_METRICS_ADAPTER_PORT=tcp://10.222.119.79:443
CONTROL_PLANE_PROXY_PORT_8080_TCP=tcp://10.222.191.230:8080
GRAFANA_DEX_AUTHENTICATOR_SERVICE_HOST=10.222.133.254
TERM=xterm
POD_NAME=prometheus-main-0
KUBERNETES_PORT_443_TCP_ADDR=10.222.0.1
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
SHARD=0
GRAFANA_DEX_AUTHENTICATOR_PORT=tcp://10.222.133.254:443
KUBERNETES_PORT_443_TCP_PORT=443
GRAFANA_DEX_AUTHENTICATOR_SERVICE_PORT=443
KUBERNETES_PORT_443_TCP_PROTO=tcp
TRICKSTER_PORT_443_TCP_ADDR=10.222.85.133
TRICKSTER_PORT_443_TCP_PORT=443
TRICKSTER_PORT_443_TCP_PROTO=tcp
KUBERNETES_SERVICE_PORT_HTTPS=443
KUBERNETES_PORT_443_TCP=tcp://10.222.0.1:443
PROMETHEUS_METRICS_ADAPTER_PORT_443_TCP_ADDR=10.222.119.79
KUBERNETES_SERVICE_HOST=10.222.0.1
PWD=/
PROMETHEUS_METRICS_ADAPTER_PORT_443_TCP_PORT=443
PROMETHEUS_METRICS_ADAPTER_PORT_443_TCP_PROTO=tcp
TRICKSTER_SERVICE_PORT_HTTPS=443
TRICKSTER_PORT_443_TCP=tcp://10.222.85.133:443
TRICKSTER_SERVICE_HOST=10.222.85.133
```

The list of valid ones are: shard, hostname, pod_name. We probably need to stick to these three variables only.

modules/300-prometheus/openapi/config-values.yaml
```diff
@@ -112,6 +112,19 @@ properties:
...
+    description: |
+      The set of external labels to add to the metrics.
+      
+      It is possible to expand [environment variables](https://prometheus.io/docs/prometheus/latest/feature_flags/#expand-environment-variables-in-external-labels) in external labels.
+    additionalProperties:
+      type: string
+      maxLength: 63
```
nabokihms, line 127: Why is this limited? AFAIK labels in Prometheus can be any length. This is also not protecting users from overcome the limit by expanding the env variable, e.g., `abcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabc $HOSTNAME`

nabokihms (changes requested):

modules/300-prometheus/templates/prometheus/prometheus.yaml
miklezzzz: 1. The flag is required, without it Prometheus doesn't do variables expansion.
2. Enabling the flag on condition was nothing more than adhering to the principle of least privilege.

modules/300-prometheus/openapi/config-values.yaml
miklezzzz: It's rather about the envs from prometheus container, variable expanding happens in runtime as I can [see](https://github.com/prometheus/prometheus/blob/86a7064dcfc0828a10520888a188c46731986b3c/config/config.go#L80-L101).
And I'm not sure I understand why user can't use any variable, even non-existent, in such case the value would be just an empty string.

modules/300-prometheus/openapi/config-values.yaml
miklezzzz: Just some sort of precaution, I guess :)) I wanted to impose some regex but wasn't able to come up with something really useful because of env var expanding. Just wonder what happens if we use something really nasty, what Prometheus limits can we reach..

nabokihms: Prometheus does not have a limit. Any string is valid. The only problem is that metrics storage on the other end may struggle with huge labels, but it should not be a real problem.

modules/300-prometheus/templates/prometheus/prometheus.yaml
nabokihms: 1. Yes, without the flag Prometheus cannot do this, but `config-reloader` sidecar does the trick. 
See [this](https://github.com/prometheus-operator/prometheus-operator/blob/main/cmd/prometheus-config-reloader/main.go) and [this](https://github.com/thanos-io/thanos/blob/main/pkg/reloader/reloader.go) for more details.
2. Restarting the Prometheus is more problematic than enabling it in the pod by default.

modules/300-prometheus/openapi/config-values.yaml
nabokihms: Answered it there
> https://github.com/deckhouse/deckhouse/pull/4968#discussion_r1234966263

They can use it, of course, but it doesn't make sense. We should not encourage users to use variables we don't want them to use.

modules/300-prometheus/templates/prometheus/prometheus.yaml
miklezzzz: The flag has been removed and now config-reloader does env substitutions.

modules/300-prometheus/openapi/config-values.yaml
miklezzzz: length constrained has been lifted

modules/300-prometheus/openapi/config-values.yaml
miklezzzz: I couldn't come up with a decent pattern for openapi schema because re2 library doesn't support negative lookahead/lookbehind syntax, so I had to remove the pattern field

modules/300-prometheus/openapi/config-values.yaml
```diff
@@ -112,6 +112,18 @@ properties:
...
+  externalLabels:
+    type: object
+    x-examples:
+      - prometheus_replica: $(POD_NAME)
+        shard: $(SHARD)
+        hostname: $(HOSTNAME)
+    description: |
+      The set of external labels to add to the metrics.
+      
+      It is possible to expand environment variables (for example: `HOSTNAME`, `POD_NAME`, `SHARD`) in external labels.
```
nabokihms, line 124: Could you please add a description for all three variables?

nabokihms (commented):
The description is not aligned with what is done. Despite this, everything looks good 👍

modules/300-prometheus/templates/prometheus/prometheus.yaml
miklezzzz: btw, I found that we use `prometheus=deckhouse` ubiquitously for alerts grouping [like](https://github.com/deckhouse/deckhouse/blob/d171e9edcc6a44dc6bf2254e9df2fae104da6543/modules/300-prometheus/monitoring/prometheus-rules/grafana.yaml#L18). Maybe it's better to leave it as it is without a way to change it value.

miklezzzz:
@nabokihms I've updated the PR description and added some extra descriptions for variables, could you please check this [concern](https://github.com/deckhouse/deckhouse/pull/4968#discussion_r1238138951) also?

modules/300-prometheus/templates/prometheus/prometheus.yaml
nabokihms: Oh, good catch. Yes, you are right.

nabokihms (approved):

nabokihms:
@miklezzzz Looks good. Thank you for this PR.

deckhouse-BOaTswain:

PR #5023, opened 2023-06-24
[dhctl] Fix parsing node index (CWE-190, CWE-681)

nabokihms:
## Description
* Use a single function to parse node index
* Convert directly to `int`

From the description of the problems
> If a string is parsed into an int using strconv.Atoi, and subsequently that int is converted into another integer type of a smaller size, the result can produce unexpected values.

> This also applies to the results of strconv.ParseInt and strconv.ParseUint when the specified size is larger than the size of the type that number is converted to.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
summary: Fix parsing node index (CWE-190, CWE-681).
```

### DIFF
--- a/dhctl/pkg/config/meta.go
+++ b/dhctl/pkg/config/meta.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/iancoleman/strcase"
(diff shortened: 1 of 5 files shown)


deckhouse-BOaTswain:

name212 (approved):

PR #5024, opened 2023-06-24
[external-module-manager] Prevent path traversal

nabokihms:
## Description
Prevent arbitrary file write during zip extraction ("zip slip")

Extracting files from a malicious zip archive without validating that the destination file path is within the destination directory can cause files outside the destination directory to be overwritten, due to the possible presence of directory traversal elements (..) in archive paths.

Zip archives contain archive entries representing each file in the archive. These entries include a file path for the entry, but these file paths are not restricted and may contain unexpected special elements such as the directory traversal element (..). If these file paths are used to determine which output file the contents of an archive item should be written to, then the file may be written to an unexpected location. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files.

For example, if a zip file contains a file entry ..\sneaky-file, and the zip file is extracted to the directory c:\output, then naively combining the paths would result in an output file path of c:\output\..\sneaky-file, which would cause the file to be written to c:\sneaky-file.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Prevent path traversal on zip unpacking
```

### DIFF
--- a/modules/005-external-module-manager/hooks/handle_sources.go
+++ b/modules/005-external-module-manager/hooks/handle_sources.go
@@ -361,8 +361,9 @@ func copyLayerToFS(rootPath string, rc io.ReadCloser) error {
 			continue
 		}
 
-		if !strings.Contains(hdr.Name, "..") {
-			continue // CWE-22 check, prevents path traversal
+		if strings.Contains(hdr.Name, "..") {
+			// CWE-22 check, prevents path traversal
+			return fmt.Errorf("path traversal detected in the module archive: malicious path %v", hdr.Name)
 		}
 
 		switch hdr.Typeflag {
--- a/modules/005-external-module-manager/hooks/handle_sources.go
+++ b/modules/005-external-module-manager/hooks/handle_sources.go
@@ -361,6 +361,10 @@ func copyLayerToFS(rootPath string, rc io.ReadCloser) error {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


modules/005-external-module-manager/hooks/handle_sources.go
```diff
@@ -361,6 +361,10 @@ func copyLayerToFS(rootPath string, rc io.ReadCloser) error {
 			continue
 		}
 
+		if !strings.Contains(hdr.Name, "..") {
```
yalosev, line 364: do you mean `strings.Contains` ?

maybe it's better to do something like
```
if strings.HasPrefix(hdr.Name, "..") {
```

yalosev: And probably we should return an error here. Cause we wouldn't get valid result otherwise

nabokihms: Error, yes, but where? Just printing it to logs also does not make it better.
Storing it in the status is much better, but this is out of the scope of this PR. Because of the same reason, I did not add a test -- there are no tests for archive parsing.

nabokihms, line 364:
```suggestion
		if strings.Contains(hdr.Name, "..") {
```

yalosev (dismissed):

PR #5027, opened 2023-06-24
[user-authn] Do not send groups header from DexAuthenticator

nabokihms:
## Description
The Group header size can grow without control, which can cause requests to fail because of the body size

This is a regression after upgrading oauth2-proxy 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Do not send groups header from DexAuthenticator
```


deckhouse-BOaTswain:

PR #5100, opened 2023-07-03
[ingress-nginx] update stick-to-maxunavailable.patch

miklezzzz:
## Description
It transpired that conditional
`		if max := len(candidatePodsToDelete); remainingUnavailable > max {
			remainingUnavailable = max
` may result in an inability to revert a `daemonset` to the previous version (similar to `kubectl rollout undo` comman).
The `candiatePodsToDelete` variable contains pods of the former version and in case we update a `daemonset` and a new pod doesn't get ready and running (for any reason, e.g. wrong container's image), breaking the update process, if we revert changes (to deal with the new broken pod), there is the situation when there is no new revision (we're reverting), thus, `candidatePodsToDelete` has length of 0, preventing the update from progressing.

It's rather a rare situation and it's possible to recover
* through deleting a failed new pod after reverting manually;
* through applying new changes to `daemonset`;
but it makes sense to fix it.

We remove the conditional  and `remainingUnavailable = max` logic because we don't separate `candidatePodsToDelete` and `Unavailable` pods in terms of calculating how many pods are allowed to be deleted on update. Instead, we apply `remainingUnavailable` limit to the whole `oldPodsToDelete` array if the array's length is bigger than `MaxUnavailable` spec setting.

It solves the case when we need to recover a broken `daemonset` after an update.

Advanved DaemonSet are able to be reverted (they don't support `rollout undo` but the idea is the same) from a failed update.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: Ingress-Nginx
type: fix
summary: Fix kruise controller update logic when reverting a failed update.
impact: Kruise controller manager will be restarted.
impact_level: default
```

### DIFF
--- a/modules/402-ingress-nginx/images/kruise/patches/README.md
+++ b/modules/402-ingress-nginx/images/kruise/patches/README.md
@@ -11,7 +11,7 @@ We can check the number of workers (concurrent reconciles) and if we have 0 work
 Remove CRD check of `BroadcastJob` and `ImagePullJob`. We don't need them for DaemonSet workflow. We don't install that CRDs.
 
 ### Stick to MaxUnavailable
-In case DaemonSet has surge == 0, Kruise Controller still able to designated more pods as allowed for replacement than
+In case DaemonSet has surge == 0, Kruise Controller still able to designate more pods as allowed for replacement than
 maxUnavailable settings allows, resulting in parallel update instead of gradual.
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:

nabokihms (approved):

deckhouse-BOaTswain:
Cherry pick PR [5122](https://github.com/deckhouse/deckhouse/pull/5122) to the branch [release-1.47](https://github.com/deckhouse/deckhouse/tree/release-1.47) successful!

PR #5110, opened 2023-07-04
[runtime-audit-engine] Mount sockets and fix webhook output

nabokihms:
## Description
* Mount docker and containerd sockets to fetch metadata.
* Mount falco config to rules-loader to enable plugins for validating webhook. Otherwise, webhook returns an error for valid rules.
* Output webhook validation error. Without this change, users have to search it in logs among all running falco pods.
* Fix converter `apiVersion` header, it was `apiversion`, which is not valid.
* Fix falcoauditrules resource name in rules. 

Various fixes of problems found while developing rules.

Falco pods will be restarted.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: runtime-audit-engine
type: fix
summary:  Mount docker and containerd sockets to fetch metadata.
---
section: runtime-audit-engine
type: fix
summary: Mount falco config to rules-loader to enable plugins for validating webhook. Otherwise, webhook returns an error for valid rules.
---
section: runtime-audit-engine
type: fix
summary:  Output webhook validation error. Without this change, users have to search it in logs among all running falco pods.
---
section: runtime-audit-engine
type: fix
summary:  Fix falcoauditrules resource name in rules. 
---
section: runtime-audit-engine
type: fix
summary:  Fix converter `apiVersion` header, it was `apiversion`, which is not valid.
impact_level: low
```

### DIFF
--- a/ee/modules/650-runtime-audit-engine/templates/daemonset.yaml
+++ b/ee/modules/650-runtime-audit-engine/templates/daemonset.yaml
@@ -146,6 +146,10 @@ spec:
           readOnly: true
         - name: rules-data
           mountPath: /etc/falco/rules.d
+        - name: dockersock
+          mountPath: /var/run/docker.sock
+        - name: containerdsock
+          mountPath: /run/containerd/containerd.sock
 
       - name: falcosidekick
         {{- include "helm_lib_module_container_security_context_read_only_root_filesystem_capabilities_drop_all" . | nindent 8 }}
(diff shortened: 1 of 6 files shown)


ee/modules/650-runtime-audit-engine/templates/daemonset.yaml
```diff
@@ -146,6 +146,10 @@ spec:
           readOnly: true
         - name: rules-data
           mountPath: /etc/falco/rules.d
+        - name: dockersock
+          mountPath: /var/run/docker.sock
```
RomanenkoDenys, line 150: We plan to remove docker support in 1.48. Does we need docker socket mount?

RomanenkoDenys (commented):
one question

PR #5130, opened 2023-07-06
[user-authn] Improve groups migration

nabokihms:
## Description
* Run migration only once
* Slugify names
* Add the group name field to the output table

The previous version of the migration can fail or overwrite manual changes.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Improve the groups migration (run only once and slugify group names)
```

### DIFF
--- a/modules/150-user-authn/hooks/migration_groups_from_user_to_group_object.go
+++ b/modules/150-user-authn/hooks/migration_groups_from_user_to_group_object.go
@@ -17,49 +17,65 @@ limitations under the License.
 package hooks
 
 import (
-	"context"
 	"fmt"
+	"hash/fnv"
+	"regexp"
 	"strings"
 
 	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
 	"github.com/flant/addon-operator/sdk"
-	"k8s.io/apimachinery/pkg/api/errors"
+	"github.com/flant/shell-operator/pkg/kube/object_patch"
+	"github.com/flant/shell-operator/pkg/kube_events_manager/types"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


modules/150-user-authn/hooks/migration_groups_from_user_to_group_object.go
```diff
@@ -75,33 +90,72 @@ func migrationGroups(input *go_hook.HookInput, dc dependency.Container) error {
...
 		}
-		newDexGroup := &DexGroup{
+		dexGroup := &DexGroup{
 			TypeMeta: metav1.TypeMeta{
 				Kind:       DexGroupKind,
 				APIVersion: DexGroupAPIVersion,
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name: strings.ToLower(groupName),
+				Name: metadataNameify(groupName),
```
shvgn, line 99: 👍😅

modules/150-user-authn/hooks/migration_groups_from_user_to_group_object_test.go
```diff
@@ -61,55 +88,94 @@ metadata:
 spec:
   email: user@example.com
   groups:
+  - admins
```
shvgn, line 91: I believe we should preserve group names which were valid metadata.names from the beginning. Thus we would not punish users who were lucky to name groups according to new concerns. They would not need to rename groups if they were named well.

E.g. 

- admins → admins
- Admins → admins-(somehash)
- -Admins → admins-(anotherhash)

nabokihms: Good point, fixed it.

deckhouse-BOaTswain:
Aye, aye, @nabokihms. I've started the workflow for label 'e2e/run/static'...

shvgn (approved):

PR #5186, opened 2023-07-12
[deckhouse] modules description in the module.yaml file

### LINKED ISSUE #4421: [deckhouse] Module CRD
### Preflight Checklist

We want to add new module CRD. When addon-operator is registering a module, new crd must be created in the cluster.
It shouldn't be updated by user, only by controller.
We need to add more information to the each module, something like `module.yaml` with extra information, which will be used or copied to the module

### Proposed Solution

I assume all of these should be implemented in the addon-operator, to work natively with Modules:
- Read `module.yaml` on a module startup
- Register module inside
- Create module CRD on registering
- Calculate the map of uniqueness
- Sort modules
- Start module with the respect of unique key and order

Also we have to realize validation webhook in the addon operator, to prohibit module crd change by user

### Additional Information

`module.yaml` structure:
```yaml
name: "my-super-module" # same name as in Chart.yaml, we have to sync or validate it with Chart.yaml
namespace: "my-ns" # instead of .namespace
disableImpact: "Some description why we can not do this"
dependencies: # skip on the first phase
  modules: ["module-1", "module-2"] 
enableOnBoostrap: true # **boolean**: can be used instead of enabled scripts
!!!Think more uniqueKey: "cni" # some unique key for the category. Only one module with this key must be enabled
labels: ["crd", "cni"] # tags for filtering, when can move it to the Module CRD label
weight: 100 # module weight, instead of number in the module name
```

`Module` CRD example:
```yaml
kind: Module
metadata:
  name: my-super-module
  labels:
    module.deckhouse.io/crd: ""
    module.deckhouse.io/cni: ""
properties:
  weight: 100
#  namespace: "my-ns" # 2 wave
#  uniqueKey: "cni" 
#  disable:
#    impact: "critical" 
```

yalosev:
## Description

- Use golang 1.20 for building (according to addon-operator and shell-operator)
- Unit tests kubernetes client moved to the 1.25 version
- Add ability to read modules weight and tags from `module.yaml` file (we can set weight to external modules now)
- Validate external modules before install. Don't apply external module if it's broken

(Close #4421)

Broken external module can block the whole deckhouse queue, we have to prevent it
Sometimes we want to apply external modules in the specific order

<details>
  <summary>ExternalModule weight and custom tags</summary>

```bash
hello-world                           987      Enabled   External: losev-test   module.deckhouse.io/testme=
```
</details>

<details>
  <summary>ExternalModuleSource contains failed module</summary>

```bash
losev-test   alpha             2       42s    Some errors occurred. Inspect status for details
```

```yaml
kind: ExternalModuleSource
metadata:
  name: losev-test
spec:
  registry:
    dockerCfg: xxx
    repo: xxx
  releaseChannel: alpha
status:
  availableModules:
  - echo
  - hello-world
  message: Some errors occurred. Inspect status for details
  moduleErrors:
  - error: 'add schemas failed: prepare module ''echo'' schemas: load global ''config''
      schema: json unmarshal: json: cannot unmarshal array into Go struct field .properties
      of type struct { spec.SchemaProps; spec.SwaggerSchemaProps }'
    name: echo
  modulesCount: 2
```

</details>

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feature
summary: Ability to specify module tags and weight in the `module.yaml` file.
```

```changes
section: chore
type: feature
summary: Use golang 1.20 for build.
```


deckhouse-BOaTswain:

go.mod
```diff
@@ -11,17 +11,17 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/deckhouse/deckhouse/dhctl v0.0.0 // use non-existent version for replace
 	github.com/fatih/color v1.13.0
-	github.com/flant/addon-operator v1.1.5-0.20230618220332-871f9192e42d // commit: 871f9192e42da6e35653f4dde81ff3f99c87e9e4
+	github.com/flant/addon-operator v1.2.0 // commit: 871f9192e42da6e35653f4dde81ff3f99c87e9e4
```
nabokihms, line 14:
```suggestion
	github.com/flant/addon-operator v1.2.0
```

modules/005-external-module-manager/crds/external-module-release.yaml
```diff
@@ -38,6 +38,9 @@ spec:
                   type: string
                   description: Module version.
                   example: 'v1.0.0'
+                weight:
+                  type: number
```
nabokihms, line 42:
```suggestion
                  type: integer
```

modules/005-external-module-manager/hooks/handle_sources.go
```diff
@@ -258,7 +300,38 @@ func fetchModuleVersion(logger *logrus.Entry, dc dependency.Container, moduleSou
...
+		return defaultExternalModuleWeight
+	}
+
+	if def.Weight > 0 && def.Weight < 100 {
+		def.Weight = defaultExternalModuleWeight + def.Weight
+	}
+
+	if def.Weight < defaultExternalModuleWeight || def.Weight >= 1000 {
+		def.Weight = defaultExternalModuleWeight
+	}
```
nabokihms, line 329: Do we have validation of module.yaml while downloading the module?
Can we validate the module.yaml against a schema?

Probably we need to introduce versioning to make schema changes like in shell operator.

nabokihms (approved):

modules/005-external-module-manager/hooks/handle_sources.go
yalosev: create an issue: https://github.com/flant/addon-operator/issues/402

Issue #5212, opened 2023-07-14
[admission-policy-engine] Wrong duplicated env variable names rule in d8containerduplicates ConstraintTemplate


verdel:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.46.11

### Expected Behavior

OperationPolicy `checkContainerDuplicates` should restrict the appearance of environment variables with the same name within the container entity.

### Actual Behavior

The policy restricts the appearance of environment variables with the same name within the pod not in individual container. If environment variables with the same name are set in different pod containers, the policy still restricts the creation of such a pod.

### Steps To Reproduce

OperationPolicy `checkContainerDuplicates` uses this [ConstraintTemplate](https://github.com/deckhouse/deckhouse/blob/main/modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/container-duplicates.yaml).

You can test the policy from ConstraintTemplate `d8containerduplicates` in [Rego Playground](https://play.openpolicyagent.org/p/Ww3GBHkTZu) with this simple pod as review object example:
```json
{
  "review": {
    "object": {
      "apiVersion": "v1",
      "kind": "Pod",
      "metadata": {
        "name": "test-pod",
        "namespace": "test-namespace"
      },
      "spec": {
        "containers": [
          {
            "env": [
              {
                "name": "SERVER_PORT",
                "value": "8081"
              }
            ],
            "name": "test-container-1"
          },
          {
            "env": [
              {
                "name": "SERVER_PORT",
                "value": "8082"
              }
            ],
            "name": "test-container-2"
          }
        ]
      }
    }
  }
}

```

### Additional Information

If we change the pod manifest for tests [here](https://github.com/deckhouse/deckhouse/blob/main/modules/015-admission-policy-engine/charts/constraint-templates/tests/container-duplicates/disallowed_envs.yaml) to:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: disallowed
  namespace: default
spec:
  containers:
    - name: nginx1
      image: nginx
      env:
        - name: key1
          value: val1
    - name: nginx2
      image: nginx
      env:
        - name: key1
          value: val1
```
and run test:
```shell
[...]

PR #5214, opened 2023-07-14
[admission-policy-engine] Fix D8ContainerDuplicates constraint template policy

verdel:
## Description
I have fixed the OPA policy to check for duplicate environment variables with the same name. Prior to the correction, the policy checked duplicates of environment variables not in individual containers, but in all containers in the pod.

The problem is described in more detail in issue #5212

At the moment, the policy disallowed the creation of a pod that has environment variables with the same name in different containers.

In my opinion, the current behavior of the policy needs to be corrected as soon as possible. I suggest adding fixes in the next patch release.

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
type: fix
summary: Fix `checkContainerDuplicates` policy. Allow env with the same name in the different containers
```


konstantin-axenov:
Closes https://github.com/deckhouse/deckhouse/issues/5212

yalosev (approved):

yalosev:
Thank you for the PR!

PR #5274, opened 2023-07-19
[global] add status 

miklezzzz:
## Description

- `set_cr_statuses` package has been added to `go_lib` for setting additional status fields.
- SecurityPolicy, CustomAlertmanager and NodeGroup CRD have been updated to include additional status fields and printer columns.
- Additional onAfterHelm hooks for the SecurityPolicy, CustomAlertmanager and NodeGroup Custom Resource have been added to set additional status fields (timestamp and generation).
- Prometheus modules has been updated to store CustomAlertmanagers' names correctly.
- Some minor changes for the relevant Custom Resources.

We have to implement some sort of a status field for deckhouse.io CustomResources that gets updated every time Deckhouse operator applies the CR in the cluster, so that it would be possible to say if latest CR version was implemented or not.

SecurityPolicy, CustomAlertmanager and NodeGroup Custom Resources should have additional status fields of the following schema:
```
deckhouse:
  synced: <true/false>
  observed:
    checkSum: <string> 
    lastTimestamp: <timestamp>
  processed:
    checkSum: <string> 
    lastTimestamp: <timestamp>
```
Also, additional columns `Last observerd` and `Last applied` should be provided in the output of `kubectl -o wide` for the aforementioned resources. 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: global
type: feature
summary: Provide additional status fields for custom resources.
```


miklezzzz:
not to forget - fix security policy issue when Last Applied generation doesn't get updated (preprocessed fields)/

go_lib/hooks/set_cr_statuses/hook.go
```diff
@@ -0,0 +1,46 @@
...
+
+func getTimeStamp() string {
+	curTime := time.Now()
+	if timeStr, ok := os.LookupEnv("TEST_CONDITIONS_CALC_NOW_TIME"); ok {
+		curTime, _ = time.Parse(time.RFC3339, timeStr)
+	}
+	return curTime.Format(time.RFC3339)
+}
+
+var SetProcessedStatus = func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
```
nabokihms, line 35:
```suggestion
func SetProcessedStatus(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
```

This is a pitfall because a user can change the value of this variable. It is better to make it a package method because it is immutable and exported at the same time.

modules/010-prometheus-crd/crds/customalertmanager.yaml
```diff
@@ -1249,3 +1267,13 @@ spec:
           name: Type
           description: 'Type of an Alertmanager.'
           type: string
+        - jsonPath: .metadata.generation
```
nabokihms, line 1270: `.metadata.generation` is not the last observed. It is just a generation. The last observed should be the one from the status. It is not guaranteed that Deckhouse actually observed the generation because it may not be running at the time.

modules/040-node-manager/hooks/update_node_group_status.go
```diff
@@ -417,6 +419,8 @@ func handleUpdateNGStatus(input *go_hook.HookInput) error {
 		)
 
 		patchNodeGroupStatus(input.PatchCollector, ngName, patch)
+		// set CR processed status
+		input.PatchCollector.Filter(set_cr_statuses.SetProcessedStatus, "deckhouse.io/v1", "nodegroup", "", ngName, object_patch.WithSubresource("/status"))
```
nabokihms, line 423: So, deckhouse reacted and set the status, but what happens after helm? Does it set the same status to the same field?

My question is, how can I distinguish when the hook observes a new resource and when the helm chart, according to this new resource, was deployed?

miklezzzz: If my memory serves me well the problem that we couldn't resolve with LastObserved was that we can't use PatchCollector for applying LastObserved status if a hook fails to converge (if a user accidentally or deliberately provides inconsistent CR config). In other words:
* if a CR is valid than we have neat status fields (LastObserved and LastProcessed would match);
* if a CR is erroneous (for example, it was possible to create an alert manager CR, filling in non-existent kubernetes service as the destination) resulting in a failed hook, LastObserved status can't get updated because `PatchCollector` applies patches only after a successful hook run.
This limitation prevents us from setting an actual status for LastObserved in case there is an error. Moreover, it renders the CR as in sync because LastObserved and LastProcessed would match.

This is why we decided to give it a try with the generation field instead of LastObserved, even though it has nothing to do with deckhouse, it is still a kind of a marker that something was changed.

miklezzzz:
image

miklezzzz:
with this [PR](https://github.com/flant/addon-operator/pull/414) for addon-operator, it would be possible to set neat statuses even in case of failed hooks, so I decided to give observed/processed statuses another try. In case I set an erroneous alertmanager CR, the hook gets status rendered correctly.
image

deckhouse-BOaTswain:

miklezzzz:
/e2e/abort yandex-cloud;WithoutNAT;containerd;1.25 6457798066 6457798066-1-con-1-25 /sys/deckhouse-oss/install:pr5274 cloud-user@158.160.117.248

deckhouse-BOaTswain:
Aye, aye, @miklezzzz. I've recognized your '/e2e/abort' command and started the workflow...

deckhouse-BOaTswain:

yalosev (approved):

nabokihms (approved):

deckhouse-BOaTswain:

deckhouse-BOaTswain:

deckhouse-BOaTswain:
Aye, aye, @miklezzzz. I've started the workflow for label 'e2e/run/aws'...

deckhouse-BOaTswain:

PR #5327, opened 2023-07-24
[candi] Remove prune-containerd-content ngc

asviel:
## Description

Cleanup oneshot job. Remove flag file.

Just housekeeping

We don't

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: chore
summary: Remove `prune-containerd-content` ngc.
impact_level: low
```


RomanenkoDenys (approved):
LGTM

PR #5332, opened 2023-07-24
[cloud-provider-openstack] Fix creating long name of backup secret in state migration

name212:
## Description
Reduce backup state secret length for openstack state migration. 

During migration backup secret can not be created for clusters with long prefix and Deckhouse queue will be stuck.

During migration backup secret can not be created for clusters with long prefix and Deckhouse queue will be stuck.

## Manual test proofs
image

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Fix creating long name of backup secret in state.
impact_level: default
```


deckhouse-BOaTswain:

RomanenkoDenys (approved):
i prefer to use hash, but this solution is ok too.

PR #5517, opened 2023-08-11
[flant-integration] Change `sum` to `avg` in `controller_metrics` hook and refactor utils

alex123012:
## Description

Change `sum` to `avg` in `controller_metrics` hook and refactor utils.

We don't want to use `sum` func because it result depends on replicas count.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: flant-integration
type: fix
summary: Change `sum` to `avg` in `controller_metrics` hook and refactor utils.
impact_level: default
```


alex123012:
`sum by (pod) (...)` were added to proper count average value not for containers, but for all pod containers resource consumption

nabokihms (changes requested):
Overall, looks good.

PR #5596, opened 2023-08-19
[igress-nginx] protobuf-exporter update

### LINKED ISSUE #5592: [ingress-nginx] there are data skips at rare metrics from protobuf-exporter
### Preflight Checklist

### Version

v1.50.1

### Expected Behavior

Sampling formulas should take this feature of the exporter into account.

### Actual Behavior

In all charts that use metrics from protobuf-exporter. metrics do not start from 0 but from 1, which gives data skipping on dashboards in grafana where "rate" is used. 

### Steps To Reproduce

1. grafana dashboards vhosts
2.  graph 4xx 
3. there is a 404 error that happens once 2h it is not visible on the charts. 

### Additional Information

https://flant.slack.com/archives/CML93AWG2/p1692116907047179
https://github.com/deckhouse/deckhouse/blob/main/modules/402-ingress-nginx/images/protobuf-exporter/rootfs/etc/protobuf_exporter/mappings.yaml

### Logs

_No response_

miklezzzz:
## Description
Protobuf exporter is updated to expose a zero value counter metric for PROMETHEUS_SCRAPE_INTERVAL time when a new counter metric is added to the collector. After PROMETHEUS_SCRAPE_INTERVAL the zero metric is updated with real accumulated counter value.
Closes #5592 

This approach allows us to smooth the Prometheus problem when in can't calculate `rate` derivatives for uninitiated metrics (when a new metric is created with non-zero value, Prometheus's rate function isn't triggered).

New counter metrics are created with zero values for the duration of PROMETHEUS_SCRAPE_INTERVAL value, then they get updated with real values (stored in temporary collector).

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: Ingress-nginx
type: feature
summary: Update protobuf-exporter collector so that new counter metrics are initiated with 0 values.
impact: All ingress nginx pods will be recreated.
impact_level: high
```

### DIFF
--- a/modules/402-ingress-nginx/images/protobuf-exporter/main.go
+++ b/modules/402-ingress-nginx/images/protobuf-exporter/main.go
@@ -34,18 +34,24 @@ func main() {
 	exporterAddress := ":8081"
 	mappingsPath := "./mappings.yaml"
 	logLevel := "info"
+	scrapeInterval, err := time.ParseDuration(os.Getenv("PROMETHEUS_SCRAPE_INTEVAL"))
+	if err != nil {
+		scrapeInterval = (time.Second * 30)
+		log.Warnf("Set default scrape interval to %s: %v", scrapeInterval, err)
+	}
 
 	flag.StringVar(&telemetryAddress, "server.telemetry-address", telemetryAddress, "Address to listen telemetry messages")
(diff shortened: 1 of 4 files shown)


deckhouse-BOaTswain:

miklezzzz:
it makes sense to stress the fact that ingress nginx pods have to be recreated every time prometheus scrape interval changes :/

PR #5641, opened 2023-08-23
[deckhouse] Set ready and converging status for module configs

nabokihms:
## Description
* Set status for modules in their normal state
* Migrate the source field from `Embedded / External: %s` to `<empty>/%s`.

It is not clear wether a module is ready or not.

Best effort alarm!

Actually, the condition for a module stage is not correct because the `CanRunHelm` status appears right before the first run.
The right approach is to check the queue for the module run task. However, there are too many addon-operator internals involved. We should consider moving these statuses to the `Module` resource, which is directly controlled by addon-operator.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature 
summary: Set ready and converging status for module configs.
```

### DIFF
--- a/go_lib/deckhouse-config/status.go
+++ b/go_lib/deckhouse-config/status.go
@@ -24,6 +24,7 @@ import (
 	"github.com/deckhouse/deckhouse/go_lib/deckhouse-config/conversion"
 	"github.com/deckhouse/deckhouse/go_lib/set"
 	d8v1alpha1 "github.com/deckhouse/deckhouse/modules/002-deckhouse/hooks/pkg/apis/v1alpha1"
+	"github.com/flant/addon-operator/pkg/module_manager"
 )
 
 type Status struct {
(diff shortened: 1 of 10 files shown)


go_lib/deckhouse-config/status.go
```diff
@@ -58,11 +60,6 @@ func (s *StatusReporter) ForConfig(cfg *d8v1alpha1.ModuleConfig, bundleName stri
 		}
 	}
 
-	moduleType := "Embedded"
-	if isExternal {
-		moduleType = fmt.Sprintf("External: %s", repo)
-	}
```
yalosev, line 64: It will break [this part](https://github.com/deckhouse/deckhouse/blob/main/modules/002-deckhouse/hooks/pkg/apis/v1alpha1/module.go#L86-L91) of a Module generation.
As result, it will break [the cleanup](https://github.com/deckhouse/deckhouse/blob/main/modules/005-external-module-manager/hooks/cleanup_releases.go#L54)

yalosev (changes requested):

go_lib/deckhouse-config/status.go
yalosev: Or not...
Anyway, we have one more place with an `external type` labels

nabokihms: Good catch! Fixed.

yalosev (approved):

PR #5664, opened 2023-08-28
[runtime-audit-engine] add read-only root for falco container

RomanenkoDenys:
## Description

Add read-only root for falco container

Read-only container fs is needed for security purposes.
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: security
type: fix
summary: add read-only root for falco container
impact: runtime-audit-engine pods should be restarted
impact_level: default
```


RomanenkoDenys:
Rules reloader works

PR #5690, opened 2023-08-31
[linstor] fixes in SPAAS usage

duckhawk:
## Description

We are currently using an external SPAAS service instead of a cluster in any case.

It is necessary to use an internal SPAAS service to save traffic especially in closed environments.

This affects our customers.

Using an internal SPAAS service when building DRBD.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: linstor
type: fix
summary: Using an internal SPAAS service instead of the external one when building DRBD.
```

### DIFF
--- a/modules/007-registrypackages/images/drbd/werf.inc.yaml
+++ b/modules/007-registrypackages/images/drbd/werf.inc.yaml
@@ -35,3 +35,4 @@ shell:
     - chmod +x /install
     - chmod +x /uninstall
     - echo 'GIT-hash:123456' > /drbd/drbd/.drbd_git_revision
+    - sed -e "s/\<curl\>/d8-curl -k/g" -i /drbd/drbd/drbd-kernel-compat/gen_compat_patch.sh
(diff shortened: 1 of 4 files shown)


modules/007-registrypackages/images/drbd/werf.inc.yaml
```diff
@@ -35,3 +35,4 @@ shell:
     - chmod +x /install
     - chmod +x /uninstall
     - echo 'GIT-hash:123456' > /drbd/drbd/.drbd_git_revision
+    - sed 's/[[:space:]]curl[[:space:]]/ d8-curl -k /g' -i /drbd/drbd/drbd-kernel-compat/gen_compat_patch.sh
```
AleksZimin, line 38: Hello! I think it is better to use sed with the -e option and use < and > as word boundaries, instead of [[:space:]]. In this case, the tab will not be removed from the last command curl in gen_compat_patch.sh script
```suggestion
    - sed -e "s/\<curl\>/d8-curl -k/g" -i /drbd/drbd/drbd-kernel-compat/gen_compat_patch.sh
```

AleksZimin (approved):

deckhouse-BOaTswain:
Cherry pick PR [5698](https://github.com/deckhouse/deckhouse/pull/5698) to the branch [release-1.50](https://github.com/deckhouse/deckhouse/tree/release-1.50) successful!

PR #5706, opened 2023-09-01
[go_lib] Change generate password hook

yalosev:
## Description
We need a flag to control basic-auth password deletion if external authentication is enabled

Prometheus alertmanager always use basic-auth event is external auth is set.
We want to keep this password forever

Alertmanager should have the `nil` password even the external auth is set

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: go_lib
type:  chore
summary: Add ability to keep basic-auth password via the flag
impact_level: low
```

```changes
section: prometheus
type:  fix
summary: Keep basic-auth password for alertmanager event is external authentication is set.
```

### DIFF
--- a/go_lib/hooks/generate_password/hook.go
+++ b/go_lib/hooks/generate_password/hook.go
@@ -37,22 +37,34 @@ const (
 	generatedPasswdLength      = 20
 )
 
-func NewBasicAuthPlainHook(moduleValuesPath string, ns string, secretName string) *Hook {
+func NewBasicAuthPlainHook(settings HookSettings) *Hook {
 	// Ensure camelCase for moduleValuesPath
-	valuesKey := addonutils.ModuleNameToValuesKey(moduleValuesPath)
+	valuesKey := addonutils.ModuleNameToValuesKey(settings.ModuleName)
 	return &Hook{
 		Secret: Secret{
-			Namespace: ns,
-			Name:      secretName,
+			Namespace: settings.Namespace,
+			Name:      settings.SecretName,
 		},
-		ValuesKey: valuesKey,
+		ValuesKey:                  valuesKey,
+		keepPasswordOnExternalAuth: settings.KeepPasswordOnExternalAuth,
 	}
+}
+
+type HookSettings struct {
+	ModuleName string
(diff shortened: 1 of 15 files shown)


go_lib/hooks/generate_password/hook.go
```diff
@@ -37,22 +37,34 @@ const (
...
 }
 
+type HookSettings struct {
+	ModuleName string
+	Namespace  string
+	SecretName string
+
+	// by default, basic-auth password will be removed if externalAuth is set
+	// you can change this behavior by this flag
+	KeepPasswordOnExternalAuth bool
```
nabokihms, line 60: I didn't get the purpose of the flag. Do we always need to keep the password or not? It should be one or another.

> Prometheus alertmanager always uses basic-auth even if external auth is set.

In this case, why the same password is used for all ingresses? It doesn't seem right.

nabokihms: I checked the alertmanager ingress settings, and there is a clear bug.
https://github.com/deckhouse/deckhouse/blob/3feea08d722b1dfd49bdc6f39138517358a59b37/modules/300-prometheus/templates/alertmanager/internal/ingress.yaml#L17-L24

This is the only Ingress where the `Values.global.enabledModules | has "user-authn"` clause is used.

nabokihms (changes requested):

go_lib/hooks/generate_password/hook.go
yalosev: Nope, usually we delete basic-auth password on external auth.
The same password is used for all ingresses in the `300-prometheus` module because all of them on the same domain, like `grafana.mydomain.com/XXX`, I don't think they would work with different passwords.

It's not a bug, because we need a basic-auth for alertmanager, it's the easiest way to authenticate for grafana/other alert systems. It would be really hard to use external authentication here.

yalosev: Discussed. Now we will remove this flag and handle external authentication
Alertmanager will be refactored later

nabokihms (approved):

PR #5714, opened 2023-09-04
[dhctl] chore: expose dhctl commands as functions of pkg/app/commands package

distorhead:
## Description

Moved kingpin run-function params from cmd/dhctl/commands to separate package pkg/app/commands.

## Why do we need it, and what problem does it solve?

This change enables embedding dhctl as a library.

PR #5715, opened 2023-09-04
[dhctl] feat(dhctl-for-commander): add ability to pass initial state and on-phase callback for dhctl operations

distorhead:
Depends on (merge only after this PR): https://github.com/deckhouse/deckhouse/pull/5714


dhctl/pkg/operations/common.go
```diff
@@ -0,0 +1,36 @@
+package operations
+
+import (
+	"errors"
+	"github.com/deckhouse/deckhouse/dhctl/pkg/state"
```
name212, line 5: sort imports

dhctl/pkg/operations/common.go
```diff
@@ -0,0 +1,36 @@
...
+
+type OnPhaseFunc func(completedPhase OperationPhase, completedPhaseState DhctlState, nextPhase OperationPhase, nextPhaseCritical bool) error
+
+type OperationPhase string
+
+const (
+	BaseInfraPhase             OperationPhase = "BaseInfra"
+	ExecuteBashibleBundlePhase OperationPhase = "ExecuteBashibleBundle"
+	InstallDeckhousePhase      OperationPhase = "InstallDeckhouse"
+	CreateResourcesPhase       OperationPhase = "CreateResources"
```
name212, line 18: I think we need one more phase - installAdditionalMastersAndStaticNodes

name212: After install deckhouse

name212 (changes requested):

dhctl/pkg/operations/common.go
distorhead: I would propose to add such phase in the separate PR on top of existing. BTW in the last PR there is small refactor and introduction of PhasedExecutionContext, which is related to the phases.

distorhead:
Merging this PR into single aggregated MR [feat/dhctl-for-commander-1](https://github.com/deckhouse/deckhouse/tree/feat/dhctl-for-commander-1)

PR #5718, opened 2023-09-05
[external-module-manager] Fix modules migration hook

yalosev:
## Description
Create CRD before modules migration

EnsureCRDs works via PatchCollector and applies changes only in the end of the script. So, we don't have new CRD when creating the Modules.

1.51 is broken atm. It will fail if you have existed resources in the cluster

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: global
type: fix 
summary: Fix modules migration webhook.
impact_level: low
```


deckhouse-BOaTswain:

Issue #5238, opened 2023-07-17
[delivery] Module does not work with CustomCertificate https mode


yalosev:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.47.3

### Expected Behavior

Custom certificate should be copied and module should start gracefully with Dex auth.

### Actual Behavior

Module `delivery` tries to copy custom certificate but fails with:
```
1. ModuleRun:main:delivery:doStartup:ReloadAllModules:failures 15:helm upgrade failed: failed to create resource: Secret "ingress-tls-argocd-customcertificate" is invalid: [data[tls.crt]: Required value, data[tls.key]: Required value]
```

module values are empty

### Steps To Reproduce

1. Create global ModuleConfig with the next settings:
```yaml
apiVersion: deckhouse.io/v1alpha1
kind: ModuleConfig
metadata:
  name: global
spec:
  settings:
    modules:
      https:
        customCertificate:
          secretName: foobar
        mode: CustomCertificate
      publicDomainTemplate: '%s.example.com'
  version: 1
```

2. Enable ModuleConfig `delivery`

PR #5762, opened 2023-09-07
[delivery] Fix customCertificate

qw1mb0:
## Description
Add CustomCertificate hook for delivery module.

Fix https://github.com/deckhouse/deckhouse/issues/5238. Delivery module fails when using CustomCertificate https mode.

This affects our customers.

Delivery module work with CustomCertificate https mode.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: delivery
type: fix
summary: Fix delivery module work in customCertificate https mode.
```

### DIFF
--- a/ee/modules/502-delivery/templates/argocd/argocd-cm-configmap.yaml
+++ b/ee/modules/502-delivery/templates/argocd/argocd-cm-configmap.yaml
@@ -13,4 +13,5 @@ data:
     issuer: https://{{ include "helm_lib_module_public_domain" (list . "dex") }}/
     clientID: dex-client-argocd@d8-{{ .Chart.Name }}
     clientSecret: $dex-client-argocd:clientSecret
+  oidc.tls.insecure.skip.verify: "true"
   {{- end }}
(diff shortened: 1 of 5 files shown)


ee/modules/502-delivery/templates/argocd/argocd-cm-configmap.yaml
```diff
@@ -13,4 +13,5 @@ data:
     issuer: https://{{ include "helm_lib_module_public_domain" (list . "dex") }}/
     clientID: dex-client-argocd@d8-{{ .Chart.Name }}
     clientSecret: $dex-client-argocd:clientSecret
+  oidc.tls.insecure.skip.verify: "true"
```
shvgn, line 16: I guess it works fine if the certificate is not a custom one. Should this line be wrapped with `if` condition?

shvgn: @nabokihms please take a look

PR #6041, opened 2023-10-04
[dhctl] fix(dhctl-for-commander): dhctl cluster config validation

Isteb4k:
## Description
Moved cluster config validation from `commands` package to separated `configure` package.
This will allow to import validation functionality (`ParseClusterConfiguration`) into DeckhouseCommander (and probably into other projects as needed)


distorhead:
@Isteb4k Rebase on actualized feat/dhctl-for-commander-improvements branch, please. Otherwise LGTM.

PR #6194, opened 2023-10-15
[local-path-provisioner] Move to distroless

name212:
## Description

local path provisioner images are now based on distroless image.

We'd like to base our images on a distroless image.

## Manual testing

Have two local path provisioners
image

Also have pvc and pv. Directory were created on node
image
image

Switch to pr branch
All are running
image

After add 2 sts, pvs were created successfully
image

Directories were created with right permission
image

After deleting sts and pvc one pv was kept another was delete as need
image

Directory was created too
image

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: local-path-provisioner
type: feature
summary: Move to distroless
impact_level: default
```


deckhouse-BOaTswain:

modules/031-local-path-provisioner/images/helper/dir-manager/main.go
```diff
@@ -0,0 +1,81 @@
...
+
+func main() {
+	flag.Parse()
+	if action != SETUP && action != TEARDOWN {
+		fmt.Fprintf(os.Stderr, "Error. Incorrect action: %s", action)
+		os.Exit(1)
+	}
+
+	if path == "" {
+		os.Stderr.WriteString("Error. Path is empty")
```
RomanenkoDenys, line 53: may be also use fmt.Printf ? To make the code the same.

RomanenkoDenys: optional

modules/031-local-path-provisioner/images/helper/werf.inc.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - '**/*'
+shell:
+  installCacheVersion: "2023-10-16.2"
+  install:
+    - export "CGO_ENABLED=0"
+    - export "GOOS=linux"
+    - export "GOARCH=amd64"
+    - cd /src
+    - echo "Build manager"
+    - go build -o /tmp/manager main.go
```
RomanenkoDenys, line 19:
```suggestion
    - GOPROXY={{ $.GOPROXY }} go build -o /tmp/manager main.go
```

RomanenkoDenys (changes requested):
some nits

modules/031-local-path-provisioner/images/helper/dir-manager/main.go
name212: good point. I've fixed it

RomanenkoDenys (approved):
LGTM

PR #6215, opened 2023-10-17
[bashible] Add Altlinux 10.0, 10.2 support

nabokihms:
## Description
Change the pattern

Both versions work fine (even if the 10.2 is still in the development stage), but Deckhouse returns an error that the version is not supported

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: bashible
type: fix
summary: Add Altlinux 10.0, 10.2 support
```


RomanenkoDenys:
tested by client

RomanenkoDenys (approved):
LGTM

PR #6254, opened 2023-10-19
[ingress-nginx] Add PDB

### LINKED ISSUE #6244: [ingress-nginx] implement /scale subresource for AdvancedDaemonSet
### Preflight Checklist

PDBs don't work without scale subresource. We have to implement it for kruise controller

### Proposed Solution

We need a simple read-only scale subresource in the kruise controller.
We can just copy `status` fields like `numberReady` and `desiredNumberScheduled` from AdvancedDaemonSet status to Scale

### Additional Information

_No response_

miklezzzz:
## Description
This PR introduces following changes in order to make `Advanced DaemonSet` support `PodDisruptionBudget`:
* Adds `.spec.Replicas` to the `Advanced DaemonsetSet` CRD to fulfill the `PDB` requirements. The field is updated by the Open Kruise Controller every time the `Advanced DaemonetSet` gets updated and the value is based on the number of the nodes currently available for scheduling the daemonset's pods (nodes are matched by the daemonset's `nodeSelector`). Cordoned nodes are also counted as possible replicas, allowing no more than the `MaxUnavailable PDB value` nodes to be cordoned/drained.
* Open Kruise reconcile loop for daemonsets also has been updated to perform periodic resyncs (once in 60 seconds, debatable magic number) for the daemonsets which have cordoned nodes counted as possible replicas. These periodic resyncs are required for updating `.spec.replicas` fields in a timely manner in case the cordoned nodes get deleted after some time. Without these periodic resyncs Open Kruise controller can't notice node deletions (it watches daemonsets - not nodes) and update `spec.replicas` accordingly.
* Open Kruise controller scheduling logic has been updated so that it doesn't recreate pods on drained and cordoned nodes(before these changes, it wasn't possible to drain a node completely as kruise controller would keep creating pods for cordoned nodes).
* Also, Open Kruise controller is granted with permissions to update PDBs' statuses. It is required in the aforementioned situation when a node gets cordoned and then deleted after awhile. PDBs watch pods by labelSelectors and if there is no change in pods number, PDB statutes and expectations aren't updated.

Some explanations what happens if a node gets cordoned and then deleted are provided below:
1. We have 3 eligible nodes for Ingress Nginx, so Kruise controller creates 3 pods and set `.spec.replicas` of the daemonset to 3, PDB `expectedPods` is also set to 3, allowing disruption for no more than one pod (in case it has MaxUnavailable set to 1).
2. We drain one of the nodes, Kruise controller doesn't recreate a pod for the cordoned node but `.spec.replicas` is still set to 3. PDB udpates its status to healthy = 2, expected = 3, so no more disruptions are allowed.
3. At this point some extra reconcile logic for the affected daemonset is kicking in. Kruise controller will be checking the daemonset every 60 seconds, reevaluating the number of eligible nodes to make sure `.spec.replicas` is up to date.
4. If the cordoned node gets finally deleted, Kruise controller sets `.spec.replicas` to 2, but PDB won't update its status because the number and statuses of the two other pods haven't changed. So, kruise controller updates the PDB's status subresource, triggering its recalculation, which results in having PDB set expectedPods to 2, allowing further disruptions.

Without the periodic syncing, mentioned in p.3, Kruise Controller wouldn't change `.spec.replicas` after the node's removal because it doesn't watch nodes, only daemonsets.

This PR adds `scale` subresource to open kruise `Advanced DaemonSets` (ADS) making it possible to apply a `PodDisruptionBudget` (PDB) to the daemonsets' pods.
Having PDBs for ADS in place allows better control over how ADSs are updated.
Closes #6244 

As of now, PDB can't calculate its status and apply limitations to Ingress Nginx controller's pods.
After the aforementioned changes It is possible to apply a PodDisruptionBudget to Ingress Nginx controller's pods.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: feature
summary: Add 
impact: Open Kruise controller will restart.
impact_level: low
```


deckhouse-BOaTswain:

yalosev (approved):

RomanenkoDenys (approved):
Looks good, but one question.
Do we create upstream PR ?

miklezzzz:
@RomanenkoDenys, the upstream hadn't been really collaborative with a much smaller patch last time, so I decided it would be in vain.

PR #6300, opened 2023-10-24
[monitoring-kubernetes] Fix dashboards breakage caused by lens-hack

meatuses:
## Description
Filter out some duplicated metrics in our grafana dashboards that are created if [lens-hack](https://deckhouse.io/documentation/v1/modules/300-prometheus/faq.html#how-do-i-get-access-to-prometheus-metrics-from-lens) is applied. Filtering is implemented by checking that `pod` tag equals `null` for impacted metrics

If the [lens-hack](https://deckhouse.io/documentation/v1/modules/300-prometheus/faq.html#how-do-i-get-access-to-prometheus-metrics-from-lens) is applied in a cluster, duplicates of some metrics are created. `node` and `nodes` dashboards using these metrics break, showing double amount of RAM or storage or in some other way.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Minor grafana dashboards fix.
impact_level: low
```

### DIFF
--- a/modules/340-monitoring-kubernetes/monitoring/grafana-dashboards/kubernetes-cluster/nodes/node.json
+++ b/modules/340-monitoring-kubernetes/monitoring/grafana-dashboards/kubernetes-cluster/nodes/node.json
@@ -323,7 +323,7 @@
       "pluginVersion": "7.4.2",
       "targets": [
         {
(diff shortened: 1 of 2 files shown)


nabokihms:
Dashboards are okay. The problem is with this hack. Probably deeply, within lens.
This should be solved by adding a feature to lens to handle custom queries / labels. See for more details: https://github.com/lensapp/lens/issues/3955

nabokihms:
@meatuses Lens closes its sources now and slowly moving from a community project to the enterprise proprietary side 
https://github.com/MuhammedKalkan/OpenLens/issues/188

It is probably worth removing the integration from our doc.

PR #6435, opened 2023-11-03
Added AWS_REGION environment variable for ClientServer mode

trublast:
- Updated trivy-operator to v1.16.4;
- Added AWS_REGION environment variable for ClientServer mode;
- Update registry secret patch to affect only standalone scan jobs.

Trivy can't check images from AWS ECR.

Trviy operator is of v1.16.4 version.
Trivy can check images from AWS ECR.
Trivy operator's scanjobs can successfully download trivy-db both in Standalone and in ClientServer modes.

## Checklist
- [ ] e2e tests passed.


miklezzzz:
@trublast Hi, after some pondering over what exactly `001-add-registry-secret-as-dockerconfigjson.patch` was supposed to do, I've come to a conclusion that we did it wrong - patching the `ClientServer` pod specs when `ClientServer` mode doesn't imply downloading trivy-db at all doesn't make any sense. Thus, I've added another commit with a patch to your patch to the patch :-D. Checked in a dev cluster and in general it should be fine.
@yalosev Could you kindly somehow review the changes?

miklezzzz (approved):

yalosev (approved):

PR #6579, opened 2023-11-17
[node-local-dns] Remove hostPorts when hostNetwork=false

asviel:
## Description

Remove `hostPorts` from manifest when `hostNetwork=false`.

We don't use `hostNetwork` if `cni-cilium` is used and use it otherwise. If you migrate, e.g. from `cni-flannel` to `cni-cilium`, the `hostPorts` added by the controller when `hostNetwork` was true will remain in the DaemonSet manifest. In some cases, this can break resolving on nodes in the cluster because the `hostPort` bind port 53 on the host and then `systemd-resolved` (which works on port 53) stops working.
The problem described here https://github.com/kubernetes/kubernetes/issues/117689

We don't

Hook MUST remove `hostPorts` from the DaemonSet manifest if `hostNetwork=false`.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-local-dns
type: feature
summary: Remove `hostPorts` from manifest when `hostNetwork=false`.
impact_level: default
```

### DIFF
--- a/ee/modules/350-node-local-dns/hooks/cleanup_daemonset_hostports.go
+++ b/ee/modules/350-node-local-dns/hooks/cleanup_daemonset_hostports.go
@@ -14,13 +14,14 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/deckhouse/deckhouse/go_lib/dependency"
 	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
 	"github.com/flant/addon-operator/sdk"
 	"github.com/flant/shell-operator/pkg/kube_events_manager/types"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
Aye, aye, @asviel. I've started the workflow for label 'e2e/run/aws'...

deckhouse-BOaTswain:
Aye, aye, @asviel. I've started the workflow for label 'e2e/run/gcp'...

ee/modules/350-node-local-dns/hooks/cleanup_daemonset_hostports.go
```diff
@@ -0,0 +1,93 @@
...
+	return false, nil
+}
+
+func handleHostNetworkFalseWithHostPorts(input *go_hook.HookInput, dc dependency.Container) error {
+	snap, ok := input.Snapshots["daemonset"]
+
+	if !ok {
+		return nil
+	}
+
```
RomanenkoDenys, line 69:
```suggestion
if len(nodes) == 0 {
		return nil
	}

```

RomanenkoDenys (approved):

PR #6617, opened 2023-11-21
[external-module-manager] Store checksums in memory

nabokihms:
## Description
* Reconcile all modules sources on startup. This affects client's registry load, but is more error pron.
* Always deploy the module source for deckhouse modules (since as of today dhctl mirror command copies the module source to private registries by default).

In the module values, checksum map looks like
```
externalModuleManager:
  internal:
    modulesChecksums:
      inspektor-gadget:
        inspektor-gadget: sha256:155628253a07fd119c179ec175658b7ce5012c5bfa9512b28323e2f24d9228b8
```
an it is dropped on restarts.

The first fix is applied to survive the situation when dechkouse the problem that module was downloaded but there is no module release in the cluster. The assumption is that the situation occurs when deckhouse is restarted in the middle of downloading the module, or hook simply fails.  With the current approaches, modules will be downloaded on deckhouse restart and checksums will only be saved if the hook executed successfully. 

The second fix is applied to make default module sources work in air-gapped environments. 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feat
summary: Store checksums in memory.
```

### DIFF
--- a/modules/005-external-module-manager/hooks/create_deckhouse_modules_source.go
+++ b/modules/005-external-module-manager/hooks/create_deckhouse_modules_source.go
@@ -79,17 +79,6 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 }, createDeckhouseModuleSource)
 
 func createDeckhouseModuleSource(input *go_hook.HookInput) error {
-	if input.Values.Get("global.modulesImages.registry.address").String() != "registry.deckhouse.io" {
-		// For now, modules are only stored in the base deckhouse registry,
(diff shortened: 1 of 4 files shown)


deckhouse-BOaTswain:

deckhouse-BOaTswain:
Cherry pick PR [6626](https://github.com/deckhouse/deckhouse/pull/6626) to the branch [release-1.54](https://github.com/deckhouse/deckhouse/tree/release-1.54) successful!

PR #6629, opened 2023-11-22
[dhctl] Validate passed credentials against registry prior to mirroring

mvasl:
## Description

`dhctl mirror` will now validate passed credentials such as Deckhouse license tokens and registry logins/passwords prior to mirroring

We got a feedback about dhctl taking long time to signal about bad registry credentials after push to the third-party registry is started due to the tar bundle unpack procedure taking a few minutes. Only after dhctl is done with it and started pushing, it will find out about bad creds and fail.

## Manual testing

https://github.com/deckhouse/deckhouse/pull/6629#issuecomment-1822893839

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Validate passed credentials against registry prior to mirroring
impact_level: default
```

### DIFF
--- a/dhctl/pkg/operations/mirror/auth.go
+++ b/dhctl/pkg/operations/mirror/auth.go
@@ -1,0 +1,78 @@
+// Copyright 2023 Flant JSC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mirror
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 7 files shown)


mvasl:
Pull, valid license token
Снимок экрана 2023-11-22 в 17 36 43

Pull, bad token
Снимок экрана 2023-11-22 в 17 29 50


Push, valid credentials
Снимок экрана 2023-11-22 в 17 26 46


Push invalid credentials
Снимок экрана 2023-11-22 в 17 29 12

dhctl/pkg/operations/mirror/auth.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err = remote.Write(ref, empty.Image, remoteOpts...); err != nil {
+		return err
+	}
+
+	if err = remote.Delete(ref, remoteOpts...); err != nil {
+		return fmt.Errorf("Could not clean up after write-testing registry permissions: %w", err)
```
name212, line 58: we need to output full path with tag which we created for manual clean up

name212 (changes requested):
one comment

dhctl/pkg/operations/mirror/auth.go
mvasl: Done

mvasl: Pushed some fixes, should be fine now

mvasl: Rewritten that to use small synthetic image instead of empty one and not to remove the image at all. Turns out that some registries do not allow images removal using registry api.

name212 (approved):

name212 (approved):

PR #6649, opened 2023-11-24
[Multitenancy manager] release migration

miklezzzz:
## Description
Add `migration_release.go` hook that cycles through multitenancy projects and annotates the projects` namespaces with helm annotation "helm.sh/resource-policy: keep" so that the namespaces aren't got deleted on next helm upgrade. Also, "meta.helm.sh/release-name" and "meta.helm.sh/release-namespace" annotations are updated so the namespaces are adopted by new per-project helm releases.

`project.go` hook is also set to run `OnAfterHelm`, otherwise we need to manually restart deckhouse for getting multitenancy projects fully recreated after deckhouse update. (On the first run of `project.go` there is an error as some project resources aren't deleted yet and helm fails to adopt them).

This PR adds migration hook without which projects' namespaces are deleted on the `deckhouse` update from 1.53 to 1.54. The deletion happens because in 1.53 projects's objects are stored in `d8-multitenancy-manager` release in `d8-system` namespace, while in 1.54 they are stored in dedicated helm releases in `d8-multitenancy-manager` namespace. Thus, we need to prevent the namespaces from deletion and prepare them for adoption. Other projects` objects (as ResourceQuota, LimitRanges, NetworkPolicies, etc) are deleted and recreated in the course of deckhouse update.

It prevents deletion of projects' namespaces.

It is expected to have all projects' namespaces in place after deckhouse update. All customers objects should be also in place. Such auxiliary objects as ResourceQuota, LimitRanges, NetworkPolicies are recreated.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: fix
summary: fix projects' namespaces deletion during deckhouse update from 1.53 to 1.54
impact_level: default
```


deckhouse-BOaTswain:

Dmitrykob (approved):

PR #6728, opened 2023-11-30
[ingress-nginx] add ingress controller deprecated version check

sh1kel:
## Description
Add version check to ingress controller hook.
Add alert if version is going to be deprecated.

In case if ingress controller version will be deprecated, it may be loosing of backward compatibility. 

User get prometheus alert if ingress controller version is 1.1

## Checklist
- [ ] e2e tests passed.

## Changelog entries
Add alerts for deprecating ingress-nginx module

```changes
section: ingress-nginx
type: chore
summary: alerts for deprecating ingress-nginx module
```

### DIFF
--- a/modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
+++ b/modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
@@ -43,10 +43,9 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 }, discoverMinimalNginxVersion)
 
 const (
-	minVersionValuesKey         = "ingressNginx:minimalControllerVersion"
-	incompatibleVersionsKey     = "ingressNginx:hasIncompatibleIngressClass"
-	disruptionKey               = "ingressNginx:hasDisruption"
-	deprecatedControllerVersion = "1.1"
+	minVersionValuesKey     = "ingressNginx:minimalControllerVersion"
+	incompatibleVersionsKey = "ingressNginx:hasIncompatibleIngressClass"
+	disruptionKey           = "ingressNginx:hasDisruption"
 )
 
 func applySpecControllerFilter(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
@@ -88,12 +87,6 @@ func discoverMinimalNginxVersion(input *go_hook.HookInput) error {
 			if ctrl.Version == "0.33" {
 				isDisruptionUpdate = true
 			}
-		}
-		if ctrl.Version == deprecatedControllerVersion {
-			input.MetricsCollector.Set("deprecated_ingress_controller", 1.0, map[string]string{
-				"ingress_class":   ctrl.IngressClass,
-				"ingress_version": ctrl.Version,
-			})
 		}
 		ctrlVersion, err := semver.NewVersion(ctrl.Version)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
```diff
@@ -88,6 +89,12 @@ func discoverMinimalNginxVersion(input *go_hook.HookInput) error {
 				isDisruptionUpdate = true
 			}
 		}
+		if ctrl.Version == deprecatedControllerVersion {
```
miklezzzz, line 92: it makes sense to add a couple of tests to check that relevant metrics get successfully created/deleted.

miklezzzz (changes requested):

modules/402-ingress-nginx/hooks/minimal_ingress_nginx_version.go
miklezzzz: and, as I remember, we already have a metric with controller version 
image. Could it be better just to add an alert checking if `controller_version` labels equals `1.1`?

modules/402-ingress-nginx/monitoring/prometheus-rules/ingress-nginx.yaml
```diff
@@ -169,3 +169,15 @@
         3. If the `Number of Nodes Scheduled with Up-to-date Pods` parameter does not match
         `Current Number of Nodes Scheduled`, check the pertinent Ingress Nginx Controller's 'nodeSelector' and 'toleration' settings,
         and compare them to the relevant nodes' 'labels' and 'taints' settings
+  - alert: NginxIngressDeprecatedVersion
+    expr: d8_ingress_nginx_controller{controller_version="1.1"}
```
miklezzzz, line 173: could you please update the expression to something like:
`count(d8_ingress_nginx_controller{controller_version="1.1"}) > 0`, otherwise we risk end up having more than one alert in case there are several ingress nginx controllers of `1.1` in the cluster.
image

modules/402-ingress-nginx/monitoring/prometheus-rules/ingress-nginx.yaml
```diff
@@ -169,3 +169,15 @@
...
+    labels:
+      severity_level: "4"
+    annotations:
+      plk_markup_format: "markdown"
+      plk_protocol_version: "1"
+      plk_ignore_labels: "name"
+      description: |
+        Ingress version 1.1 will be disabled in Deckhouse release 1.57
```
miklezzzz, line 181: Usually, we provide brief description in summary, like: There is an Ingress nginx controller of a deprecated version in the cluster; and some extra information in description, like: 
There is an Ingress nginx controller of 1.1 version in the cluster. Ingress nginx verion 1.1 support will be disabled in upcoming Deckhouse releases. Consider upgrading to a newer version of Ingress nginx controller (all the pods of the affected Ingress Nginx controllers will be consequently recreated during upgrade).
Use the following command to list the affected Ingress nginx controllers: 
`kubectl  get ingressnginxcontrollers.deckhouse.io -o json | jq '.items[] | select(.spec.controllerVersion == "1.1") | .metadata.name'`

miklezzzz:
@sh1kel swap description and summary please, summary - concise, description - wordy :)

miklezzzz (approved):

yalosev (approved):

Issue #4950, opened 2023-06-16
[operator-trivy] Add CA to operator scan jobs


alex123012:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

In the isolated environments with private container registry with self-signed CA operator trivy can't run scan jobs because it can't pull trivy-db:
```text
2023-06-16T06:19:56.708Z        INFO    Need to update DB
2023-06-16T06:19:56.708Z        INFO    DB Repository: custom.registry.com/deckhouse/security/trivy-db
2023-06-16T06:19:56.708Z        INFO    Downloading DB...
2023-06-16T06:19:56.742Z        FATAL   init error: DB error: failed to download vulnerability DB: database download error: OCI repository error: 1 error occurred:
        * Get "https://custom.registry.com/v2/": x509: certificate signed by unknown authority
```

### Proposed Solution

1. The first option is to add `--insecure` flag for http registry [in this func](https://github.com/aquasecurity/trivy-operator/blob/main/pkg/plugins/trivy/plugin.go#L1096) for trivy command (refer to `trivy --help`)
1. The second option is to add [custom CA to trusted certificates](https://stackoverflow.com/a/67232164) in trivy alpine image (for example in `initContainers` or in entrypoint in [operator-trivy deployment](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/500-operator-trivy/templates/deployment.yaml#L31)).
1. The third option is to add support for trivy-db download with custom CA in [trivy](https://github.com/aquasecurity/trivy) and then add customization of scan jobs in [trivy-operator](https://github.com/aquasecurity/trivy-operator) (preferable option)

yalosev:
Well, Have to try this in the reverse order - 3 -> 2 -> 1

PR #6747, opened 2023-12-02
[operator-trivy] Add CA to operator scan jobs #4950

libmonsoon-dev:
## Description

Added support for installing custom CA in trivy and trivy-operator

Close https://github.com/deckhouse/deckhouse/issues/4950

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: operator-trivy
type: fix
summary: Fixed trivy working with private container registry with self-signed CA
impact_level: default
```


libmonsoon-dev:
Without CA:
Screenshot_20231204_232120
With CA:
Screenshot_20231204_231229

ee/modules/500-operator-trivy/templates/configmap.yaml
```diff
@@ -60,3 +60,4 @@ data:
   TRIVY_DEBUG: "false"
   TRIVY_SKIP_DB_UPDATE: "false"
   TRIVY_DB_REPOSITORY: {{ $dbRepository }}
+  TRIVY_REGISTRY_CA: | {{ $.Values.global.modulesImages.registry.CA | nindent 4 }}
```
yalosev, line 63: Didn't get this. What for do we need this?

ee/modules/500-operator-trivy/templates/deployment.yaml
```diff
@@ -79,6 +79,8 @@ spec:
             value: "false"
           - name: OPERATOR_SCAN_JOB_TIMEOUT
             value: "5m"
+          - name: OPERATOR_SCAN_JOB_REGISTRY_CA
+            value: | {{ $.Values.global.modulesImages.registry.CA | nindent 14 }}
```
miklezzzz, line 83: idk why, but I don't have this property set in my cluster even if I switch to self-signed issuer :thinking: 
image
any ideas how it came about?
Moreover, I'm afraid of a hypothetical situation when a client might use several self-signed certificates for different parts of their infrastructure, so that the CA certificate of deckhouse cluster doesn't equal to their registry CA (an external registry like a nexus instance).
Could it be better to provide a way of specifying any custom CA via the module's properties, falling back to default value from `.global.modulesImages.registry.CA` if it isn't set?

ee/modules/500-operator-trivy/templates/configmap.yaml
libmonsoon-dev: In addition to the scan job, which needs CA to download the scanned image, trivy server also needs CA to download the vulnerability database

yalosev, line 63:
```suggestion
  {{- if $.Values.global.modulesImages.registry.CA }}
  TRIVY_REGISTRY_CA: | {{ $.Values.global.modulesImages.registry.CA | nindent 4 }}
  {{- end }}
```

miklezzzz (approved):

yalosev (approved):

PR #6776, opened 2023-12-06
[linstor] Enabling the Linstor module is prohibited when the SDS-DRBD module is enabled

duckhawk:
## Description

Enabling the Linstor module is prohibited when the SDS-DRBD module is enabled.

Simultaneously enabling the Linstor and SDS-DRBD modules is not possible, as it can lead to data corruption.

Simultaneously enabling the Linstor and SDS-DRBD modules is prohibited.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: linstor
type: chore
summary: Simultaneously enabling the Linstor and SDS-DRBD modules is prohibited.
```

### DIFF
--- a/modules/041-linstor/enabled
+++ b/modules/041-linstor/enabled
@@ -17,11 +17,11 @@
 source /deckhouse/shell_lib.sh
 
 function __main__() {
-    if values::array_has global.enabledModules "sds-drbd" ; then
-      echo "false" > $MODULE_ENABLED_RESULT
-    else
-      echo "true" > $MODULE_ENABLED_RESULT
-    fi
+  if values::array_has global.enabledModules "sds-drbd" ; then
+    echo "false" > "$MODULE_ENABLED_RESULT"
+  else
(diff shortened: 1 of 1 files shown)


modules/041-linstor/enabled
```diff
@@ -17,7 +17,11 @@
 source /deckhouse/shell_lib.sh
 
 function __main__() {
-  echo "true" > "$MODULE_ENABLED_RESULT"
+    if values::array_has global.enabledModules "sds-drbd" ; then
+      echo "false" > $MODULE_ENABLED_RESULT
+    else
+      echo "true" > $MODULE_ENABLED_RESULT
+    fi
```
apolovov, line 24:
```suggestion
  if values::array_has global.enabledModules "sds-drbd" ; then
    echo "false" > "$MODULE_ENABLED_RESULT"
  else
    echo "true" > "$MODULE_ENABLED_RESULT"
  fi
```

apolovov (approved):

AleksZimin (approved):

PR #6811, opened 2023-12-07
Preventing the creation of a new "helm" release when no changes are made

Dmitrykob:
## Description

What we've got:
When we make changes to any project or projectTemplate, we make a call to Helm.Update for all projects. This leads to unnecessary increase of Helm release version.

What we're getting at:
We compute an md5 hash based on templates and template values and store this hash in the release custom label.   When changes are made, we compare the hash and prevent helm.Update from being called.

It prevents the creation of a new "helm" release when no changes are made to it.

When changes are made to a project or project template helm.update will only be called for projects that have changes or for new projects.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: fix
summary: Prevents the creation of a new "helm" release when no changes are made to it
impact_level: default
```

### DIFF
--- a/ee/modules/160-multitenancy-manager/templates/user-resources/user-resources-templates.yaml
+++ b/ee/modules/160-multitenancy-manager/templates/user-resources/user-resources-templates.yaml
@@ -71,7 +71,7 @@ metadata:
 apiVersion: deckhouse.io/v1alpha1
 kind: AuthorizationRule
 metadata:
-  name: {{ list $kebabCaseAuthName $rootContext | include "slugify" | replace "--" "-" }}
+  name: {{ list $kebabCaseAuthName $rootContext | replace "--" "-" }}
   namespace: {{ $projectName }}
 spec:
   accessLevel: {{ $values.role }}
--- a/ee/modules/160-multitenancy-manager/templates/user-resources/user-resources-templates.yaml
+++ b/ee/modules/160-multitenancy-manager/templates/user-resources/user-resources-templates.yaml
@@ -11,6 +11,32 @@ metadata:
   {{- end }}
 {{- end }}
 
+{{- define "slugify" }}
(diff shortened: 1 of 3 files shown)


ee/modules/160-multitenancy-manager/templates/user-resources/user-resources-templates.yaml
```diff
@@ -11,6 +11,32 @@ metadata:
   {{- end }}
 {{- end }}
 
+{{- define "slugify" }}
```
yalosev, line 14: Seems redundant

go_lib/dependency/helm/helm.go
```diff
@@ -213,3 +227,21 @@ func getActionConfig(namespace string) (*action.Configuration, error) {
...
+func getMD5Hash(templates, values map[string]interface{}) string {
+	sum := make(map[string]interface{})
+	maps.Copy(sum, templates)
+	for k, v := range values {
+		sum[k] = v
+	}
+
+	// TODO is not an ideal algorithm, since the order of elements in map is not
+	// guaranteed and may lead to a different hash result.
+	byteResult, err := json.Marshal(sum)
```
yalosev, line 240: there could not be an error here

go_lib/dependency/helm/helm.go
```diff
@@ -213,3 +227,21 @@ func getActionConfig(namespace string) (*action.Configuration, error) {
 	}
 	return actionConfig, nil
 }
+
+func getMD5Hash(templates, values map[string]interface{}) string {
```
yalosev, line 231: well, on the large maps it will always generate a new hash. Probably we need something like this:
```
keys := make([]string, len(m))
	i := 0
	for k := range values {
		keys[i] = k
		i++
	}

    sort.Strings(keys)
    for _, k := range keys {
		v := values[k]
		b := sha256.Sum256([]byte(fmt.Sprintf("%v", k)))
		h.Write(b[:])
		b = sha256.Sum256([]byte(fmt.Sprintf("%v", v)))
		h.Write(b[:])
	}
```

yalosev (approved):

PR #6860, opened 2023-12-13
[candi][bb] Handle registry packages fetch errors

asviel:
## Description
Handle registry packages fetch errors.

`curl` writes file as it is loading. If the download fails, we will get corrupted file. Since we don't currently have the ability to validate packages checksum (and we don't do it), we will get an error at the unpacking stage:
```
bashible.sh[1088426]: bb-rp [INFO] Unpacking package 'kubectl'
bashible.sh[1088426]: gzip: stdin: unexpected end of file
bashible.sh[1088426]: tar: Unexpected EOF in archive
bashible.sh[1088426]: tar: Unexpected EOF in archive
bashible.sh[1088426]: tar: Error is not recoverable: exiting now
bashible.sh[1086700]: Failed to execute step ...
```

Since we don't handle this error, it will start to fill up disk space 'cause the original corrupted blob will not be refetched (there is no such logic) and the temp directory is not deleted (they do fill up disk space). 

It affects customers.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: Handle registry packages fetch errors.
impact_level: default
```

### DIFF
--- a/candi/bashible/bashbooster/50_deckhouse_registrypackages.sh
+++ b/candi/bashible/bashbooster/50_deckhouse_registrypackages.sh
@@ -20,7 +20,7 @@ bb-var BB_RP_FETCHED_PACKAGES_STORE "${TMPDIR}/registrypackages"
 # Use d8-curl if installed, fallback to system package if not
 bb-rp-curl() {
   if command -v d8-curl > /dev/null ; then
-    d8-curl --parallel "$@"
+    d8-curl --remove-on-error --parallel "$@"
   else
     curl "$@"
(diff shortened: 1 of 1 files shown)


name212 (approved):

deckhouse-BOaTswain:
Cherry pick PR [6870](https://github.com/deckhouse/deckhouse/pull/6870) to the branch [release-1.55](https://github.com/deckhouse/deckhouse/tree/release-1.55) successful!

PR #6964, opened 2023-12-20
[user-authn] OIDC allow unverified email

### LINKED ISSUE #6696: insecureSkipEmailVerified for DexAuthenticator
### Preflight Checklist

It is currently not possible to configure oauth-proxy with `--insecure-oidc-allow-unverified-email`.
As a result, you cannot log into Deckhouse modules such as grafana or dashboard if your user has unverified email.
So you have to manually path the deployment, but this will probably work only until the next module update.

### Proposed Solution

As it is already done for the DexProvider with the `spec.oidc.insecureSkipEmailVerified` parameter, the same can be done for the DexAuthenticator with the `spec.insecureSkipEmailVerified` parameter.

### Additional Information

The args of the oauth-proxy container can be changed to something like:
`- --insecure-oidc-allow-unverified-email={{ $crd.spec.insecureSkipEmailVerified }}`
https://github.com/deckhouse/deckhouse/blob/1c4dd2b160314247ca510583d87211f06ccbf0c7/modules/150-user-authn/templates/dex-authenticator/deployment.yaml#L123-L154

Then the modules DexAuthentificator should also be changed.
For example, the prometheus module:
`applicationIngressClassName: {{ .Values.prometheus.auth.insecureSkipEmailVerified }}`
https://github.com/deckhouse/deckhouse/blob/1c4dd2b160314247ca510583d87211f06ccbf0c7/modules/300-prometheus/templates/grafana/authenticator.yaml#L9-L19

qw1mb0:
## Description
Fix #6696 

It is currently not possible to configure oauth-proxy with --insecure-oidc-allow-unverified-email.
As a result, you cannot log into Deckhouse modules such as grafana or dashboard if your user has unverified email.
So you have to manually path the deployment, but this will probably work only until the next module update.

We have clients who are forced to use such users in OIDC.

A user whose email has not been verified can log into grafana.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
OIDC allow unverified email

```changes
section: user-authn
type: fix
summary: OIDC allow unverified email
impact_level: defaut
```

### DIFF
--- a/modules/150-user-authn/templates/dex-authenticator/deployment.yaml
+++ b/modules/150-user-authn/templates/dex-authenticator/deployment.yaml
@@ -152,6 +152,7 @@ spec:
         - --redis-connection-url=redis://127.0.0.1/
         - --cookie-refresh={{ $context.Values.userAuthn.idTokenTTL | default "10m" }}
         - --cookie-expire={{ $crd.spec.keepUsersLoggedInFor | default "168h" }} # 7 days
+        - --insecure-oidc-allow-unverified-email=true
         env:
         - name: OAUTH2_PROXY_CLIENT_SECRET
           valueFrom:


deckhouse-BOaTswain:
Cherry pick PR [6969](https://github.com/deckhouse/deckhouse/pull/6969) to the branch [release-1.56](https://github.com/deckhouse/deckhouse/tree/release-1.56) successful!

PR #6996, opened 2023-12-22
[node-manager] Change example NodeUser UID to avoid conflicts

meatuses:
## Description

Change default NodeUser UID to 1100 instead of 1001 in our example and the CRD description.

`NodeUser` `UID` is 1001 in our examples, which matches UID of first manually created user on a Linux machine. We should change this to avoid possible conflicts and confusion.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: chore
summary: Minor docs changes
impact_level: low
```


modules/040-node-manager/crds/nodeuser.yaml
```diff
@@ -44,8 +44,10 @@ spec:
                   description: |
                     Node user ID.
 
+                    We recommend using values `>= 1100` to avoid possible conflicts with manually created users.
```
z9r5, line 47:
```suggestion
                    We recommend using the values `>= 1100` to avoid conflicts with manually created users.
```

z9r5 (approved):

PR #7042, opened 2023-12-27
[prometheus] Set .spec.externalURL in alermanager manifest

### LINKED ISSUE #7041: [] allow the externalUrl field to be set in CustomAlertmanager
### Preflight Checklist

spec.internal does not have an externalUrl field that can be used to override the url in cases where the cluster url does not match the url to which requests should be made.

### Proposed Solution

add externalUrl field to https://github.com/deckhouse/deckhouse/blob/main/modules/300-prometheus/templates/alertmanager/internal/alertmanager.yaml

### Additional Information

[11:12:11][ssh][a.gavrilov@....-master-0] ~ $ k get crd alertmanagers.monitoring.coreos.com -oyaml | grep externalUrl -n10
2898-                      type: array
2899-                    workingDir:
2900-                      description: Container's working directory. If not specified,
2901-                        the container runtime's default will be used, which might
2902-                        be configured in the container image. Cannot be updated.
2903-                      type: string
2904-                  required:
2905-                  - name
2906-                  type: object
2907-                type: array
2908:              externalUrl:
2909-                description: The external URL the Alertmanager instances will be available
2910-                  under. This is necessary to generate correct URLs. This is necessary
2911-                  if Alertmanager is not served from root of a DNS name.
2912-                type: string
2913-              forceEnableClusterMode:
2914-                description: ForceEnableClusterMode ensures Alertmanager does not
2915-                  deactivate the cluster mode when running with a single replica.
2916-                  Use case is e.g. spanning an Alertmanager cluster across Kubernetes
2917-                  clusters with a single replica in each.
2918-                type: boolean

verdel:
## Description

When a public domain is configured the `.spec.externalURL` parameter must be set in the `Alertmanager` manifest [similar](https://github.com/deckhouse/deckhouse/blob/fdc2a2bc3072f5acae424b85eeacd053ec5eb81e/modules/300-prometheus/templates/prometheus/prometheus.yaml#L271-L273) to the Prometheus manifest

This PR fix part of problem with internal alertmanager in #7041 issue

Currently the `.spec.externalURL` parameter is absent, and as a result, in notifications where `externalURL` is used in templates, the link contains the URL of the internal k8s service

This is a minor fix, so there's no need to wait for a minor release

When a public domain is specified the `.spec.externalURL` parameter in the Alertmanager manifest should appear with the corresponding value

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: prometheus
type: chore
summary: Set .spec.externalURL in alermanager manifest when a public domain is specified

### DIFF
--- a/modules/300-prometheus/templates/alertmanager/internal/alertmanager.yaml
+++ b/modules/300-prometheus/templates/alertmanager/internal/alertmanager.yaml
@@ -71,6 +71,9 @@ spec:
             alertmanager: {{ .name }}
         topologyKey: kubernetes.io/hostname
   {{- end }}
+{{- if $.Values.global.modules.publicDomainTemplate }}
+  externalUrl: {{ include "helm_lib_module_uri_scheme" $ }}://{{ include "helm_lib_module_public_domain" (list $ "grafana") }}/alertmanager/{{ .name }}
+{{- end }}
   {{- include "helm_lib_module_pod_security_context_run_as_user_deckhouse" $ | nindent 2 }}
   {{- include "helm_lib_node_selector" (tuple $ "monitoring") | nindent 2 }}
   {{- include "helm_lib_tolerations" (tuple $ "monitoring") | nindent 2 }}

### CODE REVIEW

File: modules/300-prometheus/templates/alertmanager/internal/alertmanager.yaml [hunk 1]
```diff
@@ -71,6 +71,9 @@ spec:
             alertmanager: {{ .name }}
         topologyKey: kubernetes.io/hostname
   {{- end }}
+{{- if .Values.global.modules.publicDomainTemplate }}
+  externalUrl: {{ include "helm_lib_module_uri_scheme" . }}://{{ include "helm_lib_module_public_domain" (list . "grafana") }}/alertmanager/monitoring
```
nabokihms, line 75:
```suggestion
  externalUrl: {{ include "helm_lib_module_uri_scheme" . }}://{{ include "helm_lib_module_public_domain" (list . "grafana") }}/alertmanager/{{ .name }}
```

nabokihms (changes requested):
@verdel, thank you for the PR!

The path for the alert manager contains the up manager name, so instead of `monitoring,` it should end with the name.

modules/300-prometheus/templates/alertmanager/internal/alertmanager.yaml
verdel: Thank you very much. I didn't check the path in the Ingress manifest. I've fixed everything now.

nabokihms (approved):

modules/300-prometheus/templates/alertmanager/internal/alertmanager.yaml
```diff
@@ -71,6 +71,9 @@ spec:
             alertmanager: {{ .name }}
         topologyKey: kubernetes.io/hostname
   {{- end }}
+{{- if .Values.global.modules.publicDomainTemplate }}
+  externalUrl: {{ include "helm_lib_module_uri_scheme" . }}://{{ include "helm_lib_module_public_domain" (list . "grafana") }}/alertmanager/{{ .name }}
+{{- end }}
```
nabokihms, line 76:
```suggestion
{{- if $.Values.global.modules.publicDomainTemplate }}
  externalUrl: {{ include "helm_lib_module_uri_scheme" $ }}://{{ include "helm_lib_module_public_domain" (list . "grafana") }}/alertmanager/{{ .name }}
{{- end }}
```

Tests failed because of the scope of the variables. To fix this, it is possible to use the global context to access values.

nabokihms (changes requested):
One more thing.

modules/300-prometheus/templates/alertmanager/internal/alertmanager.yaml
verdel: Certainly, I didn't notice that the scope narrows down above. Thank you.

deckhouse-BOaTswain:

nabokihms:
The e2e test will not pass because of permission violations that do not related with the PR.

nabokihms (approved):
@verdel thank you for your contribution!

nabokihms:
/e2e/abort aws;WithoutNAT;containerd;1.25 7411690783 7411690783-1-con-1-25 /sys/deckhouse-oss/install:pr7042 centos@3.75.81.216

deckhouse-BOaTswain:
Aye, aye, @nabokihms. I've recognized your '/e2e/abort' command and started the workflow...

PR #7086, opened 2024-01-10
[linstor] add validating webhook to disable module

duckhawk:
## Description

Add a validating webhook to prevent the linstor module from being enabled.

The linstor module should not be allowed to be enabled because we are moving to sds-drbd module.

It will not be possible to enable the module if it has not been enabled before, alas to change module config.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: linstor
type: chore
summary: Add a validating webhook to prevent the linstor module from being enabled.
impact: The `linstor` module is deprecated. Please switch to [sds-drbd](https://deckhouse.io/modules/sds-drbd/stable/) module ASAP. The `linstor` module cannot be enabled but will continue to work if it was already enabled before.
impact_level: high
```

### DIFF
--- a/modules/002-deckhouse/webhooks/validating/disable_virtualization_module
+++ b/modules/002-deckhouse/webhooks/validating/disable_virtualization_module
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+# Copyright 2023 Flant JSC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# TODO: You should remove this hook as soon as the linstor module is removed from deckhouse.
+
+source /shell_lib.sh
+
+function __config__(){
+  cat <<EOF
+configVersion: v1
+kubernetesValidating:
+- name: disable-linstor-module.deckhouse.io
+  group: main
+  rules:
+  - apiGroups:   ["deckhouse.io"]
+    apiVersions: ["*"]
+    operations:  ["CREATE", "UPDATE"]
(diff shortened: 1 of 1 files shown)


modules/002-deckhouse/webhooks/validating/disable_virtualization_module
```diff
@@ -0,0 +1,52 @@
...
+function __main__() {
+  mcName=$(context::jq -r '.review.request.object.metadata.name')
+  if [[ "$mcName" == "linstor" ]]; then
+    linstorModuleEnabled="$(context::jq -r '.review.request.object.spec.enabled')"
+    if [[ "$linstorModuleEnabled" == "true" ]]; then
+      cat <<EOF > "$VALIDATING_RESPONSE_PATH"
+{"allowed":false, "message":"moduleconfigs.deckhouse.io \"linstor\" is deprecated, please use sds-drbd module instead (https://deckhouse.io/modules/)." }
+EOF
+    return 0
+    fi
```
apolovov, line 45:
```suggestion
    if context::jq -e '.review.request.object.spec.enabled' >/dev/null; then
      cat <<EOF > "$VALIDATING_RESPONSE_PATH"
{"allowed":false, "message":"moduleconfigs.deckhouse.io \"linstor\" is deprecated, please use sds-drbd module instead (https://deckhouse.io/modules/)." }
EOF
    return 0
    fi
```

modules/002-deckhouse/webhooks/validating/disable_virtualization_module
```diff
@@ -0,0 +1,52 @@
...
+    resources:   ["moduleconfigs"]
+    scope:       "Cluster"
+EOF
+}
+
+function __main__() {
+  mcName=$(context::jq -r '.review.request.object.metadata.name')
+  if [[ "$mcName" == "linstor" ]]; then
+    linstorModuleEnabled="$(context::jq -r '.review.request.object.spec.enabled')"
+    if [[ "$linstorModuleEnabled" == "true" ]]; then
```
apolovov, line 40:
```suggestion
    if context::jq -e '.review.request.object.spec.enabled' >/dev/null; then
```

apolovov (changes requested):
A small optimization.

yalosev (approved):

apolovov (approved):

AleksZimin (approved):

PR #7096, opened 2024-01-10
[deckhouse-controller] update module registry settings

### LINKED ISSUE #5107: [external-module-manager] rewrite module on dockerCfg change
### Preflight Checklist

If we change dockerCfg field in a ExternalModuleSource it will not be passed to the module registry values until a new module version comes out.

### Proposed Solution

I guess we can use not only [digest](https://github.com/deckhouse/deckhouse/blob/main/modules/005-external-module-manager/hooks/handle_sources.go#L251) for the checksum but some summary of `repository.Base`, `repository.DockerCFG` and image digest.
Then module will be rewritten on the filesystem on change

### Additional Information

Have to check if module change on filesystem is enough or we have to restart deckhouse also

miklezzzz:
## Description
The following changes are introduced:
* module release's `.status.approved` field is set to true if the module release once was approved;
* if valid registry settings are provided in a module source, their md5 checksum is calculated ad its values is stored in the `modules.deckhouse.io/registry-spec-checksum:` annotation of the module source;
* during consequent updates of the module source, `module controller` checks if registry settings were updated by means of comparison between the annotation value and the checksum of current settings;
* if a change is detected, `module controller` searches for relevant module releases in `Deployed` phase and updates their registry settings in the `/openapi/values.yaml` file so that new settings can be applied on `deckhouse` restart or the module upgrade.
* after updating relevant modules, `module controller` writes new `modules.deckhouse.io/registry-spec-checksum:` value to the module source

Closes #5107
We should update external modules' registry settings in case the registry config of their module source is updated because registry settings might be used in the external modules' charts. 

If we update registry settings of a module source and the module source has some module releases deployed, the modules from these module releases should also have their values (regarding registry configuration) updated.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: chore
summary: Update module values if the corresponding moduleSource was updated.
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
@@ -491,11 +491,14 @@ func (c *Controller) reconcilePendingRelease(ctx context.Context, mr *v1alpha1.M
 			}
 
 			// if policy mode manual
-			if policy.Spec.Update.Mode == "Manual" && !isReleaseApproved(release) {
-				if e := c.updateModuleReleaseStatusMessage(ctx, release, manualApprovalRequired); e != nil {
-					return ctrl.Result{Requeue: true}, e
+			if policy.Spec.Update.Mode == "Manual" {
+				if !isReleaseApproved(release) {
+					if e := c.updateModuleReleaseStatusMessage(ctx, release, manualApprovalRequired); e != nil {
+						return ctrl.Result{Requeue: true}, e
+					}
+					return ctrl.Result{RequeueAfter: defaultCheckInterval}, nil
 				}
-				return ctrl.Result{RequeueAfter: defaultCheckInterval}, nil
+				release.Status.Approved = true
 			}
 
 			// if policy mode auto
--- a/deckhouse-controller/pkg/controller/module-controllers/source/controller.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/source/controller.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"sort"
 	"sync"
 	"time"
@@ -265,6 +266,8 @@ const (
 
 var (
 	ErrNoPolicyFound = errors.New("no matching update policy found")
+
+	checksumAnnotation = "modules.deckhouse.io/registry-spec-checksum"
 )
 
 func (c *Controller) createOrUpdateReconcile(ctx context.Context, roMS *v1alpha1.ModuleSource) (ctrl.Result, error) {
@@ -303,6 +306,20 @@ func (c *Controller) createOrUpdateReconcile(ctx context.Context, roMS *v1alpha1
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	// check, by means of comparing registry settings to the checkSum annotation, if new registry settings should be propagated to deployed module release
+	updateNeeded, err := c.checkAndPropagateRegistrySettings(ms)
+	if err != nil {
+		return ctrl.Result{Requeue: true}, err
+	}
+	// new registry settings checksum should be applied to module source
+	if updateNeeded {
+		if _, err := c.kubeClient.DeckhouseV1alpha1().ModuleSources().Update(context.TODO(), ms, metav1.UpdateOptions{}); err != nil {
+			return ctrl.Result{Requeue: true}, err
+		}
+		// reque ms after modifying annotation
+		return ctrl.Result{Requeue: true}, nil
+	}
+
 	sort.Strings(moduleNames)
 
 	// form available modules structure
@@ -314,7 +331,7 @@ func (c *Controller) createOrUpdateReconcile(ctx context.Context, roMS *v1alpha1
 
 	md := downloader.NewModuleDownloader(c.externalModulesDir, ms, opts)
 
-	// get all policies regardless their labels
+	// get all policies regardless of their labels
 	policies, err := c.moduleUpdatePoliciesLister.List(labels.Everything())
 	if err != nil {
 		return ctrl.Result{Requeue: true}, err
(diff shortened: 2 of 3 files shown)


deckhouse-BOaTswain:

deckhouse-controller/pkg/controller/module-controllers/utils/utils.go
```diff
@@ -38,3 +43,8 @@ func GenerateRegistryOptions(ms *v1alpha1.ModuleSource) []cr.Option {
 
 	return opts
 }
+
+// CalculateChecksum calculates md5 hash of .spec.registry relevant settings of a module source
+func CalculateModuleSourceChecksum(registryCfg *v1alpha1.ModuleSourceSpecRegistry) string {
```
nabokihms, line 48: This is not a ModuleSource checksum, but a spec registry checksum. The name is misleading, the developers may think that they can use the method to find the checksum for the ModuleSource itself.

deckhouse-controller/crds/module-source.yaml
```diff
@@ -95,6 +95,9 @@ spec:
                         type: string
                       error:
                         type: string
+                configChecksum:
```
nabokihms, line 98: Usually, status is for humans and machines, not for the controllers. If you want to keep the state between configuration changes, the annotation is the best way to do this. This is how controllers operate in Kubernetes.

miklezzzz:
before
```
spec:
  registry:
    ca: ""
    dockerCfg: someconfig
    repo: example.com
    scheme: HTTPS
  releaseChannel: alpha
status:
  configChecksum: cfe6ab34f4ef11baf64fd2ebd4d2b38b
  message: illegal base64 data at input byte 8
  moduleErrors: []
  modules:
  - name: parca
```

```
cat /var/lib/deckhouse/external-modules/modules/900-parca/openapi/values.yaml 
type: object
x-extend:
  schema: config-values.yaml
properties:
  registry:
    type: object
    default: {}
    properties:
      base:
        type: string
        default: example.com
      dockercfg:
        type: string
        default: someconfig
  internal:
    default: {}
    properties:
      deckhouseAddress:
        type: string
    type: object
```

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
```diff
@@ -701,6 +710,15 @@ func isModuleExistsOnFS(symlinksDir, symlinkPath, modulePath string) bool {
 	return targetPath == modulePath
 }
 
+func moduleRegistryConfigChanged(mr *v1alpha1.ModuleRelease) bool {
```
nabokihms, line 713: The method name is `module registry config changed`, but it checks whether the restart required annotation is present inside. And there is also a side effect: it returns the boolean value and deletes the annotation.

miklezzzz:
after
```
spec:
  registry:
    ca: ""
    dockerCfg: newconfig
    repo: ya.ru
    scheme: HTTPS
  releaseChannel: alpha
status:
  configChecksum: a63e46c124169746acfa9bf687223976
  message: illegal base64 data at input byte 8
  moduleErrors: []
  modules:
  - name: parca
```

```
cat /var/lib/deckhouse/external-modules/modules/900-parca/openapi/values.yaml 
type: object
x-extend:
  schema: config-values.yaml
properties:
  registry:
    type: object
    default: {}
    properties:
      base:
        type: string
        default: ya.ru
      dockercfg:
        type: string
        default: newconfig
  internal:
    default: {}
    properties:
      deckhouseAddress:
        type: string
    type: object
```

```
{"component":"ModuleReleaseController","level":"info","msg":"Restarting Deckhouse because parca module's registry config was updated","time":"2024-01-12T09:48:22Z"}
{"component":"ModuleReleaseController","level":"info","msg":"Restarting Deckhouse because parca module's registry config was updated","time":"2024-01-12T09:48:25Z"}
{"component":"ModuleReleaseController","level":"info","msg":"Restarting Deckhouse because parca module's registry config was updated","time":"2024-01-12T09:48:28Z"}
```

looks strange, makes sense to check why it had to restart deckhouse three times

deckhouse-controller/pkg/controller/module-controllers/source/controller.go
```diff
@@ -278,6 +284,57 @@ func (c *Controller) createOrUpdateReconcile(ctx context.Context, roMS *v1alpha1
 	// Or create a copy manually for better performance
 	ms := roMS.DeepCopy()
 
+	// check if stored checksum equals to the checksum of the current config
+	if ms.Status.ConfigChecksum != controllerUtils.CalculateModuleSourceChecksum(&ms.Spec.Registry) {
```
nabokihms, line 288: Why is a checksum required in the first place? Don't you know the registry configuration you injected into the open API?

nabokihms: If you can get the current values for the module, you can avoid using checksums at all, which is a cleaner solution.

deckhouse-controller/pkg/controller/module-controllers/utils/utils.go
```diff
@@ -15,10 +15,15 @@
...
 
 import (
+	"crypto/md5"
+	"fmt"
+
 	"github.com/deckhouse/deckhouse/deckhouse-controller/pkg/apis/deckhouse.io/v1alpha1"
 	"github.com/deckhouse/deckhouse/go_lib/dependency/cr"
 )
 
+var ModuleRestartRequiredAnnotation = "modules.deckhouse.io/restart-required"
```
nabokihms, line 25: What is module restart, and how can it be required? What will happen if a user sets this annotation on a module?

nabokihms (changes requested):

deckhouse-controller/pkg/controller/module-controllers/utils/utils.go
miklezzzz: removed

deckhouse-controller/crds/module-source.yaml
miklezzzz: accepted and rewrote to use an annotations instead of .status

deckhouse-controller/pkg/controller/module-controllers/source/controller.go
miklezzzz: As I understand, we have registry settings in two places: a module source object and its related module releases (openapi/values.yaml file). So, I decided to store a checksum so as to be able to detect a relevant change of a module source object.

deckhouse-controller/pkg/controller/module-controllers/utils/utils.go
miklezzzz: it was decided not to have a go at the module restart functionality this time, so this code was removed

miklezzzz:
tested in my dev cluster, can't provide any screens as only valid registry settings can be used to check if PR does what it's supposed to do

nabokihms (approved):

yalosev (approved):

PR #7229, opened 2024-01-21
[testing][ci] Tune GolangCI Lint job

z9r5:
## Description

- Set GOCG to 50 to reduce memory usage for the GolangCI Lint job ([details](https://golangci-lint.run/usage/performance/#memory-usage))

- The GolangCI Lint job often fails because of OOM
- It seems that the timeout for unit tests is too small, as they often fail.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: ci
type: chore
summary: Fix GolangCI linter.
impact_level: low
```


Makefile
```diff
@@ -98,7 +97,7 @@ GOLANGCI_VERSION = 1.54.2
 TRIVY_VERSION= 0.38.3
 PROMTOOL_VERSION = 2.37.0
 GATOR_VERSION = 3.9.0
-TESTS_TIMEOUT="15m"
+TESTS_TIMEOUT="25m"
```
z9r5, line 100:
```suggestion
TESTS_TIMEOUT="15m"
```

Makefile
```diff
@@ -4,7 +4,6 @@ FORMATTING_BEGIN_YELLOW = \033[0;33m
 FORMATTING_BEGIN_BLUE = \033[36m
 FORMATTING_END = \033[0m
 
-TESTS_TIMEOUT="15m"
```
z9r5, line 7: This variable is defined twice (line 101), so I've removed it here.

RomanenkoDenys (approved):

PR #7266, opened 2024-01-23
[external-module-manager] restore mpo

miklezzzz:
## Description
This pr introduces additional logic for restoring module releases which can be described in the following steps:
1) on bootstrap we cycle over deployed module releases;
2) if there is a `ModulPullOverride` for a deployed release in the cluster, we skip restoring of this module release;
3) if there is no `ModulPullOverride` for the deployed release there, we check for the symlink pointing to up-to-date version of the module and restore it if necessary;
4) after cycling over deployed releases, we cycle over `ModulPullOverride` objects;
5) if there is a `ModuleSource` for a `ModulPullOverride`, we try to download relevant module dev image from the source;
6) we check for the symlink pointing to dev version of the module and restore it if necessary;

Apart from restoring correct versions of the modules deployed via module releases, we should also restore the modules deployed via `ModulPullOverride` resources and the latter should have priority over `ModuleRelease` objects.

The expected result is that deckhouse is able to restore correct version of a module, depending on if there is a deployed `ModuleRelease` or `ModulPullOverride` applied to the module.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: chore
summary: Restore modules from ModulePullOverride objects.
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
@@ -849,6 +849,17 @@ func (c *Controller) restoreAbsentSourceModules() error {
 		moduleName := item.Spec.ModuleName
 		moduleSource := item.GetModuleSource()
 
+		// if ModulePullOverride is set, don't check and restore overriden release
+		exists, err := c.isModulePullOverrideExists(moduleSource, moduleName)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


miklezzzz:
before: restoring deployed `echo` module of current version
image

miklezzzz:
tests: mpo for echo has been added and is used to restore echo module
image

nabokihms (approved):

PR #7582, opened 2024-02-21
[external-module-manager] disable module hooks on update before restart

### LINKED ISSUE #7152: [deckhouse-controller] Remove delay on ModuleRelease deploy
### Preflight Checklist

### Version

v1.56.X

### Expected Behavior

Module should run new version only after restart

### Actual Behavior

Module can run new hooks from the filesystem before the deckhouse restart

### Steps To Reproduce

_No response_

### Additional Information

1)
We have a delay to avoid multiply deckhouse reboot on ModuleRelease apply [here](https://github.com/deckhouse/deckhouse/blob/main/deckhouse-controller/pkg/controller/module-controllers/release/controller.go#L184)
But it can lead to the situation, when a new version of a module run hooks from the filesystem, while the images was not updated in the deckhouse values.
We have to remove 3 seconds delay, just restart deckhouse immediately.

2) *preffered*
Second option it too move [this hook](https://github.com/deckhouse/deckhouse/blob/main/global-hooks/discovery/modules_images_tags.go) to the deckhouse-controller and reload module images on module deploy. Then we will avoid the race

More details [here](https://flant.slack.com/archives/C05MXKP7GLR/p1705320340696269)

### Logs

_No response_

miklezzzz:
## Description
The following changes are introduced:
- Additional module state `HooksDisabled` which is inherent to a situation when a module gets its hooks disabled deliberately. This state is reflected in the module's state column as `Pending: hooks are disabled` when a new version of a module is deployed and normally shouldn't last longer that it takes from deckhouse to restart itself.
- Module release `Manual approval required` annotations for updated to contain an instruction how to approve a module release: `Waiting for manual approval (annotation modules.deckhouse.io/approved="true" required)`.
- Deckhouse restart logic on a module install/update is updated to use deferred calls to make sure a restart takes place regardless possible errors (err != nil checks).

Currently, on a module install/update, `deckhouse` controller triggers a self-restart with 3 second delay. During these 3 seconds, the module in question may run its hooks which can result in some unexpected results/disruptions.
This pr introduces additional measures to prevent a module from running its hooks on install/update before `deckhouse` restart takes place by disabling the module hooks.
As having a module with its hooks disabled may introduce an issue by itself, for example, if a restart didn't happen for some reasons, an additional module status `HooksDisabled` was introduced in `addon-operator` to reflect this problem. Also, `deckhouse` restarts are issued via deferred calls to diminish the probability of having a module in `HooksDisabled` state longer that it's needed.
Closes #7152
Relates to https://github.com/flant/addon-operator/pull/453

The expected results is that a module can be installed/updated via the module release or the module pull override without having its hooks triggered before `deckhouse` restart takes place.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feature
summary: Disable module hooks on restart.
impact_level: default
```


miklezzzz:
after
status message:
image
module update log (hooks aren't triggered on update):
image

yalosev (approved):

PR #7606, opened 2024-02-22
[extended-monitoring][image-availability-exporter] Add forceCheckDisabledControllers option

verdel:
## Description
Support force image check, even when workloads are disabled or suspended.

This feature was added in version 0.7.0 of the exporter. It would be convenient to have the ability to configure it via ModuleConfig.

We don't.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: extended-monitoring
type: feature
summary: Support force image check with image-availability-exporter, even when workloads are disabled or suspended.
impact_level: default
```


modules/340-extended-monitoring/openapi/doc-ru-config-values.yaml
```diff
@@ -26,6 +26,9 @@ properties:
               ca:
                 description: |
                   Корневой сертификат, которым можно проверить сертификат container registry при работе по HTTPS (если registry использует самоподписанные SSL-сертификаты).
+      forceCheckDisabledControllers:
+        description: |
+          Список типов контроллеров, для которых нужно проводить проверку, даже если количество реплик подов равно 0 или контроллер находится в состоянии `suspend` например `Deployment` или `*` для всех типов контроллеров.
```
z9r5, line 31:
```suggestion
          Список типов контроллеров, для которых нужно проводить проверку, даже если количество реплик подов равно 0 или контроллер находится в состоянии `suspend`.

          Тип контроллера можно указывать без учета регистра.

          Укажите `*`, для проверки всех типов контроллеров.
```

modules/340-extended-monitoring/openapi/config-values.yaml
```diff
@@ -46,6 +46,24 @@ properties:
                 type: string
                 description: |
                   Root CA certificate to validate the container registry's HTTPS certificate (if self-signed certificates are used).
+      forceCheckDisabledControllers:
+        type: array
+        x-examples:
+          - ["Deployment", "StatefulSet"]
+        description: |
+          A list of controller kinds for which image is forcibly checked, even when workloads are disabled or suspended. Acceptable values include `Deployment`, `StatefulSet`, `DaemonSet`, `Cronjob` or `*` for all kinds (this option is case-insensitive).
```
z9r5, line 54:
```suggestion
          A list of controller kinds for which image is forcibly checked, even when workloads are disabled or suspended.
          
          The option is case-insensitive.
          
          Specify `*` to check all controller kinds.
```

z9r5 (approved):

modules/340-extended-monitoring/openapi/config-values.yaml
```diff
@@ -46,6 +46,28 @@ properties:
...
+          type: string
+          enum:
+            - Deployment
+            - StatefulSet
+            - DaemonSet
+            - CronJob
+            - deployment
+            - statefulset
+            - daemonset
+            - cronjob
```
nabokihms, line 69:
```suggestion
```
These enum values are not allowed by the validation. I assume this is fine because there are other options available.

nabokihms (changes requested):
Almost good. Let's satisfy the validator, and that's it.
Great contribution, @verdel!

modules/340-extended-monitoring/openapi/config-values.yaml
verdel: We have a small issue with Enum validation. According to [this](https://github.com/deckhouse/deckhouse/blob/90136124ad01d017a302752146660a72a822fa82/testing/openapi_validation/validators/enum.go#L267-L296) code in validator, the values should start with an uppercase letter. That means we cannot specify in the documentation that case-insensitive values can be used, nor can we use the `*` symbol.

Using `string` as the array element type is also not an option because, in that case, we cannot strictly limit the values that can be passed into the options array.

The only option I see is to change the Enum values, removing the values that start with a lowercase letter and replacing `*` with `All`.

Also, in the Helm template, it will be necessary to replace `All` with `*` and to modify the tests.

If this option works for you, I will make changes to the Pull Request.

modules/340-extended-monitoring/openapi/doc-ru-config-values.yaml
```diff
@@ -26,6 +26,11 @@ properties:
               ca:
                 description: |
                   Корневой сертификат, которым можно проверить сертификат container registry при работе по HTTPS (если registry использует самоподписанные SSL-сертификаты).
+      forceCheckDisabledControllers:
+        description: |
+          Список типов контроллеров, для которых нужно проводить проверку, даже если количество реплик подов равно 0 или контроллер находится в состоянии `suspend`.
```
MariaMakeeva, line 31:
```suggestion
          Список контроллеров, которые необходимо проверять, даже если число реплик подов равняется 0 или контроллер находится в статусе `suspend` (приостановленный).
```

modules/340-extended-monitoring/openapi/doc-ru-config-values.yaml
```diff
@@ -26,6 +26,11 @@ properties:
               ca:
                 description: |
                   Корневой сертификат, которым можно проверить сертификат container registry при работе по HTTPS (если registry использует самоподписанные SSL-сертификаты).
```
MariaMakeeva, line 28:
```suggestion
                  Корневой сертификат для проверки сертификата container registry при использовании HTTPS (если registry использует самоподписанные SSL-сертификаты).
```

modules/340-extended-monitoring/openapi/doc-ru-config-values.yaml
verdel: Fixed

modules/340-extended-monitoring/openapi/doc-ru-config-values.yaml
```diff
@@ -26,6 +26,11 @@ properties:
...
+      forceCheckDisabledControllers:
+        description: |
+          Список типов контроллеров, для которых нужно проводить проверку, даже если количество реплик подов равно 0 или контроллер находится в состоянии `suspend`.
+
+          Укажите `All`, для проверки всех типов контроллеров.
   events:
     description: |
       Настройки для мониторинга event'ов в кластере Kubernetes.
```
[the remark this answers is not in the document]
verdel: The lines of code mentioned in this suggestion were not changed in this Pull Request and are not related to the functionality being added in this Pull Request.

modules/340-extended-monitoring/openapi/doc-ru-config-values.yaml
verdel: The lines of code mentioned in this suggestion were not changed in this Pull Request and are not related to the functionality being added in this Pull Request.

nabokihms (approved):

nabokihms:
Thanks @verdel. It's as awesome as always!

PR #7620, opened 2024-02-26
[dhctl] feat(dhctl-for-commander): support multiple ssh keys and passphrase

faroyam:
## Description
ClusterBootstrapper: eliminate usage of app.SSH* and add SSHClient to Params.
Add support for passphrase in Params.

Support of multiple ssh keys and passphrases in Commander.

### DIFF
--- a/dhctl/pkg/operations/bootstrap/cluster-bootstrapper.go
+++ b/dhctl/pkg/operations/bootstrap/cluster-bootstrapper.go
@@ -289,8 +289,7 @@ func (b *ClusterBootstrapper) Bootstrap() error {
 				SaveBastionHostToCache(baseOutputs.BastionHost)
 			}
 
-			app.SSHHosts = []string{masterOutputs.MasterIPForSSH}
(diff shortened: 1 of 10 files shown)


dhctl/pkg/operations/bootstrap/cluster-bootstrapper.go
```diff
@@ -319,13 +285,9 @@ func (b *ClusterBootstrapper) Bootstrap() error {
 			deckhouseInstallConfig.TerraformState = baseOutputs.TerraformState
 
 			if baseOutputs.BastionHost != "" {
-				setBastionHost(baseOutputs.BastionHost, sshClient)
```
distorhead, line 322: This func also sets sshClient.Settings.BastionHost|Port|User params. We should not forget to initialize those in the outside code where sshClient is created.

distorhead (changes requested):

distorhead (approved):

PR #7759, opened 2024-03-07
Feature/zvirt csi driver

aleksey-su:
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


ee/modules/030-cloud-provider-zvirt/images/cloud-controller-manager/src/pkg/cloudprovider/zvirt/instances.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	for _, ip := range localIP {
+		nodeAddress = append(nodeAddress, v1.NodeAddress{
+			Type:    v1.NodeInternalIP,
+			Address: ip,
+		})
+	}
+
+	vmHostname := zc.zvirtService.ComputeSvc.GetVMHostName(ctx, vm)
```
mvasl, line 135: In our case VM hostname will always match its name, which we already know at this point, there is no need to request that again

ee/modules/030-cloud-provider-zvirt/images/cloud-controller-manager/src/pkg/cloudprovider/zvirt/instances.go
```diff
@@ -0,0 +1,143 @@
...
+			Address: ip,
+		})
+	}
+
+	for _, ip := range localIP {
+		nodeAddress = append(nodeAddress, v1.NodeAddress{
+			Type:    v1.NodeInternalIP,
+			Address: ip,
+		})
+	}
```
mvasl, line 133: Internal and external IP's will always be the same in zVirt since we only attach 1 vNIC to them and that is configured with local ip. Traffic is routed from outside to the internal subnet by edge-routers/firewalls/external loadbalancers. We may as well only give them internal addresses.

ee/modules/030-cloud-provider-zvirt/templates/cloud-data-discoverer/secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cloud-data-discoverer
+  namespace: d8-cloud-provider-zvirt
+  {{- include "helm_lib_module_labels" (list . (dict "app" "cloud-data-discoverer")) | nindent 2 }}
+type: Opaque
+data:
+  host: {{ $providerClusterConfiguration.provider.server | toString | b64enc | quote }}
+  insecure: {{ "true" | b64enc | quote }}
```
mvasl, line 12: Shouldn't username and password also be there? And why is insecure hardcoded?

ee/modules/030-cloud-provider-zvirt/templates/csi/secret.yaml
```diff
@@ -0,0 +1,18 @@
+{{- define "csi_cloud_config" }}
+  {{- $providerClusterConfiguration := .Values.cloudProviderZvirt.internal.providerClusterConfiguration | required "internal.providerClusterConfiguration is required" -}}
+ovirt_url: {{ $providerClusterConfiguration.provider.server }}
+ovirt_username: {{ $providerClusterConfiguration.provider.username }}
+ovirt_password: {{ $providerClusterConfiguration.provider.password }}
+ovirt_insecure: {{ $providerClusterConfiguration.provider.insecure }}
+clusterid: NO_RDE_{{ .Values.global.discovery.clusterUUID }}
```
mvasl, line 7: RDE doesn't belong here, it's vmware thing for collecting events within infrastructure

ee/modules/030-cloud-provider-zvirt/templates/csi/storage-classes.yaml
```diff
@@ -0,0 +1,16 @@
...
+  {{- include "helm_lib_module_labels" (list $) | nindent 2 }}
+  {{- include "helm_lib_module_storage_class_annotations" (list $ $index $storageClass.name) | nindent 2 }}
+  name: {{ $storageClass.name | quote }}
+provisioner: csi.ovirt.org
+reclaimPolicy: Delete
+allowVolumeExpansion: false
+volumeBindingMode: WaitForFirstConsumer
+parameters:
+  storageDomainName: {{ $storageClass.storageDomain | quote }}
+  thinProvisioning: "true"
```
mvasl, line 15: I believe we should leave the choice to the user instead of hardcoding it

mvasl (approved):

PR #7807, opened 2024-03-14
[deckhouse-controller] Module life cycle stage

Dmitrykob:
## Description

Added ability to set and display the current life cycle stage for modules. In addition, the "description" field has been fixed.

screen-1

screen-2

Makes it more transparent to work with the life cycle stages of modules.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary:  Ability to set and display the current life cycle stage for modules.
impact_level: default
```


yalosev (changes requested):
Symlinks have to be restored

yalosev (approved):

PR #7854, opened 2024-03-20
[dhctl] fix(dhctl): meta config deep copy method does not return copy

distorhead:
## Description

This PR fixes MetaConfig::DeepCopy method, which should create a copy object and return it.
There is a typo: deep copy never creates a copy, it returns old object itself.

As a side effect of non-copy behavior we've experienced a panic in the converge op when scaling down some node-group, panic occured after abandoned nodes deletion.

### DIFF
--- a/dhctl/pkg/config/meta.go
+++ b/dhctl/pkg/config/meta.go
@@ -422,7 +422,7 @@ func (m *MetaConfig) CachePath() string {
 }
 
 func (m *MetaConfig) DeepCopy() *MetaConfig {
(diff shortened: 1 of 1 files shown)


nabokihms (approved):

name212 (approved):

deckhouse-BOaTswain:
Cherry pick PR [7855](https://github.com/deckhouse/deckhouse/pull/7855) to the branch [release-1.58](https://github.com/deckhouse/deckhouse/tree/release-1.58) successful!

PR #7887, opened 2024-03-22
[prometheus] Zero downtime rollout restart for prometheus

unreturned:
## Description
Add `preStop` hook to solve problem that some requests goes to terminated pod (lag between updates status of pods and configuration of ingress)

Zero downtime rollout restart for prometheus

Not necessarily

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: fix
summary: Zero downtime rollout restart for Prometheus.
impact_level: default
```


vladimirGuryanov:
Done in another way in this PR:  https://github.com/deckhouse/deckhouse/pull/11694

PR #7949, opened 2024-03-29
Watch crd to set webhook

Dmitrykob:
## Description

Add repeated queries to get CRD and apply conversion strategies.

Image-1

Image-2

The processing of conversion strategies occurs at the deckhouse start and when a particular module enables.
In this case, processing may be started before CRD is installed. This situation leads to an error.

The problem described above occurs in Deckhouse v1.58, and users may encounter it.

A good example to check is the multitenancy module.
It is necessary to turn off the module. Then remove the CRD of this module (such as projects.deckhouse.io, projecttemplates.deckhouse.io) and enable the module again. The first attempt to apply conversion strategies will fail, but after a few attempts, it will succeed.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Repeated queries with a limited number of attempts to get CRDs and apply conversion strategies.
impact_level: default
```

### DIFF
--- a/modules/000-common/images/shell-operator/werf.inc.yaml
+++ b/modules/000-common/images/shell-operator/werf.inc.yaml
@@ -1,4 +1,4 @@
-{{- $shellOperatorVersion := "v1.4.6"}}
+{{- $shellOperatorVersion := "v1.4.7"}}
 {{- $jqVersion := "b6be13d5"}}
 ---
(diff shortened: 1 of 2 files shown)


yalosev (approved):

deckhouse-BOaTswain:

PR #7972, opened 2024-04-02
[ingress-nginx] improve log formatting

### LINKED ISSUE #3762: [ingress-nginx] Improve logs formatting
### Preflight Checklist

Log formatting in the ingress nginx controller could be more intelligent. It uses a plain text template to compose a JSON.

1. Fields similar by the meaning represent different variable types, e.g., `status` is a number, and `upstream_status` is a string.
2. The JSON is only sometimes valid because variables representing numbers may be absent. Example:
```json
{"status": ,"upstream_status": 500}
```

### Proposed Solution

I see two possible options to solve the problem:
1. Deal with all edge cases by diving deeper. 
2. Using log_by_lua stage to print JSON messages.

### Additional Information

https://flant.slack.com/archives/C03B7S7P8FP/p1698326984309379?thread_ts=1698258105.226539&cid=C03B7S7P8FP

ipaqsa:
## Description
Provides fix for log formatting. Sets the fields `upstream_response_time` and `upstream_retries` by default.

Closes #3762

The fields `upstream_response_time` and `upstream_retries` are our own and are set to 0 in the locations (then they are calculated using lua code), but if the request is raw and does not match any location, these fields are not set and we get an invalid json log. 

We get a valid json log even if the request is raw.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: fix 
summary: improve log formatting
```


nabokihms:
This fix is no go because there are mappings in elastic search instances that expect this field to be a number, not a strings. We need another solution that will not change the changes field typing.

ipaqsa:
@miklezzzz after:
Screenshot 2024-04-03 at 2 39 08 PM

ipaqsa:
@miklezzzz after: 
Screenshot 2024-04-03 at 2 41 05 PM

nabokihms (changes requested):
Requires changes

yalosev (approved):

miklezzzz (approved):

Issue #7709, opened 2024-03-04
[deckhouse] Add cli command to show registered requirements keys/values


yalosev:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

We have a few requirements with hook-calculated values but we can't watch them during the runtime

### Proposed Solution

Add a cli command to dump in-memory requirements storage

PR #7982, opened 2024-04-02
Cli command to show registered requirements.

Dmitrykob:
## Description

Added new handler in the debug server that allows to get the list of requirements:
image-1

Added a new cli command for outputting the list of requirements and its description:
image-2

image-3

close https://github.com/deckhouse/deckhouse/issues/7709

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Added cli command for outputting the list of requirements.
impact_level: default
```


go_lib/dependency/requirements/store.go
```diff
@@ -47,3 +47,13 @@ func (m *MemoryValuesStore) Get(key string) (interface{}, bool) {
 	v, ok := m.values[key]
 	return v, ok
 }
+
+func (m *MemoryValuesStore) GetAll() interface{} {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+	var resp = make(map[string]interface{})
```
yalosev, line 54:
```suggestion
	var resp = make(map[string]interface{}, len(m.values))
```

go_lib/dependency/requirements/store.go
```diff
@@ -47,3 +47,13 @@ func (m *MemoryValuesStore) Get(key string) (interface{}, bool) {
 	v, ok := m.values[key]
 	return v, ok
 }
+
+func (m *MemoryValuesStore) GetAll() interface{} {
```
yalosev, line 51: Why `interface{}` not `map[string]interface{}`

go_lib/dependency/requirements/store.go
```diff
@@ -47,3 +47,13 @@ func (m *MemoryValuesStore) Get(key string) (interface{}, bool) {
...
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+	var resp = make(map[string]interface{})
+	for k, v := range m.values {
```
yalosev, line 55: Also don't think that we need copy of the map here

deckhouse-controller/pkg/debug/requirements.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		defer resp.Body.Close()
+
+		var requirements = make(map[string]interface{})
+		decoder := json.NewDecoder(resp.Body)
+		if err := decoder.Decode(&requirements); err != nil {
+			return fmt.Errorf("error unmarshal requirements")
+		}
+
+		for key, value := range requirements {
```
yalosev, line 42: I guess we can yaml.Marshal to the stdout here

yalosev: Or process the `output format` option and use yaml / json

go_lib/dependency/requirements/store.go
Dmitrykob: Apply

go_lib/dependency/requirements/store.go
Dmitrykob: Apply

yalosev (approved):

PR #8076, opened 2024-04-11
[deckhouse-controller] replace insecure flag with scheme flag

### LINKED ISSUE #8026: [deckhouse] Add `--scheme` flag to the change-registry helper
### Preflight Checklist

Schema is calculating automatically from a registry. If the registry has domain like `.local`, the schema is automatically set to `HTTP` (because of containerd library).
Usually, calculating the schema from registry url is ok, but we want to have an ability to set it manually.

### Proposed Solution

1. Remove the [flag `--insecure`](https://github.com/deckhouse/deckhouse/blob/1aa0877bbf29633f16f0ac626438483c6ed29e03/deckhouse-controller/pkg/helpers/register.go#L47) (replace with handling the `http` scheme)
2. Add the flag `--schema` with values `http` / `https`
2.1. If the flag `--schema` is not set - calculate the schema from the registry url (as now), otherwise - take it directly from the flag

### Additional Information

[Thread](https://flant.slack.com/archives/C78AQG5EY/p1706166332872649)

ipaqsa:
## Description
Provides replacement of insecure flag with scheme flag.

Closes #8026

The schema can be defined explicitly, rather than being calculated by path.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature 
summary: replace insecure flag with scheme flag
```

### DIFF
--- a/deckhouse-controller/pkg/helpers/change_registry/change_registry.go
+++ b/deckhouse-controller/pkg/helpers/change_registry/change_registry.go
@@ -53,7 +53,7 @@ const (
 	caKey      = "ca"
 )
 
-func ChangeRegistry(newRegistry, username, password, caFile, newDeckhouseImageTag string, insecure, dryRun bool) error {
+func ChangeRegistry(newRegistry, username, password, caFile, newDeckhouseImageTag, scheme string, dryRun bool) error {
 	ctx := context.Background()
 	logEntry := log.WithField("operator.component", "ChangeRegistry")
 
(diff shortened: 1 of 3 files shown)


deckhouse-controller/pkg/helpers/change_registry/change_registry.go
```diff
@@ -150,11 +150,8 @@ func newRemoteOptions(ctx context.Context, repo name.Repository, authConfig auth
 	}, nil
 }
 
-func newNameOptions(insecure bool) []name.Option {
+func newNameOptions() []name.Option {
 	opts := []name.Option{name.StrictValidation}
-	if insecure {
-		opts = append(opts, name.Insecure)
```
yalosev, line 156: you need this options if `scheme == http`

yalosev (approved):

PR #8143, opened 2024-04-15
[dhctl] Create DeckhouseRelease for new cluster

name212:
## Description
Create DeckhouseRelease for new cluster.

Prevent rollback on previous version if installer version > version in release channel.
In air-gaped envs deckhouse release should created for new cluster with release channel. 

Deckhouse release successfully crerated
image
image

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: feature
summary: Create `DeckhouseRelease` for a new cluster.
impact_level: default
```

### DIFF
--- a/dhctl/pkg/kubernetes/actions/deckhouse/install.go
+++ b/dhctl/pkg/kubernetes/actions/deckhouse/install.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 
+	"github.com/deckhouse/deckhouse/dhctl/pkg/apis/v1alpha1"
 	"github.com/deckhouse/deckhouse/dhctl/pkg/app"
 	"github.com/deckhouse/deckhouse/dhctl/pkg/config"
 	"github.com/deckhouse/deckhouse/dhctl/pkg/kubernetes/actions"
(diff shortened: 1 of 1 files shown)


dhctl/pkg/kubernetes/actions/deckhouse/install.go
```diff
@@ -156,11 +156,42 @@ func UnlockDeckhouseQueueAfterCreatingModuleConfigs(kubeCl *client.KubernetesCli
...
+	if tag, found := config.ReadVersionTagFromInstallerContainer(); found {
+		deckhouseRelease := unstructured.Unstructured{}
+		deckhouseRelease.SetUnstructuredContent(map[string]interface{}{
+			"apiVersion": "deckhouse.io/v1alpha1",
+			"kind":       "DeckhouseRelease",
+			"metadata": map[string]interface{}{
+				"name": tag,
+			},
+			"spec": map[string]interface{}{
+				"version": tag,
```
yalosev, line 171: Shouldn't we check, that tag is semver compatible. Otherwise Deckhouse will fail

name212: checked here
https://github.com/deckhouse/deckhouse/blob/main/dhctl/pkg/config/deckhouse_config.go#L138

name212: we use this function here. Inside function we check tag on semver compatibility if not then we return `false` from function

deckhouse-BOaTswain:
Aye, aye, @name212. I've started the workflow for label 'e2e/run/aws'...

yalosev (approved):

aleksey-su (approved):

PR #8195, opened 2024-04-18
[monitoring-custom] Set scrapeInterval via pod/service annotation

verdel:
## Description

Currently, when using the `monitoring-custom` module, it is not possible to change the `scrapeInterval` for monitoring a Pod/Service. A common value specified in the `monitoring` module settings is used.

The change presented in this PR will allow modifying this value via the annotation `prometheus.deckhouse.io/scrape-interval`.

This feature will allow for additional customization of the scrape process within application monitoring.

We don't.

In the PodMonitor `custom-pod` and ServiceMonitor `custom-service` in the `d8-monitoring` namespace, an additional relabelings will be added that modifies the `__scrape_interval__` label, substituting it with the value specified in the  `prometheus.deckhouse.io/scrape-interval` annotation.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: monitoring-custom
type: feature
summary: Set `scrapeInterval` for user application monitoring via pod/service annotation.
impact_level: default
```

### DIFF
--- a/modules/340-monitoring-custom/docs/README.md
+++ b/modules/340-monitoring-custom/docs/README.md
@@ -62,6 +62,7 @@ metadata:
     prometheus.deckhouse.io/query-param-format: "prometheus"  # '' by default
     prometheus.deckhouse.io/allow-unready-pod: "true"         # by default, NON-ready Pods are ignored
     prometheus.deckhouse.io/sample-limit: "5000"              # by default, the sample is limited to 5000 metrics for a single Pod
+    prometheus.deckhouse.io/scrape-interval: "60s"            # by default scrapeInterval from monitoring module config
 spec:
   ports:
   - name: my-app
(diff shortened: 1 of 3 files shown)


nabokihms:
@verdel sorry for the long delay. We had an idea that there should be a single scrape interval for all targets. However, we thin k about reevaluating this decision, but we need some more time.

verdel:
@nabokihms, I will explain why we needed to change the scrape_interval for a specific exporter.

We started using the prometheus stackdriver exporter to obtain metrics for Google Cloud Platform resources. Google provides data with a resolution of 1 minute.

As a result, if we poll the exporter every 30 seconds, our alert rules, such as `stackdriver_cloudsql_database_cloudsql_googleapis_com_database_postgresql_num_backends > postgresqlMaxConnections`, will not work correctly because the metric data will always lag behind the scrape_interval.

Due to this, we have to use a condition like `last_over_time(stackdriver_cloudsql_database_cloudsql_googleapis_com_database_postgresql_num_backends[5m])`.

If we could increase the scrape_interval for this exporter, it would solve the problem.

verdel:
Yes, I forgot to mention. The stackdriver exporter itself provides metrics with a timestamp, which is the cause of the problem.

verdel:
Can I help review this PR in any way?

z9r5:
@nabokihms Make a review, please.

z9r5 (approved):

PR #8216, opened 2024-04-20
aggregating-proxy: don't query longterm prometheus

lazovskiy:
## Description
Long-term Prometheus uses a federation endpoint to scrape metrics from the main Prometheus. This results in the label set discrepancy in the main and long-term series. Thus, it is not always possible for promxy to return unique series while querying these data sources. This can break queries in Grafana dashboards in certain cases, making them unusable.

This PR drops long-term Prometheus querying in aggregating-proxy. This approach turns out to be unreliable.

Multiple known configurations exist when querying long-term in aggregating-proxy breaks queries in Grafana dashboards.
Such as using `externalLabels` configuration option in the Prometheus `ModuleConfig`

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: fix
summary: Don't query longterm Prometheus in the aggregating-proxy.
impact_level: low
```

### DIFF
--- a/modules/300-prometheus/templates/aggregating-proxy/configmap.yaml
+++ b/modules/300-prometheus/templates/aggregating-proxy/configmap.yaml
@@ -48,22 +48,3 @@ data:
             bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
             tls_config:
               insecure_skip_verify: true
-          {{- if ne (int .Values.prometheus.longtermRetentionDays) 0 }}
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:
Cherry pick PR [8220](https://github.com/deckhouse/deckhouse/pull/8220) to the branch [release-1.59](https://github.com/deckhouse/deckhouse/tree/release-1.59) successful!

PR #8056, opened 2024-04-10
[monitoring] Fix aggregation with longterm prometheus

lazovskiy:
## Description
Currently, when using a `promxy-backed` data source, both main and long-term Prometheus are queried.
Long-term Prometheus metrics may have extra labels, e.g. `prometheus="deckhouse"` thus, when combining the response some unsolicited series emerge and break most of the queries in Grafana dashboards.

This PR strips the only currently known label from the long-term Prometheus responses.

## Changelog entries
```changes
section: prometheus
type: fix 
summary: Fix aggregation with longterm Prometheus
```

PR #8217, opened 2024-04-20
[prometheus] Add external_labels control HTTP-argument

lazovskiy:
## Description
The long-term Prometheus in the Deckhouse Platform employs federation to collect metrics from the main Prometheus. When exporting metrics for the federation purpose, Prometheus adds all the labels listed in externalLabels (can be configured via an eponymous setting in the Prometheus ModuleConfig) to the metrics. This makes label sets in main and long-term Prometheus differ and this is where the problem in conjunction with promxy emerges.
This PR patches Prometheus source code and adds the ability to drop (not include) configured external labels from series exported at `/federation` endpoint by passing the `drop_external_labels=1` query parameter.

A decent amount of work was done at #7757 primarily to eliminate gaps on the Grafana dashboards e.g. when Prometheus is being rolling-updated. The other prominent feature this PR tried to implement was the ability to query and merge results from both main and long-term Prometheus, thus eliminating the need for switching the data source on the Grafana dashboard.
This, however, was never achieved even though there were several attempts to fix issues in #8056, #8212. Now #8216, which disables long-term Prometheus, is waiting for approval.

This PR eliminates the label set discrepancy between main and long-term Prometheus making it possible for promxy to merge responses properly. 

We don't

Label sets for all corresponding series in main and long-term Prometheus are equal.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


nabokihms (changes requested):
Patch description must be added to README.md, and PR to upstream with the patch must be opened.

lazovskiy:
> Patch description must be added to [README.md](https://github.com/deckhouse/deckhouse/blob/main/modules/300-prometheus/images/prometheus/README.md), and PR to upstream with the patch must be opened.

Added description to the README.md, thanks. Will open PR to the upstream right after we discuss this approach with team.

vladimirGuryanov:
We have implemented this in another way.

PR #8252, opened 2024-04-24
Log shipper buffer overflow alert

borg-z:
## Description

This PR introduces a new alert in the Prometheus rules for the log-shipper-agent module. The alert is designed to notify administrators when the buffer usage exceeds 90% of its capacity, indicating a potential risk of data loss or pipeline backpressure.

This feature is necessary manage and mitigate the risks associated with high buffer usage in the log-shipper-agent. When buffer usage approaches full capacity, it can lead to data being dropped or delayed. This alert provides early warning, allowing system administrators to take preventative measures, such as adjusting buffer sizes or scaling up resources, before any data loss occurs.

After applying this change, a new alert titled D8LogShipperBufferOverflow will be active within Prometheus alerting rules. 

## Checklist
- [ ] e2e tests passed.


modules/460-log-shipper/monitoring/prometheus-rules/log-shipper-agent.yaml
```diff
@@ -117,3 +117,32 @@
 
         Consider checking logs of the pod or follow advanced debug instructions.
         `kubectl -n d8-log-shipper get pods -o wide | grep {{ $labels.node }}`
+
+  - alert: D8LogShipperBufferOverflow
```
nabokihms, line 121:
```suggestion
  - alert: D8LogShipperMemoryBufferOverflow
```

modules/460-log-shipper/monitoring/prometheus-rules/log-shipper-agent.yaml
```diff
@@ -117,3 +117,32 @@
...
+      clamp(
+          (
+            vector_buffer_events{buffer_type="memory",component_kind="sink",component_name!="prometheus_sink"}
+            /
+            vector_buffer_max_event_size{buffer_type="memory",component_kind="sink",component_name!="prometheus_sink"}
+          )* 100,
+        0,100
+      ) > 90
+    labels:
+      severity_level: "4"
```
nabokihms, line 133:
```suggestion
      severity_level: "8"
```

nabokihms (changes requested):

PR #8342, opened 2024-05-08
[Prometheus][memcached] Adding drop for memcached metrics

DaNnNiS:
## Description
Adding drop for memcached metrics with name "memcached_slab_*"

With increasing RAM consumption, the number of "memcached_slab_*" metrics increases, for this reason, "sample_limit" is exceeded. Metrics are not used in monitoring.

memcached metrics with name "memcached_slab_*" are not used in monitoring and are reason of sample_limits exceeded.

Prometheus will accept only the necessary metrics from the "memcached exporter". The sample_limit will not be exceeded.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: fix 
summary: Drop the metrics named "memcached_slab_*" from memcached
impact_level: default
```


modules/300-prometheus/templates/memcached/servicemonitor.yaml
```diff
@@ -12,6 +12,10 @@ spec:
     - port: http-metrics
       scheme: http
       honorLabels: true
+      metricRelabelings:
+        - action: drop
+          regex: (memcached_slab_.*)
+          sourceLabels:
```
vladimirGuryanov, line 18: Missing name for source label. It needs to be like this:
```yaml
       - action: drop
          regex: 'memcached_slab_.*'
          sourceLabels: [__name__]
```

vladimirGuryanov (changes requested):

modules/300-prometheus/templates/memcached/servicemonitor.yaml
DaNnNiS: My bad. I fixed a mistake.

z9r5:
@DaNnNiS Please fix the DCO test. [Here are](https://github.com/deckhouse/deckhouse/pull/8342/checks?check_run_id=24731537742) instructions.

nabokihms (approved):

Issue #8337, opened 2024-05-07
[operator-trivy] add storageClass selector to the module configs


yalosev:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

Use desired StorageClass for trivy-server

### Proposed Solution

- Add field to the [openapi/config-values.yaml](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/500-operator-trivy/openapi/config-values.yaml) as [here](https://github.com/deckhouse/deckhouse/blob/main/modules/015-admission-policy-engine/openapi/config-values.yaml#L63) for `trivyServer`
- To the [hook](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/500-operator-trivy/hooks/storage_class_change.go#L10) add support of the config storage class as [here](https://github.com/deckhouse/deckhouse/blob/main/modules/015-admission-policy-engine/hooks/storage_class_change.go#L32)

### Additional Information

[Close](https://github.com/deckhouse/deckhouse/pull/8373)

PR #8373, opened 2024-05-13
[operator-trivy] add storageClass field

ipaqsa:
## Description
It provides the storageClass field for trivy.

Solves #8337
We could not explicitly set the storage class for trivy, now we can.

The storage class can be set explicitly.
```
root@dev-master-0:~# cat trivy.yaml 
kind: ModuleConfig
metadata:
  name: operator-trivy
spec:
  version: 1
  enabled: true
  settings:
     storageClass: ceph-ssd 
```

```
root@dev-master-0:~# dcm values operator-trivy
internal:
  effectiveStorageClass: ceph-ssd
  enabledNamespaces: []
  reportUpdater: {}
```

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: operator-trivy
type: feature 
summary: add ability to explicitly set storageClass.
```

### DIFF
--- a/ee/modules/500-operator-trivy/hooks/storage_class_change.go
+++ b/ee/modules/500-operator-trivy/hooks/storage_class_change.go
@@ -8,7 +8,7 @@ package hooks
 import "github.com/deckhouse/deckhouse/go_lib/hooks/storage_class_change"
 
 var _ = storage_class_change.RegisterHook(storage_class_change.Args{
-	ModuleName:         "operator-trivy",
+	ModuleName:         "operatorTrivy",
 	Namespace:          "d8-operator-trivy",
 	LabelSelectorKey:   "name",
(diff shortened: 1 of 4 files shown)


ee/modules/500-operator-trivy/hooks/storage_class_change.go
```diff
@@ -8,7 +8,7 @@ package hooks
 import "github.com/deckhouse/deckhouse/go_lib/hooks/storage_class_change"
 
 var _ = storage_class_change.RegisterHook(storage_class_change.Args{
-	ModuleName:         "operator-trivy",
+	ModuleName:         "operatorTrivy",
```
yalosev, line 11: You also have to add param `D8ConfigStorageClassParamName` here

yalosev (approved):

PR #8417, opened 2024-05-17
Disable wait for guest network routable by default in vSphere Provider

borg-z:
Signed-off-by: borg-z <me@nikolay-z.top>

## Description
This pull request changes the default behavior of the vSphere provider by setting `wait_for_guest_net_routable` to `false`. This means that Terraform will no longer wait for the VM's network to become routable by default.

In some scenarios, virtual machines (VMs) may be launched in environments without a default gateway. This situation, although rare, can occur in isolated environments where a default route is not provided, and access to repositories and other resources is either within the network or through a proxy. Our current Terraform configuration does the following:
1. Launches the VM.
2. Waits for the VM to have the "guest.net" and "guest.ipStack" properties.
3. Waits for a default route to appear in "guest.IpRouteConfig".

If a default route does not appear, Terraform waits for 5 minutes and then times out. By setting `wait_for_guest_net_routable` to `false`, we avoid unnecessary delays and failures in such environments.

Not necessarily.

After applying these changes, Terraform will not wait for the VM's network to become routable by default, but it will still wait for an IP address to be assigned. 

## See Also

https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/resources/virtual_machine#wait_for_guest_net_routable

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: "Set `wait_for_guest_net_routable` to `false` by default in the vSphere provider to prevent unnecessary timeouts in environments without a default gateway."
impact: "VMs in isolated environments without a default gateway will no longer cause Terraform to wait for 5 minutes and timeout."
impact_level: low

```

### DIFF
--- a/ee/candi/cloud-providers/vsphere/terraform-modules/master-node/main.tf
+++ b/ee/candi/cloud-providers/vsphere/terraform-modules/master-node/main.tf
@@ -201,4 +201,5 @@ resource "vsphere_virtual_machine" "master" {
       firmware,
     ]
   }
+  wait_for_guest_net_routable = var.wait_for_guest_net_routable
 }
--- a/ee/candi/cloud-providers/vsphere/terraform-modules/master-node/main.tf
+++ b/ee/candi/cloud-providers/vsphere/terraform-modules/master-node/main.tf
@@ -201,5 +201,5 @@ resource "vsphere_virtual_machine" "master" {
       firmware,
     ]
   }
-  wait_for_guest_net_routable = var.wait_for_guest_net_routable
+  wait_for_guest_net_routable = false
 }
--- a/ee/candi/cloud-providers/vsphere/terraform-modules/master-node/variables.tf
+++ b/ee/candi/cloud-providers/vsphere/terraform-modules/master-node/variables.tf
@@ -34,6 +34,11 @@ variable "clusterUUID" {
   type = string
 }
 
+variable "wait_for_guest_net_routable" {
+  type = bool
+  default = false
+}
+
 locals {
   prefix = var.clusterConfiguration.cloud.prefix
 
--- a/ee/candi/cloud-providers/vsphere/terraform-modules/master-node/variables.tf
+++ b/ee/candi/cloud-providers/vsphere/terraform-modules/master-node/variables.tf
@@ -34,10 +34,6 @@ variable "clusterUUID" {
   type = string
 }
 
-variable "wait_for_guest_net_routable" {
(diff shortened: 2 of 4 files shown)


deckhouse-BOaTswain:
Aye, aye, @name212. I've started the workflow for label 'e2e/run/vsphere'...

name212 (approved):

ee/candi/cloud-providers/vsphere/terraform-modules/master-node/main.tf
```diff
@@ -201,4 +201,5 @@ resource "vsphere_virtual_machine" "master" {
       firmware,
     ]
   }
+  wait_for_guest_net_routable = var.wait_for_guest_net_routable
```
RomanenkoDenys, line 204: Why to not set parameter to false directly ? Why we need variable ?

RomanenkoDenys (commented):
one question

ee/candi/cloud-providers/vsphere/terraform-modules/master-node/main.tf
borg-z: I assume that someone might want to override this behavior, so to avoid redoing it twice, I moved it to variables.

asviel: How user can set this variable to true?

borg-z: Right now, this option is not available. I'm not sure if we need to complicate this now, but on the other hand, I'm afraid that during testing, we will find out that we need to add the configuration for this option.

borg-z: ok, fixed

deckhouse-BOaTswain:
Aye, aye, @borg-z. I've started the workflow for label 'e2e/run/vsphere'...

ee/candi/cloud-providers/vsphere/terraform-modules/master-node/main.tf
asviel: Approved as long as it is not a breaking change (needs to be tested).

asviel (approved):

deckhouse-BOaTswain:
Aye, aye, @RomanenkoDenys. I've started the workflow for label 'e2e/run/vsphere'...

PR #8439, opened 2024-05-21
[docs] kube-bench apply without yq

borg-z:
## Description

Updated the documentation for running kube-bench to remove dependency on yq, which is not available in the container. The new instructions use kubectl and jq only.

The change resolves the issue where yq is not available in the container, allowing users to run kube-bench using available tools only.

Not necessarily.

Users should be able to run kube-bench on specific nodes without the need for yq. The command provided in the documentation should work correctly.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: docs
type: fix
summary: Update kube-bench documentation to use `kubectl` and `jq` instead of `yq`.
impact_level: default

### DIFF
--- a/modules/002-deckhouse/docs/FAQ.md
+++ b/modules/002-deckhouse/docs/FAQ.md
@@ -21,7 +21,7 @@ Then you have to select which node you want to run kube-bench.
 * Run on specific node, e.g. control-plane node:
 
   ```shell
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
Cherry pick PR [8446](https://github.com/deckhouse/deckhouse/pull/8446) to the branch [release-1.60](https://github.com/deckhouse/deckhouse/tree/release-1.60) successful!

PR #8441, opened 2024-05-21
[user-authn] unify usage enabled instead of enable

### LINKED ISSUE #5627: Unification of enable/enabled and auto/automatic parameters
### Preflight Checklist

People see a parameter in one place, assume that it is like this everywhere, and start using it everywhere, but in some places it does not work.

### Proposed Solution

Enable/enabled and auto/automatic parameters should be unified.
Correct:
- enabled
- auto

### Additional Information

_No response_

ipaqsa:
## Description
It provides conversions and replacements for using only the `enabled` field.

Fixes #5627
People see the `enable` parameter in publishAPI field and may start to use it in some other places, but there is only the `enabled` field there.

The `publishAPI` field has the `enabled` field instead of `enable`.

```
root@dev-master-0:~# dcm values user-authn
publishAPI:
  addKubeconfigGeneratorEntry: true
  enabled: false
  https:
    mode: SelfSigned
```

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Replace the `enable` option with the `enabled` in the `publishAPI` field.
```

### DIFF
--- a/modules/150-user-authn/openapi/conversions/v2.yaml
+++ b/modules/150-user-authn/openapi/conversions/v2.yaml
@@ -1,3 +1,3 @@
 version: 2
 conversions:
-  - .publishAPI.enabled = .publishAPI.enable | del(.publishAPI.enable)
+  - if .publishAPI.enable != null then .publishAPI.enabled = .publishAPI.enable | del(.publishAPI.enable) end
--- a/modules/150-user-authn/openapi/conversions/v2.yaml
+++ b/modules/150-user-authn/openapi/conversions/v2.yaml
@@ -1,3 +1,3 @@
 version: 2
 conversions:
-  - if .publishAPI and .publishAPI.enable then .publishAPI.enabled = .publishAPI.enable | del(.publishAPI.enable) else . end
+  - .publishAPI.enabled = .publishAPI.enable | del(.publishAPI.enable)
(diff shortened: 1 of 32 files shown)


deckhouse-BOaTswain:
Aye, aye, @ipaqsa. I've started the workflow for label 'e2e/run/aws'...

z9r5 (approved):

modules/150-user-authn/openapi/conversions/v2.yaml
```diff
@@ -0,0 +1,3 @@
+version: 2
+conversions:
+  - .publishAPI.enabled = .publishAPI.enable | del(.publishAPI.enable)
```
yalosev, line 3: I don't think it works. It will generate
```yaml
spec:
  settings:
    publishAPI:
      enabled: null
```

yalosev, line 3:
```suggestion
  - if .publishAPI.enable != null then .publishAPI.enabled = .publishAPI.enable | del(.publishAPI.enable) end
```

ipaqsa: `.publishAPI.enable` has a default value

yalosev (approved):

PR #8456, opened 2024-05-22
[docs]Changing a example of creating hash of password for static user

DaNnNiS:
Changing a example of creating hash of password for static user

## Description
Changing a example of creating a password's hash for password of static user in documentation.

When creating a user, you must specify the correct password hash, otherwise it will be impossible to log in to this user.

Correct authorization for new static user.

## Checklist
- [ ] e2e tests passed.

```changes
section: docs
type: fix
summary: Changing a example of creating hash of password for static user


z9r5:
Cherry-picked into release-1.60.

PR #8606, opened 2024-06-04
[dhctl] feat(dhctl-for-commander): add validations rpc

faroyam:
## Description

Add cluster config validation gRPC services. We need them in Deckhouse Commander.

This PR adds a gRPC service for validating cluster configuration. Previously, Deckhouse Commander used these functions when calling dhctl as a library. Now validations will also be called via gRPC, like other tasks, such as Check, Bootstrap, etc.

This is change that corrects dhctl-server which was firstly introduced in the 1.60 release. So it is preferred to include this change into 1.60 release. Because after this change dhctl-server in the 1.60 release will stabilize more and will be equipped with full set of basic operations.

It is not expected to change anything outside of pkg/server scope by this PR.

```changes
section: dhctl
type: feature
summary: Add cluster config validation gRPC services.
impact_level: default
```


deckhouse-BOaTswain:
Aye, aye, @faroyam. I've started the workflow for label 'e2e/run/gcp'...

aleksey-su (approved):

name212 (approved):

deckhouse-BOaTswain:
Cherry pick PR [8728](https://github.com/deckhouse/deckhouse/pull/8728) to the branch [release-1.61](https://github.com/deckhouse/deckhouse/tree/release-1.61) successful!

PR #8612, opened 2024-06-04
[deckhouse-controller] Hide some sensitive data in debug output

asviel:
## Description

Hide cloud-providers/registry credentials and client certificates in debug output.

We don't want to expose sensitive data in debug output.

We don't.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Hide some sensitive data in debug output.
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/debug/debug.go
+++ b/deckhouse-controller/pkg/debug/debug.go
@@ -41,7 +41,7 @@ func (c *Command) Save(tarWriter *tar.Writer) error {
 
 	header := &tar.Header{
 		Name: c.File,
-		Mode: 0600,
+		Mode: 0o600,
 		Size: int64(len(fileContent)),
 	}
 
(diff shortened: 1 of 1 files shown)


deckhouse-controller/pkg/debug/debug.go
```diff
@@ -76,7 +76,7 @@ func createTarball() *bytes.Buffer {
 		{
 			File: "global-values.json",
 			Cmd:  "bash",
-			Args: []string{"-c", `deckhouse-controller global values -o json | jq .`},
+			Args: []string{"-c", `deckhouse-controller global values -o json | jq '.internal.modules.kubeRBACProxyCA = "REDACTED" | .modulesImages.registry.dockercfg = "REDACTED"'`},
```
yalosev, line 79: NIT: I would like to see some extra field in the `Command` struct, like `FilterFn` or `FilterFields` or `JSONFilter` which will be called in the `cmd.Save` function.
Then you can realize the filter in a few ways - sjson, jq, manual filtering, whatever

yalosev (approved):

deckhouse-controller/pkg/debug/debug.go
yalosev: Like:
```golang
{
			File: "global-values.json",
			Cmd:  "bash",
			Args: []string{"-c", `deckhouse-controller global values -o json`}
            FilterFunc: func(s string) string {
              val, _ := sjson.Set(s, "internal.modules.kubeRBACProxyCA", "REDACTED")
              return val
            },
```

RomanenkoDenys (approved):

PR #8665, opened 2024-06-11
[multitenancy-manager] fix: replace special characters in metadata.name

juev:
## Description

When describing the project, we fill in the block of authorisation rules.

In the authorisation rules template, the variable user.name used twice:
- in the name of the rule
- in the user name

Email can and often is used in the username. But we cannot use the special characters in the rule name due to restrictions on the part of k8s.

Now, when using a template, if the user's email is used in the username, we get a template generation error.

As a result, the necessary users are not created.

We are replacing the special characters with `-` in the rule name. And as a result, the template is generated and the necessary users are created.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: fix
summary: replace special characters in a AuthorizationRule metadata.name
impact_level: default
```

### DIFF
--- a/ee/modules/160-multitenancy-manager/templates/user-resources/_helpers.tpl
+++ b/ee/modules/160-multitenancy-manager/templates/user-resources/_helpers.tpl
@@ -1,6 +1,0 @@
-{{- define "slugify" }}
-  {{- $newName := lower . }}
-  {{- $newName = regexReplaceAll "\\W+" $newName "-" }}
-  {{- $newName = regexReplaceAll "(^-+|-+$)" $newName "" }}
-  {{- print $newName }}
-{{- end }}
--- a/ee/modules/160-multitenancy-manager/templates/user-resources/_helpers.tpl
+++ b/ee/modules/160-multitenancy-manager/templates/user-resources/_helpers.tpl
@@ -1,0 +1,6 @@
+{{- define "slugify" }}
+  {{- $newName := lower . }}
+  {{- $newName = regexReplaceAll "\\W+" $newName "-" }}
+  {{- $newName = regexReplaceAll "(^-+|-+$)" $newName "" }}
+  {{- print $newName }}
+{{- end }}
--- a/ee/modules/160-multitenancy-manager/templates/user-resources/default-project-template.yaml
+++ b/ee/modules/160-multitenancy-manager/templates/user-resources/default-project-template.yaml
@@ -128,7 +128,7 @@ spec:
     apiVersion: deckhouse.io/v1alpha1
     kind: AuthorizationRule
     metadata:
-      name: {{ $administrator.name | include "slugify" }}
+      name: {{ $administrator.name | include "normalize" }}
     spec:
       accessLevel: Admin
       subjects:
--- a/ee/modules/160-multitenancy-manager/templates/user-resources/default-project-template.yaml
+++ b/ee/modules/160-multitenancy-manager/templates/user-resources/default-project-template.yaml
@@ -128,7 +128,7 @@ spec:
     apiVersion: deckhouse.io/v1alpha1
     kind: AuthorizationRule
     metadata:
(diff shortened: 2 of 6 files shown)


deckhouse-BOaTswain:

miklezzzz (approved):

ee/modules/160-multitenancy-manager/templates/user-resources/default-project-template.yaml
```diff
@@ -128,7 +128,7 @@ spec:
     apiVersion: deckhouse.io/v1alpha1
     kind: AuthorizationRule
     metadata:
-      name: {{ $administrator.name }}
+      name: {{ $administrator.name | replace "@" "-" }}
```
nabokihms, line 131: The idea was to cover more cases and slugify other special characters like spaces.
In previous implementation, we used the slugify helper:
https://github.com/deckhouse/deckhouse/blob/cff98f6efc422621e5e9d26b4ff9930ccf5dc47f/ee/modules/160-multitenancy-manager/templates/user-resources/templates/user-resources-templates.yaml#L106

miklezzzz: does it mean it's a an eligible case to a group like below
```
root@dev-master-0:~# kubectl  get groups.deckhouse.io  
NAME             GROUP
administrators   administrator@s
```
and match it by "administrator@s", not "administrators"?

miklezzzz: we (I) thought the problem was solely about having to provide a string in a email notation  if we assign the project to a user, which would result in having ineligible, from k8s point of view, resource name with '@' character.

nabokihms: The problem is with the metadata.name field, for groups any string is valid

ee/modules/160-multitenancy-manager/templates/user-resources/_helpers.tpl
```diff
@@ -0,0 +1,6 @@
+{{- define "slugify" }}
```
yalosev, line 1: Honestly, I'm a bit confused here.
With this change we will have the `slugify` func defined twice, here and [here](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/160-multitenancy-manager/templates/user-resources/templates/user-resources-templates.yaml#L15)

Also, according to [this code](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/160-multitenancy-manager/hooks/project.go#L143-L168) module will load files only from the `templates/` dir and this file will not be loaded. Probably, you have to add this function to [this file](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/160-multitenancy-manager/templates/user-resources/templates/_helpers.tpl) but anyway we have to fix the code duplication.


I think the plan should be like this:
1.  Move [this func](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/160-multitenancy-manager/templates/user-resources/templates/user-resources-templates.yaml#L15-L39) to the `templates/_helpers.tpl` file
2. Simplify/adopt it (if necessary)
3. Use it in templates and also [here](https://github.com/deckhouse/deckhouse/blob/main/ee/modules/160-multitenancy-manager/templates/user-resources/templates/user-resources-templates.yaml#L106)

yalosev (changes requested):

ee/modules/160-multitenancy-manager/templates/user-resources/_helpers.tpl
juev: Thank you! I completely agree with you.

Made changes, transferred templates to one file. But I didn’t use `slugify` myself, since it uses two parameters. Renamed my function to `normalize`.

Conducted testing, helm charts were generated correctly.

yalosev (approved):

PR #8686, opened 2024-06-14
[user-authn] feat: update dex to v2.40.0

juev:
## Description

Updating the dex version to v2.40.0

Updating the version to fix a number of problems and improve system security.

New version of dex in the user-authn module

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: feature
summary: Update dex to `v2.40.0`.
impact_level: default
```


deckhouse-BOaTswain:

modules/150-user-authn/images/dex/Dockerfile
```diff
@@ -1,25 +1,18 @@
...
 COPY patches/client-groups.patch patches/static-user-groups.patch patches/gitlab-refresh-context.patch patches/connector-data.patch patches/oidc-ca-insecure.patch patches/robots-txt.patch patches/401-password-auth.patch /
-RUN git clone --branch v2.35.3 --depth 1 ${SOURCE_REPO}/dexidp/dex.git . \
+RUN git clone --branch v2.40.0 --depth 1 ${SOURCE_REPO}/dexidp/dex.git . \
   && git apply /client-groups.patch \
   && git apply /static-user-groups.patch \
-  && git apply /gitlab-refresh-context.patch \
-  && git apply /connector-data.patch \
-  && git apply /oidc-ca-insecure.patch \
-  && git apply /robots-txt.patch \
-  && git apply /401-password-auth.patch
```
miklezzzz, line 17: it would be better to delete the patches that that had made their way into v2.40.0 and update README.md accordingly.

juev: Completed

miklezzzz (approved):
checked in a dev cluster with a dev gitlab instance that first two patches are applied as supposed- seems good

yalosev (approved):

nabokihms:
image

modules/150-user-authn/images/dex/patches/client-groups.patch
```diff
@@ -1,64 +1,21 @@
...
--			return
--		}
- 		if err := s.templates.approval(r, w, authReq.ID, authReq.Claims.Username, client.Name, authReq.Scopes); err != nil {
- 			s.logger.Errorf("Server template error: %v", err)
- 		}
+Index: storage/kubernetes/types.go
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
```
nabokihms, line 5: Can we remove headers for patches? I think it is safe.

nabokihms:
Oh didn't know how it happend... Didn't mean to close the PR. Could you please reopen it?

modules/150-user-authn/images/dex/patches/client-groups.patch
juev: Of course I removed

juev:
> Oh didn't know how it happend... Didn't mean to close the PR. Could you please reopen it?

Reopened, it's not a problem

> After bumping the version, dex now asks for confirmation. However, we explicitly set the option to skip approval screens

Yes, this is strange, I checked the dex configuration, checked the code version 2.40.0, in this case the approval page should not be displayed. But it is there. :(

juev:
After studying the problem, I discovered that the dex-authenticator uses the `approval-prompt` flag equal to `force` by default. Which leads to the fact that all requests in dex come with the `approval_prompt=force` query parameter. This is what influenced the behavior of the system and, regardless of the configuration used, the approval page was always displayed.
Adding the `approval-prompt=basic` flag to the dex-authenticator solved the problem.

PR #8696, opened 2024-06-14
[deckhouse] feat: restrict creating system namespaces

juev:
## Description

We need to prohibit the creation of namespaces with names that begin with `kube-` and `d8-`

Only the deckhouse system user should be able to do this

Regular users, including the administrator, should not be able to create system namespaces so as not to affect important settings.

Any user, with the exception of deckhouse, should receive an error for such an attempt when trying to create a namespace whose name begins with `d8-` or `kube-`.

## Checklist
- [X] The code is covered by unit tests.
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: restrict creating system namespaces
impact_level: default
```

### DIFF
--- a/modules/002-deckhouse/templates/validation.yaml
+++ b/modules/002-deckhouse/templates/validation.yaml
@@ -21,7 +21,7 @@ spec:
   validations:
     - expression: '!(request.userInfo.username != ''system:serviceaccount:d8-system:deckhouse''
         && (object.metadata.name.startsWith(''d8-'') || object.metadata.name.startsWith(''kube-'')))'
-      messageExpression: '''Restricts create system namespaces'''
+      messageExpression: '''Creation of system namespaces is forbidden'''
 ---
 {{- if semverCompare ">= 1.28" .Values.global.discovery.kubernetesVersion }}
 apiVersion: admissionregistration.k8s.io/v1beta1
--- a/modules/002-deckhouse/templates/validation.yaml
+++ b/modules/002-deckhouse/templates/validation.yaml
@@ -1,0 +1,40 @@
+{{- if semverCompare ">= 1.26" .Values.global.discovery.kubernetesVersion }}
+{{- $policyName := "d8-create-namespace-validator" }}
+---
+{{- if semverCompare ">= 1.28" .Values.global.discovery.kubernetesVersion }}
+apiVersion: admissionregistration.k8s.io/v1beta1
+{{- else }}
+apiVersion: admissionregistration.k8s.io/v1alpha1
+{{- end }}
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: {{ $policyName }}
+  {{- include "helm_lib_module_labels" (list . (dict "app" "deckhouse") ) | nindent 2 }}
+spec:
+  failurePolicy: Fail
+  matchConditions:
+    - expression: (request.userInfo.username != "system:serviceaccount:d8-system:deckhouse")
+      name: exclude-system-user
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   ["*"]
+        apiVersions: ["*"]
+        operations:  ["CREATE"]
+        resources:   ["namespaces"]
+  validations:
+    - expression: (!object.metadata.name.startsWith('d8-') && (!object.metadata.name.startsWith('kube-')))
+      messageExpression: '''Restrics create system namespaces'''
+---
+{{- if semverCompare ">= 1.28" .Values.global.discovery.kubernetesVersion }}
+apiVersion: admissionregistration.k8s.io/v1beta1
+{{- else }}
+apiVersion: admissionregistration.k8s.io/v1alpha1
+{{- end }}
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: {{ $policyName }}
+  {{- include "helm_lib_module_labels" (list . (dict "app" "deckhouse") ) | nindent 2 }}
+spec:
+  policyName: {{ $policyName }}
+  validationActions: [Deny]
+{{- end }}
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:

modules/002-deckhouse/templates/validation.yaml
```diff
@@ -0,0 +1,76 @@
...
+{{- end }}
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: {{ $policyName }}
+  {{- include "helm_lib_module_labels" (list . (dict "app" "deckhouse") ) | nindent 2 }}
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   ["*"]
```
yalosev, line 17:
```suggestion
      - apiGroups:   [""]
```

I don't think we need all apiGroups here, `core` would be enough

modules/002-deckhouse/templates/validation.yaml
```diff
@@ -0,0 +1,76 @@
+{{- if semverCompare ">= 1.26" .Values.global.discovery.kubernetesVersion }}
+{{- $policyName := "d8-create-namespace-validator" }}
```
yalosev, line 2:
```suggestion
{{- $policyName := "system-ns.deckhouse.io" }}
```

Let's name it it a more predictable way.
Sorry for that, but it's a bit closer to [docs](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#creating-a-validatingadmissionpolicy) and looks really more readable, because we will have an error like:

```
ValidatingAdmissionPolicy 'system-ns.deckhouse.io' with binding 'system-ns.deckhouse.io' ...
```
which will definitely points to the deckhouse source

modules/002-deckhouse/templates/validation.yaml
```diff
@@ -0,0 +1,76 @@
...
+  validationActions: [Deny]
+---
+{{- if semverCompare ">= 1.28" .Values.global.discovery.kubernetesVersion }}
+apiVersion: admissionregistration.k8s.io/v1beta1
+{{- else }}
+apiVersion: admissionregistration.k8s.io/v1alpha1
+{{- end }}
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "d8-create-heritage-deckhouse-objects"
```
yalosev, line 46:
```suggestion
  name: "label-objects.deckhouse.io"
```
something like that

modules/002-deckhouse/templates/validation.yaml
```diff
@@ -0,0 +1,76 @@
...
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   ["*"]
+        apiVersions: ["*"]
+        operations:  ["CREATE"]
+        resources:   ["*"]
+  validations:
+    - expression: 'request.userInfo.username.startsWith(''system:serviceaccount:d8-'')'
+      messageExpression: '''Restricts create object as NOT service account'''
```
yalosev, line 58: We need some more obvious message here, something like:
```
Creating an object with the `heritage: deckhouse` label is forbidden
```

modules/002-deckhouse/templates/validation.yaml
```diff
@@ -0,0 +1,76 @@
...
+apiVersion: admissionregistration.k8s.io/v1beta1
+{{- else }}
+apiVersion: admissionregistration.k8s.io/v1alpha1
+{{- end }}
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "d8-create-heritage-deckhouse-objects"
+  {{- include "helm_lib_module_labels" (list . (dict "app" "deckhouse") ) | nindent 2 }}
+spec:
+  policyName: "d8-create-heritage-deckhouse-objects"
```
yalosev, line 70:
```suggestion
  policyName: "label-objects.deckhouse.io"
```

modules/002-deckhouse/templates/validation.yaml
```diff
@@ -0,0 +1,76 @@
...
+      messageExpression: '''Restricts create object as NOT service account'''
+---
+{{- if semverCompare ">= 1.28" .Values.global.discovery.kubernetesVersion }}
+apiVersion: admissionregistration.k8s.io/v1beta1
+{{- else }}
+apiVersion: admissionregistration.k8s.io/v1alpha1
+{{- end }}
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "d8-create-heritage-deckhouse-objects"
```
yalosev, line 67:
```suggestion
  name: "heritage-label-objects.deckhouse.io"
```

modules/002-deckhouse/templates/validation.yaml
juev: Done

modules/002-deckhouse/templates/validation.yaml
juev: Done

modules/002-deckhouse/templates/validation.yaml
juev: Done

modules/002-deckhouse/templates/validation.yaml
juev: Done

modules/002-deckhouse/templates/validation.yaml
```diff
@@ -0,0 +1,76 @@
...
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   [""]
+        apiVersions: ["*"]
+        operations:  ["CREATE"]
+        resources:   ["namespaces"]
+  validations:
+    - expression: '!(request.userInfo.username != ''system:serviceaccount:d8-system:deckhouse''
+        && (object.metadata.name.startsWith(''d8-'') || object.metadata.name.startsWith(''kube-'')))'
+      messageExpression: '''Restricts create system namespaces'''
```
yalosev, line 24:
```suggestion
      messageExpression: '''Creation of system namespaces is forbidden'''
```

yalosev (approved):

Issue #8403, opened 2024-05-16
[deckhouse-controller] add a convenient way of resolving which deckhouse pod is current leader


miklezzzz:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

It's required to provide our users with a universal (considering HA mode and multiple pods) kubectl one-liner to get all necessary debug information from the current running deckhouse pod.

### Proposed Solution

todo

nabokihms:
One way to solve the problem is to label the pod if it is a leader and remove this label from other pods. THen we will be able to exec using labels selector, e.g., `kubectl -n d8-system exec -l deckhouse-leader="" -it -- bash`

Another way is to use the `d8` tool.

PR #8720, opened 2024-06-18
[deckhouse-controller] add a convenient way of resolving which deckhouse pod is current leader

juev:
## Description

We need to add a new label to the leader

Previously, only one pod deckhouse was launched during deployment, and when the remote shell was launched, all commands were correctly processed.

Now it is possible to launch several pods for the deckhouse application, one is called the leader, the rest are in standby mode. As a result, when launching a remote shell, we find ourselves in an arbitrary pod, and the commands do not provide the necessary information about the system.

When using a new label, you can address the launch of a remote shell in the pod leader.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: add a convenient way of resolving which deckhouse pod is current leader
impact_level: default
```

Closes: #8403


modules/002-deckhouse/templates/deckhouse-leader-service.yaml
```diff
@@ -0,0 +1,20 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: deckhouse-leader
+  namespace: d8-system
+  {{- include "helm_lib_module_labels" (list . (dict "app" "deckhouse")) | nindent 2 }}
+spec:
```
miklezzzz, line 8:
```suggestion
spec:
  publishNotReadyAddresses: true
```
sorry, my bad, I forgot to mention this parameter at the beginning

modules/002-deckhouse/hooks/set_deckhouse_leader_pod.go
```diff
@@ -0,0 +1,75 @@
...
+	return nil
+}
+
+func emptyOrNil(b bool) any {
+	if b {
+		return ""
+	}
+
+	return nil
+}
```
miklezzzz, line 75: I tested in a cluster and it takes Deckhouse substantial amount of time to reach the point when the hook is evoked, could you please move the hook as it is to ./global-hooks/ ?

modules/002-deckhouse/templates/deckhouse-leader-service.yaml
juev: I added the parameter

modules/002-deckhouse/hooks/set_deckhouse_leader_pod.go
juev: I moved the files to global-hooks

deckhouse-BOaTswain:

miklezzzz (approved):
works as a charm

deckhouse-BOaTswain:

global-hooks/set_deckhouse_leader_pod.go
```diff
@@ -0,0 +1,71 @@
...
+
+		input.PatchCollector.MergePatch(patch, "v1", "Pod", d8Namespace, pod.Name)
+	}
+
+	return nil
+}
+
+func trueOrNil(b bool) any {
+	if b {
+		return true
```
yalosev, line 67:
```suggestion
		return "true"
```

Label could be only `string` or nil. So boolean will fail here

modules/002-deckhouse/templates/deckhouse-leader-service.yaml
```diff
@@ -0,0 +1,21 @@
...
+  publishNotReadyAddresses: true
+  selector:
+    app: deckhouse
+    leader: true
```
yalosev, line 12:
```suggestion
    leader: "true"
```

yalosev: Just to add some safety. `deckhouse` label could be interpreted only as string, while the `leader` label could be interpreted as boolean. You will get an error:
```
 Service in version "v1" cannot be handled as a Service: json: cannot unmarshal bool into Go struct field ServiceSpec.spec.selector of type string
 ```

yalosev: For example, you have to to like this with some values, like `controllerVersion`, which is string, but if you set `1.10` there, kubernetes will interpret it like a float

yalosev: At the same time, if we do `kubectl get pods -l leader=true`, kubectl will case this values to the string

global-hooks/set_deckhouse_leader_pod.go
juev: Fixed

modules/002-deckhouse/templates/deckhouse-leader-service.yaml
juev: Thank you! 
Now it has become clearer.

yalosev:
/e2e/abort yandex-cloud;WithoutNAT;containerd;1.28 9710263297 9710263297-1-con-1-28 /sys/deckhouse-oss/install:pr8720 redos@158.160.114.70

deckhouse-BOaTswain:
Aye, aye, @yalosev. I've recognized your '/e2e/abort' command and started the workflow...

deckhouse-BOaTswain:

PR #8722, opened 2024-06-18
[deckhouse] fix overwriting embedded modules' images tags

### LINKED ISSUE #8480: [deckhouse] prevent overriding of modules images tags for external modules
### Preflight Checklist

### Version

v1.60.0

### Expected Behavior

Deckhouse should keep images_tags for the first met module

### Actual Behavior

Deckhouse will set images_tags for the last met module

### Steps To Reproduce

1. Take embedded module with some image
2. Create external module with the image with the same name
3. Deckhouse will use embedded module but with the external image tag

### Additional Information

We have to modify [this]https://github.com/deckhouse/deckhouse/blob/81e07763ac0eda440bd238af3215c8ca2e901924/global-hooks/discovery/modules_images_tags.go#L114-L115 function
If we have already set images_tags for a module, we should skip a new one, as we [do](https://github.com/deckhouse/deckhouse/blob/81e07763ac0eda440bd238af3215c8ca2e901924/deckhouse-controller/pkg/controller/loader.go#L143) with the equal modules

### Logs

_No response_

miklezzzz:
## Description
This PR updates `modules_images_tags` discovery hook's logic so that if there are two modules in the cluster with the same name, images tags of first one (supposedly, an embedded module) will make their way into resulting images_digests.json file.

Closes #8480 
We need it because in opposite case images digests of an embedded module get overwritten with values of its namesake sourced module.

Embedded modules have priority in terms of charts and images digests over source modules.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: chore
summary: Fix overwriting embedded modules' images tags.
impact_level: default
```

### DIFF
--- a/global-hooks/discovery/modules_images_tags.go
+++ b/global-hooks/discovery/modules_images_tags.go
@@ -58,7 +58,10 @@ func discoveryModulesImagesDigests(input *go_hook.HookInput) error {
 
 	modulesDigestsObj := readModulesImagesDigests(input, externalModulesDir)
 	for k, v := range modulesDigestsObj {
-		digestsObj[k] = v
+		// under no circumstances do we overwrite existing digests
+		if _, found := digestsObj[k]; !found {
+			digestsObj[k] = v
+		}
 	}
 
 	input.Values.Set("global.modulesImages.digests", digestsObj)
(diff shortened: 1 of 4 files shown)


deckhouse-BOaTswain:

yalosev (approved):

PR #8744, opened 2024-06-20
[ingress-nginx] add alert for deprecated geoip envs

miklezzzz:
## Description
This pr adds a hook to check if there some geoip v1 variables in use in the cluster. If it's the case - corresponding metrics are created and prometheus alert is triggered.
Also, a couple of outdated migrations were deleted from ingress-nginx hooks.

We need it because we have introduced ingress controller v1.10 which doesn't support geoip v1 module. Moreover, we plan to deprecate v1.9 controller in the near feature so geoip v1 support is discontinued.

If there is an ingress controller of v1.6-v1.9 in the cluster and some ingress-related configuration relies on geoipv1 module's variables, there must be `d8_deprecated_geoip_version` alret in the cluster's prometheus.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: chore
summary: Add geoip deprecated version alert.
impact_level: default
```


miklezzzz:
image

miklezzzz:
image

modules/402-ingress-nginx/hooks/alert_deprecated_geoip_version.go
```diff
@@ -0,0 +1,180 @@
...
+		ingressList.GetRemainingItemCount()
+
+		for _, ingress := range ingressList.Items {
+			annotations := ingress.Annotations
+			if len(annotations) == 0 {
+				continue
+			}
+
+			for k, v := range annotations {
+				if strings.HasPrefix(k, ingressAnnotationPrefix) && geoipVarsRegexp.MatchString(v) {
```
yalosev, line 147: Why we couldn't use `... && strings.HasPrefix("$geoip_")` ?

modules/402-ingress-nginx/hooks/alert_deprecated_geoip_version.go
```diff
@@ -0,0 +1,180 @@
...
+
+	geoipVarsRegexp = regexp.MustCompile(`\$geoip_(country_(code3|code|name)|area_code|city_continent_code|city_country_(code3|code|name)|dma_code|latitude|longitude|region|region_name|city|postal_code|org)([^_a-zA-Z0-9]|$)+`)
+
+	// objectBatchSize - how many ingress objects to list from k8s at once
+	objectBatchSize = int64(30)
+	// fetchSecretsInterval pause between fetching ingress objects from apiserver
+	fetchSecretsInterval = 1 * time.Second
+)
+
+type ControllerVersion struct {
```
yalosev, line 72:
```suggestion
type controllerVersion struct {
```

don't think we have to export this

modules/402-ingress-nginx/hooks/alert_deprecated_geoip_version.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/flant/addon-operator/sdk"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/utils/pointer"
+
+	"github.com/deckhouse/deckhouse/go_lib/dependency"
+	"github.com/deckhouse/deckhouse/go_lib/dependency/k8s"
+)
+
+// TODO: Remove this migration hook after release 1.65
```
yalosev, line 38: Hm, what ?

modules/402-ingress-nginx/hooks/alert_deprecated_geoip_version.go
miklezzzz: just to avoid guessing possible variations of different annotations, I suppose. To make sure.

modules/402-ingress-nginx/hooks/alert_deprecated_geoip_version.go
miklezzzz: if we deprecate 1.6-1.9, we don't need this alert any longer, I hope it will happen somewhere around 1.6x :-D

yalosev (approved):

PR #8811, opened 2024-06-26
[testing] feat: matrix tests to check roles for wildcards

### LINKED ISSUE #8517: Minimize RBAC permissions of platform components
### Preflight Checklist

Kubernetes CIS [we recommend not using wildcard rights in Role/ClusterRole](https://github.com/aquasecurity/kube-bench/blob/0f8dfaf1156652b556c913aa8e78fccbb1dfd186/cfg/rke-cis-1.24/policies.yaml#L29-L35). 
However, the DKP platform does not take this recommendation into account everywhere. 
For example:
- dex has [verbs: *](https://github.com/deckhouse/deckhouse/blob/main/modules/150-user-authn/templates/dex/rbac-for-us.yaml#L53)
- upmeter-agent [has wildcard rights to all resources](https://github.com/deckhouse/deckhouse/blob/main/modules/500-upmeter/templates/upmeter-agent/rbac-for-us.yaml#L63-L73) in d8-upmeter namespace

Therefore, the DKP does not fully comply with the Kubernetes CIS recommendations.

### Proposed Solution

Grant specific required permissions to objects in the Kubernetes API.

### Additional Information

You might be interested in the [audit2rbac](https://github.com/liggitt/audit2rbac) tool to minimize permissions for the modules.

juev:
## Description

It is necessary to check for the presence of wildcards in Roles and ClusterRoles in the rbac-for-us.yaml files.

Kubernetes CIS [we recommend not using wildcard rights in Role/ClusterRole](https://github.com/aquasecurity/kube-bench/blob/0f8dfaf1156652b556c913aa8e78fccbb1dfd186/cfg/rke-cis-1.24/policies.yaml#L29-L35).

With each commit, template files are checked for the presence of wildcards. If no exception rule has been added, a validation error is generated.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: feature
summary: matrix tests to check roles for wildcards
impact_level: default
```

Closed #8517

### DIFF
--- a/testing/matrix/linter/rules/roles/wildcards.go
+++ b/testing/matrix/linter/rules/roles/wildcards.go
@@ -168,7 +168,7 @@ func checkRoles(object storage.StoreObject) errors.LintRuleError {
 				"WILDCARD001",
 				object.Identity(),
 				object.Path,
-				strings.Join(objs, ", ")+" contains a wildcards",
+				strings.Join(objs, ", ")+" contains a wildcards. Replace them with an explicit list of resources",
 			)
 		}
 	}
--- a/testing/matrix/linter/rules/roles/wildcards.go
+++ b/testing/matrix/linter/rules/roles/wildcards.go
@@ -17,16 +17,28 @@ limitations under the License.
 package roles
 
 import (
+	"slices"
 	"strings"
 
 	rbac "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/utils/strings/slices"
 
 	"github.com/deckhouse/deckhouse/testing/matrix/linter/rules/errors"
 	"github.com/deckhouse/deckhouse/testing/matrix/linter/storage"
 )
 
+// skipCheckWildcards is exclusion rules for wildcard verification
+//
+// the key is the file name
+// value is an array of rule names that allow wildcards
+var skipCheckWildcards = map[string][]string{
+	"admission-policy-engine/templates/rbac-for-us.yaml": {
+		"d8:admission-policy-engine:gatekeeper",
+	},
+}
+
+// ObjectRolesWildcard is a linter for checking the presence
+// of a wildcard in a Role and ClusterRole
 func ObjectRolesWildcard(object storage.StoreObject) errors.LintRuleError {
 	// check only `rbac-for-us.yaml` files
 	if !strings.HasSuffix(object.ShortPath(), "rbac-for-us.yaml") {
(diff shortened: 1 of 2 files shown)


testing/matrix/linter/rules/roles/wildcards.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+		}
+	}
+
+	converter := runtime.DefaultUnstructuredConverter
+
+	role := new(rbac.Role)
+	err := converter.FromUnstructured(object.Unstructured.UnstructuredContent(), role)
+	if err != nil {
+		panic(err)
```
juev, line 67: Here I took a sample from other linters. But in fact, we can return an error.

testing/matrix/linter/rules/roles/wildcards.go
```diff
@@ -0,0 +1,177 @@
...
+	for _, rule := range role.Rules {
+		var objs []string
+		if slices.Contains(rule.APIGroups, "*") {
+			objs = append(objs, "apiGroups")
+		}
+		if slices.Contains(rule.Resources, "*") {
+			objs = append(objs, "resources")
+		}
+		if slices.Contains(rule.Verbs, "*") {
+			objs = append(objs, "verbs")
```
yalosev, line 164: Let's separate the errors, please. We can give an advice to a user, how to fix the verbs wildcard
Something like
```
"Verb contains the wildcard. Please, use explicitly set list: `"create", "delete", "deletecollection", "get", "list", "patch", "update", "watch"`
```

testing/matrix/linter/rules/roles/wildcards.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+		if len(objs) > 0 {
+			return errors.NewLintRuleError(
+				"WILDCARD001",
+				object.Identity(),
+				object.Path,
+				strings.Join(objs, ", ")+" contains a wildcards",
```
yalosev, line 171:
```suggestion
				strings.Join(objs, ", ")+" contains a wildcards. Replace them with an explicit list of resources",
```

testing/matrix/linter/rules/roles/wildcards.go
juev: We can have wildcards in APIGroups, Resources and Verbs. If we separate the errors for each of these groups, should we provide information on the possible values ​​for each of the groups? Or will we do this only for Verbs?
Now we will have only one error for one role indicating in which groups we have wildcards.
If we divide errors into groups, we will have several times more errors. Is it worth doing?

yalosev (approved):

PR #8875, opened 2024-07-02
[dhctl] fix(dhctl-for-commander): store dhctl-server sockets in dhctl tmp dir

marriva:
## Description

Fix dhctl server sockets dir: use tmp dir.

Dhctl server creates unix sockets during its operation to communicate with its subprocesses. Use common tmp dhctl directory to store such files, rather than separate /var/run/dhctl directory, which was used before.

Otherwise we need to prepare dhctl image so that /var/run/dhctl directory exists in built image with correct permissions.

This fix is requred for dhctl server to operate normally in vanilla installer image, so it is preferred to include this fix into the patch release.

It is not expected to change any dhctl cli behaviour, only dhctl-server. 

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/dhctl/pkg/server/server/proxy.go
+++ b/dhctl/pkg/server/server/proxy.go
@@ -33,6 +33,7 @@ import (
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
+	"github.com/deckhouse/deckhouse/dhctl/pkg/app"
 	"github.com/deckhouse/deckhouse/dhctl/pkg/server/pkg/logger"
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:
Aye, aye, @distorhead. I've started the workflow for label 'e2e/run/azure'...

name212 (approved):

deckhouse-BOaTswain:
Cherry pick PR [8886](https://github.com/deckhouse/deckhouse/pull/8886) to the branch [release-1.62](https://github.com/deckhouse/deckhouse/tree/release-1.62) successful!

PR #8941, opened 2024-07-05
[ingress-nginx] add proxy connect timeout to auth locations

ipaqsa:
## Description
It provides the proxy connect timeout for auth locations.

There is no native way to configure the proxy connect timeout in auth locations.,So nginx uses default connect timeout which equals 60s, this is too long for auth servers, especially if the auth server is degraded.

The proxy connect timeout in auth locations is 15s instead of 60s. 

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: Set proxy connect timeout in auth locations to 15s.
impact: ingress-nginx controllers' pods will be recreated and in some cases, 15s would be not enough.
impact_level: low
```


deckhouse-BOaTswain:

miklezzzz (approved):

yalosev (approved):

Issue #8799, opened 2024-06-26
[deckhouse-release] suspend a module release with broken values


yalosev:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.61.X

### Expected Behavior

Module should be suspended according to [this](https://github.com/deckhouse/deckhouse/blob/main/deckhouse-controller/pkg/controller/module-controllers/release/updater.go#L149)

### Actual Behavior

Module is deployed

### Steps To Reproduce

1. create a module with invalid openapi spec / static values.yaml file
2. deploy it

### Additional Information

We have to add tests for release controller and override-controller

yalosev:
I would like to start with writing a test and then check, probably it's ok. But I was reported with this kind of a bug

PR #8989, opened 2024-07-10
[deckhouse-controller] feat: improve module validation

juev:
## Description

Modules may have an invalid configuration, and their activation may lead to a problem with the release.

As part of these changes, we are improving the verification of module configuration files. And if errors are detected, the release will be transferred to pending status.

Fixed: #8799 

Modules with incorrect configuration should be suspended

## Checklist
- [X] The code is covered by unit tests.
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Correct module validation.
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
@@ -918,12 +918,16 @@ func validateModule(def models.DeckhouseModuleDefinition) error {
 		return fmt.Errorf("cannot validate module without path. Path is required to load openapi specs")
 	}
 
-	dm, err := models.NewDeckhouseModule(def, addonutils.Values{}, nil, nil)
+	cb, vb, err := addonutils.ReadOpenAPIFiles(filepath.Join(def.Path, "openapi"))
+	if err != nil {
+		return fmt.Errorf("read open API files: %w", err)
+	}
+	dm, err := addonmodules.NewBasicModule(def.Name, def.Path, def.Weight, nil, cb, vb)
 	if err != nil {
 		return fmt.Errorf("new deckhouse module: %w", err)
 	}
 
-	err = dm.GetBasicModule().Validate()
+	err = dm.Validate()
 	if err != nil {
 		return fmt.Errorf("validate module: %w", err)
 	}
--- a/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
@@ -910,10 +910,6 @@ func (c *moduleReleaseReconciler) parseNotificationConfig(ctx context.Context) (
 }
 
 func validateModule(def models.DeckhouseModuleDefinition) error {
-	if def.Weight < 900 || def.Weight > 999 {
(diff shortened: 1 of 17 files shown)


deckhouse-controller/pkg/controller/module-controllers/release/controller.go
```diff
@@ -910,20 +910,20 @@ func (c *moduleReleaseReconciler) parseNotificationConfig(ctx context.Context) (
 }
 
 func validateModule(def models.DeckhouseModuleDefinition) error {
-	if def.Weight < 900 || def.Weight > 999 {
```
yalosev, line 913: Why have you removed this validation? We shouldn't pass external modules with some low weight, it can break everything

deckhouse-controller/pkg/controller/module-controllers/release/updater.go
```diff
@@ -134,30 +139,34 @@ func (k *kubeAPI) DeployRelease(release *v1alpha1.ModuleRelease) error {
...
+	}
 
-		return nil
+	moduleVersionPath := path.Join(k.externalModulesDir, moduleName, "v"+release.Spec.Version.String())
+	if err = os.RemoveAll(moduleVersionPath); err != nil {
+		return fmt.Errorf("cannot remove old module dir %q: %w", moduleVersionPath, err)
+	}
+	if err = copyDirectory(tmpModuleVersionPath, moduleVersionPath); err != nil {
+		return fmt.Errorf("copy module dir: %w", err)
 	}
```
yalosev, line 163: NIT:
I guess it will be secure to update the def.Path here
```
def.Path = moduleVersionPath
```

yes, it's not used below, but if someone use ModuleDefinition in this function, he will get an "interesting" error

deckhouse-controller/pkg/controller/module-controllers/release/override-controller.go
```diff
@@ -213,11 +218,15 @@ func (c *modulePullOverrideReconciler) moduleOverrideReconcile(ctx context.Conte
 	err = validateModule(*moduleDef)
 	if err != nil {
 		mo.Status.Message = fmt.Sprintf("validation failed: %s", err)
-		if e := c.updateModulePullOverrideStatus(ctx, mo); e != nil {
-			return ctrl.Result{Requeue: true}, e
-		}
+		_ = c.updateModulePullOverrideStatus(ctx, mo)
```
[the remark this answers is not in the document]
juev: fixed

deckhouse-controller/pkg/controller/module-controllers/release/updater.go
juev: Good point, corrected

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
juev: I rechecked the option with installation without overrides. In this case, the module weight is correctly determined and validation will proceed as expected.
Returned the weight check. But we need to recheck the behavior in the case of overrides. And the case when we create the module release ourselves.

deckhouse-BOaTswain:

yalosev (approved):

PR #9027, opened 2024-07-15
[monitoring] Remove trickster

Regrau:
## Description
This PR removes trickster from some modules.

Trickster and the aggregation proxy are solving the same problem. Hence removing trickster
and using the aggregation proxy helps remove redundancy.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: chore
summary: Remove Trickster.
impact: Prometheus will restart.
---
type: chore
summary: Use aggergating proxy instead of Trickster.
impact: prometheus-metrics-adapter will restart.
```


z9r5 (approved):

modules/110-istio/templates/kiali/rbac-for-us.yaml
```diff
@@ -95,10 +95,6 @@ rules:
   - tokenreviews
   verbs:
   - create
-- apiGroups: ["apps"]
```
vladimirGuryanov, line 98: If you remove only the RBAC for Trickster, Kiali won't work with the aggregator proxy. You need to update the RBAC settings for Kiali to work with the aggregator proxy.

modules/300-prometheus/templates/grafana/secret-datasources-list.yaml
```diff
@@ -19,11 +19,6 @@ secureJsonData:
 apiVersion: 1
 
 deleteDatasources:
-- name: trickster
```
vladimirGuryanov, line 22: This is a list of data sources pending deletion. Generally, this list includes data sources that have already been created and are scheduled for removal. You don't need to remove the Trickster data source from here.

modules/300-prometheus/templates/grafana/secret-datasources-list.yaml
```diff
@@ -37,14 +32,6 @@ deleteDatasources:
   {{- end }}
 
 datasources:
-- name: main-old
```
vladimirGuryanov, line 40: Needs to be moved to the 'Deleted' section, not deleted entirely.

modules/300-prometheus/templates/grafana/secret-datasources-list.yaml
```diff
@@ -54,16 +41,6 @@ datasources:
   isDefault: true
   {{- include "authorization_config" (list . (include "main_scrape_interval" . )) | nindent 2 }}
 
-  {{- if ne (int .Values.prometheus.longtermRetentionDays) 0 }}
```
vladimirGuryanov, line 57: We cannot delete this data source. If we do, users will lose the ability to retrieve data from Prometheus LTS. I suggest updating the aggregation proxy and adding a new server group for access to Prometheus LTS, similar to how it was configured in Trickster. Finally, we should update this data source accordingly.

vladimirGuryanov (changes requested):

apolovov (approved):

PR #9065, opened 2024-07-16
[network-policy-engine] Using the same iptables from registrypackages as in the host

apolovov:
## Description
Using the same iptables from registrypackages as in the host.

There is a [conflict](https://github.com/cloudnativelabs/kube-router/issues/1370) between kubelet and kube-router which leads to DROPping all the input and output traffic. They both work with the KUBE-FIREWALL iptables chain and kubelet creates the rule:
```
-A KUBE-FIREWALL -m comment --comment "kubernetes firewall for dropping marked packets" -m mark --mark 0x8000/0x8000 -j DROP
```
Meanwhile, the kube-router uses iptables-save/iptables-restore flow to sync his own rules and during restore, the `-m mark --mark 0x8000/0x8000` part is being lost. The result is lost connectivity between Nodes!

Fix is using the same iptables versions both in kubelet (host) and kube-router (image).

Also, the only affected and supported by d8 kubernetes version is v1.26, because there is `IPTablesOwnershipCleanup`  FeatureGate [disabled by default](https://github.com/kubernetes/kubernetes/blob/840d9bfe2ee5a9ca6d54435c598c2dc563f9cb70/pkg/features/kube_features.go#L1036C2-L1036C26). After v1.27 it is [enabled](https://github.com/kubernetes/kubernetes/blob/e596f2e471096ef7711b2f38529cbffb30801f14/pkg/features/kube_features.go#L1028) and [the DROP rules aren't being deployed](https://github.com/kubernetes/kubernetes/blob/e596f2e471096ef7711b2f38529cbffb30801f14/pkg/kubelet/kubelet_network_linux.go#L122).

Losing connectivity between Nodes is a great disaster. It is extremely important fix for network-policy-engine users.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: network-policy-engine
type: fix
summary: The kube-router image now use the same iptables binaries as on the host to prevent incompatibility.
```

### DIFF
--- a/modules/007-registrypackages/images/iptables/werf.inc.yaml
+++ b/modules/007-registrypackages/images/iptables/werf.inc.yaml
@@ -1,3 +1,4 @@
+{{/* Important! The iptables binaries from artifact are also used in kube-router image of network-policy-engine module. */}}
 {{- $version := "1.8.10" }}
 {{- $image_version := $version | replace "." "-" }}
 ---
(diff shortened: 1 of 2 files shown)


RomanenkoDenys (approved):
LGTM

deckhouse-BOaTswain:
Cherry pick PR [9068](https://github.com/deckhouse/deckhouse/pull/9068) to the branch [release-1.62](https://github.com/deckhouse/deckhouse/tree/release-1.62) successful!

Issue #9056, opened 2024-07-16
Use shell-operator image with customizable ports


yalosev:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

We need to change default ports for addon-operator

### Proposed Solution

After resolving the [issue](https://github.com/flant/shell-operator/issues/630)
We have to change image [here](https://github.com/deckhouse/deckhouse/blob/main/modules/000-common/images/shell-operator/werf.inc.yaml#L1)

PR #9083, opened 2024-07-18
feat: update shell-operator version

juev:
## Description

We need to change default ports for shell-operator.

We already have a new version of shell-operator with this functionality, and now we need to use it

Closed: #9056

This is necessary to be able to configure the standard ports that are used in the shell-operator.

After changing the version of shell-operator we are using, we have the opportunity to configure standard shell-operator ports through environment variables.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: tools
type: chore
summary: update shell-operator version
impact_level: low
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/fatih/color v1.13.0
 	github.com/flant/addon-operator v1.4.1
 	github.com/flant/kube-client v1.1.0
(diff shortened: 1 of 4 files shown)


miklezzzz:
go.mod?

yalosev (approved):

PR #9148, opened 2024-07-25
[node-manager] Alert for static nodes not managed by CAPS

borg-z:
Signed-off-by: borg-z <me@nikolay-z.top>

## Description

Alert that node is not controlled via CAPS.

We want all our k8s nodes to be managed via CAPS, static nodes without CAPS is an outdated approach.

If there is a Node in the cluster in Ready status and it does not have a corresponding Instance object, we consider that the node is not managed via CAPS and we need to send an alert about it

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: feature
summary: Alert about unmanaged CAPS nodes
impact: 
impact_level: default
```


deckhouse-BOaTswain:
Aye, aye, @borg-z. I've started the workflow for label 'e2e/run/gcp'...

PR #9198, opened 2024-07-29
[cilium][monitoring] Modify the CiliumAgentUnreachableHealthEndpoints alarm expression

tarabrind:
## Description
When pods start or restart in a cluster, cilium agents lose communication with each other and export the metric `cilium_unreachable_health_endpoints` to Prometheus. There are no settings in cilium to change this behavior (probe delay for just started agents, for example), and it is not possible to develop them due to internal API limitations.

It is necessary to change the conditions in the expression to avoid receiving false positive alarms from the monitoring system.

When pods are started or rebooted, the monitoring system will not register an alarm for 7 minutes after the changes. If there are no changes in the pod states, but the cilium agents have lost communication with each other, an alarm is sent through Prometheus Alarm Manager.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: cni-cilium
type: fix 
summary: Improved the CiliumAgentUnreachableHealthEndpoints metric expression to avoid false positives.
```

### DIFF
--- a/modules/021-cni-cilium/monitoring/prometheus-rules/agent.yaml
+++ b/modules/021-cni-cilium/monitoring/prometheus-rules/agent.yaml
@@ -1,7 +1,7 @@
 - name: cni-cilium
   rules:
   - alert: CiliumAgentUnreachableHealthEndpoints
-    # Show all current `cilium_unreachable_health_endpoints` if the condition is met:
+    # Select all current `cilium_unreachable_health_endpoints` if the condition is met:
     # no changes in pod states for the last N minutes (number of pods including changes - number of pods N minutes ago)
     expr: |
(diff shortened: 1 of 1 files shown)


modules/021-cni-cilium/monitoring/prometheus-rules/agent.yaml
```diff
@@ -1,7 +1,17 @@
 - name: cni-cilium
   rules:
   - alert: CiliumAgentUnreachableHealthEndpoints
-    expr: max by (namespace, pod) (cilium_unreachable_health_endpoints) > 0
+    # Show all current `cilium_unreachable_health_endpoints` if the condition is met:
```
apolovov, line 4:
```suggestion
    # Select all current `cilium_unreachable_health_endpoints` if the condition is met:
```

PR #9220, opened 2024-07-30
[cloud-provider-vcd] Added the ability to specify an organisation as the first part of the template path for master and cloud permanence node groups

name212:
## Description

Now template can contains 3 parts separated by `/`.
For example:
`MySuperOrg/Catalog/Template`

Also, 2 parted template is valid
`Catalog/Template`

Template can be shared from another organisation and we should handle this case.
For example, cloud provider can prepare one vm template and share it across all it clients.
In this case we should specify organisation where template is located. 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cloud-provider-vcd
type: feature
summary: Add the ability to specify an organisation as the first part of the template path for master and cloud permanence node groups.
impact_level: default
```


deckhouse-BOaTswain:
Aye, aye, @name212. I've started the workflow for label 'e2e/run/vcd'...

aleksey-su (approved):

z9r5 (approved):

PR #9234, opened 2024-07-31
[flannel] Removing the cilium configuration file on flannel startup

tarabrind:
## Description
When switching CNI from **cilium** to **flannel**, the **cilium** configuration file `/etc/cni/net.d/05-cilium.conflist` remains in the host file system.

Stale configuration file prevents CNI from working properly.

When starting **flannel**, the **cilium** configuration file `/etc/cni/net.d/05-cilium.conflist` will be deleted.
The file will not be deleted if both CNI **cilium** and **flannel** are enabled at the same time. 

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
section: flannel
type: chore
summary: Added deletion of stale cilium cni configuration file from the host file system when starting flannel.
impact_level: default
```

### DIFF
--- a/modules/035-cni-flannel/images/flanneld/entrypoint/main.go
+++ b/modules/035-cni-flannel/images/flanneld/entrypoint/main.go
@@ -77,6 +77,7 @@ func main() {
 	}
 
 	ciliumEnabledStr := os.Getenv("MODULE_CNI_CILIUM_ENABLED")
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
Aye, aye, @tarabrind. I've started the workflow for label 'e2e/run/yandex-cloud'...

apolovov (approved):

Issue #9074, opened 2024-07-17
[deckhouse-controller] Restart controller if a ModuleRelease is Deployed phase is deleted


miklezzzz:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.62

### Expected Behavior

If a deployed module release gets deleted, corresponding module should also be deleted from the ModuleManager and DeckhouseController to reflect new state of the cluster.

### Actual Behavior

If a deployed module release get deleted, it's only deleted from the file system. If it happens during initial module converge, ModuleManager knows nothing about it and tries to run helm upgrade, etc for the module.

### Steps To Reproduce

1. deploy a module from the ModuleRelease
2. enable the module
3. restart deckhouse
4. delete the deployed ModuleRelease as soon as possible

### Logs

```shell
{"event.id":"8058a167-25a8-458b-b260-9defa9214998","level":"info","msg":"head task ModuleRun for module 'echo', phase '' with doModuleStartup, trigger is ReloadAllModules","source":"DymicallyEnabledExtenderChanged","task.flow":"add","task.id":"a7c30160-3d20-4850-991a-47cafcf8f689","time":"2024-07-17T11:36:50Z","type":"ModuleScheduler event"}
{"event.id":"8058a167-25a8-458b-b260-9defa9214998","level":"info","module":"echo","msg":"ModuleRun task for module 'echo', phase 'Startup' with doModuleStartup, trigger is ReloadAllModules","source":"DymicallyEnabledExtenderChanged","task.flow":"start","task.id":"afe390f9-8664-46d1-80fa-4122e74107eb","time":"2024-07-17T11:39:20Z","type":"ModuleScheduler event"}
{"level":"info","msg":"module \"echo\" has neither Chart.yaml nor templates/ dir, is't not a helm chart","time":"2024-07-17T11:39:20Z"}
```
```

PR #9241, opened 2024-08-01
[deckhouse-controller] Restart controller if a ModuleRelease is Deployed phase is deleted

juev:
## Description

After removing the module release, we need to restart the controller to completely remove the module from the system.

Fixed: #9074

If a deployed module release gets deleted, corresponding module should also be deleted from the ModuleManager and DeckhouseController to reflect new state of the cluster.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Restart controller if a ModuleRelease is Deployed phase is deleted
impact_level: default
```


juev:
Tested:

1. Create ms and mup for echo module
2. Waiting for mr echo to be created
3. We are deleting mup and mr of the echo module
4. We observe the controller restart, after which echo disappears from the list of modules

Screenshot 2024-08-01 at 12 31 54

yalosev (approved):

PR #9254, opened 2024-08-02
[registrypackages][cni-simple-bridge] adding kernel version check for installing iptables and using this iptables bins in the simple-bridge image

sprait:
## Description

Add kernel version check for installing iptables. According to the documentation nftables is available upstream since Linux kernel 3.13. Also cni-simple-bridge use the same iptables binaries as on the host to prevent incompatibility.

On older kernels (e.g. centos7 with 3.10), there are problems with nftables, despite having the necessary kernel modules.

This issue may affect systems with kernel version < 3.14.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: registrypackages
type: fix
summary: Add kernel version check for installing iptables.
impact_level: default 
```
```changes
section: cni-simple-bridge
type: fix
summary: cni-simple-bridge use the same iptables binaries as on the host to prevent incompatibility.
impact_level: default 
```

### DIFF
--- a/modules/007-registrypackages/images/iptables/scripts/install
+++ b/modules/007-registrypackages/images/iptables/scripts/install
@@ -22,6 +22,9 @@ if [ -f /opt/deckhouse/bin/iptables ]; then
   exit 0
 fi
 
+NFT_KERNEL_LIMIT=3.14 # nf_tables_ipv6, nf_tables_bridge, nf_tables_arp allow since Linux kernel >= 3.14.
+CURRENT_KERNEL_VERSION=$(uname -r | awk -F"-" '{print $1}') #version kernel view format: a.b.c-generic to a.b.c.
+
 function checkNFTSupportInKernelBuildConfiguration() {
   local configNFTablesRegex='CONFIG_NF_(TABLES|TABLES_IPV4)=(y|m)'
   local configFiles=(
@@ -66,6 +69,22 @@ function checkNFTSupportByAddingNFTRules() {
   return 1
 }
 
+function check_python() {
+  for pybin in python3 python2 python; do
+    if command -v "$pybin" >/dev/null 2>&1; then
+      python_binary="$pybin"
+      return 0
+    fi
+  done
+  echo "Python not found, exiting..."
+  return 1
+}
+
+function isLegacyKernel() {
+  check_python
+  $python_binary -c "exit(0) if tuple(map(int, '$CURRENT_KERNEL_VERSION'.split('.'))) < tuple(map(int, '$NFT_KERNEL_LIMIT'.split('.')))  else exit(1)"
+}
+
 function isNftSupport() {
   if checkNFTSupportInKernelBuildConfiguration; then
     return 0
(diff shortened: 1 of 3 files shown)


modules/007-registrypackages/images/iptables/werf.inc.yaml
```diff
@@ -1,4 +1,4 @@
-{{/* Important! The iptables binaries from artifact are also used in kube-router image of network-policy-engine module. */}}
+{{/* Important! The iptables binaries from artifact are also used in kube-router image of network-policy-engine module and cni-simple-bridge. */}}
 {{- $version := "1.8.10" }}
```
apolovov, line 2:
```suggestion
{{- $version := "1.8.9" }}
```

modules/035-cni-simple-bridge/images/simple-bridge/werf.inc.yaml
```diff
@@ -1,16 +1,19 @@
+{{- $iptables_version := "1.8.10" }}
```
apolovov, line 1:
```suggestion
{{- $iptables_version := "1.8.9" }}
```

modules/007-registrypackages/images/iptables/werf.inc.yaml
apolovov, line 2:
```suggestion
{{- $version := "1.8.9" }}
```

modules/035-cni-simple-bridge/images/simple-bridge/werf.inc.yaml
```diff
@@ -25,7 +28,7 @@ import:
 docker:
   ENTRYPOINT: ["/sbin/iptables-wrapper"]
 ---
-{{ $simpleBridgeBinaries := "/bin/awk /bin/cat /bin/rm /bin/echo /usr/bin/tee /bin/sleep /bin/hostname /usr/bin/python3 /usr/bin/curl /usr/bin/jq /bin/bash /bin/grep /sbin/ip /usr/sbin/bridge /usr/bin/iptables* /sbin/iptables* /sbin/ip6tables* /sbin/xtables* /sbin/arptables*" }}
+{{ $simpleBridgeBinaries := "/bin/awk /bin/cat /bin/rm /bin/echo /usr/bin/tee /bin/sleep /bin/hostname /usr/bin/python3 /usr/bin/curl /usr/bin/jq /bin/bash /bin/grep /sbin/ip /usr/sbin/bridge "}}
```
apolovov, line 31:
```suggestion
{{ $simpleBridgeBinaries := "/bin/awk /bin/cat /bin/rm /bin/echo /usr/bin/tee /bin/sleep /bin/hostname /usr/bin/python3 /usr/bin/curl /usr/bin/jq /bin/bash /bin/grep /sbin/ip /usr/sbin/bridge" }}
```

apolovov (changes requested):

modules/007-registrypackages/images/iptables/scripts/install
```diff
@@ -66,6 +69,22 @@ function checkNFTSupportByAddingNFTRules() {
   return 1
 }
 
+function check_python() {
```
apolovov, line 72: seems overengineered.
mb bash version?
https://stackoverflow.com/a/4025065

sprait: This bash solution is bigger and less clear (IMHO). Python is a required dependency for DKP and I think it is better to use it.

apolovov (approved):

RomanenkoDenys (approved):

PR #9293, opened 2024-08-06
[deckhouse-controller] fix: improve module validation

juev:
## Description

In PR #8989, we made improvements to the validation of modules before installing them in the system. After the release, we found out that we had not foreseen a situation when a module config is created for the module to work, which should be used to validate the configuration. And if it is not loaded, an error occurs when checking the openapi configuration.

In this PR, the behavior is corrected. We check that the module has already been installed in the system. And if so, we load its module config and load this information into the module for validation.

The module must be checked in all cases. Including if the module was previously installed and the module config is already present.

Fixed: https://github.com/deckhouse/deckhouse/issues/8799

The module check is performed if the cluster already has a module config for it.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Improve module validation.
impact_level: default
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,6 +63,7 @@ require (
 	github.com/ettle/strcase v0.2.0
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/go-logr/logr v1.4.1
+	github.com/go-openapi/errors v0.19.7
 	github.com/gofrs/uuid/v5 v5.3.0
 	github.com/itchyny/gojq v0.12.15
 	github.com/jonboulle/clockwork v0.4.0
@@ -121,7 +122,6 @@ require (
 	github.com/go-gorp/gorp/v3 v3.1.0 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-openapi/analysis v0.19.10 // indirect
-	github.com/go-openapi/errors v0.19.7 // indirect
 	github.com/go-openapi/jsonpointer v0.19.6 // indirect
 	github.com/go-openapi/jsonreference v0.20.2 // indirect
 	github.com/go-openapi/loads v0.19.5 // indirect
--- a/tools/go.sum
+++ b/tools/go.sum
@@ -1061,10 +1061,6 @@ golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f/go.mod h1:oPkhp1MJrh7nUepCBc
 golang.org/x/sys v0.0.0-20220811171246-fbc7d0a398ab/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.2.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.6.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
-golang.org/x/sys v0.22.0 h1:RI27ohtqKCnwULzJLqkv897zojh5/DwS/ENaMzUOaWI=
-golang.org/x/sys v0.22.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
-golang.org/x/sys v0.20.0 h1:Od9JTbYCk261bKm4M/mw7AklTlFYIa0bIp9BgSm1S8Y=
-golang.org/x/sys v0.20.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
 golang.org/x/sys v0.23.0 h1:YfKFowiIMvtgl1UERQoTPPToxltDeZfbj4H7dVUCwmM=
 golang.org/x/sys v0.23.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
 golang.org/x/term v0.0.0-20201117132131-f5c789dd3221/go.mod h1:Nr5EML6q2oocZ2LXRh80K7BxOlk5/8JxuGnuhpl+muw=


juev:
After creating ms and map for module echo:
Screenshot 2024-08-07 at 10 08 21
Screenshot 2024-08-07 at 10 08 38

juev:
When creating an mpo for the virtualization module, the object is created and validation is correct:
Screenshot 2024-08-08 at 15 19 39
Screenshot 2024-08-08 at 16 34 51

Now we analyze errors that come from the validation side and ignore errors related to `required`.

miklezzzz:
@juev conflicts

deckhouse-controller/pkg/controller/module-controllers/release/updater.go
```diff
@@ -146,30 +153,39 @@ func (k *kubeAPI) DeployRelease(ctx context.Context, release *v1alpha1.ModuleRel
...
-		if e := k.UpdateReleaseStatus(release, "validation failed: "+err.Error(), release.Status.Phase); e != nil {
-			return e
-		}
+		_ = k.UpdateReleaseStatus(release, "validation failed: "+err.Error(), release.Status.Phase)
+		return fmt.Errorf("module '%s:v%s' validation failed: %s", moduleName, release.Spec.Version.String(), err)
+	}
 
-		return nil
+	moduleVersionPath := path.Join(k.downloadedModulesDir, moduleName, "v"+release.Spec.Version.String())
+	if err = os.RemoveAll(moduleVersionPath); err != nil {
```
miklezzzz, line 176: what's it? Why remove anything from /deckhouse/downloaded/<module_name> directory?
Frankly speaking, I didn't get the idea of using tmpDir instead of downloading right to /deckhouse/downloaded/. Could you please elaborate it a bit?

deckhouse-controller/pkg/controller/module-controllers/release/override-controller.go
```diff
@@ -211,14 +218,24 @@ func (c *modulePullOverrideReconciler) moduleOverrideReconcile(ctx context.Conte
...
 		mo.Status.Message = fmt.Sprintf("validation failed: %s", err)
 		if e := c.updateModulePullOverrideStatus(ctx, mo); e != nil {
-			return ctrl.Result{Requeue: true}, e
+			return ctrl.Result{Requeue: true}, fmt.Errorf("update override status: %w", e)
 		}
+		return ctrl.Result{}, fmt.Errorf("validation failed: %w", err)
+	}
 
-		return ctrl.Result{RequeueAfter: mo.Spec.ScanInterval.Duration}, nil
+	if err = os.RemoveAll(c.downloadedModulesDir); err != nil {
```
miklezzzz, line 234: looks suspiciously :thinking:

deckhouse-controller/pkg/controller/module-controllers/release/updater.go
juev: If you look at how we load the module into the file system now, you can see that we first delete the existing directory. And only after that we unpack it into the newly created directory.

[deckhouse-controller/pkg/controller/module-controllers/downloader/module_downloader.go](https://github.com/deckhouse/deckhouse/blob/main/deckhouse-controller/pkg/controller/module-controllers/downloader/module_downloader.go#L184)

This is done in order to correctly update the existing version and not get a module with the remaining files from the previous version variation.

Since we switched to using a temporary directory, this logic had to be pulled higher, from the function here.

deckhouse-controller/pkg/controller/module-controllers/release/override-controller.go
juev: Similar to the comment above,`(md *ModuleDownloader) storeModule` is also used with directory cleaning before copying. And because of the temporary directory, we need to do this ourselves.

miklezzzz: c.downloadedModulesDir is `/var/lib/deckhouse/downloaded` [directory](https://github.com/deckhouse/deckhouse/pull/9293/files/7279fc35d747bcddf9f1ee10b54c415237a895cd#diff-4fcd64acf898a13f2e50768163269ec98a0c0cd0f93bef40fa24515d107bed8cL76), I definitely don't want to have it cleaned every time an MPO is downloaded.

deckhouse-controller/pkg/controller/module-controllers/release/updater.go
miklezzzz: it's rather abnormal case when we already have a directory for x.y.z version of a module and the same version of module release is applied again. I suppose, that cleanup was added just to make sure we wouldn't have any intersections. Still can't wrap my head around the idea of using tmp directories.

juev: Are there any other ideas for implementing module validation without using temporary directories?

deckhouse-controller/pkg/controller/module-controllers/release/override-controller.go
juev: At the moment this is the implementation that already works. Do you want to change this behavior?

deckhouse-controller/pkg/controller/module-controllers/release/updater.go
miklezzzz: Using tmp directories in general is fine, I don't mind, just wanted to wrap my head around it, thanks.

miklezzzz (commented):
some questions

deckhouse-controller/pkg/controller/module-controllers/release/override-controller.go
juev: I apologize, it took me a long time to figure out what I did wrong. Today I figured it out. Thank you very much for noticing such a gross mistake on my part.

I fixed it and tested it on the dev cluster. Now validation in case of using mpo goes as expected.

miklezzzz:
let's fix some conflicts @juev  :)

juev:
DONE

miklezzzz (approved):

yalosev (approved):

deckhouse-BOaTswain:

yalosev:
/e2e/abort yandex-cloud;WithoutNAT;containerd;1.29 10826043152 10826043152-1-con-1-29 /sys/deckhouse-oss/install:pr9293 redos@89.169.128.161

deckhouse-BOaTswain:
Aye, aye, @yalosev. I've recognized your '/e2e/abort' command and started the workflow...

deckhouse-BOaTswain:

z9r5 (approved):

PR #9312, opened 2024-08-08
[tools] add ds_prometheus checks to grafana dashboard linter

ghostinsoba:
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```

### DIFF
--- a/tools/validation/grafana_dashboard.go
+++ b/tools/validation/grafana_dashboard.go
@@ -75,7 +75,9 @@ func isGrafanaDashboard(fileName string) bool {
 func validateGrafanaDashboardFile(fileName string, fileContent []byte) *Messages {
 	msgs := NewMessages()
 
-	dashboardPanels := extractDashboardPanels(fileContent)
+	dashboard := gjson.ParseBytes(fileContent)
+	dashboardPanels := extractDashboardPanels(dashboard)
+	dashboardTemplates := extractDashboardTemplates(dashboard)
 
 	for _, panel := range dashboardPanels {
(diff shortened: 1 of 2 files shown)


tools/validation/grafana_dashboard.go
```diff
@@ -135,13 +137,47 @@ func validateGrafanaDashboardFile(fileName string, fileContent []byte) *Messages
 				),
 			)
 		}
+		for _, datasourceUID := range invalidPrometheusDatasourceUIDs {
+			msgs.Add(
+				NewError(
+					fileName,
+					"invalid prometheus datasource uid",
+					fmt.Sprintf("Panel %s contains invalid datasource uid: '%s', required to be one of: %s",
```
vladimirGuryanov, line 145: Error message for Prometheus datasource UIDs currently describes certain UIDs as "invalid," even though they are valid but do not comply with our recommended standards (like `$ds_prometheus` or `${ds_prometheus}`). The use of the term "invalid" could be misleading, as it implies the UID is incorrect or non-functional, whereas it might just be non-standard.
I suggest rephrasing the error message to something like "non-recommended Prometheus datasource UID" instead of "invalid," to better communicate the issue to users. This would clarify that the UID is indeed valid but not in alignment with the best practices or conventions we've outlined.

ghostinsoba: fixed

deckhouse-BOaTswain:

vladimirGuryanov (approved):

PR #8867, opened 2024-07-01
[dhctl] add preflight check for python and his friends

mvasl:
## Description

Added a preflight check to validate that required python modules are installed on the node

## What is the expected result?

Installer will not begin bootstrapping nodes without required python modules

## Checklist
- [ ] The code is covered by unit tests.
- [x] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: dhctl
type: feature
summary: dhctl will now check if required python modules are installed on the node before bootstrapping it
impact_level: default
```

PR #9359, opened 2024-08-13
[dhctl] Fix static installation consume 100% of cpu

name212:
## Description
Add cpu profiling while DHCTL_TRACE=yes env provided.

While bootstrap static cluster dhctl start consume all cpu cores.
This will begin to appear after adding additional prefly checks https://github.com/deckhouse/deckhouse/pull/8867/files.
It happens  because reader in executor returns error which not handled.
Error raised because we was check interfaced variable on nil itself only, but should check on interfaced typed value.
Also we had leaking goroutines because we did not close pipes.

DHCTL should not consume a lot of cpu during work.

DHCTL should not consume a lot of cpu during work.
Affecter >= 1.62 DKP version

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Fix static installation consume 100% of cpu
impact_level: default
```

### DIFF
--- a/dhctl/cmd/dhctl/main.go
+++ b/dhctl/cmd/dhctl/main.go
@@ -15,7 +15,11 @@
 package main
 
 import (
+	"context"
 	"fmt"
+	"io"
+	"net/http"
+	"net/http/pprof"
 	"os"
 	"path"
 	"runtime/trace"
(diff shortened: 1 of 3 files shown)


deckhouse-BOaTswain:

deckhouse-BOaTswain:
Aye, aye, @name212. I've started the workflow for label 'e2e/run/yandex-cloud'...

RomanenkoDenys (approved):

z9r5:
Cherry-picked into release-1.63

PR #9478, opened 2024-08-24
[loki] Add support for config override

trublast:
## Description
This PR release introduces the ability to add a custom loki configuration that will be merged with the default configuration.
Currently, Loki runs with nearly default configuration, which is irrelevant to almost every real-world demand.

We don't
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: loki
type: feature
summary: Ability to add loki configuration parameters
impact: Loki pod will be restarted. Short disruption will occur.
impact_level: default
```


modules/462-loki/openapi/doc-ru-config-values.yaml
```diff
@@ -1,4 +1,8 @@
 properties:
+  configCustom:
+    type: string
+    description: >
+      Пользовательская конфигурация loki в формате yaml. Будет проведено слияние с конфигурацией по умолчанию.
```
z9r5, line 5:
```suggestion
      Пользовательская конфигурация loki в формате YAML.
      
      Конфигурация, указанная в параметре `configCustom`, будет объединена с конфигурацией по умолчанию.
```

modules/462-loki/openapi/config-values.yaml
```diff
@@ -1,5 +1,14 @@
 type: object
 properties:
+  configCustom:
+    type: string
+    description: >
+      Custom loki config on yaml format. Will be merged with default config.
```
z9r5, line 6:
```suggestion
      Custom loki config on YAML format.
      
      The configuration specified in the `configCustom` parameter will be merged with the default configuration.
```

z9r5 (approved):

Issue #6928, opened 2023-12-19
[user-authz] Forbidden access for logged in user if CAR has changed


meatuses:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.55

### Expected Behavior

If you apply changes to `ClusterAuthorizationRule`, new permissions should update accordingly for users logged in through `kubelogin`

### Actual Behavior

Changes to `ClusterAuthorizationRule` result in forbidden access for user until manual re-login.

### Steps To Reproduce

1. Create `user`, `group`, add `user` to `group`
2. Create ClusterAuthorizationrule, add `group` and set some permissions
3. Create config for the `user` via `kubeconfig-generator`
3. Use the config on a local machine, login with `kubelogin`+`dex`
4. Expected permissions are working
5. Make some edits: create `group2`, add the `user` to `group2`
6. `kubectl edit clusterauthorizationrule` edit the CAR, change `group` to `group2`
7.  No relogin via kubelogin, any action result in `forbidden` error, no permissions are applied.
8. Wait 10 minutes - indefinitely, still no permissions, no prompt to re-login via dex

### Additional Information

Cleaning up tokens from `~/.kube/cache/oidc-login/* ` helps to trigger a login and permissions would work again.

unreturned:
DKP v1.62.6 problem exists

PR #9598, opened 2024-09-04
[user-authn] feat: refresh groups on updating tokens

juev:
## Description

When we added groups to users, we did not add groups when refreshing the token. As a result, the groups were not updated if the token was refreshed.

Fixed: #6928

When updating a token, the user's groups are updated

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: feature
summary: refresh groups on updating tokens
impact_level: default
```

### DIFF
--- a/modules/150-user-authn/images/dex/patches/static-user-groups.patch
+++ b/modules/150-user-authn/images/dex/patches/static-user-groups.patch
@@ -1,112 +1,8 @@
-diff --git a/storage/kubernetes/types.go b/storage/kubernetes/types.go
---- a/storage/kubernetes/types.go	(revision 23efe9200ccd9e0a69242bf61cd221462370d1f4)
-+++ b/storage/kubernetes/types.go	(date 1718355089197)
-@@ -431,9 +431,10 @@
- 	// This field is IMMUTABLE. Do not change.
- 	Email string `json:"email,omitempty"`
-
--	Hash     []byte `json:"hash,omitempty"`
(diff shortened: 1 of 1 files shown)


juev:
# How I tested

1. First, we check the user's groups by retrieving data from the cache
2. We are deleting one of the groups
3. After 10 minutes, we make another request to k8s and check the groups in the cache

Screenshot 2024-09-06 at 18 03 27
Screenshot 2024-09-06 at 18 12 20

Screenshot 2024-09-06 at 18 02 24
Screenshot 2024-09-06 at 18 02 51

juev:
# How I tested

1. First, we check the user's groups by retrieving data from the cache
2. We are deleting one of the groups
3. After 10 minutes, we make another request to k8s and check the groups in the cache

Screenshot 2024-09-06 at 18 12 20

Screenshot 2024-09-06 at 18 02 24

miklezzzz (commented):
Conducted a couple of tests with documentation/dashboard endpoints and every time wasn't getting desired changes in the user's permissions via refresh tokens, even though the tokens themselves got updated with new group sets.

For example, after adding the user to some relevant group, corresponding refresh token gets new list of groups, but the current user's session still has no access, reporting no groups:
image
and if I start a new session via an incognito tab, it works as expected:
image

miklezzzz (approved):
tested two more times and everything went smoothly :thinking:

yalosev (approved):

PR #9732, opened 2024-09-16
[monitoring] Fix grafana home dashboard, panel `Controllers`

Suselz:
## Description
I have made a few changes in the `Controllers` panel in our Grafana home dashboard.
Previously, it was showing the number of pods managed by each controller type instead of the number of each controller type present. 
I have corrected this and now the dashboard shows accurate data: the amount of each type of controllers.

This change is important to get an accurate overview of the controllers present in our system. 
The previous misrepresentation might have led to confusion or misguided decisions. 
Now the dashboard will reflect the true state of our cluster by portraying the correct number of each type of controllers.

After the implementation of these changes, the `Controllers` panel in our Grafana home dashboard should provide a correct count of each type of controllers. 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: fix
summary: Fix grafana home dashboard, panel "Controllers", change promQL expression
impact_level: low 
```

### DIFF
--- a/modules/300-prometheus/images/grafana-v10/grafana_home_dashboard.json
+++ b/modules/300-prometheus/images/grafana-v10/grafana_home_dashboard.json
@@ -464,7 +464,7 @@
         {
           "editorMode": "code",
           "exemplar": true,
-          "expr": "count by (controller_type) (kube_controller_pod)",
+          "expr": "count by (controller_type) (count by (controller_type, controller_name) (kube_controller_pod))",
           "interval": "",
           "legendFormat": "__auto",
           "range": true,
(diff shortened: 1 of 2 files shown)


modules/300-prometheus/images/grafana-v10/grafana_home_dashboard.json
```diff
@@ -464,7 +464,7 @@
         {
           "editorMode": "code",
           "exemplar": true,
-          "expr": "count by (controller_type) (kube_controller_pod)",
+          "expr": "count by (controller_type) (count by (controller_type, controller_name) (kube_controller_pod))",
```
vladimirGuryanov, line 467: Hi! Thanks for the contribution.

The current PromQL might return incorrect results if there are controllers with the same name but in different namespaces. To fix this, please update the query like this:

```
count by (controller_type) (count by (controller_type, controller_name, namespace) (kube_controller_pod))
```

This ensures names are grouped by namespace to avoid collisions.

Thanks!

vladimirGuryanov (changes requested):

vladimirGuryanov (approved):

Issue #9582, opened 2024-09-03
[cert-manager] Split certificate status alerts and provide additional certificate status metric


Suselz:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

Our customers have experienced issues with the current alerting for certificate statuses in cert-manager. 

Currently, deckhouse two alerts set up - one for an expiring certificate, and one for an already expired certificate.
The alert for the expired certificate is designed in such a way that it's fired when a certificate transitions to the `notReady` state for a reason other than expiration. This is because, by design, cert-manager will set the lifespan metric of a certificate to 0 when the certificate state changes from `Ready` for any reason. Cert-manager do not have an additional metric for the textual status of the certificate.

Compounding this, when the alert for an expired certificate is fired, the alert for an expiring certificate simultaneously fires.

### Proposed Solution

In view of this, we would like to request the following:

1. Modify the PromQL queries so the alerts for an expiring certificate and an expired certificate do not fire simultaneously.
2. Implement an additional metric for the textual status of a certificate providing more granularity than the binary options of `Ready` or `notReady`. 
3. Split the alert for an expired certificate from the alert for a certificate being in the `notReady` state, leveraging the new metric from item 2 above.

deckhouse-BOaTswain:
This issue has been automatically put in the triage queue because it has not had
recent activity. The team will reconsider the status of this issue. Thank you
for your contributions.

raabdullaev:
https://github.com/deckhouse/deckhouse/pull/9751

PR #9751, opened 2024-09-17
[prometheus] Fix alerts for expiring and expired Certificates

Suselz:
## Description

This pull request is in relation to amendments made to promql queries for alerts indicating expiring and expired certificates. 
Previously, the `CertmanagerCertificateExpiredSoon` alert was triggered simultaneously with the alert indicating the certificate had expired (`CertmanagerCertificateExpired`).
Additional correction includes modifying the description of the alert that signifies an expired certificate. As it turns out, this alert was not only triggered when the certificate expired, but also when the condition became false for other reasons. 

Partially solves the problems highlighted in #9582

The prior misrepresentation of these alerts could lead to misinformation, and potentially, neglect of an urgent issue. This pull request solves the problem by ensuring the alerts are triggered at appropriate times and carry accurate descriptions.

The correction of these issues should result in more precise and relevant alerts.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: fix 
summary:  Fix alerts for expiring and expired Certificates.
impact_level: low
```


z9r5 (approved):

docs/documentation/_data/deckhouse-alerts.yml
```diff
@@ -15,9 +15,9 @@ alerts:
       moduleUrl: 101-cert-manager
       module: cert-manager
       edition: ce
-      description: Certificate {{$labels.exported_namespace}}/{{$labels.name}} expired
+      description: Certificate {{$labels.exported_namespace}}/{{$labels.name}} expired or has False condition due to another reason.
       summary: |
-        Certificate expired
```
nabokihms, line 20: What does it mean? Are there other reasons the condition can be False? Telling user that the condition is False is to broad, we have to be specific in alerts.

nabokihms (changes requested):

docs/documentation/_data/deckhouse-alerts.yml
Suselz: The reasons are partly described in #9582.

The main problem with the current alerts is that they light up not only when the certificate has expired. It also lights up when the certificate could not be ordered at all (and went into the `False` state). cert-manager does not have an additional label in the metric to distinguish these states.
As a quick fix, it was suggested to at least change the text of the alert to avoid misleading users about the reason for its triggering. For this reason, the alert description does not indicate the exact reason

miklezzzz: according to the code, the manager either sets the expiry metric to 0.0 or to the actual date from `NotAfter` status field
```
// updateCertificateExpiry updates the expiry time of a certificate
func (m *Metrics) updateCertificateExpiry(crt *cmapi.Certificate) {
        expiryTime := 0.0

        if crt.Status.NotAfter != nil {
                expiryTime = float64(crt.Status.NotAfter.Unix())
        }
```
could we alter the alert definition to exclude 0.0 value from matching the `< 1209600` condition  and introduce a new alert like `certificate isn't provided` or whatsoever for the `0,0` case?

miklezzzz (commented):
:thinking:

docs/documentation/_data/deckhouse-alerts.yml
miklezzzz: no, a description like 'Certificate has False condition' is too obscure.

Suselz: Okay, I'll change the alert title to the suggested option `certificate isn't provided`

Suselz: Done

miklezzzz: uh, it seems you didn't get my idea :(
I suggested having not one, but two alerts:
the first one, for the certificates that are about to expire - we check that reported value `< 1209600` and `!= 0.0` in this alert, it's description should stay as it is now.
the second one, that checks that the metric reports `0.0` value and it should have a description explaining that the certificate `wasn't provided` or whatsoever.
This way we have two not crossing conditions and two distinctive alerts.
Does it make sense?

Suselz: Run `make generate` and rebase changes

miklezzzz (approved):

deckhouse-BOaTswain:
Failure: cherry-pick commit 4654b9a30a52f4b6d94af2d90fb93ec1280a7a55 to the branch [release-1.67](https://github.com/deckhouse/deckhouse/tree/release-1.67) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/12880545678) for details.

z9r5:
Cherry-picked into release-1.67

Issue #4339, opened 2023-04-10
[ci] add a linter for detecting copy_custom_certificate if there is Ingress resource


apolovov:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

There was [a bug](https://github.com/deckhouse/deckhouse/pull/3985).

### Proposed Solution

Add a linter:
* If there is Ingress resource configured for a module — check if there is copy_custom_certificate.go hook.

deckhouse-BOaTswain:
This issue has been automatically put in the triage queue because it has not had
recent activity. The team will reconsider the status of this issue. Thank you
for your contributions.

PR #9825, opened 2024-09-24
[testing] feat: check copy_custom_certificate hook for ingresses

juev:
## Description

Fixed: #4339

We need to check that when using ingress, the copy_custom_certificate hook is used in the module.

When creating a PR in the repository, check the launch of matrix tests. If the module uses the ingress definition, check that the module hooks use copy_custom_certificate.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: feature
summary: check copy_custom_certificate hook for ingresses
impact_level: default
```

### DIFF
--- a/testing/matrix/linter/rules/modules/ingress.go
+++ b/testing/matrix/linter/rules/modules/ingress.go
@@ -52,7 +52,7 @@ func IngressHooksCheck(m utils.Module, object storage.StoreObject) errors.LintRu
 			"INGRESS",
 			m.Name,
 			nil,
-			"Ingress does not contains copy_custom_certificate hook",
+			"Ingress does not contain copy_custom_certificate hook",
 		)
 	}
 


testing/matrix/linter/rules/modules/ingress.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		maps.Copy(imports, p)
+	}
+
+	if _, ok := imports[copyCustomCertificateImport]; !ok {
+		return errors.NewLintRuleError(
+			"INGRESS",
+			m.Name,
+			nil,
+			"Ingress does not contains copy_custom_certificate hook",
```
miklezzzz, line 55:
```suggestion
			"Ingress does not contain copy_custom_certificate hook",
```

miklezzzz: :)

miklezzzz (approved):

PR #9943, opened 2024-10-02
[deckhouse-controller] feat: adding an alert that manual confirmation is required to install mr

juev:
## Description

If the module is set to manual installation mode, we need to approve its installation when a new version appears.

In order to know that a new version has appeared and that we need to check and approve these changes, we need to add an alert to Prometheus.

If the cluster contains modules that require confirmation of installation, then when a new version appears, an alert will appear in Prometheus.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: adding an alert that manual confirmation is required to install mr
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/release/controller.go
@@ -455,7 +455,7 @@ func (c *moduleReleaseReconciler) reconcilePendingRelease(ctx context.Context, m
 	}
 
 	k8 := newKubeAPI(ctx, c.logger, c.client, c.downloadedModulesDir, c.symlinksDir, c.moduleManager, c.dc)
-	releaseUpdater := newModuleUpdater(c.logger, nConfig, policy.Spec.Update.Mode, k8, c.moduleManager.GetEnabledModuleNames())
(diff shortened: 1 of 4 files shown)


juev:
# How I tested:

Created ms and mup for echo module in manual installation mode.

I checked that mr was created for the module, in approval waiting mode.

Then I checked that in Prometheus we had values ​​for the metric and an alert lit up.

Screenshot 2024-10-02 at 17 52 31
Screenshot 2024-10-02 at 17 52 47
Screenshot 2024-10-02 at 17 53 30

After that, I set the approval for MR, made sure that the module was installed and checked that in Prometheus, the values ​​for the metrics had disappeared and the alert had disappeared.

Screenshot 2024-10-02 at 18 00 14

miklezzzz (commented):
regarding the bottom part of the picture:
image
it shows $labes.name value for the d8_module label, possibly some variable resolving issue?

and as of the value of the metric, why is it 20? it should be 1 (gauge type of the metric?)

juev:
Another one:
Screenshot 2024-10-03 at 12 42 33
Screenshot 2024-10-03 at 12 43 16

Also changed the alert type to gauge and now the metric value is 1 if manual intervention is required.

miklezzzz (approved):

yalosev (approved):

PR #9974, opened 2024-10-02
[user-authn] Add claimMappingOverride option

nabokihms:
## Description
Add the new `claimMappingOverride` option that configures Dex's behavior.

With this option enabled, dex will use claims from claimMapping even if the token has a standard claim.
By default (when the option is disabled), dex tries to get the value of the default claim, e.g., email, and only if the claim is not present, falls back to claim mapping.

This is the case of one of our clients. They have no emails in tokens, but the email claim is present in the token and equals the `(unverified)` value, which does not work.

https://dexidp.io/docs/connectors/oidc/

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: feature
summary: Add claimMappingOverride option for OIDC Dex provider
```

### DIFF
--- a/modules/150-user-authn/crds/dex-provider.yaml
+++ b/modules/150-user-authn/crds/dex-provider.yaml
@@ -269,6 +269,13 @@ spec:
                           default: preferred_username
                           description: |
                             The [claim](https://openid.net/specs/openid-connect-core-1_0.html#Claims) to use as the user preferred username.
+                    claimMappingOverride:
+                      type: boolean
+                      description: |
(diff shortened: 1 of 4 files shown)


juev (approved):

deckhouse-BOaTswain:
Cherry pick PR [10074](https://github.com/deckhouse/deckhouse/pull/10074) to the branch [release-1.64](https://github.com/deckhouse/deckhouse/tree/release-1.64) successful!

PR #9994, opened 2024-10-04
chore: change e2e key

KraMorK:
## Description

Periodic replacement of the e2e key
!!Don't merge without change secret `LAYOUT_SSH_KEY`

Protect the stands from unauthorized access

Not related to release

The old key is out of service

```changes
section: ci
type: chore
summary: Periodic replacement of the e2e key
impact_level: low
```


name212:
Thank you for your conversation, but it is unneeded in current time

KraMorK:
it's time to rotate these keys

PR #10013, opened 2024-10-07
disable extended monitoring for standby-holder

borg-z:
## Description
Disabled the alert for `standby-holder-nodes` as it was generating unnecessary noise. It is recommended to use `d8_node_group-*` metrics for nodeGroup monitoring.

Fixes the issue of unnecessary alerts when the node group reaches maximum capacity.

The alert for `standby-holder-nodes` will no longer trigger when the node group reaches maximum capacity.

## Changelog entries

section: node-manager  
type: fix  
summary: "Disabled unnecessary alert for `standby-holder-nodes` when node group reaches maximum capacity."  
impact_level: low


deckhouse-BOaTswain:
Aye, aye, @borg-z. I've started the workflow for label 'e2e/run/yandex-cloud'...

name212 (approved):

Issue #5830, opened 2023-09-14
Running containerd alongside containerd-deckhouse causes problems


mvasl:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v.1.50.6

### Expected Behavior

Only `containerd-deckhouse.service` is allowed to run on provisioned Nodes

### Actual Behavior

Installation of `containerd` alongside `containerd-deckhouse` breaks CRI, causes undefined behaviour.

### Steps To Reproduce

1. Install Deckhouse as usual
2. Install Containerd on any node of running Deckhouse cluster and launch it

konstantin-axenov:
Connected with https://github.com/deckhouse/deckhouse/issues/6365

PR #10025, opened 2024-10-07
[bashible] Disable old or another containerd.service

trofimovdals:
## Description

We have some [issue](https://github.com/deckhouse/deckhouse/issues/5830) where two containerd services in single node. Since deckhouse only use `containerd-deckhouse.service` in actual version (since v1.50), we can turn off the other `containerd.service` before restarting this in step [033_restart_containerd_if_needed.sh
](https://github.com/deckhouse/deckhouse/blob/9228fea4b0083a84b83b813627dca8b53e940277/candi/bashible/common-steps/node-group/033_restart_containerd_if_needed.sh.tpl)
In the current situation, another running `containerd.service` breaks step `033_restart_containerd_if_needed` and the update freezes
## Changelog entries

```changes
section: bashible
type: fix
summary: Disable old or another containerd.service
impact_level: default
```

### DIFF
--- a/candi/bashible/common-steps/all/000_preflight_checks.sh
+++ b/candi/bashible/common-steps/all/000_preflight_checks.sh
@@ -19,9 +19,11 @@ if [[ "$FIRST_BASHIBLE_RUN" == "yes" ]]; then
     exit 1
   fi
 else
-  if systemctl is-enabled containerd >/dev/null 2>&1; then
-    bb-log-error "containerd.service is enabled on $HOSTNAME. Deckhouse use only containerd-deckhouse.service. Please disable/uninstall containerd.service to avoid further conflicts."
-    exit 1
+  if systemctl list-unit-files containerd.service >/dev/null 2>&1; then
+    if systemctl is-enabled containerd >/dev/null 2>&1 || systemctl is-active containerd >/dev/null 2>&1; then
+      bb-log-error "containerd.service is enabled or running on $HOSTNAME. Deckhouse use only containerd-deckhouse.service. Please disable/stop/uninstall containerd.service to avoid further conflicts."
+      exit 1
+    fi
   fi
 fi
 {{- end }}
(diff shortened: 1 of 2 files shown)


candi/bashible/common-steps/node-group/033_restart_containerd_if_needed.sh.tpl
```diff
@@ -17,6 +17,9 @@ if bb-flag? containerd-need-restart; then
   bb-log-warning "'containerd-need-restart' flag was set, restarting containerd."
   {{- if ne .runType "ImageBuilding" }}
   if out=$(containerd config dump 2>&1); then
+    if [[ -f /etc/systemd/system/containerd.service ]] || [[ -f /usr/lib/systemd/system/containerd.service ]] || [[ -f /lib/systemd/system/containerd.service ]]; then
```
name212, line 20: Better way move this code above, before `if bb-flag? containerd-need-restart; then`

name212 (changes requested):
one thing

candi/bashible/common-steps/node-group/033_restart_containerd_if_needed.sh.tpl
trofimovdals: We discussed what to leave like this

trofimovdals: Or we can check that the `containerd` process is running, turn it off and turn on `containerd-deckhouse` _always_ and not only when the flag is set

RomanenkoDenys: What do you think about move this code to the bashible first steps ?

RomanenkoDenys (changes requested):
one question

deckhouse-BOaTswain:
Aye, aye, @sprait. I've started the workflow for label 'e2e/run/openstack'...

sprait (approved):

name212 (approved):

PR #10033, opened 2024-10-08
[docs] Added description of overriding containerd parameters

alexey-gavrilov-flant:
## Description
Added description of overriding containerd parameters

I added an example with parameter overrides and a warning that such configurations will have to be maintained independently.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: docs
type: chore
summary: Added description of overriding containerd parameters.
impact_level:  low
```


modules/040-node-manager/docs/FAQ.md
```diff
@@ -784,6 +784,43 @@ Adding custom settings causes a restart of the `containerd` service.
 
 Bashible on nodes merges main Deckhouse containerd config with configs from `/etc/containerd/conf.d/*.toml`.
 
+You can override the values of parameters that are set in `/etc/containerd/deckhouse.toml`, but the work of such configurations will have to be provided independently. It is also better not to affect nodeGroup `master` c control-plane with such configurations.
```
name212, line 787: It will mark as warning. I do not know how we can do it. @z9r5 help us please

name212 (changes requested):
mark as warning

PR #10086, opened 2024-10-10
[cni-cilium] wiping unwanted iptables-legacy rules. cherry-pick to release-1.64

AndreyPavlovFlant:
## Description

The init-container for cilium-agent pod checks iptables-legacy for any unwanted rules and deletes them if found.

Previously, due to a configuration error, cilium-agent was always creating its rules in the iptables-legacy. 
The configuration error has now been fixed, and cilium-agent is creating its own rules in the correct table. 
However, previously created rules remain in the iptables-legacy table, even if the system is using iptables-nft.

Such behavior does not have a negative impact.
But having the rules in the wrong table can be confusing.

The rules created by the cilium-agent are present only in those iptables tables (nft or lagacy) that are used by the entire node.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cni-cilium
type: fix
summary: wiping unwanted iptables-legacy rules
impact: All cilium-agent pods will be restarted.
impact_level: default
```

### DIFF
--- a/modules/021-cni-cilium/images/agent/check-n-cleaning-iptables.sh
+++ b/modules/021-cni-cilium/images/agent/check-n-cleaning-iptables.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+# Copyright 2024 Flant JSC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


deckhouse-BOaTswain:
Aye, aye, @AndreyPavlovFlant. I've started the workflow for label 'e2e/run/openstack'...

apolovov (approved):

PR #10124, opened 2024-10-15
[operator-trivy] add support insecure registries

### LINKED ISSUE #9742: [operator-trivy] add support for scanning images from insecure registries
### Preflight Checklist

For some reasons, it might be required to scan images from registries with self-signed certificates.

### Proposed Solution

Update operator-trivy module's settings to support defining a list of insecure registries. This list should make its way to the trivy-operator configmap in the following form:
```
  ...
  trivy.insecureRegistry.1: registryHostName1
  trivy.insecureRegistry.2: registryHostName2
  ...
```

### Additional Information

_No response_

ipaqsa:
## Description
It provides support insecure registries. 

[This function](https://github.com/aquasecurity/trivy-operator/blob/18e40db60a957cf233afeb442bddca62a03e45e9/pkg/plugins/trivy/config.go#L461) looks for the prefix in the configmap data and creates a map with insecure registries. Then this map is used when [creating a job](https://github.com/aquasecurity/trivy-operator/blob/18e40db60a957cf233afeb442bddca62a03e45e9/pkg/plugins/trivy/jobspec.go#L105), if the image registry is present on this map, the operator adds TRIVY_INSECURE to the job.

Fixes #9742 

Insecure registries can be specified in the module config.

Example of the module config:
```
---
kind: ModuleConfig
metadata:
  name: operator-trivy
spec:
  enabled: true
  settings:
    insecureRegistries:
    - example.io
  version: 1
```

The configmap:

```
root@dev-master-0:~# kubectl get cm -n d8-operator-trivy trivy-operator-trivy-config -oyaml
apiVersion: v1
kind: ConfigMap
data:
  ...
  trivy.insecureRegistry.0: example.io
  ...
```

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: operator-trivy
type: feature
summary: Add support insecure registries.
```


ee/modules/500-operator-trivy/openapi/config-values.yaml
```diff
@@ -84,3 +84,12 @@ properties:
     x-example:
       - app
       - env
+  insecureRegistries:
+    type: object
+    additionalProperties:
+      type: string
```
yalosev, line 90: I'm totally can't understand what is `key` here. Is it some abstract string?
I guess it would be better to make it an array of strings. And use array index as `key`

ipaqsa: Yes, it is just abstract: https://github.com/aquasecurity/trivy-operator/blob/18e40db60a957cf233afeb442bddca62a03e45e9/pkg/plugins/trivy/config.go#L461

ipaqsa: i replaced it:

Screenshot 2024-10-16 at 12 54 39 AM

yalosev (approved):

miklezzzz (approved):
lacks examples, hate it so much :)

ee/modules/500-operator-trivy/openapi/config-values.yaml
```diff
@@ -84,3 +84,9 @@ properties:
     x-example:
       - app
       - env
+  insecureRegistries:
+    type: array
```
yalosev, line 88:
```suggestion
    type: array
    x-doc-example: [["my.registry.com", "http-only.registry.io"]]
```

yalosev (approved):

yalosev:
Waiting for documentation

z9r5 (approved):

PR #10198, opened 2024-10-18
[extended-monitoring] Add alert for hunged CronJobs

scaps1:
## Description
We noticed that when a CronJob is created in a namespace with resource quotas set, but resources for Job pods are not set, they start to accumulate in the status of pod creation and we don't have an alert for this case
To get alerts that we have stucked Jobs which consumes resources of cluster
No we don't
If CronJob has stucked Job pods in creating status it will trigger the new alert CronJobHungs

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: extended-monitoring
type: chore
summary: Add alert for hanged CronJobs
impact_level: low
```


deckhouse-BOaTswain:

deckhouse-BOaTswain:

z9r5 (approved):

vladimirGuryanov (approved):

PR #10208, opened 2024-10-21
[cni-cilium] Disable the metrics server in the egress-gateway-agent.

AndreyPavlovFlant:
## Description

Disable the metrics server in the `egress-gateway-agent` because we don't use it.

If the metrics server has not been explicitly turned off, it will by default be bound to *:8080. 
This can cause confusion, as if another application is already using this port, our application will not be able to start.

The `egress-gateway-agent` starts correctly even if port 8080 is already occupied by another application.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: cni-cilium
type: fix
summary: Disable the metrics server in the "egress-gateway-agent" because we don't use it.
impact: The pods of the egress-gateway-agent will be restarted.
impact_level: default 
```

### DIFF
--- a/ee/modules/021-cni-cilium/images/egress-gateway-agent/cmd/main.go
+++ b/ee/modules/021-cni-cilium/images/egress-gateway-agent/cmd/main.go
@@ -11,6 +11,8 @@ import (
 	"flag"
 	"os"
 
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
Aye, aye, @AndreyPavlovFlant. I've started the workflow for label 'e2e/run/vsphere'...

Beastlex (approved):

apolovov (approved):

Issue #7097, opened 2024-01-10
[upmeter] make the ingressClass configuration prettier


apolovov:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

There is only configuration section `smokeMini.ingressClass`, but the `smoke-mini` component hasn't an Ingress. But there are two Ingresses for `webui` and `status` without ingressClass configuration.

### Proposed Solution

To configure a common `upmeter.ingressClass` openapi section.

vladimirGuryanov:
Done in [PR](https://github.com/deckhouse/deckhouse/pull/10266)

PR #10266, opened 2024-10-23
[upmeter] Move ingressClass parameter from smokeMini section to root section.

kozharskyad:
## Description
Move ingressClass parameter from smokeMini section to root section.

The current implementation of the webui and status components includes ingress objects, yet lacks the ability to parametrize the ingressClassName. Conversely, the smokeMini component provides ingressClassName parametrization but does not offer an ingress object template. This pull request addresses the issue outlined in [https://github.com/deckhouse/deckhouse/issues/7097 ](https://github.com/deckhouse/deckhouse/issues/7097)and introduces a more streamlined approach to configuring the upmeter module's parameters. By relocating the ingressClass parameter from the smokeMini component to the root section of the module values, this change eradicates the redundant ingressClass parameter from smokeMini and harmonizes the configuration by adding this parameter to the webui, status, and any other potential module components requiring ingress. This enhancement not only simplifies the architecture but also enhances flexibility and consistency across the module components, ultimately delivering a more intuitive setup experience for end users.

-

Existence of ability to change ingress class name for upmeter module.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
Added ability to set ingress class for upmeter module.

```changes
section: upmeter
type: chore
summary: Added ability to set ingress class for upmeter module.
impact_level: low
```


z9r5 (approved):

modules/500-upmeter/openapi/conversions/v3.yaml
```diff
@@ -0,0 +1,4 @@
+version: 3
+conversions:
+  - if .smokeMini.ingressClass != null then .ingressClass = .smokeMini.ingressClass end
```
nabokihms, line 3:
```suggestion
  - if .smokeMini.ingressClass != null then .ingressClass = .smokeMini.ingressClass end
```
it seems like unneeded action because it was applied only for some mini. Making it the module ingress class may lead to unexpected side effects like unavailability of status or ui pages beacuse of the ingress class change.

My advise will be just to drop the smokeMini.ingressClass value in the new version of the config.

vladimirGuryanov (approved):

deckhouse-BOaTswain:

PR #10283, opened 2024-10-24
[candi][bashible] Added explicit addition of PACKAGES_PROXY_ADDRESSES into the NO_PROXY parameter

Suselz:
## Description

In the current implementation of DH, when installing in a closed environment, it is required to explicitly specify all single (not a network with a mask) IP addresses of master nodes in the no_proxy parameter, since the python library used in the `bb-package-fetch-blob` function cannot handle specifying a network with a mask (for example, `192.168.0.0/16`)

Since the `bb-package-fetch-blob` function is used only to receive packets from `bashible-apiserver`, I suggest explicitly adding the `bashible-apiserver` address to the `no_proxy` parameter within the function context.

With this change, it is no longer necessary to specify single (not a network with a mask) IP addresses of master nodes in the `no_proxy` parameter.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi 
type: fix 
summary:  Added explicit addition of basible-server address into the NO_PROXY parameter
impact_level: low
```

### DIFF
--- a/candi/bashible/bashbooster/50_registrypackages.sh
+++ b/candi/bashible/bashbooster/50_registrypackages.sh
@@ -58,7 +58,9 @@ function check_python() {
 bb-package-fetch-blob() {
   local REPOSITORY="${REPOSITORY:-}"
   local REPOSITORY_PATH="${REPOSITORY_PATH:-}"
-
+  local no_proxy=${PACKAGES_PROXY_ADDRESSES}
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:

sprait (approved):

deckhouse-BOaTswain:

name212 (approved):

PR #10486, opened 2024-11-06
[user-authn] Fix offline session updates

nabokihms:
## Description
Add a patch to fix the problem with offline sessions that are not created/updated properly, which causes random refresh problems.

https://github.com/dexidp/dex/pull/3828

It affects token refreshing in all clusters.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Add a patch to fix the problem with offline sessions that are not created/updated properly, which causes random refresh problems.
```


deckhouse-BOaTswain:

PR #10586, opened 2024-11-12
[deckhouse-controller] feat: compare crd spec, labels and annotations before update

juev:
## Description

Before installing crd into a cluster, we need to compare it with the existing one. Previously, we only compared spec and one of the labels. Now we will compare spec, the full set of labels and annotations.

Overview
Before installing crd into a cluster, we need to compare it with the existing one. Previously, we only compared spec and one of the labels. Now we will compare spec, the full set of labels and annotations.

What this PR does / why we need it
In order to backup the CRD cluster, we will install certain labels on the necessary resources. To be able to install labels, we need to check their presence in the cluster. So as not to delete them during the next update.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: compare crd spec, labels and annotations before update
impact_level: default
```


ldmonster (approved):

yalosev:
Bumped in the another PR

PR #10589, opened 2024-11-12
[dhct] Add human readable error on dhctl converge except

d2285:
## Description

this pr add human readable error when ssh host flag is not provided on dhctl converge

this will print human readable error if --ssh-host flag is not provided

output error is 
```
Not enough flags were passed to perform the operation.
Use dhctl converge --help to get available flags.
Ssh host flag is not provided. Need to pass --ssh-host
```
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix 
summary: Add human readable error on dhctl converge except
impact_level: default
```

### DIFF
--- a/dhctl/pkg/operations/converge/converger.go
+++ b/dhctl/pkg/operations/converge/converger.go
@@ -18,11 +18,10 @@ import (
 	"context"
 	"fmt"
 	"time"
-
+        "github.com/deckhouse/deckhouse/dhctl/pkg/kubernetes"
 	"github.com/google/uuid"
 
 	"github.com/deckhouse/deckhouse/dhctl/pkg/app"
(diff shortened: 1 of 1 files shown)


dhctl/pkg/operations/converge/converger.go
```diff
@@ -18,11 +18,10 @@ import (
 	"context"
 	"fmt"
 	"time"
-
+        "github.com/deckhouse/deckhouse/dhctl/pkg/kubernetes"
```
name212, line 21: imports are not sorted

name212 (changes requested):
one thing

dhctl/pkg/operations/converge/converger.go
d2285: fixed

PR #10688, opened 2024-11-15
[dhctl] Add tasks for moduleconfigs routines for post bootstrap and creating with resources phases

name212:
## Description

There is a situation where the cluster bootstrap terminates with an error. For this, we need to add module config `global` with parameter `modules.https.mode` `CustomCertificate` and CloudPermanent nodegroup `system` and for example `worker` with at least one replica. We have the following sequence:
- creating module configs
- cloud permanent node creation
- creating resources (such as the secret for `CustomCertificate` mode)

After applying module configs `global` with the `modules.https.mode` `CustomCertificate` parameter and creating nodegroup system, deckhouse hangs with the error

```
Queue 'main': length 30, status: 'sleep after fail for 32s (17s left of 32s delay)'
 1. ModuleHookRun:main:kubernetes:150-user-authn/hooks/https/copy_custom_certificate.go:custom_certificates:OperatorStartup:failures 7602:module hook '150-user-authn/hooks/https/copy_custom_certificate.go' failed: custom certificate secret name is configured, but secret with this name doesn't exis
```

To prevent this behaviour, dhctl introduced three types of specialised actions with module configs: preparing module config before applying it to the cluster, post action with moduleconfig at the stage of
resource creation and installation completion (the second case is needed to apply `releaseChannel` to module config deckhouse).
Now in the preparation phase for moduleconfig deckhouse `releaseChannel` is removed and remembered, for moduleconfig global `modules.https` is removed and remembered.
At the resource creation stage, `modules.https` is returned to moduleconfig global. Why is this necessary exactly at the resource creation stage? So that once the secret is created, the settings are immediately applied.
At the finalisation stage of the deployment, `releaseChannel` is returned to moduleconfig deckhouse. This is to ensure that deckhouse is not updated during the installation process. This mechanism was already implemented, but rewritten to the new approach for a better Developer Experience.

Also, the following has been done in this  pull request:
- the logic for populating deckhouse module config from configOverrides was removed, this method was removed and dead code remained
- some tests were crashing locally with panic because the /deckhouse/version and /deckhouse/candi/images_digests.json files were not created. This has been fixed
- also added a custom number of wait attempts for NodeGroup kind resources, as deckhouse-controller does not have time to set `status` in 1 attempt

Cluster bottstrap can failed in some cases.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Add tasks for moduleconfigs routines for post bootstrap and creating with resources phases.
impact_level: default
```

### DIFF
--- a/dhctl/pkg/config/deckhouse_config.go
+++ b/dhctl/pkg/config/deckhouse_config.go
@@ -119,6 +119,11 @@ func PrepareDeckhouseInstallConfig(metaConfig *MetaConfig) (*DeckhouseInstaller,
 	if metaConfig == nil {
 		return nil, fmt.Errorf("Internal error. Metaconfig is nil")
 	}
+
+	if len(metaConfig.DeckhouseConfig.ConfigOverrides) > 0 {
+		return nil, fmt.Errorf("Support for 'configOverrides' was removed. Please use ModuleConfig's instead.")
+	}
+
 	clusterConfig, err := metaConfig.ClusterConfigYAML()
 	if err != nil {
 		return nil, fmt.Errorf("Marshal cluster config failed: %v", err)
--- a/dhctl/pkg/config/deckhouse_config.go
+++ b/dhctl/pkg/config/deckhouse_config.go
@@ -70,7 +70,6 @@ type DeckhouseInstaller struct {
 	KubeadmBootstrap   bool
 	MasterNodeSelector bool
 
-	ReleaseChannel   string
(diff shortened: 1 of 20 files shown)


deckhouse-BOaTswain:
Aye, aye, @name212. I've started the workflow for label 'e2e/run/yandex-cloud'...

sprait (approved):

Issue #7552, opened 2024-02-19
[ingress-nginx] Clean up a host's iptables rules related to HostWithFailover inlet if such an inlet is removed


miklezzzz:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.57.0

### Expected Behavior

if a HostWithFailover inlet is removed, its iptables remainders also should be removed somehow. (a hook with OnAfterDeleteHelm + a job?)

### Actual Behavior

Proxy-failover-iptabes app of HostWithFailover inlet installs additional iptables rules and an ingress-failover chain required for handling traffic in case main controller isn't available. If we switch from HostWithFailover to something else, these rules remain in the host's iptables causing traffic disruption on ports 80 443. Possible workarounds are to reboot affected nodes or to flush their iptables and restart kube-proxy.

### Steps To Reproduce

1. create an ingress-nginx with inlet HostWithFailover
2. delete the created ingress-nginx
3. create an ingress-nginx with inlet HostPort using ports 80 and 443

### Additional Information

It also would be nice to improve proxy-failover-iptables app's logic so that it would check its iptables chain more thoroughly, reapplying missing rules and deleting excessive ones.

deckhouse-BOaTswain:
This issue has been automatically put in the triage queue because it has not had
recent activity. The team will reconsider the status of this issue. Thank you
for your contributions.

PR #10746, opened 2024-11-20
[ingress-nginx] feat: remove iptables rules

juev:
## Description

Fixed: #7552

Proxy-failover-iptabes app of HostWithFailover inlet installs additional iptables rules and an ingress-failover chain required for handling traffic in case main controller isn't available. If we switch from HostWithFailover to something else, these rules remain in the host's iptables causing traffic disruption on ports 80 443. Possible workarounds are to reboot affected nodes or to flush their iptables and restart kube-proxy.

If a HostWithFailover inlet is removed, its iptables remainders also should be removed somehow.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: remove iptables rules after deleting HostWithFailover inlet
impact_level: default
```


yalosev:
We don't want to realize it

PR #10747, opened 2024-11-20
Fix ssh become root

alladinattar:
## Description
When setting up a cluster, a problem occurs if the argument is --ssh-user=root and the flag --ask-become-pass is set. In this case, dhctl returns an error in preflight checks for static-cluster (Checking sudo is allowed for user)

This patch will allow preflight sudo-allowed check to pass without errors if cluster setup is performed by root user

## Checklist
- [ ] e2e tests passed.

section: dhctl
type: fix
summary: Added root user check when reading --ask-become-pass flag
impact_level: default


alladinattar:
Remove

PR #10751, opened 2024-11-20
[dhctl] Add root user check for ssh-become-pass flag

alladinattar:
## Description
Added user check for `--ask-become-pass` flag. If the user is root, the parameter value is set to false

When setting up a cluster as the root user and with the `--ssh-become-pass` parameter set, the preflight check fails, and the utility cannot check sudo availability.

Now the installer will not try to execute sudo as root and the installation will not fail with both the `--ask-become-pass` parameter and the --ssh-user=root parameter set simultaneously.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```
section: dhctl
type: fix
summary: Added root user check when reading --ask-become-pass flag
impact_level: low
```

### DIFF
--- a/dhctl/pkg/app/ssh.go
+++ b/dhctl/pkg/app/ssh.go
@@ -160,6 +160,7 @@ func DefineBecomeFlags(cmd *kingpin.CmdClause) {
 		_ = fmt.Errorf("flag --ssh-become-pass cannot be set for --ssh-user=root")
 		os.Exit(1)
 	}
+
 }
 
 func processConnectionConfigFile(sshFlagSetByUser bool, parser connectionConfigParser) error {


dhctl/pkg/app/ssh.go
```diff
@@ -154,6 +155,13 @@ func DefineBecomeFlags(cmd *kingpin.CmdClause) {
 		Envar(configEnvName("ASK_BECOME_PASS")).
 		Short('K').
 		BoolVar(&AskBecomePass)
+
+	cmd.PreAction(func(c *kingpin.ParseContext) error {
+		if SSHUser == "root" {
+			AskBecomePass = false
+		}
+		return nil
```
name212, line 163: I think we need to fail with error instead. Because for root user this flag should not set

name212 (changes requested):
one thing

dhctl/pkg/app/ssh.go
alladinattar: Agreed, made a new commit

sprait (approved):

deckhouse-BOaTswain:
Aye, aye, @alladinattar. I've started the workflow for label 'e2e/run/yandex-cloud'...

deckhouse-BOaTswain:
Aye, aye, @alladinattar. I've started the workflow for label 'e2e/run/gcp'...

miklezzzz:
wow

deckhouse-BOaTswain:
Aye, aye, @alladinattar. I've started the workflow for label 'e2e/run/yandex-cloud'...

dhctl/pkg/app/ssh.go
```diff
@@ -154,6 +155,12 @@ func DefineBecomeFlags(cmd *kingpin.CmdClause) {
 		Envar(configEnvName("ASK_BECOME_PASS")).
 		Short('K').
 		BoolVar(&AskBecomePass)
+
+	if SSHUser == "root" {
+		_ = fmt.Errorf("flag --ssh-become-pass cannot be set for --ssh-user=root")
```
sprait, line 160: 1.`--ssh-become-pass` flag does't exist, do you mean `--ask-become-pass`
2.  error message will never be displayed, because you don't send created error to output
3.  does this condition work exactly when the user is specified via `--ssh-user=root` (without ENV USER=root)? please, check it.

sprait (commented):
few things

dhctl/pkg/app/ssh.go
alladinattar: 1. I'll fix it.
2. I'll fix it.
3. As I understand it, the `DefineSSHFlags` function processes arguments. If `--ssh-user` is specified, the `SSHUser` variable will receive the value from the argument. That is, when checking the `SSHUser `variable, I always get either the `USER` environment variable or the value from `--ssh-user`. Or does it not work that way?

sprait: Looks like to me that the value of `SSHUser` will always contain value from environment variable. The value from argument `--ssh-user` will only be available after parsing flags, which has not yet happened at this phase.

PR #11028, opened 2024-12-10
[ingress-nginx] fix problem in controller with default backend for services with externalName type

Beastlex:
## Description
We are fixing a problem in the controller with incorrect attempts to resolve the endpoints addresses of the `Service` specified as the default backend.

We are fixing a problem with the controller when Ingress specifies `Service` with the `ExternalName` type as the main backend, and the default backend (using an annotation `nginx.ingress.kubernetes.io/default-backend `) - with the ClusterIP type. You can see the detailed cases here:
https://github.com/kubernetes/ingress-nginx/issues/12158
https://github.com/kubernetes/ingress-nginx/issues/12173

Successful operation of ingress, where service of external Name type is specified as the main backend, and service of ClusterIP type is specified as the default backend. It is often used to handle errors (404, 502) on a service with the ExternalName type.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ingress-nginx
type: fix
summary: Fixed controller behavior for Ingress with default backend
impact_level: default
```

### DIFF
--- a/modules/402-ingress-nginx/images/controller-1-9/patches/default-backend-fix.patch
+++ b/modules/402-ingress-nginx/images/controller-1-9/patches/default-backend-fix.patch
@@ -1,0 +1,12 @@
+diff --git a/internal/ingress/controller/controller.go b/internal/ingress/controller/controller.go
+index c1d4e4f..f105b9c 100644
+--- a/internal/ingress/controller/controller.go
++++ b/internal/ingress/controller/controller.go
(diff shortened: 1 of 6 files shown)


modules/402-ingress-nginx/images/controller-1-9/patches/default-backend-fix.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/internal/ingress/controller/controller.go b/internal/ingress/controller/controller.go
```
apolovov, line 1: README.md?

apolovov (approved):

Issue #7306, opened 2024-01-26
[prometheus] deckhouse_exported_domains creates invalid protocol on OnlyInURI


alexey-gavrilov-flant:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.57.0

### Expected Behavior

links with https protocol are generated

### Actual Behavior

links with http protocol are generated

### Steps To Reproduce

1. set https.mode: OnlyInURI

### Additional Information

https://github.com/deckhouse/deckhouse/blob/a4d39f52935e9415faa2d267f78fd15cc1c96048/modules/300-prometheus/hooks/metrics_web_interfaces.go#L94-L97
now the protocol depends only on ingress, but with OnlyInURI it is necessary to make https regardless of ingress.

PR #11077, opened 2024-12-13
[prometheus] Make metrics-web-interfaces hook consider  https.mode: OnlyInURI

ykav-flant:
## Description

The hook that generates metrics for service links 'Web Interfaces' Grafana panel now considers the https.mode: OnlyInURI parameter from the global configuration. If this parameter is set, the scheme for links is forced to https, even when TLS is not enabled in the ingress spec.

- This change ensures accurate generation of https links in metrics when https.mode: OnlyInURI is used
- Fixes https://github.com/deckhouse/deckhouse/issues/7306

Not necessarily.

- Metrics will correctly generate https links when https.mode: OnlyInURI is set, regardless of the ingress TLS spec.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: prometheus
type: fix 
summary: Ensure grafana `Web Interfaces` panel use 'https' in links when 'https.mode: OnlyInURI' is enabled
impact_level: low
```

<!---
`impact_level: default` adds to changelog as usual, this is the default that can be omitted
`impact_level: high`    something important for users, the impact will be copied to "Know Before Update" section
`impact_level: low`     omitted in changelog YAML; note there is `type:chore` for chores

### DIFF
--- a/modules/300-prometheus/hooks/metrics_web_interfaces.go
+++ b/modules/300-prometheus/hooks/metrics_web_interfaces.go
@@ -150,7 +150,7 @@ func filterIngress(obj *unstructured.Unstructured) (go_hook.FilterResult, error)
 func domainMetricHandler(input *go_hook.HookInput) error {
 	snap := input.Snapshots["ingresses"]
 	input.MetricsCollector.Expire("deckhouse_exported_domains")
-	globalHTTPSMode := input.ConfigValues.Get("global.https.mode").String()
+	globalHTTPSMode := input.ConfigValues.Get("global.modules.https.mode").String()
 
 	for _, sn := range snap {
 		if sn == nil {
--- a/modules/300-prometheus/hooks/metrics_web_interfaces.go
+++ b/modules/300-prometheus/hooks/metrics_web_interfaces.go
@@ -24,12 +24,9 @@ import (
 	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
 	"github.com/flant/addon-operator/pkg/module_manager/go_hook/metrics"
 	"github.com/flant/addon-operator/sdk"
-	"github.com/flant/shell-operator/pkg/kube_events_manager/types"
 	netv1 "k8s.io/api/networking/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


lazovskiy (changes requested):
Seems, we don't need a subscription to a ModuleConfig here.
Instead, we can rely on the `global.modules.https.mode` value which is available in any module and reflects the https mode selected in the cluster.

modules/300-prometheus/hooks/metrics_web_interfaces.go
```diff
@@ -43,28 +44,30 @@ import (
...
 					},
 				},
+				FilterFunc: filterIngress,
 			},
-			FilterFunc: filterIngress,
 		},
-	},
-}, domainMetricHandler)
+	}, domainMetricHandler)
+)
```
lazovskiy, line 70: This is a redundant change.

modules/300-prometheus/hooks/metrics_web_interfaces.go
```diff
@@ -147,13 +150,20 @@ func filterIngress(obj *unstructured.Unstructured) (go_hook.FilterResult, error)
...
 	for _, sn := range snap {
 		if sn == nil {
 			continue
 		}
 
 		domain := sn.(exportedWebInterface)
+
+		if globalHTTPSMode == "OnlyInURI" {
+			re := regexp.MustCompile(`^http://`)
+			domain.URL = re.ReplaceAllString(domain.URL, "https://")
```
lazovskiy, line 164: Do we need regular expressions here or can we handle it by using `strings.Replace` / `strings.ReplaceAll`?

ykav-flant: Fixed.

modules/300-prometheus/hooks/metrics_web_interfaces.go
ykav-flant: Fixed.

lazovskiy (approved):

vladimirGuryanov (approved):

deckhouse-BOaTswain:

deckhouse-BOaTswain:

deckhouse-BOaTswain:

deckhouse-BOaTswain:

PR #11112, opened 2024-12-17
chore: add dmt lint

KraMorK:
## Description

Current matrix tests do not cover all needs

For matrix tests an alternative solution is used

Not related to release

An alternative solution is used for tests
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ci
type: chore
summary: For matrix tests an alternative solution is used.
impact_level: low
```

### DIFF
--- a/werf.yaml
+++ b/werf.yaml
@@ -262,17 +262,11 @@ shell:
 
 ---
 artifact: golangci-lint-artifact
-from: {{ $.Images.BASE_GOLANG_23_ALPINE }}
+fromImage: common-base
 shell:
-  beforeInstall:
-  - apk add --no-cache git openssh-client
-  - mkdir -p ~/.ssh && echo "StrictHostKeyChecking accept-new" > ~/.ssh/config
   install:
-  # Use v1.60.1 for go1.23 - https://github.com/golangci/golangci-lint/issues/4837
-  - export GOPROXY={{ .GOPROXY }}
-  - git clone --depth 1 {{ .SOURCE_REPO }}/golangci/golangci-lint --branch v1.60.1
-  - cd golangci-lint/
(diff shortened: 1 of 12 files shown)


werf.yaml
```diff
@@ -262,17 +262,11 @@ shell:
...
-  beforeInstall:
-  - apk add --no-cache git openssh-client
-  - mkdir -p ~/.ssh && echo "StrictHostKeyChecking accept-new" > ~/.ssh/config
   install:
-  # Use v1.60.1 for go1.23 - https://github.com/golangci/golangci-lint/issues/4837
-  - export GOPROXY={{ .GOPROXY }}
-  - git clone --depth 1 {{ .SOURCE_REPO }}/golangci/golangci-lint --branch v1.60.1
-  - cd golangci-lint/
-  - CGO_ENABLED=0 GOOS=linux go build -ldflags '-s -w -extldflags "-static"' -o /usr/local/bin/golangci-lint cmd/golangci-lint/main.go
+  - curl -sL https://github.com/golangci/golangci-lint/releases/download/v1.60.1/golangci-lint-1.60.1-linux-amd64.tar.gz | tar -zx --strip-components 1 -C /tmp
```
name212, line 268: why do we change to download from github?

It did for sameness with cse and oss

name212 (changes requested):
one thing

werf.yaml
KraMorK: I will reply with a private message

KraMorK: We decided to take utilities that do not participate in the final images ready-made

name212 (approved):

yalosev (approved):

PR #11176, opened 2024-12-19
[CI] Add CIS label to e2e

YuryLysov:
## Description

Update CI templates to use label `e2e/use/cis` in e2e tests

If label set, we'll be able to deploy operator-trivy module into deployed cluster, to get CIS benchmark checks report

By setting label `e2e/use/cis` in PR, installation container in e2e test will start with environment variable `CIS_ENABLED` set to `true`

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ci
type: feature
summary: Update CI templates to set CIS_ENABLED in e2e tests to true if e2e/use/cis label set.
impact_level: low
```

### DIFF
--- a/.github/ci_templates/e2e_tests.yml
+++ b/.github/ci_templates/e2e_tests.yml
@@ -98,6 +98,9 @@
   REPOSITORY: ${{ github.repository }}
   DHCTL_LOG_FILE: ${{ steps.setup.outputs.dhctl-log-file}}
   GITHUB_TOKEN: ${{secrets.BOATSWAIN_GITHUB_TOKEN}}
+{!{- if eq $script_arg "run-test" }!}
+  CIS_ENABLED: ${{ needs.check_e2e_labels.outputs.cis }}
+{!{- end }!}
 run: |
 {!{- if eq $provider "eks" }!}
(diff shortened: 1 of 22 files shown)


.github/ci_templates/e2e_tests.yml
```diff
@@ -292,6 +302,7 @@ check_e2e_labels:
 {!{- end }!}
     edition: ${{ steps.check.outputs.edition }}
     multimaster: ${{ steps.check.outputs.multimaster }}
+    cis: ${{ steps.check.outputs.cis }}
```
Taior, line 305: You need to check for script_arg "run-test" here too

.github/ci_templates/e2e_tests.yml
```diff
@@ -98,6 +98,9 @@
   REPOSITORY: ${{ github.repository }}
   DHCTL_LOG_FILE: ${{ steps.setup.outputs.dhctl-log-file}}
   GITHUB_TOKEN: ${{secrets.BOATSWAIN_GITHUB_TOKEN}}
+{!{- if eq $script_arg "run-test" }!}
```
Taior, line 101: You should also check for `coll.Has $ctx "manualRun"`, here and everywhere else, as it shouldn't be present in automatic daily tests

name212 (approved):

PR #11304, opened 2024-12-25
[dhctl] Add imagesRegistry variable during the bootstrap

yalosev:
## Description
Add `imagesRegistry` to the `deckhouse-registry` secret

During the bootstrap we don't have this field. It's set in the [deckhouse](https://github.com/deckhouse/deckhouse/blob/main/modules/002-deckhouse/templates/registry-secret.yaml#L25) module.
So, we have a small period of time. If deckhouse will reboot in this interval, it can lead that we will have no value here and it will be passed to the [release controller ](https://github.com/deckhouse/deckhouse/blob/44b9f34c9acf4de67c9775fcf20a3a50edc04614/deckhouse-controller/pkg/controller/deckhouse-release/controller.go#L274). This case will change the deckhouse image to the `:v1.XX.Y` without any registry.

It would be better to set it on boostrap, because we can :)

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Fix the deckhouse-registry secret on bootstrap.
impact_level: low 
```

### DIFF
--- a/dhctl/pkg/kubernetes/actions/manifests/manifests.go
+++ b/dhctl/pkg/kubernetes/actions/manifests/manifests.go
@@ -527,11 +527,13 @@ func DeckhouseRegistrySecret(registry config.RegistryData) *apiv1.Secret {
 			apiv1.DockerConfigJsonKey: data,
 			"address":                 []byte(registry.Address),
 			"scheme":                  []byte(registry.Scheme),
+			"imagesRegistry":          []byte(registry.Address),
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


name212 (approved):

deckhouse-BOaTswain:
Cherry pick PR [11308](https://github.com/deckhouse/deckhouse/pull/11308) to the branch [release-1.67](https://github.com/deckhouse/deckhouse/tree/release-1.67) successful!

PR #11373, opened 2024-12-27
[monitoring-ping] cve update

alexey-gavrilov-flant:
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


modules/340-monitoring-ping/images/monitoring-ping/werf.inc.yaml
```diff
@@ -20,14 +37,15 @@ shell:
...
 image: {{ .ModuleName }}/{{ .ImageName }}
 fromImage: common/python-static
-git:
-- add: /{{ $.ModulePath }}modules/340-{{ $.ModuleName }}/images/{{ $.ImageName }}/
-  to: /src
-  stageDependencies:
-    install:
-    - '**/*'
+mount:
+- from: tmp_dir
```
RomanenkoDenys, line 41: What the tmp_dir ?

RomanenkoDenys (changes requested):
one nit

modules/340-monitoring-ping/images/monitoring-ping/werf.inc.yaml
alexey-gavrilov-flant: This is a temporary folder for the duration of the build. Its contents are not included in the image. This addition speeds up the build because less unnecessary information is crammed into the registry. 
It may be worth making generalized include for go, python and node.

RomanenkoDenys: we cannot use this solution, and build image only one time. I think we cannot use that hack.

RomanenkoDenys: btw, we install only one wheel. I think we should prebuild it and put to the https://fox.flant.com/deckhouse/3p/python-modules/wheels.
We have prebuild wheel - https://fox.flant.com/deckhouse/3p/python-modules/wheels/-/blob/main/prometheus_client-0.5.0-py3-none-any.whl?ref_type=heads. Looks to another modules which uses python please.

RomanenkoDenys (changes requested):
some nit

RomanenkoDenys (approved):

PR #11417, opened 2025-01-09
[prometheus] Patch out grafana CVEs

lazovskiy:
## Description
Patch out grafana vulnerabilities to pass certification.

This PR fixes all possible vulnerabilities in grafana image found by trivy scanner. The only vulnerability that is not fixed is [CVE-2024-8986](https://avd.aquasec.com/nvd/cve-2024-8986). This issue however has no impact on a Deckhouse because the build process does not imply passing any credentials via the git URL.

We don't

The number of vulnerabilities is reduced to a minimum

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: fix
summary: Fix grafana CVEs
impact: grafana in cluster will be restarted as the image has changed
impact_level: default
```

### DIFF
--- a/modules/300-prometheus/images/grafana-v10/werf.inc.yaml
+++ b/modules/300-prometheus/images/grafana-v10/werf.inc.yaml
@@ -2,7 +2,8 @@
 {{ $grafanaVersion := "10.4.14" }}
 {{ $bundledPlugins := "petrslavotinek-carpetplot-panel,vonage-status-panel,btplc-status-dot-panel,natel-plotly-panel,savantly-heatmap-panel,grafana-piechart-panel,grafana-worldmap-panel" }}
 ---
-artifact: {{ $.ModuleName }}/{{ $.ImageName }}-src-files
+image: {{ $.ModuleName }}/{{ $.ImageName }}-src-files
+final: false
 from: {{ $.Images.BASE_ALPINE_DEV }}
 git:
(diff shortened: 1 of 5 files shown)


modules/300-prometheus/images/grafana-v10/werf.inc.yaml
```diff
@@ -79,14 +70,10 @@ import:
   add: /usr/src/app/bin/linux-amd64/grafana
   to: /usr/share/grafana/bin/grafana
   before: install
-- artifact: {{ $.ModuleName }}/{{ $.ImageName }}-js-builder
-  add: /usr/src/app/public
+- artifact: {{ $.ModuleName }}/{{ $.ImageName }}-src-files
```
RomanenkoDenys, line 73:
```suggestion
- image: {{ $.ModuleName }}/{{ $.ImageName }}-src-files
```

RomanenkoDenys (changes requested):
one nit

modules/300-prometheus/images/grafana-v10/werf.inc.yaml
lazovskiy: Changed to `image` + `final: false`

vladimirGuryanov (approved):

RomanenkoDenys (approved):

PR #11659, opened 2025-01-22
[deckhouse-controller] Recreate missing deployed release

### LINKED ISSUE #11596: [deckhouse-controller] Recover Deployed DeckhouseRelease
### Preflight Checklist

There's can be situation, when we have no DeckhouseRelease resource in cluster with Deployed state.
Without it, we will have unpredictable consequences.

f.e. 
1) we do not see actual version in cluster and automatic deckhouse-release reconcile can stuck
2) we can accidentally force release and corrupt resources in cluster
3) ...

### Proposed Solution

It will be good to try to recover this information from:
1) Actual deckhouse Deployment
2) From filesystem
and create DeckhouseRelease resource with corresponding version.

### Additional Information

_No response_

miklezzzz:
## Description

Implements the idea that having a deployed DeckhouseRelease object in the cluster is mandatory.
Closes #11596 

Always having a deployed DeckhouseRelease object in the cluster facilitates:
- understanding which version the cluster is running on (from the users' point of view);
- calculating which deckhouse release should be applied next (from the controller's point of view).

If there is no deployed DeckhouseRelease in the cluster, it's recreated either using the corresponding release metadata from the registry or the current version of the deckhouse controller, in case the corresponding metadata can't be fetched.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: chore
summary: Recreate missing deployed release.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/pkg/controller/deckhouse-release/controller.go
+++ b/deckhouse-controller/pkg/controller/deckhouse-release/controller.go
@@ -583,7 +583,7 @@ func (r *deckhouseReleaseReconciler) ApplyRelease(ctx context.Context, dr *v1alp
 func (r *deckhouseReleaseReconciler) runReleaseDeploy(ctx context.Context, dr *v1alpha1.DeckhouseRelease, deployedReleaseInfo *d8updater.ReleaseInfo) error {
 	r.logger.Info("applying release", slog.String("name", dr.GetName()))
 
-	err, dryrun := r.bumpDeckhouseDeployment(ctx, dr)
+	dryrun, err := r.bumpDeckhouseDeployment(ctx, dr)
 	if err != nil {
 		return fmt.Errorf("deploy release: %w", err)
 	}
@@ -642,19 +642,21 @@ func (r *deckhouseReleaseReconciler) runReleaseDeploy(ctx context.Context, dr *v
 
 var ErrDeploymentContainerIsNotFound = errors.New("deployment container is not found")
 
-func (r *deckhouseReleaseReconciler) bumpDeckhouseDeployment(ctx context.Context, dr *v1alpha1.DeckhouseRelease) (error, bool /* dryrun? */) {
+func (r *deckhouseReleaseReconciler) bumpDeckhouseDeployment(ctx context.Context, dr *v1alpha1.DeckhouseRelease) (bool /* dryrun? */, error) {
+	var dryrun bool
 	key := client.ObjectKey{Namespace: deckhouseNamespace, Name: deckhouseDeployment}
 
 	depl := new(appsv1.Deployment)
 
 	err := r.client.Get(ctx, key, depl)
 	if err != nil {
-		return fmt.Errorf("get deployment %s: %w", key, err), false
+		return dryrun, fmt.Errorf("get deployment %s: %w", key, err)
 	}
 
 	// dryrun for testing purpose
 	val, ok := dr.GetAnnotations()[v1alpha1.DeckhouseReleaseAnnotationDryrun]
 	if ok && val == "true" {
+		dryrun = true
 		go func() {
 			r.logger.Debug("dryrun start soon...")
 
@@ -704,10 +706,10 @@ func (r *deckhouseReleaseReconciler) bumpDeckhouseDeployment(ctx context.Context
 			}
 		}()
 
-		return nil, true
+		return dryrun, nil
 	}
 
-	return ctrlutils.UpdateWithRetry(ctx, r.client, depl, func() error {
+	return dryrun, ctrlutils.UpdateWithRetry(ctx, r.client, depl, func() error {
 		if len(depl.Spec.Template.Spec.Containers) == 0 {
 			return ErrDeploymentContainerIsNotFound
 		}
@@ -715,7 +717,7 @@ func (r *deckhouseReleaseReconciler) bumpDeckhouseDeployment(ctx context.Context
 		depl.Spec.Template.Spec.Containers[0].Image = r.registrySecret.ImageRegistry + ":" + dr.Spec.Version
 
 		return nil
-	}), false
+	})
 }
 
 func (r *deckhouseReleaseReconciler) getDeckhouseLatestPod(ctx context.Context) (*corev1.Pod, error) {
--- a/deckhouse-controller/pkg/controller/deckhouse-release/controller.go
+++ b/deckhouse-controller/pkg/controller/deckhouse-release/controller.go
@@ -221,7 +221,7 @@ func (r *deckhouseReleaseReconciler) createOrUpdateReconcile(ctx context.Context
 				return res, err
 			}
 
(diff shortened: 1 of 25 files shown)


deckhouse-controller/pkg/controller/deckhouse-release/controller.go
```diff
@@ -84,12 +85,20 @@ type deckhouseReleaseReconciler struct {
...
 	moduleManager moduleManager, updateSettings *helpers.DeckhouseSettingsContainer, metricStorage *metricstorage.MetricStorage,
-	preflightCountDown *sync.WaitGroup, logger *log.Logger,
+	preflightCountDown *sync.WaitGroup, deckhouseVersion string, logger *log.Logger,
 ) error {
+
+	parsedVersion, err := semver.NewVersion(deckhouseVersion)
+	if err != nil {
+		return fmt.Errorf("parse deckhouse version: %w", err)
+	}
+
```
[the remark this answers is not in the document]
miklezzzz: thought about it, but we also need to cut off the Metadata, so the way becomes quite wordy, that's why I decided to stop on using Sprintf :)

deckhouse-BOaTswain:

deckhouse-controller/pkg/controller/deckhouse-release/controller.go
```diff
@@ -571,18 +576,22 @@ func (r *deckhouseReleaseReconciler) ApplyRelease(ctx context.Context, dr *v1alp
...
 	r.logger.Info("applying release", slog.String("name", dr.GetName()))
 
-	err := r.bumpDeckhouseDeployment(ctx, dr)
+	dryrun, err := r.bumpDeckhouseDeployment(ctx, dr)
 	if err != nil {
 		return fmt.Errorf("deploy release: %w", err)
 	}
 
+	if dryrun {
+		return nil
```
ldmonster, line 592: now we can't test full update cycle with dryrun.
looks like we need another annotation

deckhouse-controller/pkg/controller/deckhouse-release/controller.go
```diff
@@ -633,19 +642,21 @@ func (r *deckhouseReleaseReconciler) runReleaseDeploy(ctx context.Context, dr *v
...
 
 	err := r.client.Get(ctx, key, depl)
 	if err != nil {
-		return fmt.Errorf("get deployment %s: %w", key, err)
+		return dryrun, fmt.Errorf("get deployment %s: %w", key, err)
 	}
 
 	// dryrun for testing purpose
-	val, ok := dr.GetAnnotations()[deckhouseReleaseAnnotationDryrun]
+	val, ok := dr.GetAnnotations()[v1alpha1.DeckhouseReleaseAnnotationDryrun]
```
ldmonster, line 657: if we need bool - it will be better to add getter to deckhouse release resource.
looks like we can make code more clear with it

deckhouse-controller/pkg/controller/deckhouse-release/controller.go
```diff
@@ -633,19 +642,21 @@ func (r *deckhouseReleaseReconciler) runReleaseDeploy(ctx context.Context, dr *v
 
 var ErrDeploymentContainerIsNotFound = errors.New("deployment container is not found")
 
-func (r *deckhouseReleaseReconciler) bumpDeckhouseDeployment(ctx context.Context, dr *v1alpha1.DeckhouseRelease) error {
+func (r *deckhouseReleaseReconciler) bumpDeckhouseDeployment(ctx context.Context, dr *v1alpha1.DeckhouseRelease) (bool /* dryrun? */, error) {
```
ldmonster, line 645: can we parse dryrun before calling this function and do not return it here?

deckhouse-controller/pkg/controller/deckhouse-release/controller.go
```diff
@@ -633,19 +642,21 @@ func (r *deckhouseReleaseReconciler) runReleaseDeploy(ctx context.Context, dr *v
...
+	var dryrun bool
 	key := client.ObjectKey{Namespace: deckhouseNamespace, Name: deckhouseDeployment}
 
 	depl := new(appsv1.Deployment)
 
 	err := r.client.Get(ctx, key, depl)
 	if err != nil {
-		return fmt.Errorf("get deployment %s: %w", key, err)
+		return dryrun, fmt.Errorf("get deployment %s: %w", key, err)
```
ldmonster, line 653: if we make return values like this - it looks more clear.
(because we do not need to calculate function result)
```suggestion
		return false, fmt.Errorf("get deployment %s: %w", key, err)
```

deckhouse-BOaTswain:

deckhouse-controller/pkg/controller/deckhouse-release/controller.go
```diff
@@ -206,20 +211,33 @@ func (r *deckhouseReleaseReconciler) createOrUpdateReconcile(ctx context.Context
...
+		if dr.GetCurrentRestored() {
+			dr.Status.Approved = true
+			dr.Status.Phase = v1alpha1.DeckhouseReleasePhaseDeployed
+			dr.Status.TransitionTime = metav1.NewTime(r.dc.GetClock().Now().UTC())
+			dr.Status.Message = "Release object was restored"
+			if err := r.client.Status().Update(ctx, dr); err != nil {
+				return res, err
+			}
+
+			return ctrl.Result{Requeue: false}, nil
```
ldmonster, line 224:
```suggestion
			return ctrl.Result{}, nil
```

deckhouse-controller/pkg/controller/deckhouse-release/controller.go
```diff
@@ -206,20 +211,33 @@ func (r *deckhouseReleaseReconciler) createOrUpdateReconcile(ctx context.Context
 	switch dr.Status.Phase {
 	// these phases should be ignored by predicate, but let's check it
 	case "":
+		// set current restored release as deployed
+		if dr.GetCurrentRestored() {
+			dr.Status.Approved = true
```
ldmonster, line 216: can we move all con inside if statement to function like
```go
	return r.proceedRestoredRelease(ctx, dr)
```

deckhouse-controller/pkg/controller/deckhouse-release/controller.go
```diff
@@ -644,8 +662,8 @@ func (r *deckhouseReleaseReconciler) bumpDeckhouseDeployment(ctx context.Context
 	}
 
 	// dryrun for testing purpose
-	val, ok := dr.GetAnnotations()[deckhouseReleaseAnnotationDryrun]
-	if ok && val == "true" {
+	dryrun := dr.GetDryRun()
+	if dryrun {
```
ldmonster, line 666:
```suggestion
	if dr.GetDryRun() {
```

ldmonster (approved):

yalosev (approved):

PR #11694, opened 2025-01-23
[prometheus] Improve prometheus rolling update

lazovskiy:
## Description
Deckhouse prefers to run two Prometheus replicas when possible. In conjunction with the aggregation proxy, it gives some redundancy. It delivers the end user gapless charts in a cluster's Grafana even if one Prometheus replica is down for some period.

When the Prometheus StatefulSets is being rollout-restarted gaps can still occur.
The Prometheus readiness probe only checks if it can serve the requests: when one replica is ready, Kubernetes can terminate the other immediately. So we end up in the situation where the first replica has just started but never scraped any targets and the other one is being terminated and already has stopped the scraping process, hence having scraping gaps.

We should postpone the second replica stop until the first replica scrapes all the targets. The simplest approach is to employ a `minReadySeconds` feature. Waiting for two scrape intervals should be enough.

We don't

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: fix
summary: avoid gaps in charts when Prometheus StatefulSet is being rollout restarted
impact: Prometheus StatefulSet will be rollout restarted
impact_level: default
```

### DIFF
--- a/ee/fe/modules/300-prometheus/templates/prometheus/prometheus.yaml
+++ b/ee/fe/modules/300-prometheus/templates/prometheus/prometheus.yaml
@@ -8,6 +8,12 @@ cpu: 10m
 memory: 25Mi
 {{- end }}
 
+{{- define "toSeconds" -}}
+{{- $duration := . }}
+{{- $now := now }}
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:

vladimirGuryanov (approved):

PR #11913, opened 2025-02-05
[dhctl] fixed ssh static build

alexey-gavrilov-flant:
## Description
Made the use of ssh more transparent.
Corrected the build to the path that will be in the final image.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type:  chore
summary: fixed ssh static build
impact_level:  low
```


name212 (approved):

name212 (changes requested):
Please check it with standalone installer

alexey-gavrilov-flant:
> Please check it with standalone installer

I didn't take into account this startup option. I returned the parameters for scp because otherwise there will be an error if there is no ssh on the server at the moment of installation or scp from the system will be used instead of from our delivery.

sprait (approved):

name212 (approved):

PR #11989, opened 2025-02-11
Refactor dmtlint files

yalosev:
## Description
Refactor dmtlint rules

Keep excludes for every module is more readable and predictable then have a global list of excludes

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: tools
type: chore
summary: Refactor dmtlint excludes.
impact_level: low
```


modules/040-control-plane-manager/.dmtlint.yaml
```diff
@@ -0,0 +1,13 @@
...
+      resources:
+        - kind: DaemonSet
+          name: d8-control-plane-manager
+          container: image-holder-etcd
+      security-context:
+        - kind: DaemonSet
+          name: d8-control-plane-manager
+          container: image-holder-etcd
+      service-port:
+        - d8-control-plane-apiserver
```
ldmonster, line 13: we need to ingore by name in service-port check?

z9r5 (approved):

PR #12077, opened 2025-02-13
[vertical-pod-autoscaler] recommender memory-save

borg-z:
## Description

As described here https://github.com/kubernetes/autoscaler/issues/6368 vpa recommender can consume a lot of RAM. Resource consumption increases with a large number of pods and with frequent deployment rollouts.

This commentary details the mechanisms for storing and cleaning data inside the VPA https://github.com/kubernetes/autoscaler/issues/6368#issuecomment-1963800975.

The simplest solution is to enable the memory-saver option. This should reduce RAM consumption, since not all pods use VPA

```
--memory-saver 	Bool 	false 	If true, only track pods which have an associated VPA
```

In large clusters, VPA recommender can consume a large amount of RAM over time

Example memory usage:

- 10 Deployment
- every Deployment has 10 Pod replicas
- Rollout every 30 sec

Memory utilization delta ~50 MB per hour - 1.2 GB in 24 hours

Снимок экрана 2025-02-14 в 10 17 01

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: vertical-pod-autoscaler
type: fix
summary: VPA recommender memory-save option enable
impact: 
impact_level: default
```

### DIFF
--- a/modules/302-vertical-pod-autoscaler/templates/recommender/deployment.yaml
+++ b/modules/302-vertical-pod-autoscaler/templates/recommender/deployment.yaml
@@ -63,6 +63,7 @@ spec:
         - --memory-aggregation-interval=1h
         - --recommender-interval={{ printf "%vs" $.Values.global.discovery.prometheusScrapeInterval }}
         - --stderrthreshold=0
+        - --memory-saver=true
         - --v=4
         ports:
         - containerPort: 8080


deckhouse-BOaTswain:
Aye, aye, @borg-z. I've started the workflow for label 'e2e/run/yandex-cloud'...

sprait (approved):

name212 (approved):

PR #12034, opened 2025-02-12
[user-authn] Dex-authenticator add redis connection retries

meatuses:
## Description
Patch to oauth2-proxy that adds a loop to `sendRedisConnectionTest` to retry these connection. 
fixes https://github.com/deckhouse/deckhouse/issues/4644

<details>
<summary>Patched function:</summary>

```go
func sendRedisConnectionTest(client redis.Client, key string, val string) []string {
	msgs := []string{}
	ctx := context.Background()
	var err error
	for i := 0; i < 10; i++ {
		logger.Printf("testing connection to redis...")
		err = client.Set(ctx, key, []byte(val), time.Duration(60)*time.Second)
		if err == nil {
			break
		}
		time.Sleep(2 * time.Second)
	}
	if err != nil {
		msgs = append(msgs, fmt.Sprintf("unable to set a redis initialization key: %v", err))
	} else {
		gval, errGet := client.Get(ctx, key)
		if errGet != nil {
			msgs = append(msgs,
				fmt.Sprintf("unable to retrieve redis initialization key: %v", err))
		}
		if string(gval) != val {
			msgs = append(msgs,
				"the retrieved redis initialization key did not match the value we set")
		}
	}

	err = client.Del(ctx, key)
	if err != nil {
		msgs = append(msgs, fmt.Sprintf("unable to delete the redis initialization key: %v", err))
	}
	return msgs
}
```
</details>

Usually takes a couple of retries to wait for Redis:
```
root@ob-pr12034:~# kubectl -n d8-monitoring logs grafana-dex-authenticator-5b75489655-4bwf7
Defaulted container "dex-authenticator" out of: dex-authenticator, redis, self-signed-generator (init)
[2025/02/14 15:28:37] [sessions.go:70] testing connection to redis...
[2025/02/14 15:28:39] [sessions.go:70] testing connection to redis...
[2025/02/14 15:28:39] [proxy.go:83] mapping path "/dev/null" => file system "/dev/null"
[2025/02/14 15:28:39] [oauthproxy.go:166] OAuthProxy configured for OpenID Connect Client ID: grafana-d8-monitoring-dex-authenticator
[2025/02/14 15:28:39] [oauthproxy.go:172] Cookie settings: name:_oauth2_proxy secure(https):true httponly:true expiry:168h0m0s domains: path:/ samesite: refresh:after 10m0s
```

Case if Redis is specifically broken (wrong port):
```
[...]

PR #12127, opened 2025-02-15
[dex] init delay probe

maxidrom-admin:
## Description
Hey, everybody!
a small edit with a readinessProbe

The dex-authenticator feed has a sidecar container with redis and it often lacks a fraction of a second to start. Because of this the dex-authenticator container restarts on a trial basis. It seems to me that it would be enough to increase the delay a bit.


maxidrom-admin:
sorry, just found some better pr for that: https://github.com/deckhouse/deckhouse/pull/12034

PR #12129, opened 2025-02-15
[cert-manager] Rename webhooks

yalosev:
## Description
Restore the [original](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/templates/webhook-validating-webhook.yaml#L17) webhook name

Cert-manager has the hardcoded regexp [inside](https://github.com/cert-manager/cert-manager/blob/v1.16.2/pkg/util/cmapichecker/cmapichecker.go#L59). Some feature couldn't work without this check.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cert-manager
type: fix
summary: Restore the original webhook name to match the cert-manager's library regexp.
```


deckhouse-BOaTswain:

PR #10218, opened 2024-10-21
[istio] Getting rid of self-made IP address allocation for public service's ServiceEntry'es in Federation

skurbatov:
## Description
Enabling automatic DNS name allocated the Istio service and excluding self-made IP address allocation for public service's ServiceEntry'es in Federation.

## Why do we need it, and what problem does it solve?
The self-made IP address iterator is no longer needed after enabling the automatic DNS name allocation of Istio service.

## What is the expected result?
Istio Federation no longer contains information about the IP addresses of public services and interaction with them takes place using DNS names and internal IP addresses allocated by istio.

## Checklist
- [ ] The code is covered by unit tests.
- [x] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries
```changes
section: istio
type: chore
summary: Got rid of self-made IP address allocation for public service's ServiceEntries in Federation
impact_level: default
```

PR #12133, opened 2025-02-17
[istio] Add ServiceEntry ports validation

Paramoshka:
## Description

* Added ServiceEntry ports validation to warn the user about the consequences of creating a resource without ports.

After [changing](https://github.com/deckhouse/deckhouse/pull/10218/files) the assignment of ServiceEntry addresses for services in a federation, a routing error may occur on existing clusters after an update and here's why:
  * Previously, applications got the real service IP's after resolving and used to access it.
  * Now they got the fake IP.
  * If there aren't any ports in ServiceEntry, istio can't find corresponding listener and proxy-passes the request. Before the [change](https://github.com/deckhouse/deckhouse/pull/10218/files), the requests were proxy-passed to the real IP, but now they are being proxy-passed to the fake one. So, the requests fail.

There is no sense in ServiceEntry without ports, so we made the validation.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: feature
summary: Added a validation and an alert when creating a ServiceEntry resource without service ports.
```


ee/modules/110-istio/templates/federation/validation-service-entries.yaml
```diff
@@ -0,0 +1,37 @@
...
+---
+kind: ValidatingAdmissionPolicy
+apiVersion: admissionregistration.k8s.io/v1beta1
+metadata:
+  name: validation-serviceentry-ports
+  {{- include "helm_lib_module_labels" (list $) | nindent 2 }}
+spec:
+  validations:
+    - expression: "has(object.spec.ports)"
+      message: "A ServiceEntry must define at least one port to ensure proper traffic routing. Without ports, Istio cannot determine which protocols to use or how to direct traffic, potentially leading to connectivity issues. This may cause failures in service discovery and communication."
```
apolovov, line 11:
```suggestion
      message: "A ServiceEntry must define at least one port to ensure proper traffic routing. Without ports, Istio cannot determine which service traffic belongs to and just proxy-passes it. The cluster uses ISTIO_META_DNS_AUTO_ALLOCATE feature which generates fake IPs for ServiceEntries, so proxy-passed traffic is being sent to fake IP. This may cause failures in service discovery and communication."
```

ee/modules/110-istio/templates/federation/validation-service-entries.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if .Values.istio.federation.enabled }}
+---
+kind: ValidatingAdmissionPolicy
+apiVersion: admissionregistration.k8s.io/v1beta1
+metadata:
+  name: validation-serviceentry-ports
```
apolovov, line 6:
```suggestion
  name: d8:istio:serviceentry-ports
```

ee/modules/110-istio/templates/federation/validation-service-entries.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if .Values.istio.federation.enabled }}
```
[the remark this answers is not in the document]
Paramoshka: done

ee/modules/110-istio/templates/federation/validation-service-entries.yaml
Paramoshka: done

ee/modules/110-istio/templates/validations.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  policyName: d8-istio-serviceentry-ports
+  matchResources:
+    resourceRules:
+      - apiGroups: ["networking.istio.io"]
+        apiVersions: ["v1beta1", "v1"]
+        resources: ["serviceentries"]
+        operations: ["CREATE", "UPDATE"]
+  validationActions: ["Warn"]
+---
```
apolovov, line 33:
```suggestion
---
```

apolovov (approved):

PR #12163, opened 2025-02-18
[control-plane-manager] Stale service account alert

borg-z:
## Description
This PR adds a new alert rule: **D8KubernetesStaleTokensDetected**.  
The alert fires when stale service account tokens are detected in the cluster.

Currently these tokens are not blocked because the `--service-account-extend-token-expiration` flag is enabled by default (Default: true). With this flag enabled, admission-injected tokens are extended up to 1 year during token generation to facilitate a safe transition from legacy tokens to the bound service account token feature, ignoring the value of `service-account-max-token-expiration`.

For security reasons, this should be fixed.

Stale tokens can cause security risks and authentication failures because some applications read the token only at startup and do not refresh it later.  

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
type: feature
summary: add alert for detecting stale service account tokens
impact: 
impact_level: default

### DIFF
--- a/docs/documentation/_data/deckhouse-alerts.yml
+++ b/docs/documentation/_data/deckhouse-alerts.yml
@@ -1634,11 +1634,10 @@ alerts:
         - Ensure your application is configured to periodically reload the token from the file system.
         - Verify that you are using an up-to-date client library that supports automatic token rotation.
 
-        **Note:**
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
Aye, aye, @borg-z. I've started the workflow for label 'e2e/run/openstack'...

deckhouse-BOaTswain:
Aye, aye, @borg-z. I've started the workflow for label 'e2e/run/openstack'...

Zhbert (approved):

sudoroot-null (approved):

Issue #12070, opened 2025-02-13
[user-authn] Dex-authenticator add the ability to change the seccompProfile


aamorozovv:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

Encountered a problem with applying custom resource DexAuthenticator when podSecurityStandards is configured in Restricted mode in admission-policy-engine.
The dex-authenticator-* application pods, which are deployed after applying a custom resource, do not start because the containers lack seccompProfile context.

After editing the deployment itself (added seccompProfile: type: RuntimeDefault) the pods started, but this is a workaround. 

### Proposed Solution

Add the ability to set seccompProfile context in dex-auth deployment

PR #12197, opened 2025-02-19
feat: add security context with RuntimeDefault seccomp profile to dex container

juev:
## Description

fixed: #12070

Change seccomp profile to RuntimeDefault instead of Unlimited

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: add security context with RuntimeDefault seccomp profile to dex container
impact_level: default
```

### DIFF
--- a/helm_lib/charts/deckhouse_lib_helm/templates/_module_security_context.tpl
+++ b/helm_lib/charts/deckhouse_lib_helm/templates/_module_security_context.tpl
@@ -59,8 +59,8 @@ securityContext:
 securityContext:
   allowPrivilegeEscalation: false
   capabilities:
-  drop:
-  - all
+    drop:
+    - all
   runAsGroup: 64535
   runAsNonRoot: true
   runAsUser: 64535
--- a/helm_lib/charts/deckhouse_lib_helm/templates/_module_security_context.tpl
+++ b/helm_lib/charts/deckhouse_lib_helm/templates/_module_security_context.tpl
@@ -149,6 +149,8 @@ securityContext:
 securityContext:
   readOnlyRootFilesystem: true
   allowPrivilegeEscalation: false
+  seccompProfile:
+    type: RuntimeDefault
   capabilities:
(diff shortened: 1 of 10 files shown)


juev:
Initially, all inspections were carried out without current changes.

Checked the work in the namespace d8-monitoring.

Screenshot 2025-02-20 at 15 12 13
Screenshot 2025-02-20 at 15 12 48
Screenshot 2025-02-20 at 15 13 35
Screenshot 2025-02-20 at 15 14 07

I checked the application of the new policy, saw warnings on a number of pods. The change in policy did not lead to the fact that any of the pods did not work. Conducted a forced restart of the development in the names space. I saw warnings about problems in the security configuration. But after the restart, all the pods continued to work. No problems were noticed in the work.

Then I switched the deployment to use the current changes. Switched the policy to the baseline. And then he checked how the pods would behave when using the restricted policy.

Screenshot 2025-02-20 at 15 22 52
Screenshot 2025-02-20 at 15 24 03

You can see that there are fewer warnings, and at least grafana-dex-authenticator continues to work without warnings.

helm_lib/charts/deckhouse_lib_helm/templates/_module_security_context.tpl
```diff
@@ -197,3 +197,17 @@ securityContext:
...
+{{- /* Template context with .Values, .Chart, etc */ -}}
+securityContext:
+  readOnlyRootFilesystem: true
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+    - ALL
+  seccompProfile:
+    type: RuntimeDefault
+{{- end }}
```
sprait, line 213: changes in helm_lib need add in repo with bump version via make update-lib-helm:
https://github.com/deckhouse/lib-helm

sprait (changes requested):
one comment

name212 (approved):

PR #12282, opened 2025-02-24
[deckhouse] change the method of connecting deckhouse-controller to API-server

tarabrind:
## Description
Add conditions for using direct connection, API-proxy and Service.

Deckhouse stucks when cni is malfunctioned.
Also, if you switch CNIs in a cluster, the network becomes unavailable, including all services. Deckhouse works with API through the service, so it loses access to it.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Changed the method of connecting deckhouse-controller to API-server.
```

### DIFF
--- a/modules/002-deckhouse/templates/deployment.yaml
+++ b/modules/002-deckhouse/templates/deployment.yaml
@@ -147,10 +147,15 @@ spec:
           image: "{{ .Values.deckhouse.internal.currentReleaseImageName }}"
           imagePullPolicy: Always
           env:
-{{- if (not (.Values.global.enabledModules | has "control-plane-manager")) }}
-# control-plane is managed by third-party team (EKS/GKP/...): use Service by default
-{{- else }}
-  {{- if not .Values.global.clusterIsBootstrapped }}
+{{- if (.Values.global.enabledModules | has "control-plane-manager") }}
+# control-plane is NOT managed by third-party team (EKS/GKP/...)
+  {{- if .Values.global.clusterIsBootstrapped }}
+  # managed by Deckhouse: use API-proxy
+            - name: KUBERNETES_SERVICE_HOST
+              value: 127.0.0.1
+            - name: KUBERNETES_SERVICE_PORT
+              value: "6445"
+  {{- else }}
   # bootstrap phase: use direct connection
             - name: KUBERNETES_SERVICE_HOST
               valueFrom:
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
Aye, aye, @tarabrind. I've started the workflow for label 'e2e/run/gcp'...

nabokihms (changes requested):
This will break setups:
- in managed kubernetes
- when deckhouse is deployed on a dedicated node

deckhouse-BOaTswain:
Aye, aye, @tarabrind. I've started the workflow for label 'e2e/run/yandex-cloud'...

deckhouse-BOaTswain:
Aye, aye, @tarabrind. I've started the workflow for label 'e2e/run/yandex-cloud'...

modules/002-deckhouse/templates/deployment.yaml
```diff
@@ -147,15 +147,25 @@ spec:
           image: "{{ .Values.deckhouse.internal.currentReleaseImageName }}"
           imagePullPolicy: Always
           env:
-# KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT are needed on the bootstrap phase to make Deckhouse work without kube-proxy
-{{- if not .Values.global.clusterIsBootstrapped }}
+{{- if (not (.Values.global.enabledModules | has "control-plane-manager")) }}
```
apolovov, line 150: I don't like the denial. Make it 
```
if has control-plane-manager
  if clusterbootstrapped
    ...
  else
    ...
  end
else
  # # control-plane is managed by third-party team (EKS/GKP/...): use service by default
end
```

modules/002-deckhouse/templates/deployment.yaml
```diff
@@ -147,15 +147,25 @@ spec:
...
+# control-plane is managed by third-party team (EKS/GKP/...): use service by default
```
apolovov, line 151:
```suggestion
# control-plane is managed by third-party team (EKS/GKP/...): use Service by default
```

modules/002-deckhouse/templates/deployment.yaml
tarabrind: I think the last "else" should be deleted as well.

apolovov (approved):

PR #12357, opened 2025-02-27
[user-authn] feat: update dex to v2.42.0

juev:
## Description

Update dex to v2.42.0 using existing patches

The new version of dex already contains a number of patches we need. And fixes for other problems.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: feature
summary: update dex to v2.42.0
impact_level: default
```


juev:
How did I test?

I have a local user admin created, which was not part of any of the local groups. After installing this pr in the developer cluster, I tried to log in to grafana, the authorization was successful.

Then in DexAuthenticator for grafana I added options to spec:

```yaml
spec:
  allowedGroups:
  - admins
```

When I tried to log in again, I received an error that the user is not part of the group.

Added admin to the group by creating a new group:

```yaml
kind: Group
metadata:
  name: admins
spec:
  members:
  - kind: User
    name: admin
  name: admins
```

Authorization was successful.

PR #12368, opened 2025-02-28
[dhctl] fix: sudoPassword is not passed for dhctl-server

distorhead:
## Description

Related https://github.com/deckhouse/deckhouse/pull/12099
Fixes dhctl-server use-case.

Allow sudoPassword field in the SSHConfig kind.

Fixes dhctl-server use-case of sudoPassword field. Previous PR added only ability to pass param using dhctl-cli.


deckhouse-BOaTswain:
Aye, aye, @distorhead. I've started the workflow for label 'e2e/run/aws'...

name212 (approved):

PR #12399, opened 2025-03-03
[prometheus] Increase scrape timeout seconds

scaps1:
## Description
Increase scrape timeout seconds

We need to increase this parameter to avoid timeout issues while scraping metrics and It also removes the need to adjust this parameter, as it is now calculated automatically. 

Not necessarily

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: chore
summary: Increase scrape timeout seconds
impact_level: default
```


deckhouse-BOaTswain:

PR #12411, opened 2025-03-04
[prometheus] Add scrape timeout 

scaps1:
## Description
Add scrape timeout for prometheus. 

We need to add this parameter to avoid timeout issues while scraping metrics in places where it's not set.

Not necessarily

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: chore
summary: Increase scrape timeout seconds
impact_level: default
```


deckhouse-BOaTswain:

modules/300-prometheus/templates/prometheus/prometheus.yaml
```diff
@@ -140,6 +148,7 @@ spec:
         topologyKey: kubernetes.io/hostname
 {{- end }}
   scrapeInterval: {{ .Values.prometheus.scrapeInterval | default "30s"}}
+  scrapeTimeout: {{ include "calcTimeout" (default "30s" .Values.prometheus.scrapeInterval) | default "24s" }}
```
vladimirGuryanov, line 151: You don't need `default "30s"` it's default value from open api spec.

vladimirGuryanov (changes requested):

modules/300-prometheus/templates/prometheus/prometheus.yaml
scaps1: I made it so that the template tests wouldn't fail, moved it to a more appropriate place

deckhouse-BOaTswain:

ee/fe/modules/300-prometheus/templates/prometheus/prometheus.yaml
```diff
@@ -8,6 +8,74 @@ cpu: 10m
 memory: 25Mi
 {{- end }}
 
+{{- define "calcTimeout" -}}
```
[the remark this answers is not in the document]
scaps1: Yes, you're partially right, it does look smaller, but the problem is that this helper can't translate days, weeks, years, etc. 
I reduced the size of my helper, I will leave it translated in seconds and applied your version with finding 80 percent, thanks for the advice

deckhouse-BOaTswain:

deckhouse-BOaTswain:

vladimirGuryanov (approved):

PR #12419, opened 2025-03-04
[dhctl] fix: sudoPassword is not passed for dhctl-server and cannot be used non-interactively (1.68)

distorhead:
## Description

Related old PR: https://github.com/deckhouse/deckhouse/pull/12099
Related PR to master: https://github.com/deckhouse/deckhouse/pull/12368

Fixes dhctl-server use-case and dhctl-cli and connection-config use-case when passing sudoPassword in a non-interactive way.

Allow sudoPassword field in the SSHConfig kind.

Fixes dhctl-server use-case of sudoPassword field.
Fixes dhctl-cli use-case of sudoPassword usage (previously sudoPassword field in the connection-config has no effect and not used actually, this PR fixes non-interactive way of passing password).

## Changelog entries

```changes
section: dhctl
type: fix
summary: sudoPassword from the connection-config not used in dhctl CLI and dhctl server
impact_level: default
```

### DIFF
--- a/dhctl/pkg/preflight/sudoers.go
+++ b/dhctl/pkg/preflight/sudoers.go
@@ -34,7 +34,7 @@ func (pc *Checker) CheckSudoIsAllowedForUser() error {
 		return callSudo(pc.nodeInterface, app.BecomePass)
 	}
 
-	return callSudo(pc.nodeInterface, "")
+	return callSudo(pc.nodeInterface, app.BecomePass)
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
Aye, aye, @distorhead. I've started the workflow for label 'e2e/run/aws'...

name212 (approved):

PR #12485, opened 2025-03-10
[user-authn] update apiVersion for User kind in validating webhook

juev:
## Description

We have crd users version v1, the previously used v1alpha1 was transferred to the deprecated status. And in order to completely abandon it, you need to switch to the new version of api.

Rejection of deprecated versions of api in the application

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: chore
summary: update apiVersion for User kind in validating webhook
impact_level: low
```


juev:
How I tested.

After the development in the developer cluster, I checked the creation of a group with the inclusion of an existing user in it.

```bash
 # kubectl get users.deckhouse.io test -oyaml
```

```yaml
apiVersion: deckhouse.io/v1
kind: User
metadata:
  name: test
status:
  groups:
  - "916531"
  - everyone
  - testgroup
```

```bash
# kubectl get groups.deckhouse.io testgroup -oyaml
```

```yaml
kind: Group
metadata:
  name: testgroup
spec:
  members:
  - kind: User
    name: test
  name: testgroup
```

yalosev (changes requested):
Guess we need it also here:
- https://github.com/deckhouse/deckhouse/blob/98460db85abf17a2a9feb0e220f4b31a2e645e55/modules/150-user-authn/hooks/migration_groups_from_user_to_group_object.go#L48

nabokihms (changes requested):
There was a logic behind using old version, because there are no groups in the new API version.

modules/150-user-authn/hooks/migration_groups_from_user_to_group_object.go
```diff
@@ -44,7 +44,7 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 	Kubernetes: []go_hook.KubernetesConfig{
 		{
 			Name:                         "users",
-			ApiVersion:                   "deckhouse.io/v1alpha1",
+			ApiVersion:                   "deckhouse.io/v1",
```
yalosev, line 47: Aftermeet: remove this hook

juev: done

yalosev (approved):

PR #12498, opened 2025-03-10
[node-manager] the `get_crds` hook will fail with error if the node groups count in cluster is not equal to the calculated one

name212:
## Description
The `get_crds` hook will fail with error if the node groups count in cluster is not equal to the calculated one.

We had an indecent with dropping all machinedeployments and cloud nodes and we decide to add error if count calculated nodegroups not equals node groups in the cluster 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: chore
summary:  The `get_crds` hook will fail with error if the node groups count in cluster is not equal to the calculated one.
impact_level: default
```


deckhouse-BOaTswain:
Aye, aye, @name212. I've started the workflow for label 'e2e/run/aws'...

RomanenkoDenys (approved):

PR #12039, opened 2025-02-12
[deckhouse] trigger the module config when a module created

ipaqsa:
## Description
It triggers the module config when a module created.

## Why do we need it, and what problem does it solve?
A module config can be created before the module is discovered.

## Why do we need it in the patch release (if we do)?
It can help people with CI/CD.

Delete the module to test and then check that the module does not exist:
```
root@dev-master-0:~# k get module commander
Error from server (NotFound): modules.deckhouse.io "commander" not found
```

Create the module config(its normal that it is ignored now):
```
root@dev-master-0:~# kubectl get mc
NAME                         ENABLED   VERSION   AGE     MESSAGE
commander                    true      1         3m23s   Ignored: unknown module name
```

Wait for 3 minutes to source resync:
```
root@dev-master-0:~# k get module commander -oyaml
apiVersion: deckhouse.io/v1alpha1
kind: Module
metadata:
  creationTimestamp: "2025-02-12T18:05:01Z"
  generation: 3
  name: commander
  resourceVersion: "305624798"
  uid: 65c97bf5-0b93-4463-b1f9-9d31cc61343f
properties:
  availableSources:
  - deckhouse
  releaseChannel: Stable
  source: deckhouse
status:
  conditions:
  - lastProbeTime: "2025-02-12T18:05:01Z"
    lastTransitionTime: "2025-02-12T18:05:01Z"
    status: "True"
    type: EnabledByModuleConfig
  - lastProbeTime: "2025-02-12T18:05:01Z"
    lastTransitionTime: "2025-02-12T18:05:01Z"
    status: "False"
    type: EnabledByModuleManager
  - lastProbeTime: "2025-02-12T18:05:02Z"
    lastTransitionTime: "2025-02-12T18:05:01Z"
    status: "True"
    type: IsReady
  phase: Ready
```

## Checklist
- [ ] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: deckhouse
type: fix 
summary: Trigger the module config when a module created.
```

PR #12681, opened 2025-03-20
[deckhouse] module config ignored

ipaqsa:
## Description
It fixes the case when a module config ignored because the module is discovered after the config created.

Prev fix: https://github.com/deckhouse/deckhouse/pull/12039

If a module is discovered after the module config created, the config is ignored because the controller cannot find the module. When the module created - nothing triggers the config to reconcile(watcher mechanism doesn't always work) so we set interval in 3 minutes to trigger the config by it. 

Flaky error, it can appear when a cluster is setting up by commander. Only reboot or trigger config fixed it.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Fix module config ignored.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/pkg/controller/module-controllers/config/controller.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/config/controller.go
@@ -51,6 +51,8 @@ const (
 
 	maxConcurrentReconciles = 3
 
+	moduleNotFoundInterval = 3 * time.Minute
+
 	moduleDeckhouse = "deckhouse"
 	moduleGlobal    = "global"
 )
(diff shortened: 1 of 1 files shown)


deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -360,13 +362,13 @@ func (r *reconciler) deleteModuleConfig(ctx context.Context, moduleConfig *v1alp
...
 			r.log.Error("failed to update the module", slog.String("module", module.Name), log.Err(err))
-			return ctrl.Result{Requeue: true}, nil
+			return ctrl.Result{}, err
 		}
 	}
 
 	if err := r.removeFinalizer(ctx, moduleConfig); err != nil {
-		r.log.Error("failed to remove finalizer from ModuleConfig", slog.String("module", moduleConfig.Name), log.Err(err))
-		return ctrl.Result{Requeue: true}, nil
+		r.log.Error("failed to remove finalizer from config", slog.String("module", moduleConfig.Name), log.Err(err))
```
yalosev, line 370: I like `ModuleConfig` more. word `config` is more common

ipaqsa: done

PR #12809, opened 2025-03-31
[upmeter] increase upmeter PVC capacity and scale down retention

scaps1:
## Description
Introduce a PR to extend storage capacity and reduce default retention for Upmeter. As part of this process, the existing StatefulSet is re-created with updated settings (without cascading deletion of pods).

Upmeter has a 1GB disk size limit, which is no longer sufficient for older clusters where stored data exceeds this limit. Previously, we reduced the retention period from 2 years to 1.5 years, but it did not fully resolve the issue.
This PR increases the PVC size, scales down the default retention period, and safely re-creates the Upmeter StatefulSet (with cascade: false) to apply the new storage configuration.

Not necessarily

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: upmeter
type: fix
summary:  Add a hook for replacing old sts, increase storage capacity, and scale down retention to 13 months.
impact_level: default
```


deckhouse-BOaTswain:

deckhouse-BOaTswain:

modules/500-upmeter/hooks/retention_resize.go
```diff
@@ -0,0 +1,193 @@
...
+
+	retentionDays := 548
+
+	if usagePercent > 80 {
+		scaling := float64(100-usagePercent) / 20.0
+		retentionDays = int(548.0 * scaling)
+		if retentionDays < 30 {
+			retentionDays = 30
+		}
+	}
```
lazovskiy, line 130: This logic does not seem to be valid.
Let's do some unit testing:

Current usagePercent = 90, then scaling is (100-90) / 20 = 0.5, then retentionDays equals to 548 * 0.5 = 274.
In other words, for 10% overuse we cut retention in half. This is never correct.

The problem is that we don't know how much disk space is required to retain 1 day of metrics, as we don't have the actual retention, so I cannot see the obvious way to enforce disk-based retention. If upmeter cleans the metrics right after the start, we can try to decrease the retention days, for example, by 7 days on each iteration. However, this approach has a certain drawback: after several iterations, the disk usage will stabilize below the threshold, but after the deckhouse controller restart we will loose retentionDays value as we don't store it as a state. This, however, can also be fixed by implementing more complex logic, e.g., by reading the current UPMETER_RETENTION_DAYS environment variable value.

modules/500-upmeter/hooks/retention_resize.go
```diff
@@ -0,0 +1,193 @@
...
+	Namespace string
+	Ready     bool
+}
+
+var _ = sdk.RegisterFunc(&go_hook.HookConfig{
+	Queue: "/modules/upmeter/adjust_retention",
+	Schedule: []go_hook.ScheduleConfig{
+		{
+			Name:    "adjust_retention_every_15min",
+			Crontab: "*/15 * * * *",
```
lazovskiy, line 50: This might be too often, especially considering the following comments.

modules/500-upmeter/hooks/retention_resize.go
```diff
@@ -0,0 +1,193 @@
...
+
+	isReady := false
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.Name == "upmeter" {
+			isReady = cs.Ready
+			break
+		}
+	}
+
+	return PodRef{
```
lazovskiy, line 89: Here we can extend the structure and get the current UPMETER_RETENTION_DAYS value. See below

modules/500-upmeter/hooks/retention_resize.go
```diff
@@ -0,0 +1,193 @@
...
+			if len(fields) >= 5 {
+				pctStr := strings.TrimSuffix(fields[4], "%")
+				pct, err := strconv.Atoi(pctStr)
+				if err == nil {
+					return pct
+				}
+			}
+		}
+	}
+	return -1
```
lazovskiy, line 151: The magic return value.
Instead, it would be better to take the standard Go approach and return, in this case, `(int, error)`

modules/500-upmeter/hooks/retention_resize.go
```diff
@@ -0,0 +1,193 @@
...
+	if !pod.Ready {
+		return fmt.Errorf("upmeter pod is not ready")
+	}
+
+	kubeClient, err := dc.GetK8sClient()
+	if err != nil {
+		return fmt.Errorf("unable to get Kubernetes client: %w", err)
+	}
+
+	stdout, _, err := execToPod(kubeClient, "df -PB1 /db", "upmeter", pod.Name, pod.Namespace)
```
lazovskiy, line 112: You can use
`df -B1 --output=pcent /db`

which will return a much easier to parse string:
```
Use%
 87%
```

modules/500-upmeter/hooks/retention_resize.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, line := range lines {
+		if strings.Contains(line, "/db") || strings.Contains(line, "/dev/") {
+			fields := strings.Fields(line)
+			if len(fields) >= 5 {
+				pctStr := strings.TrimSuffix(fields[4], "%")
+				pct, err := strconv.Atoi(pctStr)
+				if err == nil {
+					return pct
+				}
+			}
```
lazovskiy, line 148: There are too many code blocks nesting.
You may use the `continue` statement to skip the iteration when the condition is not met.

Also you can avoid looping here, return error if there is less (or even not equal) than two lines. And pick the second one. If you use `--output=pcent`, you can even do `Trim` (for spaces) and `TrimSuffix` for the percent sign, and you get the parsible integer string.

lazovskiy (changes requested):
There are several pitfalls to be avoided.

lazovskiy (approved):

vladimirGuryanov (approved):

Zhbert (approved):

shvgn (approved):

Issue #12546, opened 2025-03-12
[user-authn] DexAuthenticator with digit name fails


yalosev:
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.68.3

### Expected Behavior

DexAuthenticator must renders gracefully.

### Actual Behavior

```
ModuleRun:main:user-authn:doStartup:ReloadAllModules:failures 1:run helm install: helm upgrade failed: failed to create resource: PodDisruptionBudget in version "v1" cannot be handled as a PodDisruptionBudget: json: cannot unmarshal number into Go struct field LabelSelector.spec.selector.matchLabels of type string
```

### Steps To Reproduce

1. Create DexAuthenticator with digital name
```yaml
apiVersion: deckhouse.io/v1
kind: DexAuthenticator
metadata:
  name: 111111
  namespace: default
spec:
  applicationDomain: foo.bar.com
  applicationIngressCertificateSecretName: ingress-tls
  applicationIngressClassName: nginx
  highAvailability: false
```

2. Deckhouse will fail

### Additional Information

Guess we have to add ` | quote` here:
- [one](https://github.com/deckhouse/deckhouse/blob/main/modules/150-user-authn/templates/dex-authenticator/deployment.yaml#L55)
- [two](https://github.com/deckhouse/deckhouse/blob/main/modules/150-user-authn/templates/dex-authenticator/deployment.yaml#L76)
- [three](https://github.com/deckhouse/deckhouse/blob/main/modules/150-user-authn/templates/dex-authenticator/deployment.yaml#L81)

### Logs

```shell

```

PR #12902, opened 2025-04-07
[user-authn] DexAuthenticator with digit name fails

juev:
## Description

fixed: #12546

Get the ability to create a DexAuthenticator with numbers in the name.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: DexAuthenticator with digit name fails
impact_level: default
```

### DIFF
--- a/modules/150-user-authn/templates/dex-authenticator/deployment.yaml
+++ b/modules/150-user-authn/templates/dex-authenticator/deployment.yaml
@@ -52,7 +52,7 @@ spec:
   selector:
     matchLabels:
       app: dex-authenticator
-      name: {{ $crd.name }}
+      name: {{ $crd.name | quote }}
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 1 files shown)


juev:
I couldn't reproduce the problem.

```plain
root@dev-master-0:~/src/test# cat dex.yaml
apiVersion: deckhouse.io/v1
kind: DexAuthenticator
metadata:
  name: 111111
  namespace: default
spec:
  applicationDomain: foo.bar.com
  applicationIngressCertificateSecretName: ingress-tls
  applicationIngressClassName: nginx
  highAvailability: false
root@dev-master-0:~/src/test# kubectl apply -f dex.yaml
error: unable to decode "dex.yaml": json: cannot unmarshal number into Go struct field ObjectMeta.metadata.name of type string
```

But I made changes to deployment according to the description.

yalosev (approved):

yalosev:
```yaml
apiVersion: deckhouse.io/v1
kind: DexAuthenticator
metadata:
  name: "111111"
  namespace: default
spec:
  applicationDomain: foo.bar.com
  applicationIngressCertificateSecretName: ingress-tls
  applicationIngressClassName: nginx
  highAvailability: false
```

quotes should be used in the CR to reproduce that

PR #12950, opened 2025-04-09
[cloud-provider-huaweicloud] Add the `--cluster-name` CLI flag to the `CCM`

artbin:
## Description

- Add the `--cluster-name` CLI flag to the `cloud-controller-manager`
- Add a patch to `cloud-controller-manager` to fix the handling of the `--cluster-name` CLI flag

The `cloud-controller-manager` uses the `--cluster-name` CLI flag to generate load balancer names. If this flag is not set, separate clusters may generate identical load balancer names.

For each `Service`, two load balancers were provisioned: one with the correct cluster name and another with the default cluster name. It is incorrect to have two load balancers for the same service. The patch to `cloud-controller-manager` fixes the issue by ensuring that only one load balancer is created with the correct cluster name.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cloud-provider-huaweicloud
type: fix
summary: Add the `--cluster-name` CLI flag to the `cloud-controller-manager`.
impact_level: default
```


artbin:
Manual testing
==============

Before patch:
Screenshot 2025-04-18 at 03 18 56

After patch:
Screenshot 2025-04-18 at 03 47 19

PR #13257, opened 2025-04-28
[testing][e2e][autoscaler] Add condition for cordon event during scaling

name212:
## Description
Added caching cordon node event.

It is important for us to know whether the node was cordoned before deletion or not. It is also important for us whether the node was successfully drained, but for now this functionality (creating an event about a successful drain) is only available in cluster-api (https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/controllers/machine/machine_controller.go#L518) but not in mcm. Therefore, we will limit ourselves to cordon events only and assume that if the node was cordoned, then it was drained.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: feature
summary: Add condition for cordon event during scaling
impact_level: low
```


deckhouse-BOaTswain:

RomanenkoDenys (approved):

PR #13393, opened 2025-05-10
[testing] Improve autoscaler e2e tests

name212:
## Description
Improve autoscaler e2e tests.

Now we save nodes names and nodes count of autoscaler in variables. We grep cordon and draining events for each node in nodes list and grep events by node name instead of nodegroup prefix. This allows you to more accurately collect events this allows restarting testing without side effects (when we have separate events for each nodes by test attempt and we wait 1 one event for test). Also, during testing, we found that k8s not always set cordon event, but autoscaler always. We begin to check k8s events and if we did not find them check autoscaler events as fallback.  

Restarting e2e tests not working correctly and we summarized tests for adding additional nodes in tests.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: testing
type: fix
summary: Improve autoscaler e2e tests.
impact_level: low
```


deckhouse-BOaTswain:

PR #13391, opened 2025-05-09
[node-manager][autoscaler] Remove additional cordon node by mcm provider

name212:
## Description
Remove additional cordon node by mcm provider.

Continue of https://github.com/deckhouse/deckhouse/pull/13180

## Why do we need it, and what problem does it solve?
MCM provider of cluster autoscaler can cordon nodes out of autoscaler flow.

## Checklist
- [ ] The code is covered by unit tests.
- [x] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [ ] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Autoscaler - remove additional cordon node by mcm provider.
impact_level: default
```

PR #13500, opened 2025-05-17
[node-manager][autoscaler] Patch autoscaler 1.30-1.32 to support our MCM

name212:
## Description
Patch autoscaler 1.30-1.32 to support our MCM.

New autoscaler works with new version MCM which select nodes for deleting from annotation `node.machine.sapcloud.io/trigger-deletion-by-mcm` This annotation does not support by our MCM, and we should set deleting priority with annotation `machinepriority.machine.sapcloud.io` for right deletion order.
We set priority for machines and keep `node.machine.sapcloud.io/trigger-deletion-by-mcm` annotation for calculation replicas, but we need to clean deleted machines from annotation in refresh function for keeping up to date annotation value to avoid drizzling replicas count in machine deployment.

After our previous patch https://github.com/deckhouse/deckhouse/pull/13391 cluster autoscaler periodically scale up and scale down machine deployments, because after our patch CA cannot right calculated replicas, because we deleted `node.machine.sapcloud.io/trigger-deletion-by-mcm` annotation.

Because new CA is oriented on `node.machine.sapcloud.io/trigger-deletion-by-mcm` annotation for deleting nodes, but our mcm is oriented on `machinepriority.machine.sapcloud.io` annotation for calculate nodes for delete. we can have situation when nodes for deleting calculated by CA do not delete, but delete another nodes and some nodes stuck in Scheduling Disable status.  

Also   

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Patch autoscaler 1.30-1.32 to support our MCM
impact_level:  low
```


deckhouse-BOaTswain:

PR #13623, opened 2025-05-26
[node-manager] Make NodeUser passwordHash parameter optional

d2285:
## Description

The PR allows creating NodeUser CR without passwordHash field.

this allow to create node user without password to avoid auth issues from pam, ldap, etc.
ex.
```
---
apiVersion: deckhouse.io/v1
kind: NodeUser
metadata:
  name: user
spec:
  uid: 20029
  sshPublicKey: "ssh-rsa  ..."
  isSudoer: true
  nodeGroups:
  - master

kubectl apply -f nodeuser.yaml
Warning: Password hash is empty. The user might not be able to log in.
nodeuser.deckhouse.io/user created

cat /etc/shadow | grep user
...
user:!:20235:0:99999:7:::
```
## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type:  chore
summary: passwordHash in nodeuser cr is not required
impact_level: default
```

### DIFF
--- a/modules/040-node-manager/webhooks/validating/node_user
+++ b/modules/040-node-manager/webhooks/validating/node_user
@@ -52,7 +52,6 @@ function __main__() {
   nodeGroups=$(context::jq -r '.review.request.object.spec.nodeGroups[]')
   operation=$(context::jq -r '.review.request.operation')
   passwordHash=$(context::jq -r '.review.request.object.spec.passwordHash')
-  warnings=()
 
   snapshots=$(context::jq -cr '[.snapshots.nodeUsers[].filterResult]')
   if [ "$operation" = "UPDATE" ]; then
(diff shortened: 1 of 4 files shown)


deckhouse-BOaTswain:
Aye, aye, @d2285. I've started the workflow for label 'e2e/run/static'...

modules/040-node-manager/webhooks/validating/node_user
```diff
@@ -80,6 +81,13 @@ EOF
     fi
   done
 
+if [ -z "$passwordHash" ] || [ "$passwordHash" = "null" ]; then
+    cat <<EOF > "$VALIDATING_RESPONSE_PATH"
+{"allowed": true, "warnings": ["Password hash is empty. The user might not be able to log in."] }
```
sprait:
```suggestion
{"allowed": true, "warnings": ["Password hash is empty. This may not be secure and it may be prohibited by PAM settings."] }
```

sprait (commented):
one suggestion

name212 (approved):

Issue #10834, opened 2024-11-26
[install] Add Local Registry Configuration Check to Installation Process


cds111:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Version

v1.65.6

### Expected Behavior

Alert the user with an error about the CA certificate when the local registry is configured with the HTTP scheme.

### Actual Behavior

A cluster can be configured with the HTTP scheme while utilizing a CA certificate.

### Steps To Reproduce

1. Create installation configuration.
2. Add the CA certificate for the local registry.
3. Select the HTTP scheme.
4. Initiate the installation process.
5. Verify the cluster configuration.

deckhouse-BOaTswain:
This issue has been automatically put in the triage queue because it has not had
recent activity. The team will reconsider the status of this issue. Thank you
for your contributions.

PR #13645, opened 2025-05-26
[dhctl] Add Local Registry Configuration Check to Installation Process

trofimovdals:
## Description

Alert the user with an error about the CA certificate when the local registry is configured with the HTTP scheme on `dhctl bootstrap` phase
Fix [issue](https://github.com/deckhouse/deckhouse/issues/10834) 

When run `dhctl bootstrap` with the HTTP scheme and CA on InitConfiguration, an unobvious behavior occurs (containerd have different [configuration](https://github.com/deckhouse/deckhouse/blob/e38ab2ee95f102629f0c93145f4f74fbd05c7b83/candi/bashible/common-steps/node-group/032_configure_containerd.sh.tpl#L148-L155) for both params)

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Add Local Registry Configuration Check to Installation Process
impact_level: default
```


trofimovdals:
It might be worth doing it as a preflight check.

name212 (approved):

PR #13646, opened 2025-05-26
[deckhouse-controller]  Convert MUP CRD v1alpha2 storage

RottenRat:
## Description

This action we are gradually getting rid of the deprecated version (`ModuleUpdatePolicy - v1alpha1`).

In this case the backup mechanism has been modified so that it retains the current resource versions.

The `v1alpha1` lifecycle is coming to an end, so the validation webhook prohibits these resources from being deployed to the cluster. Since the old resource is deprecated, we must backup the updated `v1alpha2` resource and keep data of actual version.

After successful data migration, we observe in the CRD:
```
storedVersions:
  - v1alpha2
```
 
Also, I attach `d8 backup` logs before and after the change (only related to `ModuleUpdatePolicy`):
### Before
```
# d8 logs, old resource in backup
W0527 14:54:16 [warnings.go:70] ModuleUpdatePolicy v1alpha1 deprecated (repeated x30)
```

### After
```
# d8 logs, new resource in backup
I0527 15:07:04 [request.go:729] Client throttling: GET v1alpha2/moduleupdatepolicies (2.78s)
```

</details>

The problem:
- The cluster uses `ModuleUpdatePolicy` version `deckhouse.io/v1alpha2`
- But the backup retains `deckhouse.io/v1alpha1` version, which is deprecated
- `v1alpha1` deployment by validation webhook is not allowed in cluster.

The problem is that when a backup is created, the deprecated version `v1alpha1` for the `ModuleUpdatePolicy` resource is retained, even though the cluster actually uses version `v1alpha2`.

Because of this, problems may occur with:

- **Compatibility**: The v1alpha1 version is deprecated and may not contain a mandatory field that is required in v1alpha2 and vice versa
- **Recovery**: Backups in their current form may be impossible to apply due to a validation error.
- **Consistency**: The backup must accurately reflect the current state of the cluster, including the correct versions of resources

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: convert MUP CRD v1alpha2 storage
impact_level: low
```

### DIFF
--- a/deckhouse-controller/crds/module-update-policy.yaml
+++ b/deckhouse-controller/crds/module-update-policy.yaml
@@ -20,7 +20,7 @@ spec:
   versions:
     - name: v1alpha1
       served: true
-      storage: true
+      storage: false
       deprecated: true
       schema:
         openAPIV3Schema:
@@ -212,7 +212,7 @@ spec:
           description: Module release update windows.
     - name: v1alpha2
       served: true
-      storage: false
+      storage: true
       schema:
         openAPIV3Schema:
           type: object


deckhouse-BOaTswain:
Aye, aye, @ldmonster. I've started the workflow for label 'e2e/run/aws'...

ldmonster:
/e2e/abort aws;WithoutNAT;containerd;1.30 15295212248 15295212248-1-con-1-30 /sys/deckhouse-oss/install:pr13646

deckhouse-BOaTswain:
Aye, aye, @ldmonster. I've started the workflow for label 'e2e/run/aws'...

Issue #8418, opened 2024-05-17
[prometheus] Add DeckhouseModuleUsesNonExistentStorageClass alert


meatuses:
### Preflight Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/deckhouse/deckhouse/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/deckhouse/deckhouse/issues) for an issue that matches the one I want to file, without success.

### Use case. Why is this important?

If a StorageClass that was used in a module's PVC gets deleted, this can lead to issues with volume migration and potential data loss. This condition must be tracked.

### Proposed Solution

Add `DeckhouseModuleUsesNonExistentStorageClass` that checks if modules have PVCs in non-existent storage classes

deckhouse-BOaTswain:
This issue has been automatically put in the triage queue because it has not had
recent activity. The team will reconsider the status of this issue. Thank you
for your contributions.

PR #13694, opened 2025-05-29
[prometheus] Alert if D8 PVC uses deleted SC

meatuses:
## Description
Add alert if storageclass is used in a PVC but actually missing.
Only for `d8-*` namespaces for the time being. Unfortunately, could not be tied to any metric could export list of D8 owned PVCs or PVs, we don't have any.

`
kube_persistentvolumeclaim_info{namespace=~"d8-.*"} unless(kube_persistentvolumeclaim_info{namespace=~"d8-.*"} * on(storageclass) group_left() kube_storageclass_info)
`

Feature request https://github.com/deckhouse/deckhouse/issues/8418

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: feature
summary: Add DeckhouseModuleUsesDeletedStorageClass alert
impact_level: low
```

### DIFF
--- a/modules/300-prometheus/monitoring/prometheus-rules/deletedsc.yaml
+++ b/modules/300-prometheus/monitoring/prometheus-rules/deletedsc.yaml
@@ -1,0 +1,23 @@
+- name: d8.deletedsc
+  rules:
+    - alert: DeckhouseModuleUsesDeletedStorageClass
+      expr: |
+        kube_persistentvolumeclaim_info{namespace=~"d8-.*"} unless(kube_persistentvolumeclaim_info{namespace=~"d8-.*"} * on(storageclass) group_left() kube_storageclass_info)
+      for: 10m
+      labels:
+        tier: cluster
+        severity_level: "4"
+      annotations:
+        plk_markup_format: markdown
+        plk_protocol_version: "1"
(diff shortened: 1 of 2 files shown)


z9r5 (approved):

maxinsky (approved):

modules/300-prometheus/monitoring/prometheus-rules/deletedsc.yaml
```diff
@@ -0,0 +1,25 @@
+- name: d8.deletedsc
+  rules:
+  - alert: DeckhouseModuleUsesDeletedStorageClass
+    expr: |
+      kube_persistentvolumeclaim_info{namespace=~"d8-.*"} unless(kube_persistentvolumeclaim_info{namespace=~"d8-.*"} * on(storageclass) group_left() kube_storageclass_info)
```
vladimirGuryanov: There’s a simpler way to do this:

```
sum(kube_persistentvolumeclaim_info{namespace=~"d8-.*"}) by (namespace, persistentvolumeclaim, storageclass) unless on(storageclass) kube_storageclass_info
```

vladimirGuryanov (changes requested):

PR #13711, opened 2025-05-30
[ci] Add listing of changed modules for release branch changes

Taior:
## Description
Add listing of changed modules for release branch changes.

Example:
2025-05-30-194131_485x609_scrot

This change allows to detect potentially destructive image changes for minor release.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ci
type: feature
summary: Add listing of changed modules for release branch changes.
impact_level: low
```

### DIFF
--- a/.github/scripts/python/compare_version_modules.py
+++ b/.github/scripts/python/compare_version_modules.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+
+# Copyright 2025 Flant JSC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


.github/scripts/python/compare_version_modules.py
```diff
@@ -0,0 +1,92 @@
...
+        if len(digests_set) > 1:
+            results.append(f"{module}.{image}")
+
+results.sort()
+
+dangerous_results = [i for i in results if any(m in i for m in dangerous_modules)]
+
+if len(dangerous_results != 0):
+    print("Found possibly dangerous changes in following module images:")
+    print(*dangerous_results, sep="\n")
```
Nikolay1224: "if" condition here the same as line 91, so just move this 2 lines to lint 92 to let script print out all changes at line 87 and then print out dangerous changes and fail

Taior: I believe it is more useful to have a list of possibly dangerous changes on top of the log.

PR #13868, opened 2025-06-07
move kubelet inhibitor patch to ee

name212:
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


werf.yaml
```diff
@@ -37,6 +37,21 @@ cleanup:
 {{- if not (hasKey $editionsDict .Env) }}
   {{ fail (printf "Editions name not found %v" .Env) }}
 {{- end }}
+
+{{- $eeEdition := false }}
+
+{{- range $_, $edition := $.Editions }}
+  {{- if eq $edition.name "EE" }}
```
sprait, line 44: what about FE, SE and other edition?

sprait (commented):
one question

werf.yaml
name212: only >= EE needs here

diafour:
OK.. let's see it in action ;)

PR #14111, opened 2025-06-23
[tools] Fix audit policy generator

yalosev:
## Description
Fix generator

It takes wrong namespaces for service accounts, we get records that have never exist:
 - 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


tools/audit_policy/main.go
```diff
@@ -71,104 +141,106 @@ func cwd() string {
...
+}
+
+func handleServiceAccounts(templatesDir string, modulePath string, name string, ns string, sas map[string]struct{}) error {
+	serviceAccounts, err := processServiceAccounts(templatesDir)
+	if err != nil {
+		return err
+	}
+
+	for _, saName := range serviceAccounts {
+		// ServiceAccounts for istiod will be added to audit rules in
```
name212, line 227: will NOT?

because in the next line we continue and skip processing

tools/audit_policy/main.go
```diff
@@ -183,52 +255,76 @@ func unquote(s string) (string, error) {
...
+	name, namespace, err := processModuleDefinition(moduleYamlPath)
 	if err != nil {
-		panic(err)
+		return err
 	}
 
-	type Data struct {
-		Namespace, ServiceAccount []string
+	if namespace == "" {
+		panic("Empty namespace for module " + moduleYamlPath)
```
name212: can we have a module without namespace at all? For example, a module contains only hooks and no templates at all?

in the future for example

tools/audit_policy/main.go
```diff
@@ -245,3 +341,42 @@ func uniqueNonEmptyElementsOf(s []string) []string {
...
+
+	checkTemplates := func(modulePath string) error {
+		templatesPath := filepath.Join(modulePath, "templates")
+
+		if info, err := os.Stat(templatesPath); err == nil {
+			if info.IsDir() {
+				templatesDirs = append(templatesDirs, templatesPath)
+			}
+		} else if !os.IsNotExist(err) {
+			return fmt.Errorf("error checking templates dir %s: %v", templatesPath, err)
```
name212: same question

https://github.com/deckhouse/deckhouse/pull/14111/files#r2161430577

tools/audit_policy/main.go
```diff
@@ -46,6 +47,75 @@ var auditPolicyBasicServiceAccounts = []string{
...
+
+	// 1. find all modules
+	res, err := findModules(workDir)
+	if err != nil {
+		panic(err)
+	}
+
+	// 2. find templates for each module and fill ServiceAccounts info
+	for moduleYamlPath, moduleDirName := range res {
+		err = processModule(workDir, moduleYamlPath, moduleDirName, sasMap, namespacesMap)
```
name212, line 81: I don't understand why we need to pass these two maps as arguments, judging by the code we don't use any caching mechanism between modules. Can we just return new maps via return values?

tools/audit_policy/main.go
yalosev: It's a handler for any error EXCEPT the `Not Found`. So, absent templates directory is ok. Non-readable directory will give an error

tools/audit_policy/main.go
yalosev: Have no idea about this logic, honestly. It's just a paste from the [current file](https://github.com/deckhouse/deckhouse/blob/main/tools/audit_policy/main.go#L140-L144)

tools/audit_policy/main.go
yalosev: Good point. Linked it with the `templates` dir length

tools/audit_policy/main.go
yalosev: Nope, we have to deduplicate values between modules.
For example, cni-cilium and cilium-hubble use the `d8-cni-cilium` namespace.
A few modules use `kube-system` namespace.

So, we have to keep only unique values.
Build array and then deduplicate them via the map... what for? if we can use the map from the beginning

tools/audit_policy/main.go
yalosev: Fixed

deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:fix/audit-policy-generator` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/15825709176) in 35m9s.



Yandex.Cloud-WithoutNAT-Containerd-1.30 - Connection string: `ssh  redos@84.201.173.247`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.30` succeeded in 34m6s.

</details>

PR #14189, opened 2025-06-26
[deckhouse] Change deckhouse pod resource and add alert

yalosev:
## Description
According to the research: https://loop.flant.ru/flant/pl/1qzum1k5gbdpjqeutkeq5in3za
Deckhouse need around 300-400Mi for normal usage.
Also, based on the monitoring metrics 
image
heavy usage is around 1.5-2Gi of memory.

So, I guess we can change requests for the deckhouse. And set limit x3 of maximum requests.
Also, add an alert to have a change to get memory dump before the OOM.
It's much better to restart deckhouse while node will get the global OOM

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: chore
summary: Change requests and limits for the pod, based on the medium usage. Prevent node OOM in the corner cases.
```


modules/002-deckhouse/templates/deployment.yaml
```diff
@@ -26,10 +26,11 @@ spec:
     kind: Deployment
     name: deckhouse
   updatePolicy:
-    updateMode: "Initial"
+    updateMode: "Auto"
```
yalosev:
```suggestion
    updateMode: "Initial"
```

no reason to evict a pod. Deckhouse will not migrate to a new node

ldmonster (approved):

nabokihms (approved):

PR #14589, opened 2025-07-22
[node-manager] The partedConfig field is required in mig in NodeGroup

alexey-gavrilov-flant:
If the partedConfig value is not specified, undefined behavior occurs in the migration scripts.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: chore
summary: The partedConfig field is required in mig in NodeGroup
impact: 
impact_level: default
```


alexey-gavrilov-flant:
If the field is not filled in, this error message appears.
```
# nodegroups.deckhouse.io "worker-gpu" was not valid:
# * <nil>: Invalid value: "": "spec.gpu" must validate one and only one schema (oneOf). Found none valid
# * spec.gpu.sharing: Unsupported value: "MIG": supported values: "TimeSlicing"
# * <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
#
```

z9r5 (approved):

PR #12241, opened 2025-02-21
[node-manager] kubelet wait for external inhibitors to unlock

diafour:
## Description

- Add overrides for poweroff/shutdown/reboot/halt commands to always check shutdown inhibitors.
- Add d8-shutdown-inhibitor.service which delays shutdown while Pods with "pod.deckhouse.io/inhibit-node-shutdown" label are present on the Node.
- Patch kubelet to postpone shutdown if Node has condition GracefulShutdownPostpone with status True.
- Add gracefulShutdown field to node-manager configuration. Inhibitor will be started if gracefulShutdown.mode is "BlockByPodLabel".

## Why do we need it, and what problem does it solve?

We want to delay system shutdown so cluster administrator can make a decision about Pods with specific label on this Node. For example, we want to have time to migrate virtual machines manually even on accidental system shutdown event.

Also, alert about node in shutdown state is possible. d8-shutdown-inhibitor will set this condition:

```
# kubectl get no NODENAME -oyaml | grep GracefulShutdown -B4
  - lastHeartbeatTime: null
    lastTransitionTime: null
    reason: PodsWithLabelAreRunningOnNode
    status: "True"
    type: GracefulShutdownPostpone
```

## Checklist
- [x] The code is covered by unit tests.
- [ ] e2e tests passed.
- [x] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: node-manager
type: feature
summary: Add systemd shutdown inhibitors to delay system shutdown until Pods with the specific label are gone from the Node.
impact_level: default
```

PR #14636, opened 2025-07-23
[node-manager] fix initial value for GracefulShutdownPostpone condition

diafour:
## Description
GracefulShutdownPostpone condition status should be false on start. 

True value is only for situation when Pods with label present on node when shutdown signal is received.

This fix enables creating Prometheus alert about Nodes in postponed shutdown state.

It is a small fix for #12241 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: chore
summary: Set status value to false for GracefulShutdownPostpone condition on start.
impact_level: default
```


deckhouse-BOaTswain:
Cherry pick PR [14700](https://github.com/deckhouse/deckhouse/pull/14700) to the branch [release-1.71](https://github.com/deckhouse/deckhouse/tree/release-1.71) successful!

PR #14669, opened 2025-07-24
[cert-manager] recursive settings

### LINKED ISSUE #390: [cert-manager] acme dns01 recursive nameservers
### Preflight Checklist

In some cases, we have to specify recursive nameservers for a cert-manager controller.
It can be useful when:

1. PublicDomain and ClusterDomain are equal. Cert-manager self-check will not be able to resolve public DNS records then. And DNS01 ACME challenge will not work
2. private DNS servers have restricted access and you need to specify them directly

### Proposed Solution

Cert-manager controller has a couple of flags:
```
--dns01-recursive-nameservers strings [ex. 8.8.8.8:53,8.8.4.4:53]
--dns01-recursive-nameservers-only
```
which can fix the situations described above. 

We can add these flags to the [config](https://github.com/deckhouse/deckhouse/blob/main/modules/101-cert-manager/openapi/config-values.yaml) and pass them to a cert-manager controller

### Additional Information

_No response_

ipaqsa:
## Description
It provides recursive settings for cert-manager.

You can specify recursive nameservers via module config:
```
kind: ModuleConfig
metadata:
  name: cert-manager
spec:
  enabled: true
  settings:
    recursiveSettings:
      nameservers:
         - 8.8.8.8:53
         - https://1.1.1.1
  version: 1
```

It adds ```--dns01-recursive-nameservers``` flag to the cert-manager deployment:
```
...
- --acme-http01-solver-resource-request-cpu=0
- --dns01-recursive-nameservers=8.8.8.8:53,https://1.1.1.1
...
```

If you want use only recusrive without fallback, you can specify use only recursive:
```
kind: ModuleConfig
metadata:
  name: cert-manager
spec:
  enabled: true
  settings:
    recursiveSettings:
      useOnlyRecursive: true
      nameservers:
         - 8.8.8.8:53
        - https://1.1.1.1:443
  version: 1
```

It add ```--dns01-recursive-nameservers-only``` flag to the cert-manager deployment:
```
...
- --dns01-recursive-nameservers=8.8.8.8:53,https://1.1.1.1:443
- --dns01-recursive-nameservers-only
...
```

Closes #390

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cert-manager
type: feature
summary: Add recursive settings.
```

### DIFF
--- a/modules/101-cert-manager/openapi/config-values.yaml
+++ b/modules/101-cert-manager/openapi/config-values.yaml
@@ -144,7 +144,10 @@ properties:
       DNS recursion settings for ACME DNS-01 self-check.
     x-examples:
       - nameservers: ["8.8.8.8:53", "https://1.1.1.1:443"]
-        onlyRecursive: true
+        useOnlyRecursive: true
+    x-kubernetes-validations:
+      - rule: "self.useOnlyRecursive == false || size(self.nameservers) >= 1"
+        message: "At least one nameserver must be specified when useOnlyRecursive is true"
     properties:
       nameservers:
         type: array
@@ -158,14 +161,3 @@ properties:
         description: |
           When set to true, cert-manager restricts DNS-01 challenge validation to the specified recursive nameservers and will not fall back to authoritative name servers.
         default: false
-    allOf:
-      - if:
-          properties:
-            useOnlyRecursive:
-              const: true
-        then:
-          properties:
-            nameservers:
-              minItems: 1
-          required:
-            - nameservers
--- a/modules/101-cert-manager/openapi/config-values.yaml
+++ b/modules/101-cert-manager/openapi/config-values.yaml
@@ -137,3 +137,35 @@ properties:
     description: |
       The name of the `ingressClass` used to confirm ownership of domain using the [ACME HTTP-01](https://cert-manager.io/docs/configuration/acme/http01/) challenges method.
       If the parameter is omitted, the default `ingressClass` is used.
+
+  recursiveSettings:
+    type: object
+    description: |
+      DNS recursion settings for ACME DNS-01 self-check.
+    x-examples:
+      - nameservers: ["8.8.8.8:53", "https://1.1.1.1:443"]
+        onlyRecursive: true
+    properties:
+      nameservers:
+        type: array
+        items:
+          type: string
+          description: |
+            List of recursive DNS servers (IP:port or HTTPS URLs) that cert-manager uses to validate ACME DNS-01 challenges via a self-check.
+          pattern: "^(?:[0-9]{1,3}(?:\\.[0-9]{1,3}){3}:\\d{1,5}|https?://[^\\s]+)$"
+      useOnlyRecursive:
+        type: boolean
+        description: |
+          When set to true, cert-manager restricts DNS-01 challenge validation to the specified recursive nameservers and will not fall back to authoritative name servers.
+        default: false
+    allOf:
+      - if:
+          properties:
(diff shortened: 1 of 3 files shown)


modules/101-cert-manager/openapi/config-values.yaml
```diff
@@ -136,4 +136,28 @@ properties:
...
+        type: array
+        items:
+          type: string
+          description: |
+            List of recursive DNS servers (IP:port or HTTPS URLs) that cert-manager uses to validate ACME DNS-01 challenges via a self-check.
+          pattern: "^(?:[0-9]{1,3}(?:\\.[0-9]{1,3}){3}:\\d{1,5}|https?://[^\\s]+)$"
+      useOnlyRecursive:
+        type: boolean
+        description: |
+          When set to true, cert-manager restricts DNS-01 challenge validation to the specified recursive nameservers and will not fall back to authoritative name servers.
```
Suselz: The description uses the term "authoritative server". Does this mean kube-dns in the cluster? 
If so, it is better not to add the second term "authoritative", but simply write that requests are sent only to dns servers from `recursiveSettings.nameservers`.

I suggest this because the authoritative server is the server that stores a specific zone. If you request the "deckhouse.io" zone from kube-dns, it will forward this request to the next server, i.e. it will act as a recursive server.

modules/101-cert-manager/openapi/config-values.yaml
```diff
@@ -136,4 +136,28 @@ properties:
...
-      If the parameter is omitted, the default `ingressClass` is used.
\ No newline at end of file
+      If the parameter is omitted, the default `ingressClass` is used.
+
+  recursiveSettings:
+    type: object
+    description: |
+      DNS recursion settings for ACME DNS-01 self-check.
+    x-examples:
+      - nameservers: ["8.8.8.8:53", "https://1.1.1.1:443"]
```
Suselz: This flag, according to the cert-manager documentation, only supports the formats `<ip address>:<port>` or `https://<DoH RFC 8484 server address>` (without port).
[Cert-manager source](https://github.com/cert-manager/cert-manager/blob/v1.17.1/cmd/controller/app/options/options.go#L153)

modules/101-cert-manager/openapi/config-values.yaml
```diff
@@ -136,4 +136,28 @@ properties:
...
+      - rule: "self.useOnlyRecursive == false || size(self.nameservers) >= 1"
+        message: "At least one nameserver must be specified when useOnlyRecursive is true"
+    properties:
+      nameservers:
+        type: array
+        items:
+          type: string
+          description: |
+            List of recursive DNS servers (IP:port or HTTPS URLs) that cert-manager uses to validate ACME DNS-01 challenges via a self-check.
+          pattern: "^(?:[0-9]{1,3}(?:\\.[0-9]{1,3}){3}:\\d{1,5}|https?://[^\\s]+)$"
```
Suselz, line 158: Regex is correct, so the problem is only in the example

Suselz (commented):
A few questions about the documentation

z9r5 (approved):

modules/101-cert-manager/openapi/config-values.yaml
ipaqsa: done

modules/101-cert-manager/openapi/config-values.yaml
ipaqsa: done

Suselz (approved):

PR #14706, opened 2025-07-28
[docs] improve cache for docs-watcher

juev:
## Description

This PR implements significant performance improvements to the documentation generation process on the website by enhancing the cache mechanism in the registry-modules-watcher component.

#### **Problem - Before this change:**
- We downloaded all release channels sequentially without checking if a similar image was already cached
- Redundant downloads occurred when the same module version existed across multiple release channels
- This led to unnecessary network traffic and slower documentation generation times
- No deduplication mechanism for identical content across different channels

#### **Performance Impact:**
The documentation site generation was inefficient due to repeated downloads of identical module artifacts, causing:
- Increased build times
- Higher bandwidth usage
- Redundant storage of duplicate content

#### **Smart Cache with Digest-based Deduplication:**
- We now check cache by **digest/checksum** before downloading
- The system searches through all releases across all release channels for identical digests
- If a matching digest is found, we reuse the equivalent cached data without downloading from the registry
- This eliminates redundant downloads of identical content

#### **Performance Improvements:**
- ✅ **Reduced Download Time**: Skip downloads for duplicate content
- ✅ **Lower Bandwidth Usage**: Eliminate redundant network requests
- ✅ **Faster Documentation Generation**: Reuse cached artifacts across channels
- ✅ **Improved Resource Efficiency**: Less storage and processing overhead

#### **Developer Experience:**
- ✅ **Faster Local Development**: Quicker docs builds during development
- ✅ **More Reliable Builds**: Less dependency on network conditions
- ✅ **Better Cache Utilization**: Intelligent reuse of existing data

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: docs
type: feature
summary: improve cache for docs-watcher
impact_level: low
```


docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
```diff
@@ -113,67 +113,74 @@ func (s *registryscanner) processReleaseChannels(ctx context.Context, registry,
...
-
 			return versionData, nil
 		}
 	}
 
-	// Extract version from image
-	version, err := getVersionFromImage(versionData.Image)
+	// Only download image if cache miss
+	s.logger.Info("cache miss, downloading image", slog.String("module", module), slog.String("channel", releaseChannel))
+	releaseImage, err := s.registryClients[registry].ReleaseImage(ctx, module, releaseChannel)
```
ldmonster: this function is not download image. we receive remote image from registry.

docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
```diff
@@ -113,67 +113,74 @@ func (s *registryscanner) processReleaseChannels(ctx context.Context, registry,
...
 		TarFile:        make([]byte, 0),
-		Image:          releaseImage,
 	}
 
 	// Check if we already have this release in cache
 	releaseChecksum, ok := s.cache.GetReleaseChecksum(versionData)
 	if ok && releaseChecksum == versionData.Checksum {
 		version, tarFile, ok := s.cache.GetReleaseVersionData(versionData)
 		if ok {
+			s.logger.Info("using cached data", slog.String("module", module), slog.String("version", version))
```
ldmonster: we need to check all channels in cache. if we found release with equal digest - return tar file for it

docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
juev: I don't quite understand what the difference is between downloading and receiving.

docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
juev: Understood. So, we're suggesting creating a separate cache where the image itself will be stored by digest? And accordingly, we remove the image from the current cache. Is that correct?

docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
ldmonster: np
https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/image.go#L64
accordingly to code, we receive just a reference to image, not complete image

content of image we download in cr.Extract(image) method

docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
ldmonster: we can do it by making new function like IsReleaseAlreadyCached(versionData) and check all release channels in cache for same checksum.

docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/cache/cache.go
```diff
@@ -173,13 +183,13 @@ func (c *Cache) SyncWithRegistryVersions(registryVersions []internal.VersionData
 	createTasks := RemapFromMapToVersions(RemapFromVersionData(versionsToCreate), backends.TaskCreate)
 
 	// Combine and sort all tasks
-	result := append(createTasks, versionsToDelete...)
-	sortDocumentationTasks(result)
+	createTasks = append(createTasks, versionsToDelete...)
```
juev: Just fix from golangci-lint

docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/cache/cache.go
```diff
@@ -202,10 +212,10 @@ func cleanupEmptyMaps(cache map[registryName]map[moduleName]moduleData, reg regi
 
 // Helper function to deep copy the cache map
 func copyCache(original map[registryName]map[moduleName]moduleData) map[registryName]map[moduleName]moduleData {
-	copy := make(map[registryName]map[moduleName]moduleData)
+	cacheCopy := make(map[registryName]map[moduleName]moduleData)
```
juev: Another one fix from golangci-lint

PR #14757, opened 2025-07-29
[docs-builder] Hugo config error handling

kvm999:
## Description
This PR fixes issue during calling HugFromConfig causing the panic. To get the panic it was enough to:
1) [docs/site/backends/docs-builder-template/layouts/baseof.html](https://github.com/deckhouse/deckhouse/blob/main/docs/site/backends/docs-builder-template/layouts/baseof.html) - edit this file and broke something (syntax or invalid variable name)
2) run server locally:
```
mkdir -p ~/test-docs-builder/{src,dst}

cp -r docs/site/backends/docs-builder-template/* ~/test-docs-builder/src/

go run main.go --src ~/test-docs-builder/src --dst ~/test-docs-builder/dst --address :8081
```
3) generate the docs:
```
curl -X POST http://localhost:8081/api/v1/build \
     -H "Content-Type: application/json" \
     -d '{}'
```

4) Ensure that there is no panic in logs after fix(previously it was)

With this fix I added nil pointer checks and error handling in HugFromConfig body.

To avoid potential issue with panic.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: docs-builder
type: fix
summary: error handling to avoid panicking during HugFromConfig invocation
impact: more stable behavior
impact_level: default
```

### DIFF
--- a/docs/site/backends/docs-builder/pkg/hugo/commandeer.go
+++ b/docs/site/backends/docs-builder/pkg/hugo/commandeer.go
@@ -15,6 +15,7 @@
 package hugo
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	stdlog "log"
@@ -104,6 +105,19 @@ type commonConfig struct {
 	configs *allconfig.Configs
 	cfg     config.Provider
 	fs      *hugofs.Fs
+}
+
+func (c *commonConfig) validate() error {
+	if c == nil {
+		return errors.New("commonConfig is nil")
+	}
+	if c.fs == nil {
+		return errors.New("commonConfig: no fs provided")
+	}
+	if c.configs == nil {
+		return errors.New("commonConfig: no config provided")
+	}
+	return nil
 }
 
 // This is the root command.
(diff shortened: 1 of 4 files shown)


docs/site/backends/docs-builder/pkg/hugo/commandeer.go
```diff
@@ -264,13 +264,21 @@ func (c *command) ConfigFromProvider(key int32, cfg config.Provider) (*commonCon
 }
 
 func (c *command) HugFromConfig(conf *commonConfig) (*hugolib.HugoSites, error) {
+	if c.hugologger == nil {
+		return nil, fmt.Errorf("HugFromConfig: hugologger is not initialized")
+	}
+
```
ldmonster:
```suggestion
```

ldmonster: there is no cases when we have no logger

kvm999: - [x] done

PR #14835, opened 2025-08-05
[cloud-provider-huaweicloud] Make huawei autoscaler "scale from zero" 

unholyFigaro:
## Description

Adds discovery logic so Cluster Autoscaler can create nodes starts with zero replicas

Huawei clusters can not autoscale from zero.
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cloud-provider-huaweicloud
type: feature
summary: discovery logic so Cluster Autoscaler can create nodes starts with zero replicas
impact_level: default
```


unholyFigaro:
Sceenshots of manual cluster testing:
image
image

PR #15094, opened 2025-08-20
[deckhouse] dont ensure mrs of disabled modules

ipaqsa:
## Description
It fixes ensuring mr process.

If a user disabled a module via its config, we should not ensure mr.

Without it deckhouse may stuck in restarting loop 

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Dont ensure releases of disabled modules.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
@@ -187,7 +187,7 @@ func (r *reconciler) needToEnsureRelease(
 		return false
 	}
 
-	if module.ConditionUnknown(v1alpha1.ModuleConditionEnabledByModuleManager) {
+	if module.ConditionUnknown(v1alpha1.ModuleConditionEnabledByModuleConfig) {
 		enabledByBundle := false
 		if meta.ModuleDefinition != nil {
 			enabledByBundle = meta.ModuleDefinition.Accessibility.IsEnabled(r.edition.Name, r.edition.Bundle)
@@ -200,9 +200,7 @@ func (r *reconciler) needToEnsureRelease(
 		if len(module.Properties.AvailableSources) > 1 && !source.IsDefault() {
 			return false
 		}
-	}
-
-	if module.ConditionStatus(v1alpha1.ModuleConditionEnabledByModuleConfig) {
+	} else if !module.ConditionStatus(v1alpha1.ModuleConditionEnabledByModuleConfig) {
 		return false
 	}
 
--- a/deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
@@ -25,6 +25,7 @@ import (
 	"slices"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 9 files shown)


deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
```diff
@@ -187,8 +187,7 @@ func (r *reconciler) needToEnsureRelease(
 		return false
 	}
 
-	// check the module enabled
-	if !module.ConditionStatus(v1alpha1.ModuleConditionEnabledByModuleConfig) {
+	if module.ConditionUnknown(v1alpha1.ModuleConditionEnabledByModuleConfig) {
 		enabledByBundle := false
 		if meta.ModuleDefinition != nil {
```
juev, line 193: It's also necessary to check that meta.ModuleDefinition.Accessibility is not nil, given that Accessibility is also an object reference.

ipaqsa: It is checked [inside](https://github.com/deckhouse/deckhouse/blob/e808fe252767c622f9dc69d069fba0d59e3b0b6e/deckhouse-controller/pkg/controller/moduleloader/types/definition.go#L93)

deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
```diff
@@ -201,6 +200,8 @@ func (r *reconciler) needToEnsureRelease(
 		if len(module.Properties.AvailableSources) > 1 && !source.IsDefault() {
 			return false
 		}
+	} else if !module.ConditionStatus(v1alpha1.ModuleConditionEnabledByModuleConfig) {
```
juev: Should we perhaps perform the check without a redundant 'else' statement? Given that we are verifying the same status.

And since status checks are common here, maybe it makes sense to implement the check differently—so that the same switch could be used?

For example, instead of returning true/false, return a specific status.

deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
juev: It's not exactly straightforward, but thanks anyway.

deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
ipaqsa: I thought about it, but this method is used in several places, and changing it all for a backport fix seems to be a bad idea.

deckhouse-controller/pkg/controller/module-controllers/source/helpers.go
```diff
@@ -187,8 +187,7 @@ func (r *reconciler) needToEnsureRelease(
 		return false
 	}
 
-	// check the module enabled
-	if !module.ConditionStatus(v1alpha1.ModuleConditionEnabledByModuleConfig) {
+	if module.ConditionUnknown(v1alpha1.ModuleConditionEnabledByModuleConfig) {
```
ldmonster, line 190: purpose of this replacement?
```suggestion
	if module.ConditionUnknown(v1alpha1.ModuleConditionEnabledByModuleConfig) {
```

deckhouse-BOaTswain:
Cherry pick PR [15179](https://github.com/deckhouse/deckhouse/pull/15179) to the branch [release-1.72](https://github.com/deckhouse/deckhouse/tree/release-1.72) successful!

PR #15119, opened 2025-08-21
[deckhouse] Removed the use of ~/.docker during build

alexey-gavrilov-flant:
## Description
Using files from the user folder is not secure; its contents may differ from those used to assemble images in werf. This change uses the user and password from the same variables.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: chore
summary: Removed the use of ~/.docker during build
impact: 
impact_level: low
```


alexey-gavrilov-flant:
added as pr of another change

PR #15156, opened 2025-08-23
[deckhouse-controller] coordinate restart after all concurrent ApplyR…

kvm999:
## Description

This PR implements coordination logic for Deckhouse restart operations during concurrent module deployments. The changes introduce atomic counters to track active `ApplyRelease` operations and ensure that Deckhouse restarts only after all concurrent module releases have completed their deployment process.

**Key changes:**
- Added atomic counters (`activeApplyCount`, `hasAppliedReleases`, `needsRestart`) to track module deployment state
- Modified `ApplyRelease()` method to increment/decrement counters atomically using defer
- Updated `restartLoop()` to check all restart conditions before triggering restart

**Critical component impact:** This change affects the Deckhouse controller restart behavior. The restart mechanism now waits for all concurrent module deployments to complete before restarting, which may slightly delay restarts but ensures system stability during module updates.

The current implementation restarts Deckhouse immediately after any single module completes its `ApplyRelease` operation, even when other modules are still being processed concurrently. This leads to several critical issues:

1. **Premature restarts during concurrent deployments:** When multiple modules (up to 3 concurrent) are being deployed simultaneously, the first completed module triggers an immediate restart, potentially interrupting other ongoing deployments.

2. **Inconsistent deployment state:** Interrupted module deployments can leave the system in an inconsistent state where some modules are fully deployed while others are partially applied.

3. **Resource contention and race conditions:** Restarting while other modules are still applying releases can cause resource conflicts and unpredictable behavior.

4. **Poor user experience:** Users deploying multiple modules simultaneously experience multiple restarts instead of a single coordinated restart.

This PR solves these problems by implementing a coordination mechanism that tracks all active `ApplyRelease` operations and triggers restart only when:
- At least one module has completed `ApplyRelease` (`hasAppliedReleases = 1`)
- No modules are currently processing `ApplyRelease` (`activeApplyCount = 0`)
- A restart has been requested (`needsRestart = 1`)

Not necessarily. While this improvement enhances system stability and user experience, it's not a critical bug fix that requires immediate backporting. The current behavior, though suboptimal, doesn't cause data loss or system failures. This change can be included in the next minor release.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Coordinate Deckhouse restart to wait for all concurrent module deployments to complete
impact: Deckhouse will now restart only after all concurrent module releases finish their ApplyRelease operations, reducing the number of restarts during bulk module updates and improving deployment reliability.
impact_level: default
```


deckhouse-controller/pkg/controller/module-controllers/release/controller.go
```diff
@@ -410,8 +411,8 @@ func (r *reconciler) handleDeployedRelease(ctx context.Context, release *v1alpha
 			return res, fmt.Errorf("run release deploy: %w", err)
 		}
 
-		modulesChangedReason = "module release reloaded"
-
+		r.log.Info("module release reloaded, waiting for Deckhouse restart", slog.String("release", release.GetName()))
+		r.readyForRestart.Store(true)
```
ldmonster: this case will not restart loop

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
```diff
@@ -199,31 +204,37 @@ func (r *reconciler) preflight(ctx context.Context) error {
...
+					slog.Int("active_apply_count", int(r.activeApplyCount.Load())))
+				r.readyForRestart.Store(false)
+			}
+
+			if r.releaseWasProcessed.Load() && r.readyForRestart.Load() {
+				r.log.Info("restarting Deckhouse...")
 				if err := syscall.Kill(1, syscall.SIGUSR2); err != nil {
 					r.log.Fatal("send SIGUSR2 signal failed", log.Err(err))
 				}
+				r.releaseWasProcessed.Store(false)
```
ldmonster: purpose of setting false here?

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
ldmonster: looks like we need use another one var

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
```diff
@@ -1398,12 +1400,11 @@ func (r *reconciler) deleteRelease(ctx context.Context, release *v1alpha1.Module
 			r.log.Error("failed to remove module in downloaded symlinks dir", slog.String("release", release.GetName()), slog.String("path", modulePath), log.Err(err))
 			return ctrl.Result{}, err
 		}
+
+		r.log.Info("module release reloaded, waiting for Deckhouse restart", slog.String("release", release.GetName()))
```
ldmonster: are you sure that module release is reloaded here?

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
```diff
@@ -783,7 +778,8 @@ func (r *reconciler) handlePendingRelease(ctx context.Context, release *v1alpha1
 		return ctrl.Result{}, nil
 	}
 
-	modulesChangedReason = "a new module release deployed"
+	r.log.Info("a new module release deployed, waiting Deckhouse to restart", slog.String("module", release.GetModuleName()))
+	r.readyForRestart.Store(true)
```
ldmonster: purpose of that?

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
```diff
@@ -641,12 +635,13 @@ func (r *reconciler) handlePendingRelease(ctx context.Context, release *v1alpha1
...
 		// deploy forced release without any checks (windows, requirements, approvals and so on)
-		if err = r.ApplyRelease(ctx, release, task); err != nil {
+		if err = r.applyRelease(ctx, release, task); err != nil {
 			logger.Error("apply forced release", log.Err(err))
 			return res, fmt.Errorf("apply forced release: %w", err)
 		}
 
-		modulesChangedReason = "a new module release deployed"
+		r.log.Info("a new module release deployed, waiting Deckhouse to restart", slog.String("module", release.GetModuleName()))
+		r.readyForRestart.Store(true)
```
ldmonster: purpose of that?

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
```diff
@@ -110,6 +111,8 @@ func RegisterController(
 		return fmt.Errorf("add preflight: %w", err)
 	}
 
+	r.readyForRestart.Store(true)
```
ldmonster: why store true here?

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
kvm999: done, removed it after discussion in DM

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
kvm999: Fixed by introducing a separate moduleReloadedForRestart atomic variable to handle module reload scenarios independently from the normal restart logic, preventing endless restart loops.

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
kvm999: You're absolutely right - fixed the misleading log message from 'reloaded' to 'deleted' since this function removes the module release from the filesystem

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
kvm999: The purpose is to signal a Deckhouse restart after successful module deployment, since the new module needs to be loaded into the running system

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
kvm999: This has been fixed - forced releases now use moduleReloadedForRestart.Store(true) instead of readyForRestart.Store(true) to avoid interfering with the normal restart logic

deckhouse-controller/pkg/controller/module-controllers/release/controller.go
kvm999: Removed the initial `readyForRestart.Store(true)` because the system should only be ready for restart after processing releases, not during initialization

deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:feat/coordinated-module-restart` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/17766890352) in 43m49s.



GCP-WithoutNAT-Containerd-1.30 - Connection string: `ssh  user@35.234.76.189`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.30` succeeded in 43m15s.

</details>

PR #15269, opened 2025-08-28
[user-authn] Hotfix USER CR ValidatingAdmissionPolicy oldObject == null

s10dev:
## Description
Fix the `ValidatingAdmissionPolicy` for the `User` CR by updating the validation expression.  
Now the rule checks:  

```rego
oldObject == null || object.spec.password == oldObject.spec.password
```  

This ensures that validation passes on `CREATE` (where `oldObject` is `null`) and still enforces password immutability on `UPDATE`.  

Previously, the policy tried to compare `object.spec.password` with `oldObject.spec.password` even during `CREATE` operations, where `oldObject` does not exist.  
As a result, creating new `User` resources failed.  
With this fix, user creation works correctly, while password immutability is still enforced for updates.  

Yes, because the bug directly prevents creating `User` resources in clusters with password policy enabled.  
This is a functional regression and must be fixed in the nearest patch release.  

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: user-authn
type: fix
summary: Fix `ValidatingAdmissionPolicy` for `User` CR to skip password check on create
impact: Fixes a bug where creating new User resources failed due to missing oldObject in validation; password immutability is still enforced on update
impact_level: high
```

### DIFF
--- a/modules/150-user-authn/templates/validation.yaml
+++ b/modules/150-user-authn/templates/validation.yaml
@@ -44,7 +44,7 @@ spec:
         operations:  ["CREATE", "UPDATE"]
         resources:   ["users"]
   validations:
-    - expression: "object.spec.password == oldObject.spec.password"
+    - expression: "oldObject == null || object.spec.password == oldObject.spec.password"
       message: "password policy is enabled, user password field cannot be changed manually"
 ---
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:
Cherry pick PR [15271](https://github.com/deckhouse/deckhouse/pull/15271) to the branch [release-1.72](https://github.com/deckhouse/deckhouse/tree/release-1.72) successful!

PR #15310, opened 2025-09-02
[dhctl] Enhance resource creation timeout message

neermilov:
## Description
Enhanced error messages when resource creation times out in dhctl. When dhctl fails to create resources within the timeout period, it now provides a clear, user-friendly message explicitly stating that this is not an installation failure and explaining the necessary next steps.

In reality, this timeout is expected behavior when certain resources require worker nodes that haven't been created yet. The new error message explicitly indicates that the installation is proceeding normally, and the user only needs to add a worker node or remove taints from master nodes to continue the process.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: feature 
summary: Added clearer error messages when resource creation times out
impact: Improved user experience when dhctl cannot create resources due to missing worker nodes
impact_level: default 
```

### DIFF
--- a/dhctl/pkg/kubernetes/actions/resources/resources.go
+++ b/dhctl/pkg/kubernetes/actions/resources/resources.go
@@ -36,7 +36,7 @@ import (
 	"github.com/deckhouse/deckhouse/dhctl/pkg/util/retry"
 )
 
-var ErrNotAllResourcesCreated = fmt.Errorf("Not all resources were creatated")
+var ErrNotAllResourcesCreated = fmt.Errorf("not all resources were created")
 
 // apiResourceListGetter discovery and cache APIResources list for group version kind
 type apiResourceListGetter struct {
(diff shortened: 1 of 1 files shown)


dhctl/pkg/kubernetes/actions/resources/resources.go
```diff
@@ -36,7 +36,7 @@ import (
 	"github.com/deckhouse/deckhouse/dhctl/pkg/util/retry"
 )
 
-var ErrNotAllResourcesCreated = fmt.Errorf("Not all resources were creatated")
+var ErrNotAllResourcesCreated = fmt.Errorf("not all resources were created")
```
name212: Please keep upper case here. We using uppercase everywhere because errors will display to users

dhctl/pkg/kubernetes/actions/resources/resources.go
```diff
@@ -341,6 +341,15 @@ func CreateResourcesLoop(ctx context.Context, kubeCl *client.KubernetesClient, r
 
 		select {
 		case <-endChannel:
+			if len(resources) > 0 {
+				return fmt.Errorf(
+					"creating resources timed out after %s: resources cannot become ready. "+
```
name212:
```suggestion
					"Creating resources timed out after %s: resources cannot become ready. "+
```

dhctl/pkg/kubernetes/actions/resources/resources.go
```diff
@@ -341,6 +341,15 @@ func CreateResourcesLoop(ctx context.Context, kubeCl *client.KubernetesClient, r
...
+						"This could be due to lack of worker nodes in the cluster. "+
+						"Add at least one worker node or remove taints from master nodes (for single-node cluster) ",
+					app.ResourcesTimeout,
+				)
+			}
+
 			return fmt.Errorf("creating resources failed after %s waiting", app.ResourcesTimeout)
```
name212:
```suggestion
			return fmt.Errorf("Creating resources failed after %s waiting", app.ResourcesTimeout)
```

name212 (changes requested):
Some nits and DCO

name212 (approved):

PR #15359, opened 2025-09-05
[user-authn] Hotfix password connectors issue

s10dev:
## Description
This patch fixes a critical bug in the password connector. Previously, the login logic assumed that all password connectors must have a locally stored password in the database, which is incorrect. As a result, local connectors without a passwordDB (such as LDAP, Crowd, and Keystone) could fail during login. The patch ensures that login and password policy validation logic is executed only for password connectors with a passwordDB implementation, preventing login failures for other connectors.

The bug affected **all users of user-authn module** who use LDAP, Crowd, or Keystone connectors. Login attempts could fail because the code incorrectly required a local stored password. This patch prevent login errors and ensures proper handling for connectors without a local password store.

This should be included in a patch release because it fixes a critical bug that impacts all users relying on Dex with LDAP, Crowd, or Keystone connectors.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: user-authn
type: fix
summary: "Fix critical bug in password connector that caused login failures for LDAP, Crowd, and Keystone connectors."
```

### DIFF
--- a/modules/150-user-authn/images/dex/patches/010-fix-password-connector.patch
+++ b/modules/150-user-authn/images/dex/patches/010-fix-password-connector.patch
@@ -1,0 +1,234 @@
+diff --git a/server/handlers.go b/server/handlers.go
+index b3289261..1cfb0848 100644
+--- a/server/handlers.go
++++ b/server/handlers.go
+@@ -373,22 +373,32 @@ func (s *Server) handlePasswordLogin(w http.ResponseWriter, r *http.Request) {
+ 		password := r.FormValue("password")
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
Failure: cherry-pick commit f859efbb7eb2f410ce5eb9d18582a964b6a81291 to the branch [release-1.72](https://github.com/deckhouse/deckhouse/tree/release-1.72) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/17494339792) for details.

PR #15349, opened 2025-09-04
[node-manager] Upgrade CAPI manager. Increase and add leader elect timeouts

name212:
## Description
Upgrade patch version of CAPI manager due to logic fixes and fix cve's.

Increase leader elect leases timeout in CAPI manager.

Add leader elect leases timeout in CAPS manager.

## Why do we need it, and what problem does it solve?
We have some incorrect logical solution when caps crashes due bootstrap node.
We will fix it in 1.73 version.
Probably from version 1.71 (where we upgrade etcd to 3.6) etcd increase restart timeout and we have multiple restarts of CAPI and CAPS controllers when renew lease due to bootstrap cluster and we do not continue of bootstrap static nodes.
As temporary solution we add lease duration timeouts to CAPS and set and increase lease duration for CAPI.

image

## Why do we need it in the patch release (if we do)?

Probably from version 1.71 (where we upgrade etcd to 3.6) etcd increase restart timeout and we have multiple restarts of CAPI and CAPS controllers when renew lease due to bootstrap cluster and we do not continue of bootstrap static nodes.

## Checklist
- [ ] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [ ] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: node-manager
type: feature
summary: Add and increase lease duration timeouts to CAPS.
impact_level: default
---
section: node-manager
type: fix
summary: Upgrade CAPI version and increase lease duration timeouts.
impact_level: default
```

PR #15377, opened 2025-09-07
[node-manager] Add rbac permission for getting DaemonSets for CAPI. Decrease lease timeouts and increase log verbosity for CAPI

name212:
## Description

Add rbac permissions for getting daemonsets for capi-manager.
Increase log verbosity for capi-manager.
Decrease lease lock timeouts for capi-manager. 

Tested that ds cilium pods does not evicted.
image

When we upgrade capi version we forget to add permissions for getting daemonset. In new version capi uses its own drain helper and this helper gets daemonset  for pod if owner reference of pod is daemonset. If capi gets error while getting daemon set it marks pod for evict with warning. When pod with cilium agent get evict, then break connection to node.

Also we increase lease timeouts for capi in this pr https://github.com/deckhouse/deckhouse/pull/15349
But when recreate pod with capi it stuck in waiting lease for long time. It can be worse behavior for cloud installation with capi.

Increase log verbosity for capi-manager for better observability.

In release 1.72 we upgrade capi version and forget to add permission for getting daemonset. It can break connectivity with nodes due to drain

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Add rbac permission for getting DaemonSets for CAPI.
impact_level: default
---
section: node-manager
type: fix
summary: Decrease lease timeouts and increase log verbosity for CAPI.
impact_level: low
```

### DIFF
--- a/modules/040-node-manager/templates/capi-controller-manager/deployment.yaml
+++ b/modules/040-node-manager/templates/capi-controller-manager/deployment.yaml
@@ -93,7 +93,7 @@ spec:
         - --health-addr=127.0.0.1:9442
         - --webhook-port=4200
         - --skip-crd-migration-phases=StorageVersionMigration,CleanupManagedFields # disabled for now
-        - -v=4
+        - -v=3
         livenessProbe:
           httpGet:
             path: /healthz


deckhouse-BOaTswain:
Cherry pick PR [15383](https://github.com/deckhouse/deckhouse/pull/15383) to the branch [release-1.72](https://github.com/deckhouse/deckhouse/tree/release-1.72) successful!

PR #15421, opened 2025-09-10
[user-authn] Refactor DEX rate limit setup

s10dev:
## Description
Increase flexibility of Dex authentication requests for clients behind a single IP by adjusting Ingress rate limits and adding a global AuthRequest quota.
This prevents user requests from being blocked unnecessarily while still protecting the cluster from excessive AuthRequest objects.
- Raised per-IP request limits using a token-bucket approach in the Dex Ingress.
- Added a ResourceQuota for the total number of AuthRequest objects to protect the database.

Current limit of 20 AuthRequest/min per IP is insufficient for clients where many users share a single IP (e.g., corporate NAT).
Increasing the burst in rate-limiting allows multiple concurrent users without triggering 429 errors.
The ResourceQuota ensures that the total number of AuthRequest objects stays within safe limits, preventing database overload.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: user-authn
type: feature
summary: Increase Dex AuthRequest flexibility with token-bucket rate-limiting and global ResourceQuota
impact_level: default
```

### DIFF
--- a/modules/150-user-authn/templates/dex/ingress.yaml
+++ b/modules/150-user-authn/templates/dex/ingress.yaml
@@ -38,7 +38,7 @@ metadata:
     nginx.ingress.kubernetes.io/backend-protocol: HTTPS
     nginx.ingress.kubernetes.io/configuration-snippet: |
       {{- include "helm_lib_module_ingress_configuration_snippet" $ | nindent 6 }}
-    nginx.ingress.kubernetes.io/limit-rpm: "20"
+    nginx.ingress.kubernetes.io/limit-rpm: "1000"
     # Works only for ingress-controllers >=0.40. It is here to not forget to add the annotation after upgrading ingress controller.
     nginx.ingress.kubernetes.io/limit-burst-multiplier: "2"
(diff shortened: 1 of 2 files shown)


modules/150-user-authn/openapi/config-values.yaml
```diff
@@ -92,6 +92,10 @@ properties:
             A CA for accessing the API:
             * If the parameter is not set, Kubernetes CA is used.
             * We recommend using a self-signed certificate (and specify it as masterCA) if an HTTP proxy (that terminates HTTPS traffic) is used for exposing.
+  authRequestsLimit:
+    type: integer
+    default: 3000
+    description: Maximum number of AuthRequest objects.
```
nabokihms:
```suggestion
  authRequestsLimit:
    type: integer
    default: 3000
    description: Maximum number of AuthRequest objects.
```

We can just hardcode it, there is no reason for users to think about our hardlimits. And let's make it something like 10000.

PR #15543, opened 2025-09-19
feat: update DeckhouseHighMemoryUsage alert

timmilesdw:
## Description
- Update DeckhouseHighMemoryUsage alert:
  - Add namespace to grouping: max by (pod, namespace, node)
  - Remove hardcoded namespace filter to cover all namespaces
  - Lower threshold to > 0.85
  - Reduce for duration to 30s
- Sync documentation to include namespace in alert description.

This change only affects Prometheus alerting rules and documentation.

- Increases alert accuracy and visibility across namespaces by grouping on namespace instead of hardcoding a single namespace.
- Makes the alert more sensitive and faster to surface memory pressure on the Deckhouse container, improving MTTR.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: deckhouse-controller
type: feature
summary: Widen and tune DeckhouseHighMemoryUsage alert (namespace grouping, 0.85 threshold, 30s for)
```

### DIFF
--- a/modules/340-monitoring-deckhouse/monitoring/prometheus-rules/alerting.yaml
+++ b/modules/340-monitoring-deckhouse/monitoring/prometheus-rules/alerting.yaml
@@ -366,15 +366,15 @@
         ```
 
   - alert: DeckhouseHighMemoryUsage
-    expr: max by (pod, node) (container_memory_working_set_bytes{namespace="d8-system", container="deckhouse"} / container_spec_memory_limit_bytes{namespace="d8-system", container="deckhouse"} > 0.9)
-    for: 1m
+    expr: max by (pod, namespace, node) (container_memory_working_set_bytes{container="deckhouse"} / container_spec_memory_limit_bytes{container="deckhouse"} > 0.85)
+    for: 30s
     labels:
       severity_level: "3"
     annotations:
       plk_markup_format: "markdown"
       plk_protocol_version: "1"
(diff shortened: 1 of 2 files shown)


z9r5 (approved):

modules/340-monitoring-deckhouse/monitoring/prometheus-rules/alerting.yaml
```diff
@@ -366,15 +366,15 @@
         ```
 
   - alert: DeckhouseHighMemoryUsage
-    expr: max by (pod, node) (container_memory_working_set_bytes{namespace="d8-system", container="deckhouse"} / container_spec_memory_limit_bytes{namespace="d8-system", container="deckhouse"} > 0.9)
```
lazovskiy, line 369: I suggest keeping the namespace label.
Otherwise, it will alert on any container named `deckhouse`. Or at least limit the namespace to `d8-.+` if you have the intention to monitor other containers with that name (however, in this case, it will not match the alert name)

lazovskiy (changes requested):

modules/340-monitoring-deckhouse/monitoring/prometheus-rules/alerting.yaml
timmilesdw: done

PR #15546, opened 2025-09-19
[deckhouse-controller] Update release information fetching process

brileyyyy:
## Description

The changes ensure that when YAML parsing of changelog data fails, the metadata extraction process continues and processes other critical components like module.yaml instead of terminating early.

Release metadata extraction functions would fail completely when encountering malformed changelog YAML.

**Solution:**
- Logs changelog parsing errors as warnings instead of fatal errors
- Sets an empty changelog as fallback when parsing fails
- Continues processing other metadata components like module.yaml

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Update release information fetching process
impact_level: low
```

### DIFF
--- a/deckhouse-controller/pkg/controller/deckhouse-release/check_release.go
+++ b/deckhouse-controller/pkg/controller/deckhouse-release/check_release.go
@@ -780,7 +780,7 @@ func (f *DeckhouseReleaseFetcher) fetchReleaseMetadata(ctx context.Context, img 
 
 			changelog = make(map[string]any)
 		}
-		
+
 		meta.Changelog = changelog
 	}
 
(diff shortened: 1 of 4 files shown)


deckhouse-controller/pkg/controller/deckhouse-release/check_release.go
```diff
@@ -763,22 +763,16 @@ func (f *DeckhouseReleaseFetcher) fetchReleaseMetadata(ctx context.Context, img
 			f.logger.Warn("Unmarshal CHANGELOG yaml failed", log.Err(err))
 
 			meta.Changelog = make(map[string]any)
-
-			return meta, nil
+		} else {
```
ldmonster: just move moduleReader upper. do not use else

PR #15567, opened 2025-09-22
[admission-policy-engine] Add allow rbac wildcards SecurityPolicy flag and Gatekeeper template to restrict

Jabejixo:
## Description

This PR introduces a new Gatekeeper `ConstraintTemplate` and related Helm template logic for `D8AllowRbacWildcards`.  
It prohibits the use of `*` wildcards in `Role` and `RoleBinding` objects when the corresponding SecurityPolicy flag `allowRbacWildcards` is set to `false`.  
By default, wildcards remain allowed to preserve backward compatibility.

RBAC wildcards (`*` in role references or subject names) can unintentionally grant excessive privileges to users or workloads.  
This change allows cluster administrators to explicitly disallow wildcard usage via `SecurityPolicy.spec.policies.allowRbacWildcards: false`.  
It helps reduce the blast radius of misconfigured RBAC rules and increases cluster security posture.

If the flag is not specified or set to `true`, the current behavior remains unchanged.

Additionally, the PR contains two important fixes:

- Fixed reconciliation error caused by missing RBAC permissions for deleting `validatingadmissionpolicybindings`.
    
- Updated Gatekeeper from **v3.18.1** to **v3.18.3** to resolve the webhook error:
    
    ```
    error while excluding namespace: Object 'Kind' is missing in 'null'
    ```
    

Not necessarily

## Checklist

* [x] The code is covered by unit tests.
* [ ] e2e tests passed.
* [x] Documentation updated according to the changes.
* [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
type: feature
summary: Added `allowRbacWildcards` SecurityPolicy flag and Gatekeeper template to restrict `*` in Role and RoleBinding.
```

### DIFF
--- a/modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/allowed-repos.yaml
+++ b/modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/allowed-repos.yaml
@@ -31,12 +31,7 @@ spec:
       rego: |
         package d8.operation_policies
 
-        not_delete {
-          input.review.operation != "DELETE"
-        }
-
         violation[{"msg": msg}] {
-          not_delete
           container := input.review.object.spec.containers[_]
           satisfied := [good | repo = input.parameters.repos[_] ; good = startswith(container.image, repo)]
           not any(satisfied)
@@ -44,7 +39,6 @@ spec:
         }
 
         violation[{"msg": msg}] {
-          not_delete
           container := input.review.object.spec.initContainers[_]
           satisfied := [good | repo = input.parameters.repos[_] ; good = startswith(container.image, repo)]
           not any(satisfied)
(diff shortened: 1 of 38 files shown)


modules/015-admission-policy-engine/charts/constraint-templates/tests/allow-rbac-wildcards/rolebinding-invalid-roleref.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: invalid-rb-roleref
+  namespace: default
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-*-viewer
```
mvasl: Is this a valid dns name?

PR #15585, opened 2025-09-23
[global] Exclusion from PSS checks based on labels

Suselz:
## Description

Currently, all namespace deckhouses are excluded from `admission-policy-engine` checks. 
This is because there is no convenient tool for setting specific exceptions to checks for specific services. In this PR, I'm adding these mechanisms as `lib_helm` library and also demonstrating how to use these new features using a single node-exporter service.

Security improvements

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feature 
summary: security improvements  #CHANGE ME
impact_level: default
```


Suselz:
The concept has changed. It is no longer required.
Transform to changes #16162

PR #15826, opened 2025-10-06
[istio][metrics-exporter] Fix D8IstioRemoteClusterNotSynced alert

Paramoshka:
## Description

Added metric cleanup for non-existent Istio instances. Updated the description of the alert `D8IstioRemoteClusterNotSynced`, and made minor code improvements.

After restarting the Istiod instances, expired metrics remained, which led to the incorrect triggering of the "D8IstioRemoteClusterNotSynced" alert.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section:  istio
type: fix
summary: Fixed false-positive alert `D8IstioRemoteClusterNotSynced` and improved its description.
```

### DIFF
--- a/docs/documentation/_data/deckhouse-alerts.yml
+++ b/docs/documentation/_data/deckhouse-alerts.yml
@@ -1689,7 +1689,7 @@ alerts:
         - The remote API endpoint is not reachable.
         - The remote ServiceAccount token is invalid or expired.
         - There is a TLS or certificate issue between the clusters.
-        - If you are using Istio version 1.25, this alert will continue to be fired until the neighboring cluster has been updated to the same version.
+        - Remote and local Istio versions mismatch.
       summary: |
         Istio remote cluster {{$labels.cluster_id}} is not synchronized.
       severity: "4"


docs/documentation/_data/deckhouse-alerts.yml
```diff
@@ -1689,6 +1689,7 @@ alerts:
         - The remote API endpoint is not reachable.
         - The remote ServiceAccount token is invalid or expired.
         - There is a TLS or certificate issue between the clusters.
+        - If you have the latest version of Istiod installed, we recommend upgrading your remote clusters to the current version to ensure full resource synchronization.
```
AndreyPavlovFlant:
```suggestion
        - If you are using Istio version 1.25, this alert will continue to be fired until the neighboring cluster has been updated to the same version.
```

ee/modules/110-istio/monitoring/prometheus-rules/multicluster.yaml
```diff
@@ -75,3 +75,4 @@
           - The remote API endpoint is not reachable.
           - The remote ServiceAccount token is invalid or expired.
           - There is a TLS or certificate issue between the clusters.
+          - If you have the latest version of Istiod installed, we recommend upgrading your remote clusters to the current version to ensure full resource synchronization.
```
AndreyPavlovFlant:
```suggestion
          - If you are using Istio version 1.25, this alert will continue to be fired until the neighboring cluster has been updated to the same version.
```

ee/modules/110-istio/images/metrics-exporter/src/prometheus.go
```diff
@@ -174,3 +182,14 @@ func (p *PrometheusExporterMetrics) setClusterStatus(istiod, clusterID, secretNa
 	}
 	p.clusterUp.WithLabelValues(istiod, clusterID, secretName).Set(val)
 }
+
+func (p *PrometheusExporterMetrics) DeleteIstiodMetrics(podName string) {
+	if podName == "" {
+		return
+	}
+
+	removed := p.clusterUp.DeletePartialMatch(prometheus.Labels{"istiod": podName})
```
AndreyPavlovFlant, line 191: Will this lead to the complete removal of the series for this metric, or just to the end of its future publication while keeping historical data?

If it leads to complete removal, then this option would not be desirable.

z9r5 (approved):

ee/modules/110-istio/images/metrics-exporter/src/prometheus.go
Paramoshka: image

This does not affect the metrics in prometheus, only the metrics in the container's memory, it is equivalent to restarting it.

docs/documentation/_data/deckhouse-alerts.yml
Paramoshka: Done

ee/modules/110-istio/images/metrics-exporter/src/watcher.go
```diff
@@ -7,25 +7,30 @@ package main
...
-	pods map[string]*v1.Pod
+	metrics   *PrometheusExporterMetrics
+	pods      map[string]*IstioPodInfo
 }
 
-func NewWatcher(clientSet *kubernetes.Clientset) *Watcher {
+func NewWatcher(clientSet *kubernetes.Clientset, metrics *PrometheusExporterMetrics) *Watcher {
 	return &Watcher{
-		pods: make(map[string]*v1.Pod),
+		pods:      make(map[string]*IstioPodInfo),
```
skurbatov: It seems to me that it is necessary to remove tabs in this PR, where they are.

Paramoshka: Done

ee/modules/110-istio/monitoring/prometheus-rules/multicluster.yaml
```diff
@@ -75,3 +75,4 @@
           - The remote API endpoint is not reachable.
           - The remote ServiceAccount token is invalid or expired.
           - There is a TLS or certificate issue between the clusters.
+          - If you are using Istio version 1.25, this alert will continue to be fired until the neighboring cluster has been updated to the same version.
```
apolovov:
```suggestion
          - Remote and local Istio versions mismatch.
```

PR #15944, opened 2025-10-10
DO NOT MERGE [dhctl][1.69]chore(dhctl): update dhctl to use kube-client with in-memory cache option for Commander 1.11 CSE

distorhead:
## Description

Update dhctl to use kube-client with in-memory cache option for Commander 1.11 CSE
This PR uses kube-client  v1.2.3.

Commander 1.11 uses 1.69 dhctl-lib. So we need to actualize 1.69 codebase by adding necessary fix from the flant/kube-client 1.2.3.


name212:
only for cse DO NOT MERGE

PR #16007, opened 2025-10-15
[user-authn] Add the enabled flag for the Dex provider

Jabejixo:
## Description
Added DexProvider enable flag and kubectl printer columns in user-authn:
- CRD: introduced `spec.enabled: boolean` (default `true`) for all DexProvider versions; added `additionalPrinterColumns` (Type, Display Name, Enabled, Age).
- Hook: Dex config renderer now skips providers with `spec.enabled: false` (absence is treated as `true`).
- Docs: updated DexProvider docs and usage example to mention `spec.enabled`.

This feature may trigger a Dex rollout when providers are enabled/disabled, as the connectors list changes.

- Provides a safe “toggle” to temporarily disable a provider without deleting its CR, preventing accidental login options exposure.
- Improves observability with kubectl: operators can quickly see whether a provider is enabled.
- Keeps backward compatibility: existing CRs without `spec.enabled` continue working as enabled.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: user-authn
type: feature
summary: Add DexProvider .spec.enabled flag (default true) and kubectl printer columns; skip disabled providers in Dex connectors.
impact_level: default
```


modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -12,6 +12,10 @@ spec:
           properties:
             spec:
               properties:
+                enabled:
+                  description: |
+                    Если значение `false`, провайдер будет проигнорирован при формировании конфигурации Dex и не появится на странице логина.
+                    По умолчанию `true`.
```
denmaloyreb:
```suggestion
                    Если значение `false`, провайдер будет проигнорирован при формировании конфигурации Dex и не появится на странице логина.
```

The default value is specified in the en version of the CRD. It will be automatically added to the RU version. I suggest removing the duplication.

Image

PR #16054, opened 2025-10-17
[dhctl] fix(dhctl): add context to gossh client start func

faroyam:
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```

### DIFF
--- a/dhctl/pkg/system/node/gossh/client.go
+++ b/dhctl/pkg/system/node/gossh/client.go
@@ -91,6 +91,10 @@ func (s *Client) OnlyPreparePrivateKeys() error {
 }
 
 func (s *Client) Start() error {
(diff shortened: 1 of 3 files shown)


dhctl/pkg/system/node/gossh/client.go
```diff
@@ -445,3 +449,31 @@ func (s *Client) stopKubeproxy() {
...
+		case <-ctx.Done():
+			if conn != nil {
+				conn.Close()
+			}
+		}
+	}()
+	select {
+	case res := <-ch:
+		return res.conn, res.err
+	case <-ctx.Done():
```
distorhead, line 476: Maybe do conn.Close() here instead, and do not catch ctx.Done in subroutine.

PR #16535, opened 2025-11-18
[node-local-dns] Changing stale_cache depending on HA mode

Paramoshka:
## Description

The settings of the cache plugin have been adopted depending on the cluster mode.

* When HA mode is effective (module setting, global setting, or auto‑discovered), node-local-dns sets CoreDNS serve_stale to 1h immediate. This keeps DNS responsive by serving cached answers while background refresh happens if the control plane is temporarily unavailable.

* In non‑HA environments, serve_stale is set to 1h verify to prefer fresh answers and fall back to the cache only on upstream failure. This reduces the risk of serving outdated records in frequently updated dev/stage clusters.

On development clusters, stage environments often roll over, and as a result, jobs sometimes fail, receiving an invalid response from the cache.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-local-dns
type: feature
summary: Optimized the cache plugin for clusters in dev mode.
```


ee/be/modules/350-node-local-dns/openapi/doc-ru-config-values.yaml
```diff
@@ -2,3 +2,8 @@ properties:
   enableLogs:
     description: |
       Включает логирование запросов для экземпляров node-local-dns.
+  highAvailability:
+    description: |
+      Ручное управление режимом отказоустойчивости.
```
apolovov:
```suggestion
      Ручное управление режимом отказоустойчивости. В рамках данного модуля опция влияет на механизм кеширования DNS-записей на узлах:
      * `true`: по истечении TTL у отдельной DNS-записи, node-local-dns продолжит возвращать "устаревшее" значение немедленно и параллельно с этим будет запрашивать свежее значение у вышестоящего kube-dns. Таким образом, при проблемах с узлами, на которых работает kube-dns, приложения продолжат получать ответы на DNS-запросы с устаревшим TTL, но без задержки.
      * `false`: по истечении TTL у отдельной DNS-записи, node-local-dns продолжит возвращать "устаревшее" значение, но только после запроса к вышестоящему kube-dns. Таким образом, при проблемах с узлами, на которых работает kube-dns, приложения продолжат получать DNS-ответы с устаревшим TTL, но с возможной большой задержкой. Данный подход уместен в высокодинамичных окружениях, где часто пересоздаются Service и Namespace.
```

ee/be/modules/350-node-local-dns/openapi/config-values.yaml
```diff
@@ -5,3 +5,10 @@ properties:
       Enables query logging for node-local-dns instances.
     type: boolean
     default: false
+  highAvailability:
+    type: boolean
+    x-examples: [true]
+    description: |
+      Manually enable the high availability mode.
```
apolovov:
```suggestion
      Manual failover mode management. Within this module, the option affects the DNS records caching mechanism on nodes:
      * `true`: upon the expiration of the TTL for an individual DNS record, node-local-dns will continue to return the "stale" value immediately while simultaneously requesting a fresh value from the upstream kube-dns. Thus, in the event of issues with the nodes where kube-dns is running, applications will continue to receive responses to DNS queries with expired TTL, but without delay.
      * `false`: upon the expiration of the TTL for an individual DNS record, node-local-dns will continue to return the "stale" value, but only after querying the upstream kube-dns. Thus, in the event of issues with the nodes where kube-dns is running, applications will continue to receive DNS responses with expired TTL, but with potentially significant delay. This approach is suitable in highly dynamic environments where Services and Namespaces are frequently recreated.
```

PR #16694, opened 2025-11-26
[deckhouse-controller] fix: readiness bug

timmilesdw:
## Description

Refactored `searchModuleBatchHooks()` to be a pure function that returns `hasReadiness` as a separate value instead of modifying the struct field directly. The `bm.hasReadiness` field is now set only in `RegisterHooks()` after successful registration.

This change does not affect critical cluster components.

When `RegisterHooks()` fails after `searchModuleBatchHooks()` has already found a readiness hook (e.g., error in `AssembleEnvironmentForModule()`), the `hasReadiness` flag remains `true` while `hooks.registered` stays `false`. On retry, `searchModuleBatchHooks()` finds the same readiness hook again but sees `hasReadiness` already set to `true`, incorrectly reporting "multiple readiness hooks found".

**Root cause**: `searchModuleBatchHooks()` was modifying `bm.hasReadiness` as a side effect during the search, before registration was complete.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Fixed false-positive "multiple readiness hooks found" error on module hook registration retry.
impact_level: default 
```


deckhouse-BOaTswain:
Failure: cherry-pick commit 6b56ca8b1ff7285fa22a22033a3b1e07f3c70066 to the branch [release-1.74](https://github.com/deckhouse/deckhouse/tree/release-1.74) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/19823519343) for details.

PR #16771, opened 2025-12-01
[docs-watcher] Fix: registry watcher diff error

timmilesdw:
## Description

When a registry returns an error (e.g., 401 Unauthorized), the scanner now immediately returns `nil` instead of continuing with incomplete data. This prevents the cache from being synchronized with partial registry data, which would cause false deletions of module documentation.

Changes:
- Return `nil` immediately when `registry.Modules()` fails
- Return `nil` immediately when `registry.ListTags()` fails

When registry-modules-watcher receives a 401 error from a container registry, it previously continued processing with an empty module list for that registry. This caused `SyncWithRegistryVersions()` to interpret all modules from that registry as deleted, generating delete tasks for all documentation.

**Impact**: Upon temporary registry unavailability (auth issues, network problems), all module documentation would be deleted and then recreated when access was restored. This caused unnecessary churn and potential downtime for documentation.

**Root cause**: The scanner was calling `cache.SyncWithRegistryVersions()` with incomplete data when any registry failed.

**Fix**: On any registry error, skip cache synchronization entirely and retry on the next scan interval.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: docs
type: fix
summary: Fixed registry-modules-watcher deleting all documentation when registry returns an error.
impact_level: default
```

### DIFF
--- a/docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
+++ b/docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
@@ -63,21 +63,27 @@ func (s *registryscanner) processRegistries(ctx context.Context) []backends.Docu
 			s.logger.Error("registry is unavailable",
 				slog.String("registry", registry.Name()),
 				log.Err(err))
-			continue
+			return nil
 		}
 
 		s.logger.Debug("found modules",
 			slog.Any("modules", modules),
 			slog.String("registry", registry.Name()))
 
-		vers := s.processModules(ctx, registry, modules)
+		vers, err := s.processModules(ctx, registry, modules)
+		if err != nil {
+			s.logger.Error("failed to process modules",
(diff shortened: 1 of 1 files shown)


docs/site/backends/registry-modules-watcher/internal/backends/pkg/registry-scanner/process.go
```diff
@@ -89,7 +89,7 @@ func (s *registryscanner) processModules(ctx context.Context, registry Client, m
 				slog.String("module", module),
 				slog.String("registry", registry.Name()),
 				log.Err(err))
-			continue
+			return nil
```
ldmonster: we must return error here, and check it on the top

timmilesdw: fixed

PR #16919, opened 2025-12-08
Add project name length validation

RapidNode777:
## Description
This PR adds validation for the Project resource name to ensure it does not exceed 53 characters.
Project names are used as prefixes for generated Kubernetes resource names. These names later get a hashed suffix, and exceeding the Kubernetes 63-character limit would cause resource creation failures.

## Why do we need this:
Without this validation, users can create Project objects with names too long to be safely used as base names for generated resources. This results in broken deployments and hard-to-debug errors.

## What was changed:
- Added x-kubernetes-validations rule to CRD projects.deckhouse.io (v1alpha2).
- Validation rule: size(self.metadata.name) <= 53.
- Added user-facing message explaining the restriction.

### DIFF
--- a/modules/160-multitenancy-manager/crds/projects.yaml
+++ b/modules/160-multitenancy-manager/crds/projects.yaml
@@ -85,6 +85,9 @@ spec:
       schema:
         openAPIV3Schema:
           type: object
+          x-kubernetes-validations:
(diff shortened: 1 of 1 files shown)


RapidNode777:
Closing this PR. I will recreate the same change from a branch in the main deckhouse repo (now that I have org access).

PR #17123, opened 2025-12-18
[node-manage] fencing agent, detect dead node without control plane

W1seKappa:
## Description

During traffic distribution across worker nodes in a cluster, one of them may become unavailable.
In the current implementation, to monitor the availability of a node for traffic processing, we use a fencing agent, which is represented by a daemonset on each node.
Its logic is based on the network availability of the kube-apiserver.
That is, if the control plane fails, all worker nodes start to think that they are unavailable, and the watchdog mechanism used inside the fencing agent restarts the node.
This is logically incorrect behavior, since the worker nodes can continue to process traffic even without the control plane.
But how to assess the actual availability of a node if the control plane is down?
The Gossip protocol could be a solution.
An implementation provided by the [memberlist library](https://github.com/hashicorp/memberlist) was chosen.

There are two mechanism, which we used to detect "dead" node:
<details>
  <summary> Images: </summary>
image
image
</details>

The speed of detecting a node failure depends on many factors. The memberlist library has many settings that need to be adjusted to suit the specific situation and cluster characteristics.
For the experiment, a NetworkPolicy was used to disable network connectivity for the pod on a specific node.
```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: block-pod-traffic
spec:
  podSelector:
    matchLabels:
      test: block 
  policyTypes:
  - Ingress
  - Egress
  ingress: []
  egress: []   
```
We have a cluster of 4 nodes, 3 of which are worker nodes running pods with memberlist.
Memberlist has the following configuration settings:
```
- PROBE_INTERVAL - The interval between direct health check probes sent to a random node. Lower values speed up failure detection but increase network traffic.
value: 200ms
- PROBE_TIMEOUT - How long to wait for a probe response before considering it failed. Lower values make detection faster but may cause false positives on slow networks.
value: 100ms
- SUSPICION_MULT - Multiplier that determines how long a node stays in "suspected" state before being declared dead. Lower values speed up failure detection but increase the risk of false positives during network hiccups.
value: 2
- INDIRECT_CHECKS - Number of indirect probes sent through other nodes when a direct probe fails. Higher values improve detection reliability but generate more network traffic.
value: 3
- GOSSIP_INTERVAL - How often gossip messages are exchanged with random nodes to propagate cluster state. Lower values speed up information propagation but increase network load.
value: 100ms
RETRANSMIT_MULT - Multiplier that determines how many times gossip messages are retransmitted. Higher values ensure message delivery in unreliable networks but increase traffic.
value: 4
- GOSSIP_TO_THE_DEAD_TIME - How long to continue gossiping about dead nodes before removing them completely. Affects how long failure information propagates through the cluster.
value: 2s
```
With these settings, given our cluster configuration (node count, network characteristics, etc.), we can achieve a specific node failure detection speed.
```
echo "Start: $(date '+%Y-%m-%d %H:%M:%S:%3N')"; time kubectl apply -f block_pod.yaml

Start: 2025-12-24 10:14:56:719

networkpolicy.networking.k8s.io/block-pod-traffic created

real	0m0.148s
user	0m0.096s
sys	0m0.023s
```

```json
{
   "level":"info",
   "timestamp":"2025-12-24T10:14:57.522Z",
   "msg":"Node left, have to notify cilium",
   "node":"bobkov-worker-aeecacd3-8gkfh-c6dpd"
}
```
2025-12-24T10:14:57.522 - 2025-12-24 10:14:56:719 = 803ms for detect dead node.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```

### DIFF
--- a/modules/040-node-manager/images/fencing-agent/src/internal/agent/controller.go
+++ b/modules/040-node-manager/images/fencing-agent/src/internal/agent/controller.go
@@ -224,11 +224,11 @@ func (fa *FencingAgent) Run(ctx context.Context) error {
 					fa.logger.Error("Unable to feed watchdog", zap.Error(err))
 				}
 			}
-
+			num := fa.gs.NumMembers() - 1
 			if !APIIsAvailable && !MaintenanceMode {
 				// except this node
-				num := fa.gs.NumMembers() - 1
-				fa.logger.Debug("Number of members in gossip", zap.Int("num", num))
+
+
 				if (num == 0 && fa.gs.IsAlone()) || num > 0{
 					fa.logger.Debug("Feeding the watchdog from gossip check", zap.Bool("alone", fa.gs.IsAlone()))
 					err = fa.watchDog.Feed()
@@ -243,7 +243,7 @@ func (fa *FencingAgent) Run(ctx context.Context) error {
 					}
 				}
 			}
-
+			fa.logger.Debug("Number of members in gossip", zap.Int("num", num))
 		case <-ctx.Done():
 			fa.logger.Debug("Finishing the API check")
 			if fa.watchDog.IsArmed() {


W1seKappa:
irrelevant

PR #17305, opened 2025-12-25
[user-authn] Forbid IP addresses in DexAuthenticator application domains

RapidNode777:
## Description

This PR adds validation to the `DexAuthenticator` custom resource to **forbid IP addresses** in domain fields (`spec.applicationDomain`, `spec.additionalApplications[].domain`, and `v2alpha1: spec.applications[].domain`).

Key changes:

- Extends the validating webhook `dexauthenticators-unique.deckhouse.io` with IP address detection logic (both IPv4 and IPv6).
- Returns a clear error message on CREATE/UPDATE if a domain field contains an IP address instead of a DNS name.
- Updates CRD documentation (`dex-authenticator.yaml`, `doc-ru-dex-authenticator.yaml`) to clarify that IP addresses are not allowed.

This change does **not** restart or affect any critical cluster components. The webhook runs in the user-authn module and only validates incoming `DexAuthenticator` resources.

Previously, the CRD's OpenAPI schema pattern allowed values like `155.111.1.22` to pass validation because the regex `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` treats IP addresses as valid "DNS-like" strings.

When a user created a `DexAuthenticator` with an IP in `spec.applicationDomain`, the module attempted to create an **Ingress** resource with that IP in `spec.rules[].host`. Kubernetes **rejects such Ingresses** ([validation.go#L330-L355](https://github.com/kubernetes/kubernetes/blob/04d87a4b6e72336ee9afb1e5b477223c96a8fcbb/pkg/apis/networking/validation/validation.go#L330-L355)) because `host` must be a DNS name, not an IP address.

This caused:
- The Ingress creation to fail silently or get stuck in the reconciliation queue.
- Confusing user experience: the CR was accepted, but authentication never worked.

**This PR moves the validation earlier** — at the time of `DexAuthenticator` creation — so users get an immediate, actionable error message:

Not necessarily.

This is a user-facing validation improvement. While it prevents a confusing failure mode, it does not fix a regression or critical bug introduced in a recent patch release. It can go into the next minor release.

## Checklist
- [x] The code is covered by unit tests. *(The webhook is a bash script; syntax validated with `bash -n`.)*
- [ ] e2e tests passed. 

## Changelog entries

```changes
section: user-authn
type: fix
summary: Forbided IP addresses in DexAuthenticator domain fields; only DNS names are allowed.
impact: DexAuthenticator resources with IP addresses in domain fields will now be rejected at creation/update time with a clear error message. Previously, such resources were accepted but failed silently during Ingress creation.
impact_level: default
```

### DIFF
--- a/modules/150-user-authn/webhooks/validating/dex_authenticator
+++ b/modules/150-user-authn/webhooks/validating/dex_authenticator
@@ -16,30 +16,19 @@
 
 source /shell_lib.sh
 
-function is_ipv4_address() {
-  local ip="$1"
+function is_ip_address() {
+  local value="$1"
 
-  [[ "$ip" =~ ^([0-9]{1,3}\.){3}[0-9]{1,3}$ ]] || return 1
+  # Use Python's stdlib to properly parse IPv4/IPv6 instead of maintaining complex regexes in bash.
+  python3 - "$value" >/dev/null 2>&1 <<'PY'
+import ipaddress
+import sys
 
-  local IFS='.'
-  local o1 o2 o3 o4
-  read -r o1 o2 o3 o4 <<<"$ip"
-
-  for o in "$o1" "$o2" "$o3" "$o4"; do
-    [[ "$o" =~ ^[0-9]{1,3}$ ]] || return 1
-    (( o >= 0 && o <= 255 )) || return 1
-  done
-
-  return 0
-}
-
-function is_ipv6_address() {
-  local ip="$1"
-
-  # DNS names must not contain ':'; presence of ':' strongly indicates IPv6 (or host:port),
-  # both are invalid for Ingress host and therefore invalid for applicationDomain.
-  [[ "$ip" == *:* ]] || return 1
-  return 0
+try:
+    ipaddress.ip_address(sys.argv[1])
+except ValueError:
+    sys.exit(1)
+PY
 }
 
 function __config__(){
@@ -109,7 +98,7 @@ function __main__() {
   for domain_pair in "${domains_to_check[@]}"; do
     IFS=$'\t' read -r domain ingressClass fieldPath <<< "$domain_pair"
 
-    if is_ipv4_address "$domain" || is_ipv6_address "$domain"; then
+    if is_ip_address "$domain"; then
       msg="DexAuthenticator '$newAuthNamespace/$newAuthName' is invalid: '$fieldPath' must be a DNS name, not an IP address ('$domain')"
       jq -n --arg msg "$msg" '{allowed:false, message:$msg}' > "$VALIDATING_RESPONSE_PATH"
       return 0


RapidNode777:
I tested by this way

How validation works (example with IP - invalid option):


Correct option:

modules/150-user-authn/crds/doc-ru-dex-authenticator.yaml
```diff
@@ -21,6 +21,7 @@ spec:
                     Адрес приложения (домен DNS для Ingress-ресурса), для которого необходима аутентификация.
 
                     **Внимание.** Адрес не должен содержать HTTP-схему.
+                    **Внимание.** Должно быть DNS-имя (RFC 1123 subdomain), а не IP-адрес.
```
maxinsky:
```suggestion
                    > **Внимание.** Укажите DNS-имя, соответствующее требованиям RFC 1123 (subdomain), без HTTP-схемы. Использование IP-адресов не допускается.
```

modules/150-user-authn/crds/doc-ru-dex-authenticator.yaml
```diff
@@ -137,6 +138,7 @@ spec:
                           Домен приложения (для использования в Ingress-ресурсе), запросы на который будут перенаправлены в Dex для аутентификации.
 
                           **Внимание.** Домен не должен содержать HTTP-схему.
+                          **Внимание.** Должно быть DNS-имя (RFC 1123 subdomain), а не IP-адрес.
```
maxinsky:
```suggestion
                          > **Внимание.** Укажите DNS-имя, соответствующее требованиям RFC 1123 (subdomain), без HTTP-схемы. Использование IP-адресов не допускается.
```

modules/150-user-authn/crds/doc-ru-dex-authenticator.yaml
```diff
@@ -227,6 +229,7 @@ spec:
...
```
maxinsky:
```suggestion
                          > **Внимание.** Укажите DNS-имя, соответствующее требованиям RFC 1123 (subdomain), без HTTP-схемы. Использование IP-адресов не допускается.
```

modules/150-user-authn/crds/dex-authenticator.yaml
```diff
@@ -74,6 +74,7 @@ spec:
                     An address for the application (DNS domain for Ingress resource) from which the user request will be redirected for authentication in Dex.
 
                     **Warning.** Must be specified without HTTP scheme.
+                    **Warning.** Must be a DNS name (RFC 1123 subdomain) and must not be an IP address.
```
maxinsky:
```suggestion
                    > **Warning.** Specify a DNS name that complies with RFC 1123 (subdomain) and does not include an HTTP scheme. IP addresses are not allowed.
```

modules/150-user-authn/crds/dex-authenticator.yaml
```diff
@@ -255,6 +256,7 @@ spec:
                           An application domain from which the user request will be redirected for authentication in Dex.
 
                           **Warning.** Must be specified without HTTP scheme.
+                          **Warning.** Must be a DNS name (RFC 1123 subdomain) and must not be an IP address.
```
maxinsky:
```suggestion
                          > **Warning.** Specify a DNS name that complies with RFC 1123 (subdomain) and does not include an HTTP scheme. IP addresses are not allowed.
```

modules/150-user-authn/crds/dex-authenticator.yaml
```diff
@@ -313,6 +315,7 @@ spec:
                           An application domain (for Ingress resources) from which the user request will be redirected for authentication in Dex.
 
                           **Warning.** Must be specified without HTTP scheme.
+                          **Warning.** Must be a DNS name (RFC 1123 subdomain) and must not be an IP address.
```
maxinsky:
```suggestion
                          > **Warning.** Specify a DNS name that complies with RFC 1123 (subdomain) and does not include an HTTP scheme. IP addresses are not allowed.
```

z9r5 (approved):

RapidNode777:
> IPv6 detection is kind of weird and parsing such addresses in bash can be problematic, maybe python will be a better choice. Have you tested IPv6 validation?

Thanks! IPv6 isn’t supposed to pass here anyway — the CRD pattern rejects anything with :, so the webhook never even sees an IPv6 value. I added the Python-based check to correctly validate IPv4 cases the regex previously missed, and to avoid the brittle “:`”-detector.
image

PR #17503, opened 2026-01-16
[testing] fix test output not showing stderr on failure

timmilesdw:
## Description

Changed `Output()` to `CombinedOutput()` in gator test execution so stderr is captured alongside stdout. Moved the output directly into the `Fail()` message instead of printing separately.

When gator policy tests fail, the test wasn't showing the actual error details. `exec.Command.Output()` only captures stdout, but gator writes failure information to stderr. This made debugging failing tests difficult since the error output was silently discarded.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: fix
summary: Fixed gator policy test output not showing error details on failure
impact_level: low
```


timmilesdw:
included in #17549

PR #17605, opened 2026-01-21
[deckhouse] display error messages in PackageRepository status

timmilesdw:
## Description

Added conditions-based status tracking to PackageRepository that syncs error messages from PackageRepositoryOperation resources. When a scan operation fails, the error message and reason are now displayed in the PackageRepository status, making it easier to diagnose issues without checking the operation resource directly.

Added a `Conditions` field to `PackageRepositoryStatus` with a `LastOperationScanFinished` condition that tracks scan operation status. The controller now watches PackageRepositoryOperation resources and syncs their status to the repository when operations complete. Updated CRDs to include a MSG column in kubectl output showing error messages.

Previously, when a PackageRepositoryOperation failed (e.g., registry connection issues, authentication failures), the error messages were only visible in the PackageRepositoryOperation resource. Users had to manually check operation resources to understand why a repository scan failed, which wasn't obvious from the PackageRepository status.

Now error messages from failed operations are automatically synced to the PackageRepository status, both in the `status.message` field and in the `status.conditions` array. This follows Kubernetes conventions for condition-based status reporting and makes failures immediately visible when inspecting the repository resource.

## Examples

- `kubectl get packagerepository`
```
➜ kubectl get packagerepository
NAME        PHASE    SYNC                   MSG
test-repo   Active   2026-01-28T18:46:26Z   Failed to create registry client: create package service: failed to get auth from docker config: read auth config: unmarshal docker config: invalid character 'a' looking for beginning of value
```
- `kubectl get packagerepositoryoperations.deckhouse.io `
```
➜  kubectl get packagerepositoryoperations.deckhouse.io 
NAME                        COUNT   PROCESSED   MSG
test-repo-scan-1769625985   5       True        
test-repo-scan-1769626009           False       Failed to create registry client: create package service: failed to get auth from docker config: read auth config: unmarshal docker config: invalid character 'a' looking for beginning of value
```

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Display error messages from failed scan operations in PackageRepository status.
impact_level: low
```


deckhouse-controller/pkg/apis/deckhouse.io/v1alpha1/package_repository.go
```diff
@@ -115,6 +119,36 @@ type PackageRepositoryStatus struct {
 	// Human-readable message about the repository status.
 	// +optional
 	Message string `json:"message,omitempty"`
+
+	// Conditions represent the latest available observations of the repository's state.
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	Conditions []PackageRepositoryStatusCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
ipaqsa: Its better to use metav1.Condition. Application conditions will be redesigned too

deckhouse-controller/pkg/controller/packages/package-repository-operation/controller.go
```diff
@@ -660,3 +677,60 @@ func (r *reconciler) setOperationTruePhase(ctx context.Context, operation *v1alp
...
+			}
+
+			repo.Status.Conditions[idx].Reason = reason
+			repo.Status.Conditions[idx].Message = message
+			break
+		}
+	}
+
+	if !conditionExists {
+		status := corev1.ConditionTrue
```
ipaqsa: It seems like duplicate, you can define it before the loop

ipaqsa (changes requested):
Provide please examples from the cluster. And the second part of the task is to show error in an operation too(just add printable column). Check the card for details

PR #17329, opened 2025-12-26
[observability] Add propagated recording rules

maksimmadzhuga:
## Description
This PR splits cluster and cluster propagated prometheus recording rules.

## Why do we need it, and what problem does it solve?
When observability module is enabled it converts some cluster wide prometheus recording rules to propagated observability rules so that their metrics will be routed to a proper project / namespace (based on alert namespace label).

## Checklist
- [ ] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: prometheus
type: feature
summary: Replace PrometheusRules with ClusterObservabilityMetricsRulesGroups or ClusterObservabilityPropagatedMetricsRulesGroups when deployed using helm_lib_prometheus_rules helper and the observability module is enabled
impact_level: default
```

PR #17638, opened 2026-01-23
[monitoring-kubernetes] Filter unbound PVCs

lazovskiy:
## Description
In #17329, the `kube_persistentvolume_is_local` recording rule was changed to have a `namespace` label to be usable on propagated dashboards. However, if the PV is not yet provisioned (e.g., in the Pending state), the `kube_persistentvolumeclaim_info` metric has series without a volumename, leading to a many-to-many operation in the expression. This change filters out `kube_persistentvolumeclaim_info` series without a volume name. If the PV is not bound (thus, not used by any payload in the cluster), we can safely skip it.

Presence of PVCs in the Pending state leads to rule evaluation failures.

It would be nice to have this change in the next patch release

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
type: fix
summary: fix kube_persistentvolume_is_local recording rule for not-bound PVCs
impact_level: default
```

### DIFF
--- a/modules/340-monitoring-kubernetes/monitoring/prometheus-rules/generic/propagated-kube-state-metrics.yaml
+++ b/modules/340-monitoring-kubernetes/monitoring/prometheus-rules/generic/propagated-kube-state-metrics.yaml
@@ -19,6 +19,6 @@
         or
           (kube_persistentvolume_info - 1)
       ) * on(persistentvolume) group_left(namespace) max by (namespace, persistentvolume) (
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:
Cherry pick PR [17640](https://github.com/deckhouse/deckhouse/pull/17640) to the branch [release-1.74](https://github.com/deckhouse/deckhouse/tree/release-1.74) successful!

PR #17659, opened 2026-01-26
[deckhouse] clean up APV usedBy when application version changes

timmilesdw:
## Description

Fixed a bug where ApplicationPackageVersion (APV) resources retained stale `usedBy` references when an Application switched to a different package version. When an app updated from v1.0.1 to v1.0.2, the old v1.0.1 APV incorrectly showed `usedByCount=1` instead of being cleared.

Changes:
- Added version change detection in Application controller
- Cleanup old APV's `usedBy` list when version changes
- Remove stale APV owner references from Application
- Refactored ApplicationPackage status: renamed `installed`/`installedOverall` to `usedBy`/`usedByCount` and changed from map-based to flat list structure (now tracks version per app)
- Fixed pre-existing bug in `setConditionFalse` (was accessing wrong array)

When users update an Application's `spec.packageVersion`, the old ApplicationPackageVersion resource should no longer show the app in its `usedBy` list. Previously, old versions retained these references, causing confusion about which versions are actually in use and potentially blocking cleanup operations.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: deckhouse
type: fix
summary: Fixed APV usedBy not being cleared when application version changes.
impact_level: low
```


deckhouse-controller/crds/doc-ru-applicationpackage.yaml
```diff
@@ -13,15 +13,20 @@ spec:
...
                   items:
                     properties:
                       name:
                         description: |
-                          Название установленного приложения.
-                installedOverall:
+                          Название экземпляра приложения.
+                      namespace:
+                        description: |
+                          Пространство имён, в котором установлено приложение.
```
z9r5:
```suggestion
                          Неймспейс, в котором установлено приложение.
```

PR #17695, opened 2026-01-28
[deckhouse] enhance nelm service

ipaqsa:
## Description
It enhances nelm service to support packages(module/app)

Before we passed an application directly to nelm, now we pass interface.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: chore
summary: Enhance nelm service.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/internal/packages/manager/nelm/errors.go
+++ b/deckhouse-controller/internal/packages/manager/nelm/errors.go
@@ -26,7 +26,6 @@ const (
 	ConditionReasonRender           status.ConditionReason = "RenderFailed"
 	ConditionReasonCheckChart       status.ConditionReason = "CheckChart"
 	ConditionReasonCreateValuesFile status.ConditionReason = "CreateValuesFile"
-	ConditionReasonRuntimeValues    status.ConditionReason = "MarshalRuntimeValues"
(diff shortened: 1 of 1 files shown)


deckhouse-controller/internal/packages/manager/manager.go
```diff
@@ -232,7 +232,7 @@ func (m *Manager) RunPackage(ctx context.Context, name string) error {
 
 	m.logger.Debug("run nelm upgrade", slog.String("name", name))
 
-	if err := m.nelm.Upgrade(ctx, app); err != nil && !errors.Is(err, nelm.ErrPackageNotHelm) {
+	if err := m.nelm.Upgrade(ctx, app.GetNamespace(), app); err != nil && !errors.Is(err, nelm.ErrPackageNotHelm) {
```
LazarenkoA, line 235: why are we calling the `GetNamespace` method here, we could add it to the Package interface and pass only the app as a parameter

ipaqsa: A Module can't have namespace(all releases are stored in d8-system), so I decided make it an argument

PR #17771, opened 2026-02-02
[node-manager] fencing agent

W1seKappa:
## Description

This PR introduces a new implementation of fencing-agent with gossip-based node failure detection and gRPC API.

### Problem

During traffic distribution across worker nodes in a cluster, one of them may become unavailable.
In the previous implementation, to monitor the availability of a node for traffic processing, we used a fencing agent based on the network availability of the kube-apiserver.
That is, if the control plane fails, all worker nodes start to think that they are unavailable, and the watchdog mechanism used inside the fencing agent restarts the node.
This is logically incorrect behavior, since the worker nodes can continue to process traffic even without the control plane.

The Gossip protocol is used to assess the actual availability of a node even when the control plane is down.
An implementation provided by the [memberlist library](https://github.com/hashicorp/memberlist) was chosen.

### Node Failure Detection Strategy

There are two mechanisms used to decide whether to feed the watchdog:

**Tier 1: Quorum-Based**
- Calculates quorum: `(totalNodes / 2) + 1`
- Feeds watchdog if current cluster members >= quorum
- Total node count is broadcast via memberlist gossip

**Tier 2: Fallback to API Access**
- If quorum not met, tries to access Kubernetes API
- Feeds watchdog if API is reachable (proves connectivity)

### gRPC API

Proto: `modules/040-node-manager/images/fencing-agent/src/api/v1/fencing.proto`

Added a gRPC server that listens on a Unix Socket created on each node.

This implementation solves the split-brain problem that occurs when using control-plane availability as the sole indicator of node health. With gossip-based detection:
- Worker nodes can correctly assess their availability even when control plane is down
- Nodes in quorum continue processing traffic normally
- Only truly isolated nodes (not in quorum AND cannot reach API) are fenced
- External consumers can subscribe to node membership events via gRPC

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: feature
summary: Added gossip-based node failure detection and gRPC API to fencing-agent.
impact: |
  The fencing-agent now uses gossip protocol (`memberlist`) for distributed node health monitoring.
  This prevents incorrect node reboots when control plane is unavailable but worker nodes are healthy.
  A new gRPC API is available via Unix socket at `/tmp/fencing-agent.sock` for querying node membership.
impact_level: default
```


nabokihms (approved):

z9r5 (approved):

z9r5:
@W1seKappa Please fill the PR description and changelog section

modules/040-node-manager/images/fencing-agent/ADR.md
```diff
@@ -0,0 +1,748 @@
+# ADR: Fencing-agent на основе memberlist с механизмом Lifeguard
```
z9r5: @W1seKappa Is it necessary to keep the ADR (in Russian) here?

W1seKappa: deleted)

z9r5 (approved):

mortelumina98 (approved):

yalosev (approved):

PR #17986, opened 2026-02-16
[descheduler] Fix converting v1alpha1 into v1alpha2

W1seKappa:
## Description

Make the `get_crds` hook in the descheduler module resilient to Descheduler CRs that have no valid v1alpha2 strategies.

When a v1alpha1 Descheduler CR contains only deprecated strategies (e.g. `removePodsViolatingNodeTaints`, `removeFailedPods`, `removePodsHavingTooManyRestarts`, `removePodsViolatingTopologySpreadConstraint`), the conversion webhook strips them during v1alpha1 -> v1alpha2 conversion but does not produce an error. The resulting v1alpha2 object is stored in etcd with an empty `spec.strategies`. When the hook processes this object, it writes empty strategies into internal values, which fails the `anyOf` OpenAPI validation — causing the hook to error out and the module queue to hang in an infinite retry loop.

Changes:
- Add `HasValidStrategies()` method to the `Strategies` struct in `hooks/internal/v1alpha2/descheduler.go`.
- In `hooks/get_crds.go`, skip Descheduler CRs with no valid strategies instead of adding them to internal values. A warning is logged with the CR name so users can identify and fix the problematic resource.

This change does not affect any critical cluster components. No restarts of ingress-controllers, control-plane, or Prometheus are triggered.

Applying a v1alpha1 Descheduler CR with a strategy that only exists in v1alpha1 (e.g. `removePodsViolatingNodeTaints`) causes the Deckhouse module queue to hang permanently.

**Root cause chain:**

1. v1alpha1 is `served: true` and `deprecated: true` in the CRD — the API server accepts the request and shows a deprecation warning.
2. Since v1alpha2 is the storage version, the conversion webhook converts the object. It deletes deprecated-only strategies but does not check whether any valid strategies remain after conversion.
3. The converted object with empty `spec.strategies` is stored in etcd (Kubernetes does not re-validate after conversion).
4. The `get_crds` hook receives the object, serializes it into internal values with `strategies: {}`.
5. The OpenAPI values schema (`openapi/values.yaml`) requires at least one strategy via `anyOf` — validation fails.
6. The hook returns an error, addon-operator retries the event, the same error occurs — **infinite loop, queue is blocked**.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: descheduler
type: fix
summary: Fixed module queue hang when a v1alpha1 Descheduler CR with deprecated-only strategies is applied.
impact_level: default
```

### DIFF
--- a/modules/400-descheduler/hooks/common_test.go
+++ b/modules/400-descheduler/hooks/common_test.go
@@ -19,11 +19,11 @@ package hooks
 import (
 	"testing"
 
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
 )
 
 func Test(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "")
+	gomega.RegisterFailHandler(ginkgo.Fail)
+	ginkgo.RunSpecs(t, "")
 }


deckhouse-BOaTswain:
:green_circle: `e2e: AWS` for `deckhouse:fix_converting_v1alpha1_into_v1alpha2` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/22058982451) in 33m6s.



AWS-WithoutNAT-Containerd-1.33 - Connection string: `ssh  ec2-user@3.78.58.81`


:green_circle: `e2e: AWS, Containerd, Kubernetes 1.33` succeeded in 31m57s.

</details>

PR #18161, opened 2026-02-26
[deckhouse] Problem when creating the config for the global

RottenRat:
## Description
`Global` module is not contained in `Module CR`, so an attempt to get it from `Module` always fails. If we bypass these cases, we should skip this module everywhere.

```
root@master-0:~/packages# kubectl get mc global
NAME     ENABLED   VERSION   AGE   MESSAGE
global   true      2         34d   
root@master-0:~/packages# kubectl get module global
Error from server (NotFound): modules.deckhouse.io "global" not found
```

This was not the correct behavior, because this module should never be contained in Module CR, but the method should not return a reject error.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Fixed global configuration generation.
```

### DIFF
--- a/deckhouse-controller/pkg/apis/deckhouse.io/validation/validate_module_config.go
+++ b/deckhouse-controller/pkg/apis/deckhouse.io/validation/validate_module_config.go
@@ -135,16 +135,18 @@ func moduleConfigValidationHandler(
 				}
 
 				// Fallback: check from Module CR (when module not downloaded, but properties synced from registry).
-				// Create: reject if module not found.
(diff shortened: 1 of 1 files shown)


ipaqsa (approved):

ldmonster (approved):

deckhouse-BOaTswain:
Cherry pick PR [19689](https://github.com/deckhouse/deckhouse/pull/19689) to the branch [release-1.75](https://github.com/deckhouse/deckhouse/tree/release-1.75) successful!

PR #18007, opened 2026-02-16
[admission-policy-engine] Fix SecurityPolicy constraints for omitted fields

RapidNode777:
## Description
Fixes `SecurityPolicy` constraints generation in `admission-policy-engine` so omitted boolean fields don’t unintentionally enable unrelated Gatekeeper constraints. Adds regression and edge-case template tests for this behavior.

## Why do we need it, and what problem does it solve?
When a `SecurityPolicy` had only `allowPrivileged: false` set, Helm templating logic used negation (`not`) on omitted boolean fields and incorrectly rendered additional constraints (e.g. `D8HostNetwork`), causing workload Pods to be denied with unexpected `hostNetwork` / `hostPort` violations.

This change makes constraint rendering depend on the explicit presence of corresponding fields (except documented defaults) and prevents accidental enforcement of unrelated policies.

## Why do we need it in the patch release (if we do)?
Fixes unexpected admission denials for workloads when `SecurityPolicy` omits unrelated boolean fields. This is a user-facing bug affecting cluster operations and should be safe to backport.

## Checklist
- [x] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: admission-policy-engine
type: fix
summary: Prevent unintended Gatekeeper constraints from being rendered for SecurityPolicy when boolean fields are omitted.
impact: Workload Pods are no longer denied by unrelated SecurityPolicy checks (e.g. hostNetwork/hostPort) when corresponding policy fields are not explicitly set.
impact_level: default
```

PR #18197, opened 2026-02-27
backport: [admission-policy-engine] Fix SecurityPolicy constraints for omitted fields

Suselz:
## Description
Fixes `SecurityPolicy` constraints generation in `admission-policy-engine` so omitted boolean fields don’t unintentionally enable unrelated Gatekeeper constraints. Adds regression and edge-case template tests for this behavior.

When a `SecurityPolicy` had only `allowPrivileged: false` set, Helm templating logic used negation (`not`) on omitted boolean fields and incorrectly rendered additional constraints (e.g. `D8HostNetwork`), causing workload Pods to be denied with unexpected `hostNetwork` / `hostPort` violations.

This change makes constraint rendering depend on the explicit presence of corresponding fields (except documented defaults) and prevents accidental enforcement of unrelated policies.

Fixes unexpected admission denials for workloads when `SecurityPolicy` omits unrelated boolean fields. This is a user-facing bug affecting cluster operations and should be safe to backport.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Prevent unintended Gatekeeper constraints from being rendered for SecurityPolicy when boolean fields are omitted.
impact: Workload Pods are no longer denied by unrelated SecurityPolicy checks (e.g. hostNetwork/hostPort) when corresponding policy fields are not explicitly set.
impact_level: default
```

### DIFF
--- a/modules/015-admission-policy-engine/template_tests/security_policies_test.go
+++ b/modules/015-admission-policy-engine/template_tests/security_policies_test.go
@@ -92,6 +92,50 @@ var _ = Describe("Module :: admissionPolicyEngine :: helm template :: security p
 			Expect(f.KubernetesGlobalResource("D8AllowRbacWildcards", testPolicyName).Exists()).To(BeTrue())
 		})
 
+		It("Minimal security policy must not render unrelated constraints", func() {
+			Expect(f.RenderError).ShouldNot(HaveOccurred())
+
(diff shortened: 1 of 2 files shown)


Suselz:
manual backport of #18007

PR #18218, opened 2026-03-02
[node-manager] fencing-controller. Add filtering by new label

W1seKappa:
## Description
Fencing controller hook now reads two additional node labels — `node-manager.deckhouse.io/fencing-mode` and `node.deckhouse.io/type` — and uses them to decide whether to delete the Node object after deleting pods.

**Notify mode**: when the fencing agent runs without an armed watchdog, there is no mechanism to reboot the node. If the hook deletes the Node object, the node cannot rejoin the cluster on its own. By preserving the Node object, we allow the node to recover and return to a healthy state.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Fencing controller no longer deletes Node objects for Notify-mode and Static/CloudStatic nodes.
impact_level: default
```

### DIFF
--- a/modules/040-node-manager/hooks/fencing_controller.go
+++ b/modules/040-node-manager/hooks/fencing_controller.go
@@ -31,7 +31,6 @@ import (
 	sdkobjectpatch "github.com/deckhouse/module-sdk/pkg/object-patch"
 
 	"github.com/deckhouse/deckhouse/go_lib/dependency"
-	"github.com/deckhouse/deckhouse/go_lib/set"
 	"github.com/deckhouse/deckhouse/pkg/log"
 )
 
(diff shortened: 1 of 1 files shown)


modules/040-node-manager/hooks/fencing_controller.go
```diff
@@ -104,8 +114,8 @@ func fencingControllerHandler(_ context.Context, input *go_hook.HookInput, dc de
 		return err
 	}
 
-	// make map with nodes to kill
 	nodesToKill := set.New()
+	nodesToDeletePod := set.New()
```
yalosev:
```
map[string]bool

nodeName: needToBeKilled<bool>

map
[node-1]: true
[node-2]: false

for k,v := range map {
  //TODO: Evict pods
  if v {
  // TODO : delete node
  }
} 
```

PR #18252, opened 2026-03-03
[control-plane-manager] Feature/kubeadm replacement etcd

borovetsav:
## Description

Refactored cluster kubeadm logic and implemented part of our own reusable library. Added two core functions: `InitCluster` and `JoinCluster` for cluster initialization and node joining.

This allows the Control Plane Manager to use a dedicated internal library instead of relying on external binaries.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feature
summary: Introduced internal library with InitCluster and JoinCluster functions
impact_level: default
```

### DIFF
--- a/go_lib/controlplane/etcd/client/client.go
+++ b/go_lib/controlplane/etcd/client/client.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2026 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


go_lib/controlplane/client/constants/constants.go
```diff
@@ -0,0 +1,49 @@
...
+	EtcdListenPeerPort = 2380
+
+	// EtcdHealthcheckClientCertName defines etcd's healthcheck client certificate name
+	EtcdHealthcheckClientCertName = "etcd/healthcheck-client.crt"
+	// EtcdHealthcheckClientKeyName defines etcd's healthcheck client key name
+	EtcdHealthcheckClientKeyName = "etcd/healthcheck-client.key"
+
+	// EtcdAdvertiseClientUrlsAnnotationKey is the annotation key on every etcd pod, describing the
+	// advertise client URLs
+	EtcdAdvertiseClientUrlsAnnotationKey = "kubeadm.kubernetes.io/etcd.advertise-client-urls"
```
trofimovdals, line 43: We planning to use this annotation `control-plane-manager.deckhouse.io/etcd.advertise-client-urls`

borovetsav: fixed it

PR #18303, opened 2026-03-04
[istio] Refactoring of federation and multicluster discovery

MrZuwZuw:
## Description
Added new istio subdomain for `istio-metadata` which will be used in metadata-exporter Ingress for metadata exchange in multicluster. While migration, we keep both old and new subdomain and alerting if old one is in use.

### Refactor

Logic of `federation_discovery.go` and `multicluster_discovery.go` moved to `metadataExporter` package and now use same interface to start, validate and proceed discovery. Almost everything in discovery process is unified - that means we can combine these two hooks in one, like alliance_discovery.go and cron-fetch for both resources of `IstioMulticluster` and `IstioFederation` - now there are many options about how-to handle it.

Both CrdInfo use embedded Common struct with shared fields for each Kind also with unique fields declared right in them.
Added Discovery struct with all needed functions to operate with entire process of discovery of certain discovery type (multicluster | federation) over each resource of `IstioMulticluster` and `IstioFederation` 

### Alerts

**Added two new alerts:**

- **D8IstioAllianceDiscoveryMetadataEndpointUsesDeprecatedSubdomain** - `istiomulticlusters` or `istiofederations` use old subdomain
- **D8IstioAllianceDiscoveryMetadataEndpointAccessedViaDeprecatedSubdomain** - another cluster uses old subdomain

**Added podMonitor**
To export `d8_istio_metadata_exporter_accessed_via_deprecated_subdomain` metric

**Added two additional labels for both federation and multicluster alerts**

- alliance_kind (IstioMulticluster|IstioFederation)
- scope (public|private)

**Renamed alert labels for both federation and multicluster alerts**

- `multicluster_name` > `name` 
- `federation_name` > `name`

So, in summary, in federation or multicluster alert we have labels like:
```
"alliance_kind": "IstioMulticluster",
"name":  "proper-multicluster-0",
"endpoint":      "https://proper-hostname-0/metadata/public/public.json",
"scope":         "public",
```
or
```
"alliance_kind": "IstioFederation",
"name":  "proper-federation-0",
"endpoint":      "https://proper-hostname-0/metadata/private/federation.json",
"scope":         "private",
```
which actually means we can delete both alerts (`D8IstioMulticlusterMetadataEndpointDoesntWork` and `D8IstioFederationMetadataEndpointDoesntWork`) and replace with just one `Alliance-like-alert`, eg `**D8IstioAllianceDiscoveryMetadataEndpointDoesntWork**` because we have entire context:

- which Kind
- which name
- public or private scope
- and exact endpoint
- _could be_ added one of the reasons firing alert (eg general error - not connected at all, error in generating auth token,  bad http code, wrong metadata format, error while patching status)

In this case we have single-named alerts for each kind of Kind and name of the cluster with full description of a problem.
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: feature
summary: Changed the default Istio subdomain for the metadata-exporter Ingress.
impact_level: default
```


modules/110-istio/openapi/doc-ru-config-values.yaml
```diff
@@ -14,6 +14,9 @@ properties:
       Класс Ingress-контроллера, который используется для Kiali, metadata-exporter и api-proxy.
 
       Опциональный параметр, по умолчанию используется глобальное значение `modules.ingressClass`.
+  customMetadataExporterDomain:
+    description: |
+      Пользовательский домен для metadata-exporter ingress.
```
z9r5:
```suggestion
      Пользовательский домен для metadata-exporter Ingress.
```

modules/110-istio/openapi/config-values.yaml
```diff
@@ -33,6 +33,14 @@ properties:
       The class of the Ingress controller used for Kiali, metadata-exporter and proxy-api.
 
       Optional. By default, the `modules.ingressClass` global value is used.
+  customMetadataExporterDomain:
+    type: string
+    pattern: '^([a-z0-9]([-a-z0-9]*[a-z0-9])?\.)+[a-z]{2,}$'
+    description: |
+      Custom domain for metadata-exporter ingress.
```
z9r5:
```suggestion
      Custom domain for Ingress metadata-exporter.
```

z9r5 (approved):

skurbatov (approved):

ee/modules/110-istio/templates/alliance/metadata-exporter/ingress.yaml
```diff
@@ -16,7 +16,7 @@ metadata:
 spec:
   ingressClassName: {{ include "helm_lib_module_ingress_class" . | quote }}
   rules:
-  - host: {{ include "helm_lib_module_public_domain" (list . "istio") }}
+  - host: {{ include "helm_lib_module_public_domain" (list . "istio-metadata") }}
```
apolovov, line 29: Nonono, the old domain have to continue working.
Also, there must be alert which checks the IstioMulticluster | IstioFederation resources and fires if there is an old domain.

skurbatov (approved):

PR #18345, opened 2026-03-06
[go_lib/controlplane] Node mark as control plane

meatuses:
## Description
Add node package that functionally replaces `kubeadm init phase mark control-plane`

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feature
summary: Replace `kubeadm mark control-plane` feature with internal library.
impact_level: low
```

### DIFF
--- a/go_lib/controlplane/node/node.go
+++ b/go_lib/controlplane/node/node.go
@@ -17,10 +17,12 @@ package node
 
 import (
 	"context"
+	"encoding/json"
 	"log/slog"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/deckhouse/deckhouse/pkg/log"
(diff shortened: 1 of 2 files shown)


go_lib/controlplane/node/node.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// Convert back to slice
+	node.Spec.Taints = make([]corev1.Taint, 0, len(existingTaints))
+	for _, taint := range existingTaints {
+		node.Spec.Taints = append(node.Spec.Taints, taint)
+	}
+
+	_, err = m.kubeClient.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
+	return err
+}
```
AmazinMax, line 141: I would suggest considering the approach of updating patches using Patch(...), something like:
```go
func (m *NodeManager) setTaints(nodeName string, taints []corev1.Taint) error {
	patchData := map[string]interface{}{
		"spec": map[string]interface{}{
			"taints": taints,
		},
	}

	patchBytes, _ := json.Marshal(patchData)

	_, err := m.kubeClient.CoreV1().Nodes().Patch(
		context.Background(),
		nodeName,
		types.StrategicMergePatchType,
		patchBytes,
		metav1.PatchOptions{},
	)
	return err
}
```

It might also be necessary to provide error handling due to patch conflicts (409) using appropriate retry logic.

As a result, this approach gives us
- Resilience to concurrent changes;
- Performance: one request instead of two (Get() + Update());
- Minimal traffic: only changed data is transferred.

go_lib/controlplane/node/node.go
```diff
@@ -0,0 +1,113 @@
...
+		return err
+	}
+
+	for k, v := range labels {
+		node.Labels[k] = v
+	}
+
+	_, err = m.kubeClient.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
+	return err
+}
```
AmazinMax, line 94: similar as [setTaints review comment](https://github.com/deckhouse/deckhouse/pull/18345/changes#r2895233597)

go_lib/controlplane/node/node.go
meatuses: For taints, there is really no way to blindly override taints by key/effect pair. Reconstructing the list of taints manually seems to be the only way. This is how kubeadm does it https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/markcontrolplane/markcontrolplane.go#L59-L66
I choose to compare every taint by key to ensure we have proper control-plane taint set and other taints are kept.
It is ok for labels though. I have switched from node.update to just plain patch()

PR #18375, opened 2026-03-10
[istio] Merge public services by hostname across federations

yshch:
## Description

When multiple `IstioFederation` CRs expose public services with the same hostname, separate `ServiceEntry` and `DestinationRule` resources were created per federation, causing undefined Istio behavior.

The `alliance_metadata_merge.go` hook now merges public services by hostname across federations into `istio.internal.serviceEntries`, aggregating endpoints and using first-wins for ports. A warning is logged when port definitions conflict. The `remote-services.yaml` template produces a single `ServiceEntry` and `DestinationRule` per unique (hostname, endpoints) combination.

**Note:** Existing resources will be recreated with new names, causing a brief traffic interruption during the first reconciliation.

In a scenario where two clusters in a multicluster setup share the same cluster domain (e.g., `cluster.local`) and are both federated with a third cluster, both expose public services with identical hostnames (e.g., `my-svc.my-ns.svc.cluster.local`). The third cluster ends up with two `ServiceEntry` objects and two `DestinationRule` objects targeting the same host. Istio does not handle this well - behavior is undefined when multiple `ServiceEntry` resources declare the same host, and conflicting `DestinationRule` objects cause routing issues.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: fix
summary: Deduplicated federation ServiceEntry and DestinationRule resources by hostname across multiple IstioFederation CRs.
impact: ServiceEntry and DestinationRule resources for federated public services will be recreated with new names. This causes a brief traffic interruption for cross-cluster federated service routing during the first reconciliation after the update.
impact_level: high
```

### DIFF
--- a/modules/110-istio/openapi/values.yaml
+++ b/modules/110-istio/openapi/values.yaml
@@ -70,6 +70,11 @@ properties:
         default: []
         x-examples:
         - [{"name": "aaa", "trustDomain": "bbb", "spiffeEndpoint": "ccc", "ingressGateways": [{"address": "1.2.3.4", "port": 1234}], "publicServices": [{"hostname": "zzz.xxx.ccc", "ports": [{"name": "ddd", "port": 2345, "protocol": "TCP"}], "virtualIP": "3.4.5.6"}]}]
+      federationsMergedPublicServices:
+        type: array
+        default: []
+        x-examples:
+        - [{"hostname": "zzz.xxx.ccc", "ports": [{"name": "ddd", "port": 2345, "protocol": "TCP"}], "endpoints": [{"address": "1.2.3.4", "port": 15443}]}]
       multiclusters:
         type: array
         default: []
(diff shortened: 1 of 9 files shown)


z9r5 (approved):

modules/110-istio/openapi/values.yaml
```diff
@@ -67,6 +67,11 @@ properties:
         default: []
         x-examples:
         - [{"name": "aaa", "trustDomain": "bbb", "spiffeEndpoint": "ccc", "ingressGateways": [{"address": "1.2.3.4", "port": 1234}], "publicServices": [{"hostname": "zzz.xxx.ccc", "ports": [{"name": "ddd", "port": 2345, "protocol": "TCP"}], "virtualIP": "3.4.5.6"}]}]
+      serviceEntries:
```
apolovov:
```suggestion
      federationServiceEntries:
```

yshch: Fixed in https://github.com/deckhouse/deckhouse/pull/18375/changes/8d946223d9dc038d903c4e2de644b26b711d7352

skurbatov (approved):

apolovov (approved):

PR #18476, opened 2026-03-13
[istio] Cleanup hook code

yshch:
## Description

Cleanup and small fixes in the Istio module hooks:

- **discovery_istiod_health.go**: Fix schedule name typo (`discovery_istiod_health` instead of `disocvery_istiod_health`).
- **ensure_crds_istio.go**: Clarify error message (reference `discovery_versions_to_install.go` instead of `discovery_versions.go`). Add validation before applying CRDs: check that the CRD glob pattern is valid and matches at least one file, and return clear errors (`invalid glob pattern` / `no CRD files found matching pattern`) instead of failing later.
- **ensure_crds_istio_test.go**: Align tests with the new error message and correct the test description (CRD version 0.991).
- **generate_ca.go**: Move error check immediately after the call that can fail (fail-fast), before using the result.
- **lib/istio_versions/istio_versions.go**: Allocate slice with correct capacity `make([]string, 0, len(vm))` and drop nolint directives.
- **migration_delete_obsolete_ingress.go**: Use `GetK8sClient()` with error handling instead of `MustGetK8sClient()`; set hook order to 15 for consistency.
- **migration_service_with_many_ports.go**: Fix argument order in `input.PatchCollector.Delete()` (namespace before name per API).

Hook logic remains mostly the same.

Improves code quality, consistency with project style (error handling, naming, slice allocation), and fixes a typo in the schedule name and the wrong argument order in `PatchCollector.Delete`. The extra validation in `ensure_crds_istio.go` gives clearer errors when CRD files are missing or the glob is invalid.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: chore
summary: Cleanup Istio module hook code (typos, error handling, validation, style).
impact_level: low
```

### DIFF
--- a/modules/110-istio/hooks/migration_service_with_many_ports.go
+++ b/modules/110-istio/hooks/migration_service_with_many_ports.go
@@ -78,8 +78,8 @@ func patchServiceWithManyPorts(_ context.Context, input *go_hook.HookInput) erro
 		input.PatchCollector.Delete(
 			"v1",
 			"Service",
-			serviceInfoObj.Namespace,
 			serviceInfoObj.Name,
+			serviceInfoObj.Namespace,
 		)
 	}
 	return nil


modules/110-istio/hooks/migration_service_with_many_ports.go
```diff
@@ -78,8 +78,8 @@ func patchServiceWithManyPorts(_ context.Context, input *go_hook.HookInput) erro
 		input.PatchCollector.Delete(
 			"v1",
 			"Service",
-			serviceInfoObj.Name,
 			serviceInfoObj.Namespace,
+			serviceInfoObj.Name,
```
apolovov, line 81: It isn't worth staining git log for a file.

yshch: This seems like a major bug, unless i'm missing something. The hook is scoped to `kiali` service, so the blast area is pretty small. I think we should either fix it, or drop the entire hook/file, since it just doesn't seem to work at all and does absolutely nothing as is.

yshch: Removed this change from PR after discussion, will handle this separately.

PR #16499, opened 2025-11-14
[deckhouse-controller] Add CNI migration mechanism

tarabrind:
## Description

This PR implements a mechanism to switch CNI plugins (e.g., from Flannel to Cilium) in an existing cluster.

Key changes include:

- Added `CNIMigration` and `CNINodeMigration` Custom Resource Definitions.
- Implemented `cni-migration-controller` consisting of a Manager (orchestrates the flow) and Agents (run on nodes to clean up network configurations).
- Added the `cni-migration-init-checker` init container to all CNI agent pods  (`cni-cilium`, `cni-flannel`, `cni-simple-bridge`, `kube-proxy`), ensuring the new CNI pods start only after the node has been cleaned up.
- Added global hooks to detect active migrations and disable conflicting webhooks during the process.
- Added comprehensive documentation describing the migration procedure.

Please note that the migration process triggers a restart of all pods in the cluster to switch them to the new network interface.

The implementation of cli capabilities is implemented in the deckhouse-cli utility with this [PR](https://github.com/deckhouse/deckhouse-cli/pull/242).

## Why do we need it, and what problem does it solve?

Changing the CNI plugin in a running Kubernetes cluster is a complex operation that usually requires manual intervention or a full cluster re-installation. This feature provides an automated, controlled way to migrate networking with minimal downtime, handling node cleanup and pod restarts automatically.

## Checklist

- [x] The code is covered by unit tests.
- [x] e2e tests passed.
- [x] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Added a mechanism to migrate between CNI plugins (e.g., Flannel to Cilium) in a running cluster.
---
section: cni-cilium
type: feature
summary: Added a mechanism to migrate between CNI plugins (e.g., Flannel to Cilium) in a running cluster.
impact: All cilium agents will restart.
impact_level: default
---
section: cni-flannel
type: feature
summary: Added a mechanism to migrate between CNI plugins (e.g., Flannel to Cilium) in a running cluster.
impact: All flannel agents will restart.
impact_level: default
---
section: cni-simple-bridge
type: feature
summary: Added a mechanism to migrate between CNI plugins (e.g., Flannel to Cilium) in a running cluster.
impact: All simple-bridge agents will restart.
impact_level: default
---
section: kube-proxy
type: feature
[...]

PR #18479, opened 2026-03-13
[deckhouse-controller] Prevent enabling multiple CNI modules simultaneously

tarabrind:
## Description

This PR assigns all CNI modules (`cni-cilium`, `cni-flannel`, and `cni-simple-bridge`) to a single `exclusiveGroup: cni`. This leverages the built-in ValidatingWebhookConfiguration to prevent users from accidentally enabling more than one CNI module at the same time.

Additionally, it fixes a bug in the ModuleConfig webhook validation logic. Previously, the exclusive group check was performed unconditionally, which blocked users from editing the configuration of a disabled CNI module if another CNI module was already active. The logic is now properly wrapped in an `if cfg.Spec.Enabled` check so the validation only triggers when a module is explicitly being enabled.

This PR prohibits the simultaneous activation of multiple CNIs, even for the purpose of switching CNIs within a cluster. CNI switching is implemented in DKP via the platform's built-in functionality and has been added via a separate [PR](https://github.com/deckhouse/deckhouse/pull/16499), which allows multiple CNIs to be activated simultaneously within a cluster during the switching process.

Enabling multiple CNI modules simultaneously can lead to severe cluster network instability and conflicts.

By assigning them to the same `exclusiveGroup`, the Deckhouse admission webhook will gracefully reject the creation or modification of a ModuleConfig if it attempts to enable a second CNI module. Users must now explicitly disable the currently running CNI module before enabling a new one. The bug fix in the webhook ensures that users can still safely update the settings (e.g., in preparation for a migration) of disabled modules without triggering false-positive rejections.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Prevented enabling multiple CNI modules simultaneously.
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/apis/deckhouse.io/validation/validate_module_config.go
+++ b/deckhouse-controller/pkg/apis/deckhouse.io/validation/validate_module_config.go
@@ -274,21 +274,24 @@ func moduleConfigValidationHandler(
 		if err != nil {
 			return allowResult(warnings)
 		}
-		exclusiveGroup := module.GetModuleExclusiveGroup()
-		if exclusiveGroup != nil {
-			modules := moduleStorage.GetModulesByExclusiveGroup(*exclusiveGroup)
 
-			for _, moduleName := range modules {
-				// if any module with same unique key enabled, return error
-				if moduleManager.IsModuleEnabled(moduleName) && moduleName != cfg.Name {
-					return rejectResult(
-						fmt.Sprintf(
-							"can't enable module %q because different module %q with same exclusiveGroup %s enabled",
-							cfg.Name,
-							moduleName,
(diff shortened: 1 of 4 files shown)


deckhouse-BOaTswain:
:green_circle: `e2e: Static` for `deckhouse:feat/cni-exclusive-group` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/23186101276) in 50m21s.



Static-Static-Containerd-1.33 - Connection string: `ssh -J astra@185.184.54.129 astra@192.168.199.209`


:green_circle: `e2e: Static, Containerd, Kubernetes 1.33` succeeded in 49m0s.

</details>

PR #18537, opened 2026-03-18
[dhctl] Add external logger form lib-dhctl

YuryLysov:
## Description

Add logger from lib-dhctl, add wrapper to pkg/log

First part of refactoring dhctl, to switch to lib-connection instead of current ssh/kube-client implementation

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: chore
summary: Add logger from lib-dhctl.
impact_level: low
```

### DIFF
--- a/dhctl/pkg/log/logger.go
+++ b/dhctl/pkg/log/logger.go
@@ -592,6 +592,26 @@ func (d *DummyLogger) Write(content []byte) (int, error) {
 	return len(content), nil
 }
 
+type ExternalProcessLogger struct {
+	logger external.ProcessLogger
+}
+
+func (e *ExternalProcessLogger) LogProcessStart(name string) {
+	e.logger.ProcessStart(name)
+}
+
+func (e *ExternalProcessLogger) LogProcessEnd() {
+	e.logger.ProcessEnd()
+}
+
+func (e *ExternalProcessLogger) LogProcessFail() {
+	e.logger.ProcessFail()
+}
+
+func newExternalProcessLogger(logger external.ProcessLogger) *ExternalProcessLogger {
+	return &ExternalProcessLogger{logger: logger}
+}
+
 type ExternalLogger struct {
 	logger external.Logger
 }
@@ -601,7 +621,7 @@ func (e *ExternalLogger) GetLogger() external.Logger {
 }
 
 func (e *ExternalLogger) ProcessLogger() ProcessLogger {
-	return newWrappedProcessLogger(e)
+	return newExternalProcessLogger(e.logger.ProcessLogger())
 }
 
 func (e *ExternalLogger) NewSilentLogger() *SilentLogger {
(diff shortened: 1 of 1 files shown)


dhctl/pkg/log/logger.go
```diff
@@ -555,6 +584,86 @@ func (d *DummyLogger) Write(content []byte) (int, error) {
...
+func (e *ExternalLogger) LogProcess(_, t string, run func() error) error {
+	return e.logger.Process("", t, run)
+}
+
+func (e *ExternalLogger) LogInfoF(format string, a ...interface{}) {
+	e.logger.InfoF(format, a...)
+}
+
+func (e *ExternalLogger) LogInfoLn(a ...interface{}) {
+	e.logger.InfoLn(a...)
```
name212: let try to use *F functions.

Ln are deprecated

dhctl/pkg/log/logger.go
```diff
@@ -72,7 +73,35 @@ func (l *klogWriterWrapper) Write(p []byte) (int, error) {
...
+		log.Default().SetOutput(defaultLogger)
+	}
+}
+
+func initExternalKlog(logger *ExternalLogger) {
+	err := ext_logger.InitKlog(logger.logger, ext_logger.WithKlogSanitizer(ext_logger.NewKeywordSanitizer()))
+	if err != nil {
+		panic(err)
+	}
+	klog.SetOutput(newKlogWriterWrapper(logger))
```
name212: InitKlog set correct output 

https://github.com/deckhouse/lib-dhctl/blob/main/pkg/log/klog.go#L97

dhctl/pkg/log/logger.go
```diff
@@ -72,7 +73,35 @@ func (l *klogWriterWrapper) Write(p []byte) (int, error) {
...
+		// Enable shell-operator log, because it captures klog output
+		// todo: capture output of klog with default logger instead
+		log.Default().SetLevel(log.LevelDebug)
+		// Wrap them with our default logger
+		log.Default().SetOutput(defaultLogger)
+	}
+}
+
+func initExternalKlog(logger *ExternalLogger) {
+	err := ext_logger.InitKlog(logger.logger, ext_logger.WithKlogSanitizer(ext_logger.NewKeywordSanitizer()))
```
name212: please init sanitazer and pass keyywords https://github.com/deckhouse/lib-dhctl/blob/main/pkg/log/klog.go#L117C28-L117C50 from here https://github.com/deckhouse/deckhouse/blob/main/dhctl/pkg/log/log_sanitizer.go#L22

dhctl/pkg/log/logger.go
name212: but be carefully, it can may produce `[]` wrap

name212: we can use *F because *F now add \n

dhctl/pkg/log/logger.go
```diff
@@ -72,7 +73,35 @@ func (l *klogWriterWrapper) Write(p []byte) (int, error) {
 }
 
 func InitLogger(loggerType string) {
-	InitLoggerWithOptions(loggerType, LoggerOptions{IsDebug: app.IsDebug})
+	InitExternalLogger(loggerType)
+}
+
+func InitExternalLogger(loggerType string) {
```
name212: and please return lib-dhctl logger from here to pass in handlers

name212: and you can retuturn err instead panics

PR #18756, opened 2026-03-26
[docs-builder] Add alert

RottenRat:
## Description
Added a Prometheus gauge metric `docs_builder_module_render_error` to the docs-builder. The metric is set to 1 for each module whose documentation failed to render and was consequently removed during the Hugo build process.

The metric uses a group-based lifecycle: at the start of every build the previous set of error gauges is expired, so modules that recover between builds no longer appear as broken in Prometheus. A Prometheus alert with severity 6 fires when the gauge is non-zero.

When the docs-builder encounters a Hugo render error for a module it silently removes that module's documentation and retries. Previously this failure was only visible in logs - there was no way to detect it via monitoring or trigger an alert.

This change makes render failures observable: operators and on-call engineers will receive an alert as soon as a module's documentation is dropped, allowing them to investigate and fix the broken module content without having to grep logs manually.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: docs
type: feature
summary: Added an alert for failed documentation renderings.
impact_level: default
```

### DIFF
--- a/docs/site/backends/docs-builder/internal/metrics/register-metrics.go
+++ b/docs/site/backends/docs-builder/internal/metrics/register-metrics.go
@@ -35,44 +35,79 @@ const (
 
 func RegisterMetrics(mStorage *metricsstorage.MetricStorage) error {
 	// Counters: count of upload/build/delete requests (ok/fail)
-	_, err := mStorage.RegisterCounter(DocsBuilderBuildTotal, []string{"status"})
+	_, err := mStorage.RegisterCounter(
+		DocsBuilderBuildTotal,
+		[]string{"status"},
+		options.WithHelp("Total number of docs builder build requests (success/failure)"),
+	)
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


docs/site/backends/docs-builder/internal/metrics/register-metrics.go
```diff
@@ -67,5 +68,11 @@ func RegisterMetrics(mStorage *metricsstorage.MetricStorage) error {
 		return err
 	}
 
+	// Gauge: module documentation render error (1.0 = broken module was removed during build)
+	_, err = mStorage.RegisterGauge(DocsBuilderModuleRenderError, []string{"module"})
```
ldmonster: we need to group and expire group before build

z9r5 (approved):

z9r5:
@ldmonster Should we have the alert not only in public site but also in a deployed cluster (as in the current implementation)?

ldmonster (approved):

PR #18769, opened 2026-03-26
fix expr in CertmanagerCertificateExpired alert

rvekaterina:
## Description

Add `!= 0` filter to CertmanagerCertificateExpired alert expression

The `CertmanagerCertificateExpired` alert produces false positives caused by upmeter's `control-plane/cert-manager` probe.

A value of 0 is a sentinel meaning "not yet issued" and can never represent a real certificate expiration (Unix timestamp 0 = January 1, 1970). Real expired certificates always have a positive timestamp, so filtering `!= 0` does not suppress legitimate alerts.

no need

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cert-manager
type: fix
summary: Fix false positive CertmanagerCertificateExpired alert
impact_level: default
```

### DIFF
--- a/modules/101-cert-manager/monitoring/prometheus-rules/propagated-certificate.yaml
+++ b/modules/101-cert-manager/monitoring/prometheus-rules/propagated-certificate.yaml
@@ -22,7 +22,7 @@
 
   - alert: CertmanagerCertificateExpired
     expr: |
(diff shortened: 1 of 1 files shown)


AlwxSin (approved):

Suselz:
Fixed in #19446

PR #19241, opened 2026-04-17
[deckhouse] replace config-values.yaml with settings.yaml

Fkuloff:
## Description

Rename `openapi/config-values.yaml` to `openapi/settings.yaml` in the package format. Both loaders accept either name for backward compatibility:
- on-disk loader (`loader/values.go`) reads `settings.yaml`, falls back to `config-values.yaml`;
- registry tar parser (`application-package-version/parser.go`) accepts either entry name when extracting metadata from the image.

No impact on cluster components — the change affects only how `deckhouse-controller` resolves the user-settings OpenAPI schema for packages.

config-values.yaml is a legacy name inherited from modules. The package format already exposes user configuration as Application.spec.settings, so the schema file is renamed to match — openapi/settings.yaml. Aligns naming across the API and the package layout, and removes the “values vs config-values” confusion for package authors.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: chore
summary: Replace config-values.yaml with settings.yaml.
```


deckhouse-controller/pkg/controller/packages/application-package-version/parser.go
```diff
@@ -145,6 +152,17 @@ func (r *metadataReader) untarMetadata(rc io.Reader) error {
 				return err
 			}
 		case settingsSchemaFile:
+			// settings.yaml is canonical and overwrites any previously read legacy entry.
+			r.settingsSchemaReader.Reset()
```
ipaqsa: Its redundant, I dont think there would be such case

deckhouse-controller/pkg/controller/packages/application-package-version/parser.go
```diff
@@ -145,6 +152,17 @@ func (r *metadataReader) untarMetadata(rc io.Reader) error {
...
+			if _, err = io.Copy(r.settingsSchemaReader, io.LimitReader(tr, maxMetadataFileSize)); err != nil {
+				return err
+			}
+			r.settingsSchemaCanonical = true
+		case legacySettingsSchemaFile:
+			// Read legacy openapi/config-values.yaml only if settings.yaml has not been seen yet.
+			if r.settingsSchemaCanonical {
```
ipaqsa: thus it is too

deckhouse-controller/internal/packages/loader/values.go
```diff
@@ -73,30 +75,47 @@ func loadValues(name, path string) (addonutils.Values, []byte, []byte, error) {
 	return static, config, values, nil
 }
 
-// loadPackageSchemas reads config-values.yaml and values.yaml from the specified directory.
-// Package schemas:
+// readOptionalFile reads the file at path. If the file does not exist, it
+// returns (nil, nil) instead of an error. Any other read error is wrapped.
+func readOptionalFile(path string) ([]byte, error) {
```
ipaqsa: its too complicated, make it simple, without new functions

deckhouse-controller/pkg/controller/packages/application-package-version/parser_test.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// tarEntry describes one regular file inside a synthetic tar archive.
+type tarEntry struct {
+	name    string
+	content []byte
+}
+
+// buildTar packs entries into a tar archive in the order given.
+func buildTar(t *testing.T, entries []tarEntry) []byte {
```
ipaqsa: dont need tests in this pr. It should a small pr

z9r5 (approved):

ipaqsa (approved):

ldmonster (approved):

PR #19303, opened 2026-04-20
[deckhouse] add lastAppliedConfiguration to Application status

Fkuloff:
## Description

Adds `status.lastAppliedConfiguration` to the `Application` CR (`deckhouse.io/v1alpha1`). The field is populated from `values.Storage.GetSettings()` (user config merged with config-schema defaults) and is committed to the status alongside `status.currentVersion.version` whenever the internal `ReadyInCluster` condition is true.

No impact on critical cluster components: only the Application controller's status reporting in `deckhouse-controller` is changed; existing fields and behavior are untouched. The field is optional.

There's currently no in-cluster way to see which settings are actually applied to an Application — only spec.settings (raw user input) and the OpenAPI config schema are available, so the effective merged config has to be reconstructed by hand. This value is already computed at runtime (GetSettings() is exposed to templates as .Application.Settings) — it's just not surfaced.

status.lastAppliedConfiguration gives operators and tooling a single source of truth for "what is currently applied", kept in sync with currentVersion.version. Useful for incident triage, diffing expected vs. actual config, and any automation built on top of the Application CR.

```
 "lastAppliedConfiguration": {
    "replicas": 2
  },
```

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Added `lastAppliedConfiguration` to Application status.
impact_level: default 
```


Lada7878 (approved):

ipaqsa:
Where is an example from your cluster ?

z9r5 (approved):

deckhouse-controller/internal/packages/runtime/tasks/configure/task.go
```diff
@@ -72,40 +76,46 @@ func NewTask(pkg packageI, settings addonutils.Values, status statusService, log
...
 	t.status.SetConditionTrue(t.pkg.GetName(), status.ConditionSettingsValid)
 
 	return nil
 }
 
-// applySettings validates and applies settings to application
-func (t *task) applySettings(ctx context.Context) error {
-	ctx, span := otel.Tracer(taskTracer).Start(ctx, "ApplySettings")
+// configure validates and applies settings to the package.
+func (t *task) configure(ctx context.Context) error {
```
ipaqsa: it should remain applySettings and error newApplySettingsErr

deckhouse-controller/internal/packages/runtime/tasks/configure/task.go
```diff
@@ -72,40 +76,46 @@ func NewTask(pkg packageI, settings addonutils.Values, status statusService, log
...
+		return fmt.Errorf("configure: %w", err)
 	}
 
+	// Propagate the effective settings (user config + config-schema defaults)
+	// to the internal status service. The CR status handler will later commit
+	// them to Application.status.lastAppliedConfiguration alongside the version
+	// when ConditionReadyInCluster becomes True.
+	t.status.UpdateSettings(t.pkg.GetName(), t.pkg.GetSettings())
+
 	t.status.SetConditionTrue(t.pkg.GetName(), status.ConditionSettingsValid)
```
ipaqsa, line 96: UpdateSettings and SetConditionTrue both will notify, it means two patches. 
Two solutions:
- Remove SetConditionTrue here, make it inside UpdateSettings
- Dont notify in UpdateSettings

Fkuloff: done

ipaqsa (approved):

ldmonster (approved):

PR #19446, opened 2026-04-27
[cert-manager] Ignore zero expiration timestamps in expired alert

IlshatKhisamov:
## Description
Update `CertmanagerCertificateExpired` alert expression to ignore zero expiration timestamps.
The alert now treats certificates as expired only when:
- expiration timestamp is in the past, and
- expiration timestamp is greater than zero.
This prevents false positives for temporary/not-yet-issued certificates (notably from upmeter cert-manager probe lifecycle).
`certmanager_certificate_expiration_timestamp_seconds` may be `0` while a certificate is not yet issued.
Current expression interprets `0 - time() < 0` as expired, which produces false `CertmanagerCertificateExpired` alerts.
This is especially visible with upmeter probe certificates that are created/deleted frequently and can hit scrape/lookback windows during API/cert-manager delays.
Issue/card: #62744440
Yes, this should be backported to patch releases, because it reduces noisy false-positive alerts in production and lowers on-call load without changing behavior for genuinely expired certificates.
## Checklist
- [ ] e2e tests passed.
## Changelog entries
```changes
section: cert-manager
type: fix
summary: Prevent false positive CertmanagerCertificateExpired alerts by excluding zero expiration timestamps.
impact_level: default

### DIFF
--- a/modules/101-cert-manager/monitoring/prometheus-rules/propagated-certificate.yaml
+++ b/modules/101-cert-manager/monitoring/prometheus-rules/propagated-certificate.yaml
@@ -22,7 +22,7 @@
 
   - alert: CertmanagerCertificateExpired
     expr: |
-      max by (name, exported_namespace) (certmanager_certificate_expiration_timestamp_seconds{job="cert-manager"} - time() < 0)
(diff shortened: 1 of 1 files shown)


AlwxSin (approved):

Suselz (approved):

deckhouse-BOaTswain:
Cherry pick PR [19515](https://github.com/deckhouse/deckhouse/pull/19515) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #19520, opened 2026-04-29
feat(prometheus): add keep_firing_for support to CustomPrometheusRules

Ruthercode:
## Description
Added the `keep_firing_for` field to the `CustomPrometheusRules` CRD schema. The existing hook logic already passes the raw `groups` array to the `PrometheusRule` resource, so no Go code changes were required. Also added a dedicated test case to ensure the field is preserved.

This feature does not influence critical cluster components and does not cause unexpected restarts.

Prometheus 2.42 introduced the `keep_firing_for` field, which allows alerts to remain active for a specified time after the triggering condition has cleared. This is extremely useful for handling flapping alerts or ensuring that temporary recoveries don't prematurely resolve an incident. 

Previously, the Deckhouse CRD wrapper `customprometheusrules.deckhouse.io` did not allow this field in its OpenAPI schema validation, preventing users from utilizing this native Prometheus feature.

Not necessarily. It's a new feature, not a critical bug fix.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: prometheus
type: feature
summary: Add support for the `keep_firing_for` field in the CustomPrometheusRules CRD.
impact_level: default

### DIFF
--- a/modules/300-prometheus/crds/customprometheusrules.yaml
+++ b/modules/300-prometheus/crds/customprometheusrules.yaml
@@ -72,6 +72,8 @@ spec:
                               x-kubernetes-int-or-string: true
                             for:
                               type: string
+                            keep_firing_for:
+                              type: string
                             labels:
                               additionalProperties:
                                 type: string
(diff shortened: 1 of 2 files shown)


z9r5:
@MargaretKozlova We need to add descriptions for rules:*, and translations

z9r5 (approved):

PR #19556, opened 2026-04-30
fix(user-authn): remove expired users from group members

IlshatKhisamov:
## Description
Fix local user TTL expiration flow in `user-authn`: when an expired `User` is deleted, the hook now also removes this user from all `Group.spec.members` entries.
Previously, after TTL expiration, a local user was removed but stale `User` references remained in groups.
This produced inconsistent authn data and “dead” group members.
The fix keeps group membership state consistent by cleaning up expired users from groups during the same expiration cycle.
This is a user-visible bug fix in existing behavior.
## Checklist
- [ ] e2e tests passed.
## Changelog entries
```changes
section: user-authn
type: fix
summary: Remove expired local users from Group members during TTL cleanup.
impact: Group membership no longer keeps stale references to deleted local users after TTL expiration.
impact_level: default

### DIFF
--- a/modules/150-user-authn/hooks/expire_dex_user_crds.go
+++ b/modules/150-user-authn/hooks/expire_dex_user_crds.go
@@ -76,6 +76,14 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 			ExecuteHookOnSynchronization: ptr.To(false),
 			FilterFunc:                   applyDexUserExpireFilter,
 		},
+		{
+			Name:                         "groups",
+			ApiVersion:                   "deckhouse.io/v1alpha1",
+			Kind:                         "Group",
+			ExecuteHookOnEvents:          ptr.To(false),
+			ExecuteHookOnSynchronization: ptr.To(false),
+			FilterFunc:                   applyDexGroupFilter,
+		},
 	},
 }, expireDexUsers)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/150-user-authn/hooks/expire_dex_user_crds.go [hunk 1]
```diff
@@ -89,7 +97,52 @@ func expireDexUsers(_ context.Context, input *go_hook.HookInput) error {
 
 		if dexUserExpire.CheckExpire && dexUserExpire.ExpireAt.Before(now) {
 			input.PatchCollector.Delete("deckhouse.io/v1", "User", "", dexUserExpire.Name)
+
+			for group, err := range sdkobjectpatch.SnapshotIter[DexGroup](input.Snapshots.Get("groups")) {
```
AlwxSin: There is a possible race condition here.
I.e. there are two users (Alice and Bob) which will expire at the same time.
Then, there are two threads which both will execute `user delete` operation, but there will be 
```yaml
members: [Alice Bob Charlie]
```
for both threads.
Then, thread `#1` will patch with
```yaml
members: [Bob Charlie]
```
and thread `#2` will patch with
```yaml
members: [Alice Charlie]
```
causing Alice still present in members.

Usually, hooks produces desired full state.
For k8s it's common to rely on `resourceVersion` with `retryOnConflict`, but you need a k8s client
```go
err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
	group := &v1alpha1.Group{}
	err := cl.Get(ctx, client.ObjectKey{Name: groupName}, group)
	if err != nil {
		return err
	}

	group.Spec.Members = removeMember(group.Spec.Members, "Alice")

	return cl.Update(ctx, group)
})
```

modules/150-user-authn/hooks/expire_dex_user_crds.go
```diff
@@ -89,7 +97,52 @@ func expireDexUsers(_ context.Context, input *go_hook.HookInput) error {
...
+						"members": members,
+					},
+				}, "deckhouse.io/v1alpha1", "Group", "", group.Name)
+			}
 		}
 	}
 	return nil
 }
+
+func groupContainsUser(group DexGroup, username string) bool {
```
AlwxSin: This function is redundant. `removeUserFromGroupMembers` do the same.

modules/150-user-authn/hooks/expire_dex_user_crds.go
AlwxSin: Consider adding log entry with `username + groupName` data

modules/150-user-authn/hooks/expire_dex_user_crds_test.go
```diff
@@ -68,15 +69,47 @@ spec:
...
+  - kind: Group
+    name: admins
 `)
 			f.BindingContexts.Set(f.GenerateScheduleContext("*/5 * * * *"))
 			f.RunHook()
 		})
 
 		When("User expired (.status.expireAt < time.Now())", func() {
-			It("Should delete user CR", func() {
+			It("Should delete user CR and remove user from groups", func() {
```
AlwxSin, line 120: Consider adding a case with two users in single group to test race condition

modules/150-user-authn/hooks/expire_dex_user_crds.go
IlshatKhisamov: I changed the hook to compute desired group members in one pass per group using the full set of expired users, so we no longer patch the same group per-user from stale snapshot data.
This removes the lost-update scenario you described.

modules/150-user-authn/hooks/expire_dex_user_crds.go
IlshatKhisamov: Agreed, removed.
Membership check is now part of removal function result (removedUsers), and we patch only when something was actually removed.

modules/150-user-authn/hooks/expire_dex_user_crds_test.go
IlshatKhisamov: Done.
Added test scenario with two expired users in one group and verified both users are removed and group.spec.members becomes empty.

modules/150-user-authn/hooks/expire_dex_user_crds.go
IlshatKhisamov: Done.
Added info log with group name and removed users list for each patched group.

AlwxSin (approved):

PR #19596, opened 2026-05-05
[istio] Add ambient waypoint support

yshch:
## Description

Add waypoint proxy support for Istio ambient mode. Introduce a full set of Helm templates (Deployment, Service, ServiceAccount, Gateway, VPA, HPA, PodMonitor) for the waypoint proxy component, along with OpenAPI configuration schema for all new parameters. This enables users to deploy a shared waypoint proxy in the `d8-istio` namespace with full resource management (VPA/Static), replicas management (Standard/Static/HPA), and monitoring out of the box.

Additionally, refactors ztunnel templates to use shared helper functions (`istioSupportsAmbient`, `istioGlobalRevision`, `istioImageSuffix`, `istioCloudPlatform`) defined in `_helpers.tpl`, reducing code duplication. Updates istiod RBAC rules for Gateway API v1.25 compatibility (broader resource access for `gateway.networking.k8s.io`).

Waypoint proxies are a core component of Istio's ambient mesh architecture. They handle L7 traffic processing (authorization policies, traffic management, observability) for workloads enrolled in ambient mode. Without a waypoint proxy, ambient mesh can only provide L4 mTLS via ztunnel — no L7 policies can be enforced.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: feature
summary: Add waypoint proxy support for ambient mesh mode with configurable resource and replica management.
impact_level: default
```

### DIFF
--- a/modules/110-istio/openapi/config-values.yaml
+++ b/modules/110-istio/openapi/config-values.yaml
@@ -765,6 +765,288 @@ properties:
                 ```
               > **Warning!** Disabling this timeout (setting the value to `0s`) is very likely to result in leaky connections due to TCP FIN packet loss, etc.
               > **Warning!** After changing this setting, a restart of the client pods is required.
+      waypoint:
+        type: object
+        description: |
+          Waypoint proxy specific settings.
+        default: {}
+        x-doc-skip: true
+        x-experimental: true
+        oneOf:
+        - properties:
+            replicasManagement:
+              properties:
+                mode:
+                  enum: ['Standard', 'Static']
+            resourcesManagement:
+              properties:
+                mode:
+                  enum: ['Static', 'VPA']
+        - properties:
+            replicasManagement:
+              properties:
+                mode:
+                  enum: ['HPA']
+            resourcesManagement:
+              properties:
+                mode:
+                  enum: ['Static']
+        properties:
+          enabled:
+            type: boolean
+            default: false
+            description: |
+              Enable shared waypoint proxy deployment to `d8-istio` namespace.
+            x-experimental: true
+            x-examples: [true]
+          replicasManagement:
+            description: |
+              Replication management settings and scaling of waypoint proxy.
+            type: object
+            default: {}
+            x-examples:
+            - mode: Standard
+            - mode: Static
+              static:
+                replicas: 3
+            x-doc-examples:
+            - mode: Standard
+            - mode: Static
+              static:
+                replicas: 3
+            - mode: HPA
+              hpa:
+                minReplicas: 2
+                maxReplicas: 5
+                metrics:
+                - type: CPU
+                  targetAverageUtilization: 80
+            properties:
+              mode:
+                type: string
+                description: |
+                  Replicas management mode:
+                  - `Standard` — replicas management and scaling mode according to the global fault tolerance mode (the [highAvailability](/products/kubernetes-platform/documentation/v1/reference/api/global.html#parameters-highavailability) parameter);
+                  - `Static` — the mode, where the number of replicas is specified explicitly (the [static.replicas](#parameters-dataplane-waypoint-replicasmanagement-static-replicas) parameter);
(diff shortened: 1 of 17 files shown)


modules/110-istio/templates/control-plane/iop/istiod/_rules_v-1-25.tpl
```diff
@@ -239,12 +231,7 @@
   - list
 - apiGroups:
   - gateway.networking.k8s.io
-  resources:
-  - gateways
-  - httproutes
-  - grpcroutes
-  - gatewayclasses
-  - referencegrants
+  resources: ["*"]
```
yshch: This is pretty loose, but matches upstream ClusterRole: https://github.com/istio/istio/blob/1.25.2/manifests/charts/istio-control/istio-discovery/templates/clusterrole.yaml#L131-L134. We can tighten it up, but i think it's better to follow upstream as close as possible.

modules/110-istio/templates/control-plane/iop/istiod/_rules_v-1-25.tpl
```diff
@@ -5,10 +5,11 @@
   - deployments
   verbs:
   - get
-  - list
   - watch
+  - list
   - update
   - patch
+  - create
```
yshch, line 12: Needed for istiod deployment controller to provision Deployments for Gateway API Gateway objects.

modules/110-istio/templates/control-plane/iop/istiod/_rules_v-1-25.tpl
```diff
@@ -253,6 +240,23 @@
...
+  - udproutes/status
+  verbs:
+  - update
+  - patch
+- apiGroups:
+  - gateway.networking.k8s.io
+  resources:
+  - gatewayclasses
   verbs:
+  - create
```
yshch, line 266: Needed for istiod gatewayclass controller to provision builtin GatewayClasses, including `istio-waypoint`.

modules/110-istio/templates/waypoint/gateway.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: waypoint
+  namespace: d8-{{ .Chart.Name }}
+  {{- include "helm_lib_module_labels" (list . (dict
+    "app" "waypoint"
+    "istio.io/rev" (include "istioGlobalRevision" .)
+    "istio.io/waypoint-for" "all"
+    "gateway.networking.k8s.io/gateway-name" "waypoint"
+  )) | nindent 2 }}
+spec:
+  addresses:
```
yshch, line 19: Setting `type: Hostname` address is a documented way to make istiod deployment controller consider this Gateway "unmanaged". So it doesn't conflict with our own Deployment/Service/ServiceAccount manifests for this Gateway.

modules/110-istio/openapi/config-values.yaml
```diff
@@ -765,6 +765,288 @@ properties:
                 ```
               > **Warning!** Disabling this timeout (setting the value to `0s`) is very likely to result in leaky connections due to TCP FIN packet loss, etc.
               > **Warning!** After changing this setting, a restart of the client pods is required.
+      waypoint:
+        type: object
+        description: |
+          Waypoint proxy specific settings.
+        default: {}
+        x-doc-skip: true
```
yshch, line 773: Docs are hidden from UI for now, since this is still an experimental feature.

modules/110-istio/templates/control-plane/iop/istiod/_rules_v-1-25.tpl
skurbatov: We need to specify the RBAC more clearly, we have a way that we cannot use the asterisk in the RBAC and must strive for this.

yshch: Fixed in https://github.com/deckhouse/deckhouse/pull/19596/changes/cf5051f12954432d45966497286ff4398da00157

skurbatov (approved):

yshch:
Decided to remove common/shared waypoint deployment in d8-system after discussion. Closing this PR in favor of much lighter version in https://github.com/deckhouse/deckhouse/pull/19841

PR #19667, opened 2026-05-07
[control-plane-manager] fix don't rebind kubeadm:cluster-admins until cluster is bootstrapped

Jabejixo:
## Description

Fixes a regression that breaks fresh cluster bootstrap when `control-plane-manager` rebinds `ClusterRoleBinding/kubeadm:cluster-admins` from `cluster-admin` to `user-authz:cluster-admin`.

The chart now does the rebind **only after** `global.clusterIsBootstrapped` becomes `true`. The reconciliation hook gets the same gate plus a target-role pre-check, best-effort rollback on failed Create, and `recover()` guard. `nodes/proxy: [get, create]` moved from the supplement role into `user-authz:cluster-admin` to make the rebind atomic.

No control-plane restarts.

On a fresh cluster kubeadm pre-creates `kubeadm:cluster-admins → cluster-admin`. Helm tried to apply the same name with a different `roleRef`, which is immutable in K8s:

```
ClusterRoleBinding "kubeadm:cluster-admins" is invalid:
roleRef: Invalid value: {"Name":"cluster-admin"}: cannot change roleRef
```

Once the hook eventually rebinds, `kube-apiserver-kubelet-client` (group `kubeadm:cluster-admins` since kubeadm 1.29+) loses `nodes/proxy create`, and `d8 s queue list` / `kubectl exec`/`logs`/`port-forward` start returning 403.

This PR removes both failures.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Skip rebind of ClusterRoleBinding/kubeadm:cluster-admins until the cluster is fully bootstrapped; harden the reconciliation hook. Fixes "cannot change roleRef" on fresh clusters.
impact_level: default
```


z9r5 (approved):

modules/040-control-plane-manager/templates/rbac-for-us.yaml
```diff
@@ -72,7 +72,25 @@ roleRef:
   kind: ClusterRole
   name: d8:control-plane-manager
   apiGroup: rbac.authorization.k8s.io
-{{- if ($.Values.global.enabledModules | default list | has "user-authz") }}
+{{- $userAuthzEnabled := has "user-authz" ($.Values.global.enabledModules | default list) }}
+{{- /*
+  Three gates avoid the immutable-roleRef trap on a fresh cluster (kubeadm pre-creates the binding
+  pointing at cluster-admin and Helm SSA cannot rewrite roleRef):
+    1. $userAuthzEnabled       — user-authz module is on.
```
AlwxSin: I'm worried that all three gates used by hook and template independently. If somebody changes in one place and forget to sync in other place...
Is there an option to keep gates in one place which forms single unified gate and use that gate in another place?

Jabejixo: done

AlwxSin (approved):

sprait (approved):

z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [19744](https://github.com/deckhouse/deckhouse/pull/19744) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #19686, opened 2026-05-08
[cni-cilium] Implement optional enable-host-firewall disabling in Cilium

D13410N3:
## Description
This feature allow cilium setting `enable-host-firewall` that is currently hardcoded to `true` be overriden with `spec.settings.hostFirewallEnabled` in ModuleConfig `cni-cilium`. Default value is not changed and set to `true`, so the only affect is cilium-agent restart on this value changing by cluster operator.

Sometimes you may want to set up additional cni in your cluster. Deckhouse supports it - otherwise, why do `exclusiveCNIPlugin` parameter exists? But additional CNI might not work, in my case it was multus + whereabouts which were implementing optional additional interface in Pod. Everything was broken once I tried to enable IPv6 in multus IPAM - and there is the place where Cilium's host-firewall blocked all requests, as it was "unknown IP family" for it.

The only way I can use it now is this flow:
* set `spec.maintenance` to `NoResourceReconciliation` in cilium ModuleConfig
* edit configmap `d8-cni-cilium` / `cilium-config` - set `enable-host-firewall` to `false`
* live without cilium updates on cluster updates at all

Not necessarily

## Checklist
- [ ] e2e tests passed.
Template tests were added, but could not be run locally because the test suite expects `/deckhouse/ee/se-plus/modules/021-cni-cilium/templates/` to exist.

## Changelog entries

```changes
section: cni-cilium
type: feature
summary: Added the ability to configure Cilium Host Firewall via the cni-cilium ModuleConfig.
impact_level: low
```

### DIFF
--- a/tools/build.go
+++ b/tools/build.go
@@ -651,6 +651,7 @@ func (e *executor) executeEdition(editionName string) {
 		writeSettingsExcludeFileName.Prefix = prefix
 		writeSettingsExcludeFileName.Dir = "modules"
 		writeSettingsExcludeFileName.ExcludePaths = defaultModulesExcludes
+		writeSettingsExcludeFileName.StageDependencies = stageDependenciesSetup
 
 		writeSettingStageDeps.Prefix = prefix
 		writeSettingStageDeps.Dir = "modules"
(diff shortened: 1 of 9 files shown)


z9r5 (approved):

D13410N3:
Closed in favor to https://github.com/deckhouse/deckhouse/pull/20852

PR #19734, opened 2026-05-12
[user-authz] Add permission-browser availability alert

IlshatKhisamov:
## Description
Adds monitoring for Permission Browser availability in `user-authz`:
- New Prometheus alert rule for `permission-browser-apiserver` deployment unavailability.
- Rule is rendered only when `userAuthz.enableMultiTenancy=true` (to avoid false alerts when component is disabled).
- Added template tests to verify:
  - alert is rendered when multi-tenancy is enabled;
  - alert is not rendered when multi-tenancy is disabled.
This change does **not** restart or reconfigure critical cluster components directly (control-plane, ingress, Prometheus runtime config beyond normal rule reload).
Permission Browser is used in UI authorization flows; when `permission-browser-apiserver` is unavailable, users can face permission-check failures in the interface.
Before this change, there was no dedicated alert to proactively signal this condition.  
Now operators get a clear alert and can react faster, reducing MTTR and improving visibility for on-call teams.
Not necessarily.
This is a low-risk monitoring improvement (new alert + tests) and not a runtime bugfix.
## Checklist
- [ ] e2e tests passed.
## Changelog entries
```changes
section: user-authz
type: feature
summary: Added alert for `permission-browser-apiserver` unavailability in `user-authz`.
impact_level: default

### DIFF
--- a/modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
+++ b/modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
@@ -1,0 +1,34 @@
+- name: d8.user-authz.permission-browser-apiserver.availability
+  rules:
+  - alert: D8UserAuthzPermissionBrowserUnavailable
+    expr: |
+      (
+        kube_deployment_spec_replicas{
+          namespace="d8-user-authz",
+          deployment="permission-browser-apiserver"
+        }
+        -
+        kube_deployment_status_replicas_available{
+          namespace="d8-user-authz",
+          deployment="permission-browser-apiserver"
+        }
+      ) > 0
+    for: 5m
+    labels:
+      severity_level: "5"
+      tier: cluster
+      d8_module: user-authz
+      d8_component: permission-browser-apiserver
+    annotations:
+      plk_protocol_version: "1"
+      plk_markup_format: "markdown"
+      summary: Permission Browser API server has unavailable replicas.
+      description: |-
+        Deckhouse has detected unavailable replicas in the `permission-browser-apiserver` Deployment in the `d8-user-authz` namespace.
+        Permission Browser requests can fail while this alert is active.
+
+        Check Deployment and Pod states:
+        ```bash
+        d8 k -n d8-user-authz get deploy permission-browser-apiserver
+        d8 k -n d8-user-authz get pods -l app=permission-browser-apiserver
+        ```
--- a/modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
+++ b/modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
@@ -15,19 +15,21 @@
       ) > 0
     for: 5m
     labels:
-      severity_level: "5"
+      severity_level: "6"
       tier: cluster
       d8_module: user-authz
       d8_component: permission-browser-apiserver
     annotations:
       plk_protocol_version: "1"
       plk_markup_format: "markdown"
+      plk_create_group_if_not_exists__d8_user_authz_permission_browser_unavailable: "D8UserAuthzPermissionBrowserUnavailable,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: modules/140-user-authz/templates/monitoring.yaml [hunk 1]
```diff
@@ -1 +1,3 @@
+{{- if .Values.userAuthz.enableMultiTenancy }}
```
AlwxSin: I'm worry, that you gated rules completely.
Currently there's only one file, so it works — but if someone later adds a rule that should fire regardless of multiTenancy, it will be silently suppressed. Consider moving the condition inside the rule YAML (e.g., wrapping only the `permission-browser-apiserver.yaml` content with a Helm  `if`)

modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
```diff
@@ -0,0 +1,34 @@
...
+        }
+      ) > 0
+    for: 5m
+    labels:
+      severity_level: "5"
+      tier: cluster
+      d8_module: user-authz
+      d8_component: permission-browser-apiserver
+    annotations:
+      plk_protocol_version: "1"
```
AlwxSin, line 23: There should be `plk_group`  fields.
@sudoroot-null correct me if I'm wrong.

modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
```diff
@@ -0,0 +1,34 @@
...
+        }
+        -
+        kube_deployment_status_replicas_available{
+          namespace="d8-user-authz",
+          deployment="permission-browser-apiserver"
+        }
+      ) > 0
+    for: 5m
+    labels:
+      severity_level: "5"
```
AlwxSin: @IlshatKhisamov why severity level is `5`? Shouldn't it be the `6`?

AlwxSin (changes requested):

modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
```diff
@@ -0,0 +1,34 @@
...
+      d8_component: permission-browser-apiserver
+    annotations:
+      plk_protocol_version: "1"
+      plk_markup_format: "markdown"
+      summary: Permission Browser API server has unavailable replicas.
+      description: |-
+        Deckhouse has detected unavailable replicas in the `permission-browser-apiserver` Deployment in the `d8-user-authz` namespace.
+        Permission Browser requests can fail while this alert is active.
+
+        Check Deployment and Pod states:
```
denmaloyreb:
```suggestion
        Deckhouse has detected unavailable replicas in the `permission-browser-apiserver` deployment in the `d8-user-authz` namespace.
        Permission Browser requests can fail while this alert is active.

        Check deployment and pod states:
```

denmaloyreb: Please don't forget to regenerate deckhouse-alerts.yml

denmaloyreb (approved):

AlwxSin (approved):

modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
IlshatKhisamov: Updated severity to 6 in both severity_level (rule labels) and generated alerts metadata.

modules/140-user-authz/monitoring/prometheus-rules/permission-browser-apiserver.yaml
IlshatKhisamov: Regenerated/updated docs/documentation/_data/deckhouse-alerts.yml and included it in the commit.

modules/140-user-authz/templates/monitoring.yaml
IlshatKhisamov: Removed the global enableMultiTenancy gate from templates/monitoring.yaml; now rules are always included, and this alert remains harmless when component metrics are absent.

sudoroot-null (approved):

z9r5 (approved):

Jabejixo (approved):

PR #19844, opened 2026-05-14
[user-authn] Validate and alert on conflicting LDAP TLS settings in DexProvider

AlwxSin:
## Description

`DexProvider`'s `spec.ldap` exposed three independent TLS boolean flags  (`insecureNoSSL`, `startTLS`, `insecureSkipVerify`) plus `rootCAData`,  without any cross-validation. The combinations below are silently  ignored by Dex (see `connector/ldap/dap.go` switch on `InsecureNoSSL`)  and therefore mask misconfiguration:

  * `insecureNoSSL` + `startTLS` — TLS is disabled (ldap://, port 389),
    `startTLS` is ignored.
  * `insecureNoSSL` + `insecureSkipVerify` — TLS is disabled, no
    certificate to verify.
  * `insecureNoSSL` + `rootCAData` — TLS is disabled, CA is ignored.

  This PR:

  1. Adds three CEL rules (`x-kubernetes-validations`) on `spec.ldap` rejecting the combinations above. The YAML anchor `&ldap` ensures both `v1alpha1` and `v1` versions get the same rules.
  2. Updates English and Russian descriptions of `insecureNoSSL`, `startTLS`, `insecureSkipVerify`, and `rootCAData` to call out the constraint.
  3. Adds a hook `alert_on_deprecated_dex_provider_tls.go` that scans existing `DexProvider` objects and emits `d8_dex_provider_ldap_tls_conflict{name,conflict}` for each conflicting LDAP combination. CRD validation only runs on create/update and CRD validation ratcheting lets legacy invalid objects survive partial updates — without the alert they would stay silent.
  4. Adds a `PrometheusRule` `D8DexProviderLDAPTLSConflict` (`severity_level: "8"`, grouped under `UserAuthnDeprecations` PLK group) with a migration runbook in the description.
  5. Removes `insecureSkipVerify: true` from the internal demo (`docs/internal/demo/dex-provider.yaml`) and from the hook test fixture (`get_dex_providers_crds_test.go`): both combined it with `insecureNoSSL: true` and would now fail CRD validation.

  No restarts required. Existing valid `DexProvider` objects are unaffected.

Today an operator can ship a `DexProvider` like `{insecureNoSSL: true, startTLS: true, rootCAData: ...}` and it will be accepted by the API server. Dex will silently fall back to plain `ldap://` and ignore the TLS-related fields, which looks like working TLS but isn't. The CRD now refuses such combinations up front, and the new alert lets cluster operators discover any pre-existing objects that were created before the validation landed.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Reject mutually exclusive LDAP TLS settings in DexProvider; alert on legacy objects.
impact: |
  spec.ldap.insecureNoSSL combined with startTLS, insecureSkipVerify, or a non-empty rootCAData is now rejected by the CRD. Pre-existing DexProvider objects with such combinations keep working but trigger
  D8DexProviderLDAPTLSConflict until updated. Audit with: kubectl get dexproviders.deckhouse.io -o yaml | yq '.items[] | select(.spec.type=="LDAP") | {name: .metadata.name, ldap: .spec.ldap | pick(["insecureNoSSL","startTLS","insecureSkipVerify","rootCAData"])}'
impact_level: default
```

### DIFF
--- a/modules/150-user-authn/monitoring/prometheus-rules/dex-deprecation.yaml
+++ b/modules/150-user-authn/monitoring/prometheus-rules/dex-deprecation.yaml
@@ -1,0 +1,36 @@
+- name: d8.user-authn.deprecation
+  rules:
+  - alert: D8DexProviderLDAPTLSConflict
+    expr: d8_dex_provider_ldap_tls_conflict == 1
+    labels:
+      severity_level: "8"
+      tier: cluster
+      d8_module: user-authn
+      d8_component: dex
+    annotations:
+      plk_markup_format: markdown
+      plk_protocol_version: "1"
+      plk_create_group_if_not_exists__d8_user_authn_deprecations: "UserAuthnDeprecations,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
+      plk_grouped_by__d8_user_authn_deprecations: "UserAuthnDeprecations,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
+      summary: DexProvider `{{ $labels.name }}` has a conflicting LDAP TLS configuration.
+      description: |
+        DexProvider `{{ $labels.name }}` combines mutually exclusive LDAP TLS settings: `{{ $labels.conflict }}`. Dex silently ignores the redundant flag, which masks misconfiguration:
+
+        * `insecureNoSSL + startTLS`: TLS is disabled (ldap://, port 389), so `startTLS` is ignored.
+        * `insecureNoSSL + insecureSkipVerify`: TLS is disabled, there is no certificate to verify.
+        * `insecureNoSSL + rootCAData`: TLS is disabled, the CA chain is ignored.
+
+        The CRD schema rejects this combination on create/update; this DexProvider was created before the validation was added and remains active.
+
+        Update the DexProvider to remove the conflict:
+
+        * Plaintext LDAP (port 389): keep `spec.ldap.insecureNoSSL: true`, remove `startTLS`, `insecureSkipVerify`, and `rootCAData`.
+        * STARTTLS on port 389: set `spec.ldap.insecureNoSSL: false`, `spec.ldap.startTLS: true`. Use `insecureSkipVerify` or `rootCAData` as needed.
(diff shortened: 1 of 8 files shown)


Suselz (approved):

modules/150-user-authn/monitoring/prometheus-rules/dex-deprecation.yaml
```diff
@@ -0,0 +1,36 @@
...
+      plk_create_group_if_not_exists__d8_user_authn_deprecations: "UserAuthnDeprecations,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
+      plk_grouped_by__d8_user_authn_deprecations: "UserAuthnDeprecations,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
+      summary: DexProvider `{{ $labels.name }}` has a conflicting LDAP TLS configuration.
+      description: |
+        DexProvider `{{ $labels.name }}` combines mutually exclusive LDAP TLS settings: `{{ $labels.conflict }}`. Dex silently ignores the redundant flag, which masks misconfiguration:
+
+        * `insecureNoSSL + startTLS` — TLS is disabled (ldap://, port 389), so `startTLS` is ignored.
+        * `insecureNoSSL + insecureSkipVerify` — TLS is disabled, there is no certificate to verify.
+        * `insecureNoSSL + rootCAData` — TLS is disabled, the CA chain is ignored.
+
```
denmaloyreb:
```suggestion
        * `insecureNoSSL + startTLS`: TLS is disabled (ldap://, port 389), so `startTLS` is ignored.
        * `insecureNoSSL + insecureSkipVerify`: TLS is disabled, there is no certificate to verify.
        * `insecureNoSSL + rootCAData`: TLS is disabled, the CA chain is ignored.

```

Please don't forget to regenerate deckhouse-alerts.yml

sudoroot-null (approved):

denmaloyreb (approved):

z9r5 (approved):

nabokihms (approved):

PR #19850, opened 2026-05-14
[cert-manager] Auto-enable cainjector when the cluster has resources annotated for CA injection

IlshatKhisamov:
## Description

Added a `cert-manager` hook that auto-enables `cainjector` when the cluster has resources annotated for CA injection:
- `cert-manager.io/inject-ca-from`
- `cert-manager.io/inject-ca-from-secret`
- `cert-manager.io/inject-apiserver-ca`

`cainjector` manifests are now rendered based on `certManager.internal.enableCAInjector`, resolved as:
- manual setting: `certManager.enableCAInjector`
- OR automatic detection from cluster resources

To reduce hook overhead, Kubernetes bindings for this hook are configured with:
- `ExecuteHookOnEvents: false`
- `ExecuteHookOnSynchronization: false`

The hook still runs on `OnBeforeHelm`, which is sufficient for this use case.

No critical control-plane components are affected; the change only adjusts `cainjector` enablement behavior inside the `cert-manager` module.

Previously, `cainjector` was enabled only via manual config (`certManager.enableCAInjector`).
This created a non-obvious failure mode: users could create resources with CA injection annotations but forget to enable `cainjector`, resulting in CA injection not being performed.

Now the module enables `cainjector` automatically when such resources are detected, while preserving full backward compatibility with manual enablement.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cert-manager
type: feature
summary: Auto-enable cainjector when cluster resources require cert-manager CA injection annotations.
impact_level: default
```


z9r5 (approved):

modules/101-cert-manager/hooks/discover_cainjector_usage.go
```diff
@@ -0,0 +1,93 @@
...
+
+const cainjectorEnabledValuesPath = "certManager.internal.enableCAInjector"
+
+var caInjectorAnnotations = map[string]struct{}{
+	"cert-manager.io/inject-ca-from":        {},
+	"cert-manager.io/inject-ca-from-secret": {},
+	"cert-manager.io/inject-apiserver-ca":   {},
+}
+
+var _ = sdk.RegisterFunc(&go_hook.HookConfig{
```
AlwxSin, line 38: Too broad filter func, there could be thousands of CRDs and APIServices.
Consider adding `ExecuteHookOnEvents: false` and `ExecuteHookOnSync: false` as the hook only needs on `BeforeHelm`

modules/101-cert-manager/hooks/discover_cainjector_usage_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	. "github.com/deckhouse/deckhouse/testing/hooks"
+)
+
+const (
+	validatingWebhookWithInjectAnnotation = `
```
AlwxSin, line 27: No tests for MutatingWebhookConfiguration and CustomResourceDefinition

modules/101-cert-manager/openapi/values.yaml
```diff
@@ -43,3 +43,7 @@ properties:
         format: email
         x-examples: ["example@example.com"]
 
+      enableCAInjector:
```
AlwxSin, line 46: Shouldn't we either delete corresponding option in `config-values.yaml` or add a description says something like "Manually enable CAInjector. When not set, it is automatically enabled if cluster resources with CA injection annotations are detected."

modules/101-cert-manager/template_tests/module_test.go
```diff
@@ -93,6 +93,7 @@ discovery:
 const certManager = `
 enableCAInjector: true
 internal:
+  enableCAInjector: true
```
AlwxSin, line 96: Consider adding test for opposite way, when `internal.enableCAInjector: false` resources cainjector doesn't render.

AlwxSin (approved):

modules/101-cert-manager/hooks/discover_cainjector_usage.go
IlshatKhisamov: Addressed. I added ExecuteHookOnEvents: false and ExecuteHookOnSynchronization: false for all Kubernetes bindings in this hook. The hook logic is evaluated on OnBeforeHelm, which is sufficient for this case and avoids unnecessary event-driven processing for large CRD/APIService sets.

modules/101-cert-manager/hooks/discover_cainjector_usage_test.go
IlshatKhisamov: Added. I included dedicated test cases for MutatingWebhookConfiguration and CustomResourceDefinition with CA injection annotations, and verified that both paths enable cainjector as expected.

modules/101-cert-manager/openapi/values.yaml
IlshatKhisamov: Updated. I kept the option for backward compatibility and clarified the description in config-values.yaml (and RU docs): enableCAInjector is a manual override, while automatic enablement is performed when resources with CA injection annotations are detected.

modules/101-cert-manager/template_tests/module_test.go
IlshatKhisamov: Added. I introduced a template test for the opposite case and verified that when internal.enableCAInjector: false, cainjector resources are not rendered.

Suselz (approved):

Zhbert (approved):

PR #19915, opened 2026-05-18
[user-authn] add kubernetes-dex-client-configuration secret for console consumption

AlwxSin:
## Description

  Phase 1 of the kubeconfig-generator deprecation plan (see `agentic_docs/plans/kubeconfig_generator_deprecate.md`).

  Renders a new internal `Secret d8-user-authn/kubernetes-dex-client-configuration` alongside the existing legacy `Secret kubeconfig-generator`. The new secret carries the same data in a cleaner shape (camelCase, dead `dex-k8s-authenticator`-specific fields dropped) and is intended for the console module — it will switch to it in a later release.

  Each entry in `userAuthn.kubeconfigGenerator[]` now also gets its own dedicated `OAuth2Client` named `kubeconfig-{slug(id)}` with label `app: kubernetes-dex-clients`, in addition to the existing legacy `kubeconfig-generator-N` client. This means after rollout `d8 k -n d8-user-authn get oauth2clients` will show roughly **twice** as many OAuth2Clients per `kubeconfigGenerator[]` entry — that's expected: the legacy `kubeconfig-generator-N` clients stay so existing kubeconfigs keep working; they will be removed in a later phase together with the UI generator itself (deadline 2026-09-01).

  `trustedPeers` in the OAuth2Client `kubernetes` are extended with the new `kubeconfig-{slug}` client_ids; legacy peers remain untouched.

  Slug transform: `@` → `-at-`, `:` → `-`, anything else from the allowed `id` pattern (`[@.:0-9a-z._-]+`) passes through. Identifiable client_id values in Dex audit logs are a side benefit.

  **No user-visible change in this phase.** The legacy generator UI continues to work; the console continues to read the legacy secret until the phase 2 PR lands in the console repo.

The deprecated `kubeconfig-generator` UI is scheduled for removal on 2026-09-01. The external `console` module was supposed to replace it, but currently piggybacks on the legacy `Secret kubeconfig-generator` to generate per-user kubeconfigs. This PR introduces a clean, dedicated secret that the console can switch to, decoupling its lifecycle from the legacy UI stack.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: common
type: chore
summary: Added rendering of a new internal `kubernetes-dex-client-configuration` Secret alongside the legacy `kubeconfig-generator` Secret in `user-authn`.
impact_level: low
```

### DIFF
--- a/modules/150-user-authn/hooks/generate_kubeconfig_encoded_names.go
+++ b/modules/150-user-authn/hooks/generate_kubeconfig_encoded_names.go
@@ -19,9 +19,11 @@ package hooks
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
 	"github.com/flant/addon-operator/sdk"
+	"github.com/tidwall/gjson"
 
 	"github.com/deckhouse/deckhouse/go_lib/encoding"
 )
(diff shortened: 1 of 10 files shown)


modules/150-user-authn/hooks/generate_kubeconfig_encoded_names.go
```diff
@@ -30,29 +31,43 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
...
+		legacy := encoding.ToFnvLikeDex(fmt.Sprintf("kubeconfig-generator-%d", i))
+		legacyEncodedNames = append(legacyEncodedNames, legacy)
 
-	for i := 0; i < kubeconfigsLength; i++ {
-		name := encoding.ToFnvLikeDex(fmt.Sprintf("kubeconfig-generator-%d", i))
-		encodedNames = append(encodedNames, name)
+		id := entry.Get("id").String()
+		clientID := fmt.Sprintf("kubeconfig-%s", slugifyKubeconfigID(id))
+		clientEncodedNames = append(clientEncodedNames, encoding.ToFnvLikeDex(clientID))
 	}
```
Jabejixo, line 76: the hook now also produces `userAuthn.internal.kubeconfigClientEncodedNames`, but the hook test still only checks the legacy `kubeconfigEncodedNames`

modules/150-user-authn/template_tests/kubernetes_oauth2client_test.go
```diff
@@ -115,4 +116,126 @@ var _ = Describe("Module :: user-authn :: helm template :: kubernetes oauth2clie
...
+  name: "no-access-dex-authenticator"
+  truncated: false
+  hash: ""
+  secretName: "dex-authenticator-no-access"
+  secretTruncated: false
+  secretHash: ""
+  ingressNames: {}
+`)
+			hec.ValuesSetFromYaml("userAuthn.internal.dexClientCRDs", `
+- id: my-app
```
Jabejixo: this writes directly to `userAuthn.internal.dexClientCRDs`, but the fixture does not match the internal shape produced by the DexClient hook. Other templates render `dexClientCRDs` too and expect fields like `name`, `namespace`, `clientSecret`, `encodedID`, etc.

z9r5 (approved):

nabokihms (approved):

Zhbert (approved):

PR #19992, opened 2026-05-20
[user-authz] Fix get ans for users with authorizationRules

Jabejixo:
## Description

After multitenancy was switched to deny-by-default, users granted access only via namespaced `AuthorizationRule` (AR) — without any `ClusterAuthorizationRule` (CAR) — got an empty `accessiblenamespaces` response, even though their ARs had already produced valid RoleBindings.

`webhook/configmap.yaml` now also serializes `authRuleCrds` into `config.json`, and the `permission-browser-apiserver` multitenancy engine consumes them: every AR adds its own namespace to the subject's `LimitNamespaces`. The user-authz-webhook still reads only `crds` and ignores unknown fields, so its behavior is unchanged.

Restores `accessiblenamespaces` visibility for users whose access is configured only through `AuthorizationRule`s. Without this fix, such users see an empty namespace list in the UI/CLI despite RBAC allowing them to operate in the AR's namespace.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: fix
summary: AuthorizationRule grants are now reflected in the accessiblenamespaces response from permission-browser-apiserver.
impact_level: low
```

### DIFF
--- a/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/authorizer/multitenancy/engine.go
+++ b/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/authorizer/multitenancy/engine.go
@@ -149,7 +149,7 @@ func (e *Engine) authorizeNamespacedRequest(attrs authorizer.Attributes, entry *
 	}
 
 	// Check system namespaces restriction
-	if !denied && !entry.AllowAccessToSystemNamespaces && isSystemNamespace(namespace) {
+	if !denied && isSystemNamespace(namespace) && !systemNamespaceAllowed(entry, namespace) {
 		denied = true
 		reason = noNamespaceAccessReason
 	}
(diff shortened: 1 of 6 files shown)


ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/authorizer/multitenancy/engine.go
```diff
@@ -424,13 +414,63 @@ func (e *Engine) renewDirectories() {
...
+			kind, name := subjectDirectoryKey(subject.Kind, subject.Name, subject.Namespace, ar.Namespace)
+
+			if _, ok := directory[kind]; !ok {
+				continue
+			}
+
+			dirEntry := directory[kind][name]
+			dirEntry.LimitNamespaces = append(dirEntry.LimitNamespaces, nsRegex)
+			if arInSystemNS {
+				dirEntry.AllowAccessToSystemNamespaces = true
```
AlwxSin: Correct me if I'm wrong. `AllowAccessToSystemNamespaces` is a global flag on the dirEntry. So, if user has CAR with `limitNamespaces: [d8-.*]` without `allowAccessToSystemNamespaces` and AR lives in `d8-monitoring` - user will have access to every `d8-*` namespace?

AlwxSin (changes requested):

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/authorizer/multitenancy/engine.go
Jabejixo: oh... fixed

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [20005](https://github.com/deckhouse/deckhouse/pull/20005) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #19592, opened 2026-05-05
[webhook-handler] Handle shell-operator exit state in main loop

riptide-01:
## Description
If the `shell-operator` stops on its own  (for example, when the cluster API is briefly unavailable) the webhook-handler operator now detects that and starts it again, instead of leaving the pod in a broken state until someone restarts it.

**Before:** We only restarted the hook runner when webhooks were added or removed. If it exited because of a startup or connectivity failure, nothing brought it back.
```
shell-operator: error: init failed: initialize ValidatingWebhookManager fail: ValidatingWebhookManager start: register validating webhook: list ValidatingWebhookConfiguration: Get "https://10.222.0.1:443/apis/admissionregistration.k8s.io/v1/validatingwebhookconfigurations?fieldSelector=metadata.name%3Dd8-deckhouse-validating-webhook-handler-hooks": dial tcp 10.222.0.1:443: connect: connection refused, try --help
{"level":"fatal","logger":"shell-operator","msg":"essemble shell operator","error":"initialize ValidatingWebhookManager fail: ValidatingWebhookManager start: register validating webhook: list ValidatingWebhookConfiguration: Get \"https://10.222.0.1:443/apis/admissionregistration.k8s.io/v1/validatingwebhookconfigurations?fieldSelector=metadata.name%3Dd8-deckhouse-validating-webhook-handler-hooks\": dial tcp 10.222.0.1:443: connect: connection refused","time":"2026-04-29T07:25:47Z"}
# then nothing happening
```

**After:** We also react when the hook runner has stopped, so it can come back without a manual pod restart.
```bash
{"level":"fatal","logger":"shell-operator","msg":"essemble shell operator","source":"src/pkg/shell-operator/bootstrap.go:101","error":"initialize ValidatingWebhookManager fail: ValidatingWebhookManager start: start webhook server: load TLS certs: open /certs/tls.crt.bak: no such file or directory","time":"2026-05-05T11:00:56Z"}
shell-operator: error: init failed: initialize ValidatingWebhookManager fail: ValidatingWebhookManager start: start webhook server: load TLS certs: open /certs/tls.crt.bak: no such file or directory, try --help
2026-05-05T11:00:59Z DEBUG msg='restarting shell-operator' source=perator/cmd/main.go:290 
2026-05-05T11:00:59Z INFO msg='killed shell-operator' source=perator/cmd/main.go:306 pid='1296' 
[...]

PR #20109, opened 2026-05-25
[webhook-handler] Fix data race and deadlock in shell-operator supervisor loop

ldmonster:
## Problem

After #19592 (`[webhook-handler] Handle shell-operator exit state in main loop`), the `webhook-handler` pod can get into a state where port `9681` (the conversion webhook endpoint) stops responding while the pod itself stays Running and Healthy. This causes kube-apiserver to time out when calling the conversion webhook:

```
conversion webhook for deckhouse.io/v1, Kind=DexAuthenticator failed:
Post "https://conversion-webhook-handler.d8-system.svc:443/…?timeout=30s":
context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

Any module that watches a CRD with a registered conversion (e.g. `user-authn` watching `DexAuthenticator` v2alpha1) is blocked at startup until the webhook is reachable, which can pin the entire Deckhouse operator queue.

### Root causes in the previous implementation

1. **Data race on `cmd`.** The runner goroutine reassigned `cmd = exec.Command(…)` on each iteration while the reload goroutine read `cmd.Process.Signal(…)` / `cmd.Process.Pid` without synchronisation. Under the Go memory model this is a data race; in practice it can nil-panic in the reload goroutine, permanently preventing any further restarts while the container keeps running.

2. **Channel deadlock window on natural exit.** `processExited` was unbuffered. When shell-operator exited on its own, the runner blocked on `processExited <- struct{}{}` until the reload goroutine drained it — which only happened up to one `reloadInterval` (5 s) later. During that window shell-operator was not running, but the pod remained in the Service endpoints and every kube-apiserver TLS connection on `:9681` hung until the 30 s API-server timeout.

3. **No backoff on rapid restarts.** A malformed hook file could cause shell-operator to crash-loop with no throttle, churning connections on `:9681` continuously.

4. **`ctrl.SetupSignalHandler()` called twice.** The original code called it once at the top of `main` and again in `mgr.Start(…)`. A second call panics at runtime. This was latent but would surface as soon as the first call was refactored out.

## Fix

Introduces a `shellRunner` type that owns the `*exec.Cmd` pointer behind a `sync.Mutex` and exposes a safe `SignalTerm()` method. `Run(ctx)` replaces both goroutines:

- Uses `cmd.Start()` + `cmd.Wait()` per iteration — natural exit and SIGTERM exit are handled identically; no channel needed.
- Applies exponential backoff (500 ms → 30 s) when shell-operator exits in under 5 s, preventing tight crash loops.
- Returns when `ctx` is cancelled (pod SIGTERM) after forwarding SIGTERM to the child, so shell-operator gets a clean shutdown instead of SIGKILL.

A separate short-lived goroutine replaces the reload ticker and drains cleanly on ctx cancellation. The `isReloadShellNeed` flag is cleared *before* SIGTERM so a reload request arriving between the signal and the next tick is not dropped.

`ctrl.SetupSignalHandler()` is now called exactly once; the returned context is shared by the manager and the supervisor.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: tools
type: fix
summary: fix data race webhook handler
impact_level: low
```

### DIFF
--- a/modules/002-deckhouse/images/webhook-handler/operator/cmd/main.go
+++ b/modules/002-deckhouse/images/webhook-handler/operator/cmd/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"flag"
@@ -26,6 +27,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


riptide-01 (approved):

deckhouse-BOaTswain:
Cherry pick PR [20120](https://github.com/deckhouse/deckhouse/pull/20120) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #20129, opened 2026-05-25
[go_lib/controlplane] add PKI/kubeconfig renewal libraries and structured expiration reports

trofimovdals:
## Description

Adds reusable renewal and expiration-report APIs under `go_lib/controlplane` and adapts `control-plane-manager` certificate observation to use them.

- introduces `pki.RenewCertificates` with per-certificate `PKIRenewReport` and `pki.RenewCertificate` for just err return, leaf filtering (`WithRenewLeafs`), optional `WithDryRun`, and typed skip/error reasons (`MissingError`, `CAMissingError`, etc.)
- on renewal, copies Subject, DNS/IP SANs, EKU, and encryption algorithm from the existing leaf cert (`certConfigFromX509`); serving certs can get an extra IP SAN via `WithRenewExtraIP` (client certs without IP SANs are unchanged)
- tightens `pki` certificate validation with a one-directional SAN subset check: every DNS/IP SAN required by the desired config must be present in the cert; extra SANs in the cert are allowed
- introduces `kubeconfig.RenewClientCerts` with `KubeconfigRenewReport` for admin/super-admin/controller-manager/scheduler client certs (kubelet excluded), preserving client cert Subject fields on re-sign
- replaces `WithIgnoreReadErrors` on expiration listing with structured `ExpirationReport` / `KubeconfigExpirationReport` entries that carry per-file `Err` (including `MissingError` for absent files)
- updates `CertObserve` in `control-plane-manager` to consume the new reports: warn per failed file, persist successfully read expirations into `ControlPlaneOperation` status, and fail the step when any targeted cert/kubeconfig could not be read
- adds unit tests for renewal and SAN validation in `pki/renew_test.go`, `pki/validate_test.go`, and `kubeconfig/renew_test.go`

Control-plane certificate renewal and expiration checks are moving out of ad-hoc / `kubeadm`-oriented code paths into `go_lib/controlplane`

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feature
summary: Add PKI/kubeconfig renewal libraries and structured expiration reports
impact_level: low

```

### DIFF
--- a/go_lib/controlplane/kubeconfig/renew.go
+++ b/go_lib/controlplane/kubeconfig/renew.go
@@ -98,9 +98,9 @@ type KubeconfigRenewReport struct {
 }
 
 type KubeconfigRenewEntry struct {
-	File   File
-	Path   string
-	Action KubeconfigRenewAction
+	File File
+	Path string
+	Err  error
 }
 
 // KubeconfigRenewAction describes what happened to a single kubeconfig file.
@@ -115,11 +115,11 @@ const (
 	KubeconfigRenewActionSkippedExternalCA
 )
 
-func (r *KubeconfigRenewReport) add(file File, path string, action KubeconfigRenewAction) {
+func (r *KubeconfigRenewReport) add(file File, path string, err error) {
 	r.Entries = append(r.Entries, KubeconfigRenewEntry{
-		File:   file,
-		Path:   path,
-		Action: action,
+		File: file,
+		Path: path,
+		Err:  err,
 	})
 }
 
@@ -227,7 +227,7 @@ func RenewClientCerts(opts ...RenewOption) (KubeconfigRenewReport, error) {
 
 		err := renewClientCert(o.kubeconfigDir, o.pkiDir, info.File)
 		if err == nil {
-			report.add(info.File, path, KubeconfigRenewActionRenewed)
+			report.add(info.File, path, nil)
 			continue
 		}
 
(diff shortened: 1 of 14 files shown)


go_lib/controlplane/kubeconfig/renew.go
```diff
@@ -0,0 +1,283 @@
...
+// All other kubeconfig fields are preserved. CA cert and key are loaded from pkiDir.
+//
+// The returned error encodes the outcome:
+//   - nil              — re-signed cleanly
+//   - *MissingError    — kubeconfig file absent (skipped)
+//   - *CAMissingError  — CA cert file absent (skipped)
+//   - *CAExternalError — CA key absent / external CA (skipped)
+//   - *CAExpiredError  — CA already expired (skipped)
+//   - any other error  — IO/permissions/signing failure (skipped)
+func RenewClientCert(file File, opts ...RenewOption) error {
```
trofimovdals, line 234: Not using at this moment(for pki/renew.go same), just for simmetry with expiration...

sprait (approved):

go_lib/controlplane/pki/expiration.go
```diff
@@ -72,40 +93,43 @@ func WithRootCertificates(names ...RootCertName) ExpirationOption {
...
-func ListCertificateExpirations(opts ...ExpirationOption) ([]CertificateExpiration, error) {
+// ListCertificateExpirations enumerates the selected certificates and returns a structured report.
+// The caller iterates report.Entries and decides per-entry what to do.
+func ListCertificateExpirations(opts ...ExpirationOption) ExpirationReport {
 	options := newExpirationOptions(opts...)
 
 	inventory, err := buildCertificateInventory(options)
 	if err != nil {
-		return nil, err
+		panic(err)
```
AmazinMax: seems like vulnerability: if user provides incorrect `rootCertificate` or `leafCertificate` through options (`WithRootCertificates(), WithLeafCertificates()`), it leads to an error in  `buildCertificateInventory`, and panic here. may be we actually should return error as a second param here with empty report

AmazinMax (changes requested):
Good job! 1 suggestion

AmazinMax (approved):

PR #20281, opened 2026-05-28
[user-authn] Add useroperations detail logs

Jabejixo:
## Description

`UserOperation` (user-authn module) gets three changes:
- new `spec.lock.permanent: true` field — indefinite lock; hook writes `lockedUntil = 9999-12-31`, cleared by a regular `Unlock` operation;
- new `deckhouse.io/initiator` annotation — email of the admin who triggered the operation from the Console UI, surfaced in the hook's structured logs;
- `spec.resetPassword.newPasswordHash` is wiped from the CR as soon as the operation reaches a terminal phase (Succeeded/Failed), so the bcrypt hash does not sit in etcd for the 24h retention window;
- richer audit logging across the hook lifecycle: `operation`, `type`, `initiator`, `initiatorType`, `targetKind`, `targetUser`/`targetConnector`/`targetEmail`, `lockedUntil`.

Backward-compatible: `spec.lock.permanent` and the annotation are both optional. No impact on control-plane, ingress-controllers, or Prometheus.

- Operators could not tell from the `deckhouse` logs *who* triggered a `ResetPassword` / `Lock` / `Unlock` from Console — only the role-level marker `initiatorType=admin`. The new annotation + log key now records the actual admin email.
- There was no way to ban a user "forever" — admins had to pass an arbitrarily large `for` duration. Explicit `permanent: true` semantics is now supported end-to-end (CRD, hook, Console UI).
- The bcrypt hash of the new password lived in `spec.resetPassword.newPasswordHash` for the full 24h retention period after the password was applied. It is now removed immediately after the hook finishes processing the operation, regardless of outcome.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: feature
summary: UserOperation supports permanent user locks via spec.lock.permanent.
---
section: user-authn
type: feature
summary: UserOperation audit logs record the initiator admin's email from the deckhouse.io/initiator annotation, along with the operation type and the target user.
---
section: user-authn
type: fix
summary: UserOperation no longer keeps the bcrypt password hash in spec.resetPassword.newPasswordHash after the operation completes.
```


modules/150-user-authn/crds/user-operation.yaml
```diff
@@ -67,10 +72,25 @@ spec:
...
+                        Days are expanded as `1d = 24h` by the hook before parsing.
+                      pattern: '^(permanent|([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h|d))+)$'
+                      x-kubernetes-validations:
+                        # Self-sufficient: re-asserts shape (so garbage like "permanent7d" or
+                        # "abc1d" cannot slip through if `pattern` ever regresses) and then
+                        # requires at least one non-zero digit, which — combined with the
+                        # all-segments-are-non-negative shape — guarantees a positive total
+                        # ("30m" passes, "0s" / "0d0h" do not). We avoid CEL's `duration()`
+                        # because it does not accept the "d" unit we surface here.
+                        - rule: 'self == "permanent" || (self.matches("^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h|d))+$") && self.matches("[1-9]"))'
```
AlwxSin: Is there any reason to allow values less than second?

AlwxSin (approved):

Zhbert (approved):

PR #20334, opened 2026-06-01
[deckhouse-controller] fix conditional dependency check

ldmonster:
## Description

- Fixes the conditional (`!optional`) module dependency check so that enabling a module no longer disables an already running module that depends on it.
- Affects only the `ModuleDependency` extender (`CheckEnabling`) and its unit tests.

### **Problem:**

- A module can declare a conditional dependency on another module, e.g. `virtualization` → `sdn: ">= 0.6.2 !optional"`: it works without `sdn`, but if `sdn` is enabled it must satisfy the version constraint.
- When the dependent module was already running and a user enabled the dependency at an older, incompatible version, the running dependent module was **disabled and torn down** (its namespace and workloads removed) instead of the incompatible enable being rejected.
- Root cause: `CheckEnabling` (the ModuleConfig admission check) returned early for modules that declare no dependencies of their own (like `sdn`). Because of that, the reverse check "does enabling this module break an already enabled dependent?" never ran. The enable was allowed, and the runtime scheduler then disabled the dependent module.

### **Fix**

- `CheckEnabling` now always validates the module being enabled against the constraints of already enabled dependent modules, even when the module itself declares no dependencies.
- Result: enabling an incompatible dependency is **rejected by the admission webhook**, and the running dependent module stays up.

### **Impact**

- An already running module (and its workloads) is no longer unexpectedly disabled when an incompatible version of its optional dependency is enabled.

---

#### Before

2026-06-01 AT 17 58@2x

#### After

2026-06-01 AT 18 02@2x

## Checklist
- [x] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: A module that conditionally depends on another is no longer disabled when an incompatible version of that dependency is enabled; the enable is rejected instead.
```

PR #20341, opened 2026-06-01
[deckhouse-controller] fix conditional dependency check

Glitchy-Sheep:
## Description

**⚠️ FIX BACKPORTED from**: #20334

- Fixes the conditional (`!optional`) module dependency check so that enabling a module no longer disables an already running module that depends on it.
- Affects only the `ModuleDependency` extender (`CheckEnabling`) and its unit tests.

### **Problem:**

- A module can declare a conditional dependency on another module, e.g. `virtualization` → `sdn: ">= 0.6.2 !optional"`: it works without `sdn`, but if `sdn` is enabled it must satisfy the version constraint.
- When the dependent module was already running and a user enabled the dependency at an older, incompatible version, the running dependent module was **disabled and torn down** (its namespace and workloads removed) instead of the incompatible enable being rejected.
- Root cause: `CheckEnabling` (the ModuleConfig admission check) returned early for modules that declare no dependencies of their own (like `sdn`). Because of that, the reverse check "does enabling this module break an already enabled dependent?" never ran. The enable was allowed, and the runtime scheduler then disabled the dependent module.

### **Fix**

- `CheckEnabling` now always validates the module being enabled against the constraints of already enabled dependent modules, even when the module itself declares no dependencies.
- Result: enabling an incompatible dependency is **rejected by the admission webhook**, and the running dependent module stays up.

### **Impact**

- An already running module (and its workloads) is no longer unexpectedly disabled when an incompatible version of its optional dependency is enabled.

---

#### Before

2026-06-01 AT 17 58@2x

#### After

2026-06-01 AT 18 02@2x

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: A module that conditionally depends on another is no longer disabled when an incompatible version of that dependency is enabled; the enable is rejected instead.
```

### DIFF
--- a/go_lib/dependency/extenders/moduledependency/moduledependency.go
+++ b/go_lib/dependency/extenders/moduledependency/moduledependency.go
@@ -17,7 +17,6 @@ limitations under the License.
 package moduledependency
 
 import (
-	"errors"
 	"fmt"
 	"log/slog"
 	"slices"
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
:white_circle: `e2e: Yandex.Cloud` for `deckhouse:fix/dependency-correctly-check-in-1.73` [cancelled](https://github.com/deckhouse/deckhouse/actions/runs/26766738614).

</details>

ldmonster (approved):

PR #20351, opened 2026-06-02
[dhctl] Add logs output to interactive mode

YuryLysov:
## Description

Add the area in interactive mode, which shows the logs of ongoing processes

Some processes are taking too long, e.g. `terraform apply` sometimes takes like 10m, and we should keep end user informed, what it's not just a lag, and process keep doing it's work.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: feature
summary: Added a log area to the interactive mode of dhctl.
impact_level: default
```

### DIFF
--- a/dhctl/pkg/util/progressbar/logbox.go
+++ b/dhctl/pkg/util/progressbar/logbox.go
@@ -1,0 +1,234 @@
+// Copyright 2026 Flant JSC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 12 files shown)


090809 (approved):

dhctl/pkg/util/progressbar/logbox.go
```diff
@@ -0,0 +1,234 @@
...
+func (b *LogBox) Update() {
+	b.started = true
+	for {
+		select {
+		case <-b.stopChan:
+			b.started = false
+			return
+		case msg, ok := <-b.logChan:
+			if !ok {
+				continue
```
borg-z: If logChan is never closed, then `if !ok` statement is useless, if it is closed, there should be a `return` statement instead of a `continue`.

dhctl/pkg/util/progressbar/logbox.go
```diff
@@ -0,0 +1,234 @@
...
+	for i := range 11 {
+		b.SpinnerPrinterArray[i].SetWriter(b.writerArray[i])
+	}
+
+	b.writerFabric.PutWriter(wr)
+}
+
+type WriterFabric struct {
+	mp     *pterm.MultiPrinter
+	witers []io.Writer
```
borg-z: writers?

borg-z (approved):

PR #20355, opened 2026-06-02
[deckhouse] fix application update summary reporting Ready while manifests still applying

Fkuloff:
## Description

In the update phase `summarize` returned `Ready` whenever the update pipeline had no
failing gate. But "no blocker" ≠ "update finished": `firstFalse` skips a transient
`ManifestsApplied=False/ApplyingManifests`, so the pipeline reads clear during every
re-apply window while the previous — possibly failed — conditions stay sticky.

Now `summarize` returns `Ready` only when `ManifestsApplied=True`, and otherwise
reports a new `Updating` summary (the new version is still being applied; the
previous version keeps serving).

Without the gate, a mid-apply retry over a failed update reported `Ready` while the
conditions a client also reads still said `ManifestsApplyFailed` — a self-contradicting
status. An update whose manifests are not yet applied is now reported as `Updating`,
never `Ready`.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Application update summary no longer reports Ready while the new version's manifests are still being applied.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/pkg/controller/packages/application/status/summary.go
+++ b/deckhouse-controller/pkg/controller/packages/application/status/summary.go
@@ -207,6 +207,17 @@ var summarySuspended = advice{
 // there is no message or tip.
 var summaryReady = advice{state: stateReady}
 
+// summaryUpdating is the fixed Summary for an update that is mid-flight with no
+// failing gate: the new version's manifests are still being applied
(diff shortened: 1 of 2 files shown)


ldmonster (approved):

deckhouse-BOaTswain:
Backport failed ⚠️.

PR #20382, opened 2026-06-02
[node-manager] Fix calculate NodeGroup status

YuryLysov:
## Description

Fix static NodeGroup readiness calculation 

Now readiness of NodeGroup calculates wrong, if static NodeGroup has not only CAPS, but static nodes

Affected 1.75, not affected main

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Fix static NodeGroup readiness calculation.
impact_level: default
```

### DIFF
--- a/modules/040-node-manager/hooks/internal/conditions/calculate.go
+++ b/modules/040-node-manager/hooks/internal/conditions/calculate.go
@@ -200,7 +200,7 @@ func CalculateNodeGroupConditions(
 		curTime, _ = time.Parse(time.RFC3339, timeStr)
 	}
 
-	staicDesired := ng.Desired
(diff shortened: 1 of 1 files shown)


modules/040-node-manager/hooks/internal/conditions/calculate.go
```diff
@@ -197,6 +200,8 @@ func CalculateNodeGroupConditions(
 		curTime, _ = time.Parse(time.RFC3339, timeStr)
 	}
 
+	staicDesired := ng.Desired
```
090809: typo

090809 (approved):

PR #20102, opened 2026-05-25
[dhctl] refactor bootstrap process with adding more opentelemetry for bashible

090809:
## Description

This PR is based on top of https://github.com/deckhouse/deckhouse/pull/19956, which contains improvements to the phases of the bootstrap process.

Two changes:

1. **Bootstrap steps refactor.** The monolithic `dhctl/pkg/operations/bootstrap/steps.go` (~600 lines) is removed and split by concern into focused files:
   - `steps_ssh.go` — SSH/bootstrap-node steps
   - `steps_cache.go` — cache handling steps
   - `steps_nodes.go` — node provisioning steps
   - `steps_deckhouse.go` — Deckhouse install steps
   - `steps_bashible.go` — bashible run steps

   `cluster-bootstrapper.go` is reworked to wire the split steps, and `phases.go` / `progressbar.go` are adjusted accordingly. No change to bootstrap behavior — pure restructuring for readability and maintainability.

2. **OpenTelemetry tracing for bashible `run-step` (opt-in).** A new `dhctl/pkg/telemetry/relay` package (`relay.go`, `server.go`) runs an OTel HTTP relay on the node over the existing SSH reverse-proxy connection. The bashible templates (`candi/bashible/bashible.sh.tpl`, `candi/bashible/lib.sh.tpl`) gain `bb-telemetry-start-span` / `bb-telemetry-end-span` helpers that emit a span per executed step back to dhctl. Tracing is **disabled by default**; it is enabled only via the `DHCTL_TRACE` env var (`dhctl/pkg/telemetry/enabled.go`), which sets `DHCTL_TELEMETRY_ENABLED=true` for bashible. When off, the helpers return immediately and node bootstrap is unchanged.

## Why do we need it, and what problem does it solve?

`steps.go` had grown into the largest, hardest-to-navigate file in the bootstrap package, mixing SSH, cache, node, Deckhouse, and bashible concerns. Splitting it by responsibility makes the bootstrap flow easier to read, review, and extend.

The OpenTelemetry tracing gives developers and operators per-step timing/visibility into the bashible `run-step` execution on bootstrapping nodes, making it possible to diagnose where node bootstrap is slow or failing without manual log archaeology.

## Checklist
- [ ] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: dhctl
type: chore
[...]

PR #20405, opened 2026-06-03
[dhctl] fix: static instance local installation fix

090809:
## Description

In this [PR](#20102), when we fixes linter issues, we added error handling for scenario, where static configuration is missed, which wrong.

Fixed wrong error handling

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Omitted an error return when the static config is missing.
impact:
impact_level: default
```


borg-z (approved):
awesome

PR #20486, opened 2026-06-04
[istio] Custom annotations for Ingress Controller

skurbatov:
## Description
This change adds support for custom annotations on ingressgateway pods via the `istio.alliance.ingressGateway.gatewayPodAnnotations` module setting.

The parameter accepts arbitrary key-value annotations that are applied to the pods created by the alliance ingressgateway DaemonSet. This lets cluster administrators customize ingressgateway pod behavior and integrate with external systems without changing module templates.

Typical use cases include:

annotations for cluster-autoscaler, monitoring, or security policies;
annotations read by controllers and admission webhooks at the Pod level;
other operational metadata that must live on ingressgateway pods rather than on the DaemonSet object or Service.
The feature is available in EE and CSE Pro editions and applies to the ingressgateway deployed when federation or multicluster is enabled and an ingress gateway is required for cross-cluster traffic.

Cluster impact: changing `gatewayPodAnnotations` triggers a rolling update of the ingressgateway DaemonSet, which recreates ingressgateway pods. It does not affect the control plane, Prometheus, or other ingress controllers (such as `IngressIstioController`).

Today, alliance ingressgateway configuration supports only Service annotations (`serviceAnnotations`), for example to configure a cloud Load Balancer. There is no way to set custom annotations on ingressgateway pods through module configuration.

Because of that, administrators cannot attach required pod-level annotations via `ModuleConfig` and must work around the limitation manually or wait for template changes.

This matters because many Kubernetes mechanisms operate on Pod annotations, not on DaemonSet metadata. The same approach is already used elsewhere in Deckhouse (for example, `controllerPodsAdditionalAnnotations` in ingress-nginx).

The new parameter closes this gap and provides a consistent way to customize alliance ingressgateway pods through the `istio` module configuration.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: feature
summary: Added support for custom annotations on the Istio Ingress Gateway controller.
impact_level: default
```

### DIFF
--- a/modules/110-istio/openapi/config-values.yaml
+++ b/modules/110-istio/openapi/config-values.yaml
@@ -135,7 +135,7 @@ properties:
               - cse-pro
             example:
               yandex.cpi.flant.com/listener-subnet-id: xyz-123
-          daemonSetAnnotations:
+          gatewayPodAnnotations:
             type: object
             additionalProperties:
               type: string
--- a/modules/110-istio/openapi/config-values.yaml
+++ b/modules/110-istio/openapi/config-values.yaml
@@ -135,6 +135,16 @@ properties:
               - cse-pro
             example:
               yandex.cpi.flant.com/listener-subnet-id: xyz-123
+          daemonSetAnnotations:
+            type: object
+            additionalProperties:
+              type: string
+            description: Additional annotations for ingressgateway DaemonSet pods.
+            x-doc-d8Editions:
+              - ee
+              - cse-pro
+            example:
+              checksum/config: xyz-123
           nodeSelector:
             type: object
             additionalProperties:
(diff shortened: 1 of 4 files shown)


modules/110-istio/openapi/config-values.yaml
```diff
@@ -135,6 +135,16 @@ properties:
               - cse-pro
             example:
               yandex.cpi.flant.com/listener-subnet-id: xyz-123
+          daemonSetAnnotations:
```
apolovov:
```suggestion
          gatewayPodAnnotations:
```

skurbatov: fixed

apolovov (approved):

ldmonster (approved):

Zhbert (approved):

deckhouse-BOaTswain:
Cherry pick PR [21118](https://github.com/deckhouse/deckhouse/pull/21118) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #20507, opened 2026-06-05
[cotrol-plane-manager] Fix error from validating root certs function during regeneration

borovetsav:
## Description

Fixed the certificate regeneration case: only non-root certificates are regenerated according to the documented behavior. Also no more error from validating root certs function.

Added a tests covering this case.

To get rid of error from validating root certs function.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Fixed an error in the root certificate validation function.
impact_level: default
```


meatuses:
lgtm

sprait (approved):

PR #20561, opened 2026-06-09
[dhctl] Fix grammar and spelling in log and user-facing messages

nabokihms:
## Description

Grammar, spelling, and phrasing fixes across dhctl's human-readable messages — log lines (`log.*`), interactive prompts/confirmations, error messages (`fmt.Errorf`), operation/phase titles, and CLI flag help texts. Only message strings were touched; variable names, flags, interfaces, and format verbs (`%s/%d/%v/%w`) were left unchanged.

Examples:
- `Cannot drain node, because process was timeout. Do we continue without full-fledged drain?` → `Cannot drain the node because the process timed out. Continue without a full-fledged drain?`
- `%s not exists. Fallback to embedded minget` → `%s does not exist. Falling back to embedded minget`
- `Lease renew was failed.` → `Lease renewal failed.`
- spelling: `applyed`→`applied`, `Gorutines`→`Goroutines`, `terrform`/`infastructure`/`opentofy`/`pipline`/`parrent`, doubled `write write`.

Also dropped a dangling `%v` in `lock.go` where `log.DebugLn` (Println-style, not a format function) printed the verb literally.

dhctl's output was often grammatically incorrect and inconsistent, which looks unpolished and can confuse users reading logs and prompts. This makes the messages read like fluent technical English without changing behavior.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

Note: text-only message changes. Verified locally with `go build ./...` and `go vet ./...` (clean) and the affected package tests pass; no behavior changes.

## Changelog entries

```changes
section: dhctl
type: chore
summary: Fix grammar and spelling in dhctl log and user-facing messages.
impact_level: low
```

### DIFF
--- a/dhctl/cmd/dhctl/action.go
+++ b/dhctl/cmd/dhctl/action.go
@@ -83,7 +83,7 @@ func (i *actionIniter) init(c *kingpin.ParseContext) error {
 	}
 
 	if i.registerOnShutdown == nil {
-		return fmt.Errorf("Internal error: action initer not initialized. Did not pass register on shutdown")
+		return fmt.Errorf("Internal error: action initer not initialized. The register-on-shutdown callback was not passed")
 	}
 
 	tmpDir := i.params.tmpDirName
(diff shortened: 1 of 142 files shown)


borg-z (approved):
looks good

PR #20716, opened 2026-06-17
[node-manager] drop Unknown shutdown condition status

borg-z:
## Description

shutdown-inhibitor no longer sets the `GracefulShutdownPostpone` node condition to `status: Unknown`. The idle (armed) state now uses `status: True` with `reason: WaitingForShutdownSignal`. So the condition now has only `True` or `False`.

`status: Unknown` looks like a problem in `kubectl describe node` and in monitoring, but it is only the normal idle state. The kubelet patch already treats `True` as "postpone", so node behavior does not change and the race on the shutdown signal stays closed.

Not necessarily.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Made shutdown-inhibitor set the GracefulShutdownPostpone condition only to True or False.
```

### DIFF
--- a/ee/modules/040-node-manager/images/d8-shutdown-inhibitor/src/pkg/app/nodecondition/condition.go
+++ b/ee/modules/040-node-manager/images/d8-shutdown-inhibitor/src/pkg/app/nodecondition/condition.go
@@ -6,7 +6,6 @@ Licensed under the Deckhouse Platform Enterprise Edition (EE) license. See https
 package nodecondition
 
 const (
-	StatusTrue    = "True"
-	StatusFalse   = "False"
-	StatusUnknown = "Unknown"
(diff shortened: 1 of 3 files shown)


090809 (approved):

deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:feature/shutdown-inhibitor-drop-unknown-status` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/27718236439) in 30m32s.



GCP-WithoutNAT-Containerd-1.34 - Connection string: `ssh  user@34.185.176.159`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.34` succeeded in 29m29s.

</details>

PR #20782, opened 2026-06-19
[admission-policy-engine] Fix constraint-template check of AssignImage crd

Suselz:
## Description

Fix mutation resource discovery in admission-policy-engine constraint-exporter for Gatekeeper mutation APIs.

The exporter now supports both` mutations.gatekeeper.sh/v1` and `mutations.gatekeeper.sh/v1alpha1`, so mutation tracking is populated even on clusters where only `v1alpha1` mutation CRDs are served (including AssignImage).

Also added unit tests for version-specific discovery scenarios.

AssignImage did not trigger rendering of MutatingWebhookConfiguration because tracked mutate resources stayed empty.

This change makes mutation discovery resilient across both versions and restores the chain:
mutation CRs -> tracked mutate resources -> webhook rules rendering.

This is a bugfix for existing installations where mutation resources are served via v1alpha1 only. Without it, mutation webhook rules are not generated, and mutation policies (including AssignImage) are effectively non-working.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Fixed the constraint-template check for the AssignImage CRD.
impact_level: default
```

### DIFF
--- a/modules/015-admission-policy-engine/images/constraint-exporter/src/pkg/gatekeeper/mutations.go
+++ b/modules/015-admission-policy-engine/images/constraint-exporter/src/pkg/gatekeeper/mutations.go
@@ -19,19 +19,26 @@ package gatekeeper
 import (
 	"context"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


AlwxSin (approved):

deckhouse-BOaTswain:
Backport failed ⚠️.

PR #20838, opened 2026-06-23
[control-plane-manager] Virtual control plane/vcpn controller and refactor variant

trofimovdals:
## Description

Adds the cpn controller for [virtual control planes](https://github.com/deckhouse/deckhouse/pull/20681) and shares the cpn planning logic for existing physical control plane.

- New controller `virtual-control-plane-node` (`internal/controllers/virtual-control-plane-node`). It runs in the virtual manager under leader election, watches `ControlPlaneNode` objects labeled `control-plane.deckhouse.io/type: virtual` (created by `virtual-control-plane-configuration` in `vcp-<name>` namespaces), creates the required `ControlPlaneOperation` objects per component, and folds completed operations back into the node `status`.
- New shared package `internal/cpn` — the mode-agnostic ControlPlaneNode planner. It computes per-component state and the operations a node needs (`converge`, `cert-renew`, `signature-renew`, `observe`), deduplicates against in-flight operations, rotates terminal operations, and builds the node status report. Both the physical `control-plane-node` controller and the new virtual controller drive their reconcile loops through it; only controller wiring and the operation executor differ (disk vs host-cluster API).
- `control-plane-manager` hardening that applies to both physical and virtual nodes:
  - maintenance mode now keeps requeuing, so a node resumes operation planning promptly after the `control-plane-manager.deckhouse.io/maintenance` label is removed (label removal does not bump `generation`, so nothing else would wake the controller);
  - operations are confirmed against a strongly-consistent uncached read only when something is about to be created, preventing duplicate `ControlPlaneOperation` objects under informer-cache lag without extra API load on steady-state reconciles;
  - operations are filtered by owner `UID`, so a same-name `ControlPlaneNode` recreated by the virtual configurator does not reconstruct state from a previous instance's not-yet-garbage-collected operations;
  - first `kube-apiserver` deploy seeds the signature keys (CSE builds), and signature-key rotation is planned as a separate operation.

This change does not change behavior for already-running physical control planes: the planning result for `type: normal` nodes is unchanged, and no control-plane component is restarted by this PR on its own.

Virtual (tenant) control planes need the same operation-driven lifecycle as physical masters — initial deploy, certificate and signature renewal, scaling masters from 1 to 3, and periodic certificate observation — but executed against the host cluster API instead of node disks.

Rather than duplicating the planning logic, this PR extracts it into `internal/cpn` and adds a thin virtual controller on top. Sharing one planner keeps physical and virtual behavior consistent, lets the upcoming virtual operation executor reuse proven planning logic, and turns the future physical refactor into a thin wiring change. The PR also closes latent issues that affect both modes: a controller that could stay asleep after maintenance mode is lifted, and duplicate operations created under informer-cache lag.

This is incremental work on the `feature/virtual-control-plane` branch; the virtual control plane is not yet enabled for end users.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feature
summary: Add the virtual ControlPlaneNode controller and the shared cpn planner reused by physical and virtual control planes.
impact_level: low
```


modules/040-control-plane-manager/images/control-plane-manager/src/internal/cpn/cpnreconcile/reconciler.go
```diff
@@ -0,0 +1,138 @@
+/*
```
sprait, line 1: rename to reconciler.go

modules/040-control-plane-manager/images/control-plane-manager/src/internal/operations/operation.go
```diff
@@ -0,0 +1,83 @@
...
+	if op.IsObserveOnlyOperation() {
+		op.Spec.Approved = true
+	} else {
+		op.Spec.DesiredConfigChecksum = intended.Config
+		op.Spec.DesiredPKIChecksum = intended.PKI
+		op.Spec.DesiredCAChecksum = intended.CA
+	}
+	op.GenerateName = generateName(op)
+	return op
+}
```
sprait: may be  separate to two func `NewApprovedOperation(...)` and `NewOperation` and add method `Build`     for `TargetSteps`? then we can use more clear logic withour extra if:

```golang
func (VirtualStepBuilder) TargetSteps(s componentState) []TargetSteps {
    var targets []TargetSteps
    switch {
    case s.needsConverge():
        steps := convergeSteps(s)
        intended := s.intended
        component := s.component
        targets = append(targets, TargetSteps{
            HasDuplicate: operations.MatchesChecksums(intended),
            Build: func(node operations.NodeRef) *controlplanev1alpha1.ControlPlaneOperation {
                return operations.NewOperation(node, component, steps, intended)
            },
        })
    case s.needsObserve():
        component := s.component
        targets = append(targets, TargetSteps{
            HasDuplicate: operations.IsAnyActiveOperation,
            Build: func(node operations.NodeRef) *controlplanev1alpha1.ControlPlaneOperation {
                return operations.NewApprovedOperation(node, component)
            },
        })
    }
...
```

modules/040-control-plane-manager/images/control-plane-manager/src/internal/cpnplanner/plan.go
```diff
@@ -0,0 +1,82 @@
...
+	if IsMaintenanceMode(cpn) {
+		return p
+	}
+	node := nodeRef(cpn)
+	for _, s := range computeComponentStates(cpn) {
+		for _, t := range builder.TargetSteps(s) {
+			if operations.HasActiveOperation(current, s.component, t.HasDuplicate) {
+				continue
+			}
+			p.Create = append(p.Create, operations.New(node, s.component, t.Steps, s.intended))
```
sprait: add `HasDuplicate: t.HasDuplicate` to PlannedOperation struct and use single call to ComputePlan. 
In reconcileOperation we will use fresh only for createOperationm something like that:
```
fresh, err := r.listOperationsForNodeUncached(ctx, cpn)
    if err != nil {
        return err
    }
    for _, p := range planned {
        if operations.HasActiveOperation(fresh, p.Component, p.HasDuplicate) {
            continue
        }
        if err := r.createOperation(ctx, cpn, p.Op); err != nil {
            return err
        }
```

sprait (commented):
some suggestions

modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-node/controller.go
```diff
@@ -0,0 +1,61 @@
...
+	cpoPreds, err := controlPlaneOperationPredicates()
+	if err != nil {
+		return fmt.Errorf("build ControlPlaneOperation predicates: %w", err)
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		WithOptions(controller.TypedOptions[reconcile.Request]{
+			MaxConcurrentReconciles: 10,
+		}).
+		Named(ControllerName).
```
AmazinMax: `ControllerName` -> `constants.VirtualControlPlaneNodeController`. Or remove constant `VirtualControlPlaneNodeController` and store controller name in `ControllerName` local const

modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-node/predicates.go
```diff
@@ -0,0 +1,67 @@
...
+	"control-plane-manager/internal/constants"
+)
+
+// controlPlaneNodePredicates filters ControlPlaneNode events for virtual nodes with spec changed.
+func controlPlaneNodePredicates() ([]predicate.Predicate, error) {
+	labelPred, err := predicate.LabelSelectorPredicate(metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			constants.ControlPlaneTypeLabelKey: string(constants.ControlPlaneTypeVirtual),
+		},
+	})
```
AmazinMax: can be removed as unnecessary duplicate of the cache filtering logic

modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-node/predicates.go
```diff
@@ -0,0 +1,67 @@
...
+	return []predicate.Predicate{labelPred, predicate.GenerationChangedPredicate{}}, nil
+}
+
+// controlPlaneOperationPredicates filters owned ControlPlaneOperation events for virtual cpo with status changed.
+func controlPlaneOperationPredicates() ([]predicate.Predicate, error) {
+	labelPred, err := predicate.LabelSelectorPredicate(metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			constants.ControlPlaneTypeLabelKey: string(constants.ControlPlaneTypeVirtual),
+		},
+	})
```
AmazinMax: can be removed as unnecessary duplicate of the cache filtering logic

AmazinMax (approved):

sprait (approved):

AmazinMax (approved):

PR #20847, opened 2026-06-24
[user-authn] fix user hook

Jabejixo:
## Description

Local Dex users (`User` CR) now have their `Password` object reconciled directly by the `get_dex_user_crds` hook instead of being rendered through Helm. On creation the hook stamps `hashUpdatedAt` with the current time and writes the object immediately, so a freshly created user can log in right away and is not forced to rotate the password on first login. Existing `Password` objects are updated with a JSON merge patch that only touches Deckhouse-owned fields (`email`, `username`, `userID`, `groups`, labels, annotations) and never clobbers Dex-managed runtime fields (`hash`, `hashUpdatedAt`, `previousHashes`, `incorrectPasswordLoginAttempts`, `lockedUntil`, `requireResetHashOnNextSuccLogin`).

The old `templates/dex/passwords.yaml` template is removed. To keep Helm from pruning previously rendered `Password` objects after the template is gone, every object carries the `helm.sh/resource-policy: keep` annotation. Orphaned module-managed `Password` objects (deleted users or changed emails) are cleaned up by the hook. Does not restart any critical cluster component.

Two UX problems when creating a local user from the Console:
- ~5 minute delay before the user could log in (the `Password` object was created indirectly via a module Helm reconcile).
- Immediate "rotate your password" screen on first login, because the `Password` object had no `hashUpdatedAt`, so the password-rotation policy treated the brand-new password as ancient.

Reconciling `Password` objects directly in the hook makes new users usable immediately and starts the rotation clock at creation time.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Reconcile Dex Password objects directly in the hook so new local users can log in immediately and are not forced to rotate their password on first login.
impact_level: low
```


modules/150-user-authn/hooks/get_dex_user_crds.go
```diff
@@ -249,14 +302,27 @@ func getDexUsers(_ context.Context, input *go_hook.HookInput) error {
...
+		// Helm. This makes a freshly created user usable immediately (no wait for a
+		// module Helm reconcile) and lets us stamp hashUpdatedAt on creation, which
+		// the password rotation policy relies on. We never touch Dex-managed runtime
+		// fields (hash after a user change, hashUpdatedAt, previousHashes,
+		// incorrectPasswordLoginAttempts, lockedUntil, requireResetHashOnNextSuccLogin)
+		// on an existing object.
+		if passwordExists {
+			reconcileExistingPassword(input, existingPassword, dexUser.Name, email, groups)
+		} else {
+			input.PatchCollector.CreateIfNotExists(newPasswordObject(encodedName, dexUser.Name, email, rawPassword, groups, now))
```
AlwxSin, line 361: The lookup key for the existing `Password` changed from `dexUser.Name` (stable) to `encoding.ToFnvLikeDex(email)` (line 239-242). Since `User.spec.email` is mutable (only its format is validated, and the immutability policy only freezes `spec.password`), an email edit makes `passwordsByName[encodedName]` miss the existing object, so the branch treats the user as brand-new and reseeds the password from `dexUser.Spec.Password` here - which is the immutable creation-time hash, not the current one. If the user changed their password at any point after account creation, this silently reverts them to their very first password on the next email edit, and the old `Password` object (with the live hash) gets deleted the same run as an "orphan". Can we key the lookup by username as a fallback when the email-encoded name doesn't match, so an existing password survives a rename?

modules/150-user-authn/hooks/get_dex_user_crds_test.go
```diff
@@ -655,4 +665,207 @@ username: admin
...
+		})
+		It("Should delete the old-name Password and create one under the new email encoding", func() {
+			Expect(f).To(ExecuteSuccessfully())
+			// Object keyed by the previous email encoding is orphaned and removed.
+			Expect(f.KubernetesResource("Password", "d8-user-authn", "oldemailencodedname").Exists()).To(BeFalse())
+			// A fresh object appears under the new email's Dex (FNV-like) encoding.
+			newPassword := f.KubernetesResource("Password", "d8-user-authn", "mfsg22loibsxqylnobwgkltdn5w4x4u44scceizf")
+			Expect(newPassword.Exists()).To(BeTrue())
+			Expect(newPassword.Field("email").String()).To(Equal("admin@example.com"))
+			Expect(newPassword.Field("hashUpdatedAt").Time()).To(BeTemporally("~", time.Now(), time.Minute))
```
AlwxSin: Could you extend this test so the pre-existing `Password` has a `hash` that differs from `User.spec.password` (like the group-change test above does with its `...MustNotOverwriteLiveHash` literal), and assert the new object's `hash` still equals the old live hash? As written this test only checks that *a* new object appears - it doesn't catch the password getting reset to the CR's original value.

modules/150-user-authn/hooks/get_dex_user_crds.go
Jabejixo: Fixed. Added a passwordsByUsername fallback: on an email rename we recreate the object under the new name via newPasswordObjectFromExisting, carrying the live hash + runtime fields instead of reseeding from spec.password. Limited to module-managed objects.

modules/150-user-authn/hooks/get_dex_user_crds_test.go
Jabejixo: Done. The pre-existing Password now has a live hash differing from spec.password; the test asserts the recreated object keeps it.

AlwxSin (approved):

PR #20890, opened 2026-06-25
[deckhouse] fix: preserve legacy bash completion flag

fuldaxxx:
## Description
After the migration of deckhouse-controller from kingpin to cobra, the legacy `--completion-script-bash` flag was removed in favor of the canonical completion bash subcommand. However, the bash completion setup shipped inside the `Deckhouse` image (`/etc/bashrc`, rendered from `deckhouse-controller/files/bashrc.tpl`) still invokes `deckhouse-controller --completion-script-bash`, so the autocompletion script stopped being generated and `deckhouse-controller` tab-completion broke inside the pod.

This change:
 - Adds a hidden `--completion-script-bash` persistent flag on the root cobra command, bound to a package-level `legacyBashCompletion` variable.
 - In `PersistentPreRunE`, when the flag is set, the bash completion script is generated via `cmd.Root().GenBashCompletionV2(os.Stdout)` and the process exits early, mirroring the previous kingpin behavior.
 - Simplifies the corresponding eval line in `bashrc.tpl`, dropping the now-unnecessary `sed -e s/deckhouse/deckhouse-controller/g` substitution (cobra generates the script for `deckhouse-controller` directly, while kingpin produced one named `deckhouse`).

The flag is hidden from help output to keep the completion subcommand as the canonical interface; the legacy flag exists solely for backward compatibility.

The cobra migration silently dropped the `--completion-script-bash` flag that the in-image `/etc/bashrc` relies on to enable bash completion for `deckhouse-controller`. As a result, every new shell inside a `Deckhouse` pod failed to load completions (the eval of the missing-flag command produced no usable script), degrading the operator experience in the pod shell with no visible error.

This PR restores the flag as a backward-compatible alias, so existing `bashrc` entries and any external callers/scripts that still use `--completion-script-bash` keep working without changes, while the completion subcommand remains the documented, canonical path going forward.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Add alias for legacy bash completion flag.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/cmd/deckhouse-controller/main.go
+++ b/deckhouse-controller/cmd/deckhouse-controller/main.go
@@ -127,8 +127,10 @@ func main() {
 			// External callers (and our own image's /etc/bashrc) still invoke
 			// `deckhouse-controller --completion-script-bash`, so keep emitting
 			// the bash completion script and exit early when the flag is set.
+			// Use GenBashCompletionV2 (with descriptions) so the output is
+			// byte-for-byte identical to the `completion bash` subcommand.
 			if legacyBashCompletion {
-				if err := cmd.Root().GenBashCompletion(os.Stdout); err != nil {
+				if err := cmd.Root().GenBashCompletionV2(os.Stdout, true); err != nil {
 					return err
 				}
 				os.Exit(0)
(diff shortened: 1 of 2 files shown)


ipaqsa (approved):

ldmonster (approved):

deckhouse-BOaTswain:
Backport failed ⚠️.

PR #20939, opened 2026-06-26
[registrypackages] update d8-ca-updater CA certificates

mary-ppv:
## Description

Updated the local Mozilla certdata.txt snapshot used by d8-ca-updater to generate trusted CA certificates.

Also updated certdata2pem.py so that trusted expired certificates are treated as a build error instead of only being printed as a warning. The script now uses not_valid_after_utc instead of the deprecated not_valid_after property.

The change does not restart or directly affect running cluster components. It affects the d8-ca-updater registrypackage artifact used to install CA certificates on nodes.

The previous Mozilla CA snapshot contained the trusted root certificate Baltimore CyberTrust Root, which expired on May 12, 2025.

During artifact build, certdata2pem.py detected the expired trusted certificate and printed:

```
Trusted but expired certificate found: "Baltimore CyberTrust Root"
```

However, the script exited successfully, so the build continued and an expired trusted CA certificate could be included into the generated d8-ca-updater artifact.

This PR updates the Mozilla CA snapshot and makes the certificate generation step fail if a trusted expired certificate is detected. This prevents stale CA snapshots from silently producing artifacts with expired trusted root certificates.

The fix should be backported to patch releases because it prevents expired trusted CA certificates from being included into d8-ca-updater artifacts.

Without this change, the problem can be missed unless someone notices the warning in build logs.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: registrypackages
type: fix 
summary: Updated the Mozilla CA snapshot used by d8-ca-updater and made the build fail on trusted expired certificates.
impact_level: default 
```


deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:fix/d8-ca-updater` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/28241557137) in 38m27s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@51.250.11.96`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 37m19s.

</details>

090809 (approved):

PR #20884, opened 2026-06-25
Remove operator-prometheus, prometheus and loki modules

lazovskiy:
## Description
`operator-prometheus`, `prometheus` and `loki` modules go external;

## Why do we need it, and what problem does it solve?
This PR removes in-tree modules. These modules become external, stored in separate repositories, and have their own development and release pace. This decoupling should dramatically reduce release cycle duration.

## Checklist
- [ ] The code is covered by unit tests.
- [x] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: deckhouse
type: chore
summary: Removed the in-tree operator-prometheus, prometheus, and loki modules; they are now sourced externally.
impact_level: default
```

PR #20961, opened 2026-06-29
[docs] Migrate observability modules to external in doc metadata files

z9r5:
## Description

Migrate observability modules to external in doc metadata files:
- Remove loki, prometheus, and operator-prometheus entries from the embedded modules tags registry as they are no longer bundled internally
- Register the same three modules in the external modules catalogue with full edition availability, external flag, path routing, and preserved tag metadata

Ref: https://github.com/deckhouse/deckhouse/pull/20884


z9r5:
@lazovskiy Hi, get my changes please

alexey-gavrilov-flant:
@z9r5 we need to make a rebase

PR #21031, opened 2026-06-30
[ci] AV scan only changed images in PR

MixBars:
## Description
This PR updates PR-time image security scan selection.
The changed image detection now uses the current Build FE report and selects final werf images whose Commit belongs to the current PR commit range.
The antivirus scan jobs now receive this changed_compact list through ONLY_IMAGES. If the list is empty, Dr.Web and Kaspersky scans are skipped with an explicit log message:
`No rebuilt images in this PR. Nothing to scan.`
Scan artifacts are not uploaded for skipped antivirus scans. The antivirus status label logic was also updated so skipped scans caused by an empty image diff remove both success and failed antivirus labels instead of marking the scan as successful.
Before this change, PR image scans could fall back to scanning all images when the changed image diff was empty. This wastes runner time and scanner capacity.
Using Rebuilt is also not reliable enough for PR diff detection. A module image can be rebuilt in an earlier PR run and then reused from cache in a later run. In that case Rebuilt can be false even though the image still belongs to the PR changes.
This PR uses the image Commit from the Build FE report instead. If a final image points to a commit from the current PR range, it is selected for scanning. This keeps scan selection stable across reruns and cached builds.
This is needed in a patch release only if we want to reduce unnecessary PR-time antivirus scans and avoid full antivirus scans when no module images changed.
The change affects CI scan selection only. It does not change Deckhouse runtime behavior.
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: ci
type: fix
summary: Fix diff images logic and add skip PR antivirus image scans when no module images were changed.
impact: PR antivirus scans now use the changed image list from the Build FE report. If no module images were changed in a PR, Dr.Web and Kaspersky scans are skipped with a clear log message instead of running a full scan.
impact_level: low
```

### DIFF
--- a/.github/ci_templates/antivirus_tests.yml
+++ b/.github/ci_templates/antivirus_tests.yml
@@ -1,19 +1,26 @@
 {!{ define "av_semver_regexp" }!}^v?[0-9]+.[0-9]+.[0-9]+${!{ end }!}
 
-{!{ define "av_scan_step" }!}
+  {!{ define "av_scan_step" }!}
 - name: Run {!{ .container_name }!} scan
   id: scan
   {!{- if .only_images_expr }!}
   env:
     # `ONLY_IMAGES` comes from needs.changed_images.outputs.changed_compact
-    # (JSON array of "<module>.<image>" keys). Wired here for PR runs of both
-    # drweb-scan and kaspersky-scan when their templates are called with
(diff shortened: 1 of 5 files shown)


MixBars:
Changes in py script is reason why scan_images fails here

Taior (approved):

Nikolay1224 (approved):

PR #21066, opened 2026-07-01
[user-authz] fix webhook discovery cache to use local apiserver endpoint

AlwxSin:
## Description

Fixes the `user-authz-webhook` internal discovery cache client to reach kube-apiserver via the node-local endpoint (`KUBERNETES_SERVICE_HOST`/`KUBERNETES_SERVICE_PORT`) instead of always resolving the `kubernetes.default` DNS name.

This does not restart or affect any critical cluster component other than the `user-authz-webhook` DaemonSet itself (rolling restart on deploy, as with any image/config change to that DaemonSet).

#17580 made the webhook's Kubernetes client use the node-local apiserver endpoint (`status.hostIP:6443`) instead of the `kubernetes` ClusterIP service, specifically to keep the authorization webhook stable when cluster DNS/ClusterIP routing is degraded. However, that fix only covered the `rest.InClusterConfig()` path (namespace informer). The webhook's own `internal/cache` package — which handles RBAC discovery cache refresh **and backs the `/healthz` endpoint used by the container's livenessProbe** — has its own hand-rolled HTTP client with a hardcoded `kubernetesAPIAddress = "https://kubernetes.default"` constant, untouched by #17580.

As a result, when cluster DNS briefly misbehaves (e.g. node-local-dns/CoreDNS SERVFAIL during a network/CNI hiccup), the webhook's liveness probe fails, kubelet restarts the container, and it stops listening on `127.0.0.1:40443`. Since kube-apiserver's authorization webhook is configured with `failurePolicy: Deny`, this makes kube-apiserver deny all requests — including from `kubeadm:cluster-admins` — until the webhook comes back up, producing a full API outage/CrashLoopBackOff on all control-plane nodes.

This was reproduced on a production cluster (DKP v1.75.11): `kube-apiserver` logs showed `dial tcp 127.0.0.1:40443: connect: connection refused`, while `user-authz-webhook` logs showed `lookup kubernetes.default on <dns-ip>:53: server misbehaving: exceeded retry limit`. The incident self-resolved once DNS recovered, ~40 minutes later.

The fix adds `resolveKubernetesAPIAddress()`, mirroring the same env-var convention `rest.InClusterConfig()` already uses, so the discovery cache client (and its liveness check) hits the same node-local apiserver endpoint the DaemonSet already points the rest of the client at. Falls back to the old DNS-based address if the env vars aren't set (e.g. running outside this DaemonSet's env).

Yes — this closes a control-plane availability gap where a transient DNS blip can escalate into a full multi-master API outage via the fail-closed authorization webhook. Low-risk, isolated to the `user-authz-webhook` image.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: fix
summary: User-authz-webhook now uses the node-local kube-apiserver endpoint for its discovery cache and liveness check, instead of resolving the `kubernetes.default` DNS name.
impact: |
  This release fixes an issue where transient cluster DNS failures could cause the user-authz-webhook to restart, temporarily denying all Kubernetes API requests until DNS recovered.
impact_level: high
```

### DIFF
--- a/ee/be/modules/140-user-authz/images/webhook/src/internal/cache/cache.go
+++ b/ee/be/modules/140-user-authz/images/webhook/src/internal/cache/cache.go
@@ -6,6 +6,7 @@ Licensed under the Deckhouse Platform Enterprise Edition (EE) license. See https
 package cache
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
@@ -34,7 +35,14 @@ const (
 	tokenPath = saPath + "token"
 	apiV1Path = "/api/v1"
 
+	// kubernetesAPIAddress is the fallback used when the caller doesn't provide one
+	// (e.g. running outside the user-authz-webhook DaemonSet's env).
 	kubernetesAPIAddress = "https://kubernetes.default"
+
+	// requestTimeout bounds a single discovery/healthz request so it fails fast and
+	// deterministically well within the container's livenessProbe budget, instead of
+	// relying on the much larger transport-level dial/handshake timeouts.
+	requestTimeout = 2 * time.Second
 )
 
 type Cache interface {
@@ -109,7 +117,16 @@ type NamespacedDiscoveryCache struct {
 	kubernetesAPIAddress string
 }
 
-func NewNamespacedDiscoveryCache(logger *log.Logger) *NamespacedDiscoveryCache {
+// NewNamespacedDiscoveryCache builds a cache client talking to apiAddress (e.g.
+// config.Host from the same rest.InClusterConfig() the caller already builds for its
+// own clientset), so this client and the caller's are provably pointed at the same
+// apiserver endpoint. Falls back to the "kubernetes.default" DNS name if apiAddress
+// is empty.
+func NewNamespacedDiscoveryCache(logger *log.Logger, apiAddress string) *NamespacedDiscoveryCache {
+	if apiAddress == "" {
+		apiAddress = kubernetesAPIAddress
+	}
+
 	c := &NamespacedDiscoveryCache{
 		logger:            logger,
 		data:              make(map[string]*namespacedCacheEntry),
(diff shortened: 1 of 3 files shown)


Jabejixo (approved):
Two things I'd add before merge:

**1. Bound the cache requests with a context/timeout (this is the part that actually stops the crashloop).**
The cache requests use `http.NewRequest` with no per-request deadline. The client only has `DialTimeout: 30s` + `TLSHandshakeTimeout: 10s`, so `/healthz → Check()` can block ~30s while the liveness probe is capped at 1s. This PR removes the *DNS* trigger, but the latent failure mode ("healthz hangs → probe fails → restart → apiserver fail-closed denies") stays: if the node-local apiserver is briefly slow/unreachable, the probe still won't fit in 1s. Add `context.WithTimeout` + `http.NewRequestWithContext` so `/healthz` fails fast and deterministically:

```go
func (c *NamespacedDiscoveryCache) Check() error {
	return Retry(func() (bool, error) {
		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
		defer cancel()

		req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.kubernetesAPIAddress+"/version", nil)
		if err != nil {
			return false, fmt.Errorf("check Kubernetes API create request: %w", err)
		}
		if err := c.execRequest(req, "check API", nil); err != nil {
			return true, err
		}
		return false, nil
	})
}
```

Apply the same `NewRequestWithContext` + timeout to the renew/discovery calls (`:197`, `:216`, `:251`, `:324`) — a small `newReq(ctx, path)` helper avoids repetition.

**2. Inject the API address instead of reading env in the constructor.**
`resolveKubernetesAPIAddress()` calls `os.Getenv` inside the constructor — a hidden dependency on global env, and it duplicates the env parsing that `rest.InClusterConfig()` already does (which is exactly what the informer uses in `NewServer`). Passing the address in makes the cache and the informer provably use the *same* endpoint, keeps it DRY, and drops the `t.Setenv` coupling from the test:

```go
// server.go — NewServer already has the config
config, err := rest.InClusterConfig()
...
c := discoverycache.NewNamespacedDiscoveryCache(logger, config.Host) // config.Host == https://<host>:<port>

// cache.go
func NewNamespacedDiscoveryCache(logger *log.Logger, apiAddress string) *NamespacedDiscoveryCache {
	if apiAddress == "" {
		apiAddress = kubernetesAPIAddress
	}
	c := &NamespacedDiscoveryCache{
		...
		kubernetesAPIAddress: apiAddress,
	}
	c.initClient()
	return c
}
```

This lets you drop `resolveKubernetesAPIAddress`; the test then just passes an address directly (no env). Optional nice-to-have: add an IPv6 case to lock in the `JoinHostPort` bracketing (`fd00::1` → `https://[fd00::1]:6443`).

nabokihms (approved):

deckhouse-BOaTswain:
Cherry pick PR [21080](https://github.com/deckhouse/deckhouse/pull/21080) to the branch [release-1.75](https://github.com/deckhouse/deckhouse/tree/release-1.75) successful!

deckhouse-BOaTswain:
Cherry pick PR [21081](https://github.com/deckhouse/deckhouse/pull/21081) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #20949, opened 2026-06-26
[deckhouse] harden migrated modules check

ldmonster:
## Description

When a module is migrating from embedded (shipped inside the Deckhouse image) to an external `ModuleSource`, this change downloads it onto the filesystem **ahead of the upgrade** that drops the embedded copy, instead of downloading it afterwards.

Mechanics (the single invariant: files are always materialized on the filesystem, but the module is *activated* — symlink for the symlink backend, mount for the erofs backend — only when no embedded copy of the same name is present on disk):

1. **A ModuleRelease is created for an embedded module** once it appears in an external source. The release source is chosen as follows: if `ModuleConfig.spec.source` selects a source that offers the module, that one is used; otherwise, if exactly one source offers it, that one is used. If the source is undecided — several sources offer it and none is selected, **or** the configured `.spec.source` no longer offers it (a stale/mistyped `ModuleConfig`) — the `ModuleAtConflict` alert is fired and no release is created (the running embedded module is left untouched).
2. **The ModuleRelease reaches `Deployed` and is downloaded to the filesystem without being activated** (no symlink/mount), so the embedded copy keeps serving the module and its hooks are not disabled.
3. **After the upgrade drops the embedded copy**, the moduleloader restore activates the already-staged module (symlink/mount), so it is mounted immediately without a registry download on the critical path.
4. **The `migratedModules` DeckhouseRelease gate is hardened**: an enabled migrated module must now have a `Deployed` ModuleRelease (i.e. be physically on disk) before the upgrade is allowed, instead of merely being listed as available in a `ModuleSource`.

`Stage`/`StageFromRegistry` are implemented for both the symlink and erofs backends.

### Alerts

Three alerts surface the states above so an operator always has a diagnostic signal instead of a silently-stuck upgrade:

| Alert | Metric | Severity | When it fires | Operator action |
| --- | --- | --- | --- | --- |
| `ModuleAtConflict` | `d8_module_at_conflict` | 4 | The release source for an embedded module is undecided: several sources offer it and none is selected via `ModuleConfig`, or the configured `.spec.source` does not (or no longer does) offer it. No release is pre-staged; the embedded module keeps running. | Set/fix `ModuleConfig.spec.source` to the correct source. |
[...]

PR #21077, opened 2026-07-01
[deckhouse-controller] 1.76 harden migrated modules check

ldmonster:
## Description

Backport: https://github.com/deckhouse/deckhouse/pull/20949

When a module is migrating from embedded (shipped inside the Deckhouse image) to an external `ModuleSource`, this change downloads it onto the filesystem **ahead of the upgrade** that drops the embedded copy, instead of downloading it afterwards.

Mechanics (the single invariant: files are always materialized on the filesystem, but the module is *activated* — symlink for the symlink backend, mount for the erofs backend — only when no embedded copy of the same name is present on disk):

1. **A ModuleRelease is created for an embedded module** once it appears in an external source. The release source is chosen as follows: if `ModuleConfig.spec.source` selects a source that offers the module, that one is used; otherwise, if exactly one source offers it, that one is used. If the source is undecided — several sources offer it and none is selected, **or** the configured `.spec.source` no longer offers it (a stale/mistyped `ModuleConfig`) — the `ModuleAtConflict` alert is fired and no release is created (the running embedded module is left untouched).
2. **The ModuleRelease reaches `Deployed` and is downloaded to the filesystem without being activated** (no symlink/mount), so the embedded copy keeps serving the module and its hooks are not disabled.
3. **After the upgrade drops the embedded copy**, the moduleloader restore activates the already-staged module (symlink/mount), so it is mounted immediately without a registry download on the critical path.
4. **The `migratedModules` DeckhouseRelease gate is hardened**: an enabled migrated module must now have a `Deployed` ModuleRelease (i.e. be physically on disk) before the upgrade is allowed, instead of merely being listed as available in a `ModuleSource`.

`Stage`/`StageFromRegistry` are implemented for both the symlink and erofs backends.

### Alerts

Three alerts surface the states above so an operator always has a diagnostic signal instead of a silently-stuck upgrade:

| Alert | Metric | Severity | When it fires | Operator action |
| --- | --- | --- | --- | --- |
| `ModuleAtConflict` | `d8_module_at_conflict` | 4 | The release source for an embedded module is undecided: several sources offer it and none is selected via `ModuleConfig`, or the configured `.spec.source` does not (or no longer does) offer it. No release is pre-staged; the embedded module keeps running. | Set/fix `ModuleConfig.spec.source` to the correct source. |
| `DeckhouseMigratedModuleNotDownloaded` | `d8_migrated_module_not_downloaded` | 6 | **(new)** At the upgrade gate, a `ModuleSource` offers the migrated module but it has not been pre-downloaded yet (no `Deployed` ModuleRelease). The DeckhouseRelease is locked until the module is on disk. | Usually transient — wait for the source controller to pull the module. If it persists, check the update policy / `ModuleSource` config and registry connectivity. |
| `DeckhouseMigratedModuleNotFound` | `d8_migrated_module_not_found` | 6 | At the upgrade gate, **no** `ModuleSource` offers the migrated module at all (terminal). The DeckhouseRelease is locked. | Publish the module in a `ModuleSource`, or fix the module's configured source, before upgrading. |

Previously a single `DeckhouseMigratedModuleNotFound` alert covered both the "no source offers it" and the "offered but not downloaded yet" cases. They are now split because they require different operator actions — the not-downloaded case is transient and self-resolves once the source controller pre-stages the module, while the not-found case needs manual intervention.

Influence on critical cluster components: no extra restarts are introduced. For migrating modules (e.g. `ingress-nginx`) this *reduces* the risk of a restart/outage window, because the module is on disk before the upgrade rather than being downloaded during it.

Today a migrating module is only downloaded **after** Deckhouse is upgraded and the embedded copy is removed. The `migratedModules` requirement check only verifies that the module is *listed* in a `ModuleSource`, not that it is actually on disk. If, right after the upgrade, the registry is unavailable or slow, the module is not yet on the filesystem when the reconcile queue reaches it — causing a temporary outage of the module (for `ingress-nginx` this means real ingress downtime).

This change pre-downloads the module while the embedded copy is still running and blocks the upgrade until the module is physically present on disk, eliminating the download-during-upgrade race.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: harden migrated modules check
impact_level: default
```

### DIFF
--- a/deckhouse-controller/internal/metrics/metrics.go
+++ b/deckhouse-controller/internal/metrics/metrics.go
@@ -33,6 +33,7 @@ const (
 	// ============================================================================
 	// Migration and experimental module tracking
 	MigratedModuleNotFoundMetricName        = "d8_migrated_module_not_found"
+	MigratedModuleNotDownloadedMetricName   = "d8_migrated_module_not_downloaded"
 	ExperimentalModulesAreAllowedMetricName = "is_experimental_modules_allowed"
 	ExperimentalModuleIsEnabled             = "is_experimental_module_enabled"
 	DeprecatedModuleIsEnabled               = "is_deprecated_module_enabled"
@@ -78,10 +79,11 @@ const (
 // Metric Groups
 // ============================================================================
 const (
-	MigratedModuleNotFoundGroup = "migrated_module_not_found"
-	D8Updating                  = "d8_updating"
-	D8ModuleUpdatingGroup       = "d8_module_updating_group"
-	ModuleReleaseGroup          = "module_release_group"
+	MigratedModuleNotFoundGroup      = "migrated_module_not_found"
+	MigratedModuleNotDownloadedGroup = "migrated_module_not_downloaded"
+	D8Updating                       = "d8_updating"
+	D8ModuleUpdatingGroup            = "d8_module_updating_group"
+	ModuleReleaseGroup               = "module_release_group"
 )
 
 // Group templates for dynamic metric names using fmt.Sprintf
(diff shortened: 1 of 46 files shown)


deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:feat/1-76-hardened-migrated-modules-check` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/28536344836) in 33m59s.



GCP-WithoutNAT-Containerd-1.34 - Connection string: `ssh  user@35.246.221.126`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.34` succeeded in 32m56s.

</details>

PR #21104, opened 2026-07-02
[deckhouse] collect application endpoint URLs

ldmonster:
## Description

Adds a new `status.urls` field to the `Application` custom resource, populated from Ingress resources of the application chart.

If a chart developer annotates an Ingress with `packages.deckhouse.io/application-endpoint` (any value except `"false"`), the controller builds endpoint entries from its `spec.rules` and `spec.tls`:

- each entry is an object `{url, description}`;
- one entry per `host`/`path` pair (`scheme://host/path`); a rule without paths produces `scheme://host/`;
- the scheme is `https` when the host is listed in `spec.tls[].hosts`, `http` otherwise;
- `description` is taken from the annotation value (`"true"` means no description);
- the resulting list is deduplicated and sorted.

Example:

```yaml
metadata:
  annotations:
    packages.deckhouse.io/application-endpoint: "Web UI"
```

```yaml
status:
  urls:
  - url: https://app.example.com/ui
    description: Web UI
```

Implementation details:

- URLs are extracted in the nelm service right after a successful chart render (`internal/packages/nelm/endpoints.go`), before the `shouldUpgrade` check, so they are refreshed even when the Helm upgrade itself is skipped.
- The internal package status service stores the URLs and notifies listeners only when the list actually changes.
- The Application status service copies the URLs to `Application.status.urls` together with `currentVersion` and `lastAppliedConfiguration`, i.e. only after manifests are successfully applied. Removing the annotation clears the URLs on the next successful apply.

No critical cluster components are affected; the change only extends the Application status.

The UI (and any API client) needs a way to show entry points of an installed application. Currently there is no machine-readable place where application endpoints are exposed. With this change, a chart author marks the relevant Ingress with a single annotation, and the platform publishes ready-to-use URLs in `Application.status.urls` — a single source of truth for the frontend, with correct scheme detection based on TLS configuration.

Not necessarily.

### Manual testing

Installed an application whose chart ships an Ingress annotated with `packages.deckhouse.io/application-endpoint: "A-K test echo endpoint"` (single host, path `/`, no TLS) into a cluster running this change. After a successful install the application is `Ready` and `Application.status.urls` is populated from the rendered Ingress.

```
$ k get application
NAME           PACKAGE    VERSION   STATE   MESSAGE   AGE
a-k-test-app   a-k-test   v0.0.3    Ready             52m
```

```yaml
kind: Application
metadata:
  name: a-k-test-app
  namespace: default
  ...
spec:
  packageName: a-k-test
  packageRepositoryName: test
  packageVersion: v0.0.3
status:
  conditions:
  - reason: Installed
    status: "True"
    type: Installed
  - reason: Ready
    status: "True"
    type: Ready
  ...
  currentVersion:
    version: 0.0.3
  summary:
    state: Ready
  urls:
  - description: A-K test echo endpoint
    url: http://a-k-test-app.example.com/
```

The `url` is built by the controller from the rule host and path, the `http` scheme reflects the absent `spec.tls`, and `description` is the annotation value verbatim.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Application status now exposes endpoints (`status.urls` with `url` and `description`) collected from chart Ingresses annotated with `packages.deckhouse.io/application-endpoint`.
```


z9r5 (approved):

z9r5:
@ldmonster I've made some minor changes to CRD and have updates translation.

ipaqsa (approved):

PR #21132, opened 2026-07-03
[node-manager] Fix static node bootstrap hang caused by empty cluster UUID

nabokihms:
## Description

Fixes a race that makes Static / CAPS-adopted nodes hang ~20 minutes during bootstrap.

On a fresh cluster, `node-manager` could render the node bootstrap bundle with an empty `global.discovery.clusterUUID`. With an empty UUID, the on-node `rpp-get` installer queries `registry-packages-proxy` at the prefix-less `/rpp-get` path; the proxy expects `/<clusterUUID>/rpp-get` and answers `404`, so bashible never installs `rpp-get` and the **first** bootstrap attempt stalls until `StaticInstanceBootstrapTimeout` (~20m) fires → the StaticInstance goes to `Cleaning` → a replacement Machine retries (by then the UUID is populated, so it succeeds). Net effect: ~20 wasted minutes per affected node, intermittently.

No restarts of critical components. The dhctl change affects the bootstrap `Create Manifests` flow only.

Root cause is a startup ordering race in `dhctl`. During `Create Manifests`, every install manifest **except** the `d8-system` Namespace was applied in a single parallel fan-out — so the `deckhouse` Deployment could be created **before its own prerequisites**: the registry pull secret, RBAC/ServiceAccount, the cluster-configuration secrets, and the `d8-cluster-uuid` ConfigMap. The `cluster_uuid` global hook only sets `global.discovery.clusterUUID` after it observes that ConfigMap, so in the window before it existed, `_bootstrap.tpl` rendered the bundle with `clusterUUID | default ""` — a broken bundle that the on-node `rpp-get` / `registry-packages-proxy` path rejects.

Fixes:
- **dhctl** — split the manifest tasks into ordered phases instead of one fan-out: `Namespace` → **all prerequisites** (RBAC, ServiceAccount, registry pull secret, cluster-configuration secrets, `d8-cluster-uuid`, queue lock, ...) applied concurrently → **deckhouse Deployment** → **ModuleConfigs**. The controller pod now always starts with its prerequisites in place (this also removes the latent races for the pull secret, RBAC and cluster-config secrets, not just the UUID).
- **node-manager** — `required` the cluster UUID in `_bootstrap.tpl` instead of `default ""`, so a bootstrap bundle is never emitted with an empty UUID (fail-fast; node-manager re-renders once discovery completes).

The two changes are complementary: dhctl removes the race at the source, the template guard prevents shipping a broken bundle for any other reason the UUID could be empty.

Worth backporting — it removes an intermittent ~20-minute stall (and the associated `Cleaning`/retry churn) from every static/hybrid cluster bootstrap and static-node adoption.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Apply all deckhouse prerequisites (registry pull secret, RBAC, cluster-configuration secrets, d8-cluster-uuid, ...) before the deckhouse Deployment during bootstrap, so the controller never starts ahead of resources it and its hooks read.
```
```changes
section: node-manager
type: fix
summary: Require a non-empty cluster UUID when rendering the node bootstrap script, preventing static nodes from stalling ~20m on a registry-packages-proxy 404.
```

### DIFF
--- a/dhctl/pkg/kubernetes/actions/deckhouse/install.go
+++ b/dhctl/pkg/kubernetes/actions/deckhouse/install.go
@@ -661,41 +661,39 @@ func CreateDeckhouseManifests(
 		tasks = append(tasks, *lockCmTask)
 	}
 
+	// The deckhouse controller Deployment is kept out of the prerequisite task
+	// list and applied only after all of them exist. The pod and its hooks read
+	// those resources on startup (registry pull secret, RBAC/ServiceAccount,
+	// cluster-configuration secrets, d8-cluster-uuid, ...); racing the Deployment
+	// against them can yield a pod that cannot pull images, or hooks that observe
+	// half-initialized state — e.g. an empty global.discovery.clusterUUID, which
+	// makes node-manager render node bootstrap scripts without the cluster-UUID
+	// prefix for registry-packages-proxy and hangs CAPS-adopted nodes ~20m.
 	deploymentTask, err := controllerDeploymentTask(kubeCl, cfg)
 	if err != nil {
 		return nil, err
 	}
-	tasks = append(tasks, deploymentTask)
 
 	result := &ManifestsResult{}
 
+	// ModuleConfigs are applied last: their CRD is installed by the running
+	// deckhouse pod, so these retry-wait for it to appear.
+	var moduleConfigTasks []actions.ManifestTask
 	if len(cfg.ModuleConfigs) > 0 {
 		prepareModuleConfig(ctx, cfg.ModuleConfigs[0], result)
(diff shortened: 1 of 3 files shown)


090809 (approved):

deckhouse-BOaTswain:
:green_circle: `e2e: Static` for `deckhouse:fix/static-node-bootstrap-cluster-uuid` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/28666916357) in 48m16s.



Static-Static-Containerd-1.34 - Connection string: `ssh -J astra@87.228.63.189 astra@192.168.199.150`


:green_circle: `e2e: Static, Containerd, Kubernetes 1.34` succeeded in 47m9s.

</details>

PR #21147, opened 2026-07-06
Virtual approver

AmazinMax:
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/operations-approver/controller.go
```diff
@@ -17,26 +17,16 @@ limitations under the License.
 package operationsapprover
```
sprait: If we use `controllers/virtual-control-plane-approver`  then I suggest using the same name scheme here as well:

` controllers/operations-approver/`  -> ` controllers/control-plane-approver/`

sprait (approved):
minor suggestion

trofimovdals (approved):

PR #20615, opened 2026-06-11
User authz/rbacv2

Jabejixo:
## Description
Umbrella branch for the full RBACv2 + multitenancy delivery. For review it is split into four focused PRs that together reassemble to this branch byte-for-byte:
- **#21157 — multitenancy-manager (160):** v1alpha3 Project, project CRDs and controller, schema-based ProjectTemplate, multi-namespace fan-out.
- **#21161 — permission-browser apiserver (EE 140):** WhoCan and RBACv2-aware AccessibleNamespaces.
- **#21162 — webhooks (140):** RBACv2 admission enforcement and the manage→use fan-out hooks.
- **#21165 — role-model migration (MERGE LAST):** the role/capability rename, the core framework and CI contract test, and the one-release compatibility layer.
Recommended merge order: #21157, #21161, #21162 in any order, then **#21165 last**. This branch is kept as the integration reference; the reviewable changes live in the child PRs.
## Why do we need it, and what problem does it solve?
Full RBACv2 delivery: a unified role/capability model, project multitenancy as first-class RBAC scopes, admission-time enforcement of the invariants the API server cannot express, reverse-RBAC introspection tooling, and a one-release backward-compatibility layer so existing clusters do not lose access on upgrade. Per-area detail is in the child PRs.

## Why do we need it in the patch release (if we do)?
Not necessarily.

## Checklist
- [x] The code is covered by unit tests.
- [x] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries
```changes
section: user-authz
type: feature
summary: RBACv2 and project multitenancy — unified role/capability model, project CRDs and controller, admission webhooks, permission-browser, with one-release backward compatibility.
impact_level: high
impact: |
  The RBACv2 role model replaces the legacy d8:manage/d8:use roles (old names kept for one release via deprecated aliases). Migrate bindings to the new names before the next release. Policies, Pod Security Standard, features and cluster-resource grants now apply to every namespace of a project.

PR #21157, opened 2026-07-06
rbacv2: multitenancy-manager (160)

Jabejixo:

## Description

Reworks the `multitenancy-manager` module (160) to implement the multitenancy half of RBACv2. Projects become first-class RBAC scopes with their own role ladder, additional namespaces, and project-wide bindings, replacing the previous parameter-driven model.

What changes:

- **Project API `v1alpha3`** (served + storage; `v1alpha1`/`v1alpha2` kept via conversion webhooks). `administrators` and `quota` are now first-class `spec` fields (previously buried in `spec.parameters`); `status.usage`, `status.namespaces` (Main/Additional), and structured conditions are exposed. The controller materializes a `ResourceQuota/d8-project-quota` and a controller-managed `ProjectRoleBinding/d8-administrators` (role `d8:project:admin`) from these fields.
- **New CRDs:** `ProjectNamespace` (namespaced, additional project namespaces `<project>-<name>`), `ProjectRoleBinding` (grants a role across every namespace of a project), and the cluster-scoped `ClusterProjectRoleBinding` (grants a role across all projects; `status.boundProjects`).
- **Project-scope RBACv2 roles/capabilities** (`d8:project:{viewer,user,manager,admin,superadmin}` + `project-capability:*`), aggregated cumulatively so `admin ⊇ manager ⊇ viewer` — including read/manage of `ProjectNamespace`/`ProjectRoleBinding`.
- **Schema-based `ProjectTemplate` (v1alpha2)** rendered by a native Go renderer (Helm is used only for release management): structured fields (`podSecurityStandard`, `networkPolicy`, `nodeSelector`, `tolerations`, `features`, `runtimeAudit`, `logShipping`, `namespaceMetadata`, `allowedUIDs/GIDs`, `grantPolicies`) with `{fromParam}` parametrization. The template admission webhook validates every `fromParam` reference for both existence and type compatibility.
- **Multi-namespace fan-out:** cluster-scoped policies select by the project label, namespaced template objects (NetworkPolicy, PodLoggingConfig) are rendered into every project namespace, and additional namespaces inherit the main namespace's policy/feature/grant labels (Pod Security Standard, monitoring, vulnerability scanning, project-template) — including re-sync when those labels change after the namespace already exists.
- **Helm release-name decoupling:** project names up to 61 characters are supported by hashing names longer than Helm's 53-character release-name limit.

No restarts of ingress-controllers, control-plane or Prometheus. Only the `multitenancy-manager` controller and its CRDs are affected.

The legacy multitenancy model configured everything through `Project.spec.parameters` and applied policies/features only to the project's main namespace. This PR:

- makes projects proper RBAC scopes with cumulative `d8:project:*` roles and dedicated binding CRDs;
- applies Pod Security Standard, features, network policy and cluster-resource grants to **all** namespaces of a project, not just the main one;
- fixes a latent bug where `d8:project:admin`/`manager` could not read/manage `ProjectNamespace` (the role ladder was not cumulative);
- moves `{fromParam}` validation to admission time (existence + type), so a mistyped binding is rejected on the template instead of silently failing when every project on it renders;
- removes the object-quota CRD that was dropped from the module, keeping only the declarative grant-policy path.

Not necessarily.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: feature
summary: RBACv2 projects — first-class v1alpha3 Project (administrators/quota/status), ProjectNamespace/ProjectRoleBinding/ClusterProjectRoleBinding CRDs, schema-based ProjectTemplate with a native renderer, and per-namespace resource fan-out.
impact_level: high
impact: |
  The multitenancy-manager controller and its CRDs are upgraded to the RBACv2 model and Projects are reconciled into the new v1alpha3 shape (older API versions remain available via conversion). Policies, Pod Security Standard, features and cluster-resource grants now apply to every namespace of a project.

### CODE REVIEW

File: modules/160-multitenancy-manager/webhooks/conversion/projects [hunk 1]
```diff
@@ -70,9 +80,92 @@ function __on_conversion::v1alpha2_to_v1alpha1() {
...
+
+# v1alpha2 -> v1alpha3 lifts administrators and resourceQuota out of spec.parameters into the
+# dedicated top-level spec.administrators and spec.quota fields.
+#   parameters.administrators[].subject -> administrators[].kind
+#   parameters.resourceQuota.{requests,limits}.* -> quota."requests.*"/"limits.*"
+#   parameters.resourceQuota.<other>       -> quota."<other>"
+# spec.quota is a flat ResourceQuota.spec.hard map, so it also holds keys that have no requests/limits
+# nesting ("pods", "count/deployments.apps"). Those are carried over verbatim; nested objects are
+# skipped because they have no flat representation.
+function __on_conversion::v1alpha2_to_v1alpha3() {
```
AlwxSin, line 99: 🟡 **The conversion webhooks were rewritten and still have no tests**

Two of the fix commits change jq that decides what survives a version round-trip — quota keys without a `requests.`/`limits.` prefix here, and the structured half of a ProjectTemplate spec in `webhooks/conversion/projecttemplates`. Neither added a test, and there is no test file for these two hooks anywhere in the module: the only tests under `modules/160-multitenancy-manager/` are `hooks/*_test.go` and `openapi/openapi-case-tests.yaml`.

The original review note was precisely "conversions written last, without round-trip tests". The logic improved; the regression guard did not appear. A table of `(v1alpha3 quota) → down → up → same object` cases would be a few dozen lines and would have caught both of the bugs that were just fixed.

Two edges I checked by hand and could not close from the code alone — which is exactly the sort of claim a test should be making instead of a reviewer:

- `.spec.quota` with a key of exactly `requests` or `limits` (no dot) takes the `else` branch of the down-conversion and lands as `resourceQuota.requests = "<string>"`. The up-conversion then feeds that string to `to_entries`, which errors and fails the whole conversion for that object.
- A nested non-`requests`/`limits` object under `resourceQuota` is silently dropped on the way up (the `(.value | type) != "object"` filter). Documented in the comment, but nothing pins it.

modules/160-multitenancy-manager/webhooks/conversion/projecttemplates
```diff
@@ -0,0 +1,103 @@
...
+    crdName: projecttemplates.deckhouse.io
+    conversions:
+    - fromVersion: deckhouse.io/v1alpha2
+      toVersion: deckhouse.io/v1alpha1
+EOF
+}
+
+# Annotation carrying the structured-only part of the spec across a v1alpha1 round-trip. See the
+# down-conversion below for why it exists.
+STRUCTURED_SPEC_ANNOTATION="conversion.deckhouse.io/v1alpha2-spec"
```
AlwxSin: 🟡 **The round-trip annotation is a sharp fix for a problem `served: false` removes**

Making v1alpha1 lossless by stashing the structured spec in `conversion.deckhouse.io/v1alpha2-spec` works, but it buys three problems:

1. **The annotation is part of the v1alpha1 view.** `kubectl get projecttemplate -o yaml` at v1alpha1 now shows the whole structured spec as a JSON blob in metadata, contrary to the down-conversion comment's "the annotation is not part of the spec".
2. **It is unbounded.** `spec.resources` plus `spec.grantPolicies` of a real template, JSON-encoded, and then kubectl's own `last-applied-configuration` after a `get -o yaml | apply -f -`, both count against the 256 KB metadata limit.
3. **A stale copy resurrects an old spec.** An object written back through v1alpha2 while still carrying the annotation keeps it in storage; the next v1alpha1 write then merges that stale structured spec over the live one via `.spec = ($structured + (.spec // {}))`.

`Project` v1alpha1 is already `served: false` in this same PR (`crds/projects.yaml:19-22`). Doing the same for `ProjectTemplate` v1alpha1 in `crds/projecttemplate.yaml:19-22` — it is already `deprecated: true` — deletes the annotation, both conversion directions and all three problems at once. If v1alpha1 must stay served, documenting it as lossy (the normal Kubernetes answer for an old version that cannot represent new fields) is still smaller than this.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/webhook/projectrolebinding/hook.go
```diff
@@ -0,0 +1,86 @@
...
+			return admission.Denied("ProjectRoleBinding cannot be created in a virtual project namespace")
+		}
+	}
+
+	return rolebindingwebhook.Validate(ctx, v.client, req, rolebindingwebhook.Input{
+		RoleRefKind: prb.Spec.RoleRef.Kind,
+		RoleRefName: prb.Spec.RoleRef.Name,
+		Subjects:    prb.Spec.Subjects,
+		Namespace:   req.Namespace,
+		ManagedBy:   prb.Labels[v1alpha3.ResourceLabelManagedBy],
```
AlwxSin: 🔵 **The `managed-by` guard reads the new object on UPDATE, so the label can be stripped in the same request**

This hook and `internal/webhook/clusterprojectrolebinding/hook.go:59` now agree with each other — that was the earlier review note and it is fixed — but they agree on reading `req.Object` for everything except DELETE. So an UPDATE that removes `projects.deckhouse.io/managed-by: controller` in the same request arrives at `internal/webhook/rolebinding/validator.go:69` with `ManagedBy == ""`, and the controller-managed protection does not fire.

`modules/140-user-authz/webhooks/validating/system_resources.py` evaluates its markings on both the new *and* the old object for exactly this reason, and says so in a comment ("so a user cannot bypass protection by stripping the marking in the same UPDATE"). The two guards now answer the same question differently.

In practice the multitenancy ValidatingAdmissionPolicy catches it: its binding selects on `heritage: multitenancy-manager` (`templates/validation.yaml`, `objectSelector`), the controller stamps that label alongside `managed-by` (`internal/manager/project/standard_fields.go:36-40`), and a binding `objectSelector` matches when *either* the old or the new object matches. Supported Kubernetes is 1.32–1.36 (`candi/version_map.yml`), where VAP is GA, so the backstop is always present — hence 🔵 rather than higher. Still, reading the label from `req.OldObject` on UPDATE too is a two-line change and stops this guard depending on a policy in another file.

modules/160-multitenancy-manager/webhooks/conversion/projects
Jabejixo: Both edges were real and both are covered now. The tests read the jq out of the shipped hook and run it with gojq, so the copy cannot drift from the original. The bare `requests`/`limits` key is fixed: only an object counts as nesting, a scalar travels flat, and a quota holding both `requests` and `requests.cpu` resolves to the nested form whatever the key order. The documented drop of a nested non-requests/limits object is pinned by a test as well. `2bf46419ce`

modules/160-multitenancy-manager/webhooks/conversion/projecttemplates
Jabejixo: Done: v1alpha1 is no longer served and the annotation is gone with all three problems. The conversion itself stays — the version is still in `spec.versions` and in `status.storedVersions`, and the CRD still declares the webhook, so the apiserver keeps asking for it; it now does the version bump and nothing else. The project webhook reads templates at v1alpha2 and projects them onto the legacy shape afterwards. `2bf46419ce`, `e07a782369`

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/webhook/projectrolebinding/hook.go
Jabejixo: Done. Both hooks resolve the marking through one function that reads the old object on UPDATE as well, so the guard no longer depends on the VAP. `2bf46419ce`

PR #21161, opened 2026-07-06
rbacv2: permission-browser apiserver (ee/be/140)

Jabejixo:

## Description
Adds to permission-browser extension API server in `user-authz`, providing read-only introspection of effective access under the RBACv2 model.
What changes:
- **`WhoCan`** (`authorization.deckhouse.io/v1alpha1`, cluster-scoped, create-only, no etcd): a reverse-RBAC query — given a verb + resource (and optional namespace/group/name/subresource), it returns the subjects (users/groups/serviceAccounts) allowed to perform it. The engine reuses the direct authorizer's matcher over ClusterRoleBindings and the target namespace's RoleBindings from informer caches (no live API calls), handling wildcards, subresources, resourceNames, aggregated ClusterRoles and dedup.
- **`SubjectAccessReport`** (same group, cluster-scoped, create-only, no etcd): the forward counterpart — given a subject (or the caller itself), everything it may do, assembled from its bindings in one pass instead of probed one `SubjectAccessReview` at a time. The report carries the resolved group graph (nested groups from the group catalog, with a cycle guard), a flat list of role assignments with the subject entry that matched each binding, and access grouped into scopes: the cluster-wide one, then one per set of namespaces whose access is identical, so a project reads as a single section. Rows say how they were derived — `viaWildcard` for an expanded resource wildcard, `viaVerbWildcard` for a `verbs: ["*"]` rule whose listed verbs are a sample — and `scopes[].caveat` reports what admission adds on top of RBAC: which verbs the system-resource webhook restricts, and whether this subject bypasses it as a superadmin, a cluster administrator or a member of a bypass group. Those role and group sets are the webhook's own, asserted by a contract test that reads them out of `system_resources.py`.
- Non-self reports are gated by the `subjectaccessreports/nonself` subresource and a dedicated ClusterRole that must be bound explicitly; the self report is available to `system:authenticated` and discloses nothing `SelfSubjectRulesReview` would not.

Under RBACv2 (roles, capabilities, projects, CAR/AR union) operators need to answer questions that plain `kubectl auth can-i` cannot: "who can do X across the cluster?", "what can this user actually do, and where did it come from?" and "which namespaces can this subject see?". `WhoCan` gives the reverse lookup (equivalent to `oc policy who-can`), `SubjectAccessReport` gives the forward one with provenance and the admission caveats RBAC alone cannot express, and the RBACv2-aware `AccessibleNamespaces` resolver drives correct, multitenancy-filtered namespace listing in the console and API.

Not necessarily.

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
section: user-authz
type: feature
summary: permission-browser API server — WhoCan (reverse RBAC lookup) and SubjectAccessReport (what a subject can do, with provenance and admission caveats).
impact_level: low
```


ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/registry/subjectaccess_storage.go
```diff
@@ -0,0 +1,241 @@
...
+	klog.V(4).Infof("SubjectAccessReport: %s %q -> %d role assignments, %d scopes",
+		req.Subject.Kind, req.Subject.Name, len(status.RoleAssignments), len(status.Scopes))
+
+	return report, nil
+}
+
+// buildRequest resolves the report input, applying defaults and the non-self
+// authorization gate.
+func (s *SubjectAccessStorage) buildRequest(ctx context.Context, caller user.Info, spec *v1alpha1.SubjectAccessReportSpec) (resolver.SubjectAccessRequest, error) {
+	req := resolver.SubjectAccessRequest{
```
AlwxSin, line 129: ### 1. 🔴 Self-mode honors `spec.groups`, bypassing the non-self gate

`ExtraGroups: spec.Groups` is filled in this struct literal, and the self branch (`spec.Subject == nil`, a few lines below) returns without clearing it. `resolveIdentity` then adds every one of those groups to the identity unconditionally (`subjectaccess.go:255`), so the report includes every binding whose subject list names them: `status.roleAssignments[].matchedBy = {kind: Group, name: <injected>}`, plus the resource/verb rows and their `sources`.

`d8:user-authz:self-subject-access-checker` is bound to `system:authenticated`, with the comment *"Self reports disclose nothing the user cannot already discover with SelfSubjectRulesReview"* (`templates/permission-browser-apiserver/rbac-for-us.yaml:136`). That no longer holds: any authenticated user can POST

```yaml
apiVersion: authorization.deckhouse.io/v1alpha1
kind: SubjectAccessReport
spec:
  groups: ["d8:some-group"]
```

and get the full permission map of any group name they know — the exact data a `Kind: Group` report guards behind `subjectaccessreports/nonself` (200 group names per request, unlimited requests). Group names have to be guessed rather than enumerated, but in DKP they are predictable. No test covers self + `spec.groups`, which is why this reads like an oversight rather than a decision.

Two ways out, whichever you prefer:

- clear `req.ExtraGroups` in the self branch, or
- require the non-self gate whenever `spec.groups` carries anything the caller's token does not already have.

> **Объяснение:** В self-режиме `spec.groups` не сбрасывается и попадает прямо в identity. ClusterRole на self-отчёт привязан к `system:authenticated`, а отчёт по `Kind: Group` намеренно закрыт сабресурсом `nonself` — но тот же самый результат достаётся через `spec.groups` без всяких грантов. Любой залогиненный юзер узнаёт полную карту прав по любому известному ему имени группы (имена надо угадывать — читать `groups.deckhouse.io` не требуется, но в DKP они предсказуемые). Тестов на self+groups нет, похоже просто забыли. Фикс: либо зануляем `ExtraGroups` в self-ветке, либо требуем nonself, если в `spec.groups` есть что-то, чего нет в токене вызывающего.

modules/140-user-authz/templates/cluster-roles.yaml
```diff
@@ -380,6 +380,19 @@
   - nodes
   verbs:
 {{- include "user_authz_verbs" "p" }}
+# Permission browser queries. The nonself subresources are what makes a review
```
AlwxSin: ### 2. 🟡 `whocans` + `subjectaccessreports/nonself` granted to an access level that supports `limitNamespaces`

These rules land in `user_authz_cluster_admin_rules`, i.e. in `user-authz:cluster-admin`. But `accessLevel: ClusterAdmin` is fully compatible with `limitNamespaces`/`namespaceSelector` — the docs state that even `SuperAdmin` **does not lift** those (`modules/140-user-authz/docs/USAGE.md:288`). Meanwhile this same PR writes, twice, that these capabilities *"must NEVER be bound to tenant-scoped (project) admins"*: `templates/permission-browser-apiserver/rbac-for-us.yaml:97-105` and the `registerAPIGroup` security note in `pkg/apiserver/apiserver.go:195-207`. Those two statements contradict each other.

It is not only theoretical. Both resources are cluster-scoped, and the authorization webhook denies cluster-scoped requests only for **namespaced** resources (`ee/be/modules/140-user-authz/images/webhook/src/internal/web/hook/handler.go:219`; the apiserver's own engine mirrors it at `pkg/authorizer/multitenancy/engine.go:236`), so nothing filters these. Such an admin already reads all ClusterRoles/ClusterRoleBindings through `cluster_editor_rules` (`cluster-roles.yaml:297`), so the cluster-wide half of `WhoCan` adds nothing new — but:

- `WhoCan{namespace: <outside limitNamespaces>}` returns the subjects of that namespace's RoleBindings, which they cannot read directly (RoleBinding reads come from `common_rules` and are namespaced, so the webhook denies them outside the limits);
- `subjectaccessreports/nonself` returns any subject's namespaced access plus their resolved group graph.

Either drop these resources from `cluster_admin_rules` and require an explicit binding of `who-can-checker` / `subject-access-checker`, or reject the query when the caller's ClusterAuthorizationRule carries namespace filters. Worth an explicit call either way, since the PR currently documents the opposite of what it ships.

> **Объяснение:** Правила добавлены в `user_authz_cluster_admin_rules`, т.е. в уровень `ClusterAdmin`, а у него спокойно бывает `limitNamespaces` — по докам даже `SuperAdmin` их не снимает. При этом сам же PR в двух местах пишет «никогда не привязывать к tenant-scoped админам». Оба ресурса cluster-scoped, а authz-вебхук режет cluster-scoped запросы только для namespaced-ресурсов, так что фильтрация не срабатывает. CRB/ClusterRole такой админ и так читает (`cluster_editor_rules`), но `WhoCan` с чужим namespace отдаёт субъектов RoleBinding'ов, которые ему читать нельзя, а `nonself`-отчёт — карту прав любого субъекта плюс граф групп. Нужно осознанное решение: либо убрать из `cluster_admin_rules`, либо денаить запрос, если у CAR вызывающего есть namespace-фильтры.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/roledescriptor.go
```diff
@@ -0,0 +1,214 @@
...
+// user-authz:super-admin, which a ClusterAuthorizationRule with accessLevel
+// SuperAdmin binds to. They carry no rbac.deckhouse.io labels, so the descriptor
+// says nothing about them and they have to be recognised by name.
+var clusterAdminRoles = map[string]struct{}{
+	"cluster-admin":          {},
+	"user-authz:super-admin": {},
+}
+
+// IsClusterAdminRole reports whether holding this role bypasses the
+// system-resource protection the same way a superadmin does. Keep in sync with
```
AlwxSin, line 215: ### 3. 🟡 Superadmin-bypass detection diverges from the webhook it models

This comment promises to stay in sync with `CLUSTER_ADMIN_ROLES` in `modules/140-user-authz/webhooks/validating/system_resources.py`, and states the invariant correctly: *"a report that claims a restriction the webhook does not apply is worse than no report at all"*. That file arrives in #21162, so it is not in this branch or its base yet — but comparing against `pr/c-webhooks` shows drift in both directions.

**Groups are ignored entirely.** The webhook skips its checks for `BYPASS_GROUPS` = `system:masters`, `kubeadm:cluster-admins`, `superadmins`, `system:sudousers`. `superadminIndex` (`pkg/resolver/subjectaccess.go:411`) only walks `report.assignments`, i.e. role names and descriptors from bindings, so a member of `system:masters` gets stamped with `caveat.restrictedNamespaces` while the webhook waves them through — exactly the failure direction this comment calls unacceptable.

**`IsSuperadminRole` is broader than `SUPERADMIN_ROLES`.** The webhook matches three literal names (`d8:namespace:superadmin`, `d8:project:superadmin`, `d8:system:superadmin`). The Go side accepts any role whose descriptor says `level=superadmin` and `scope ∈ {namespace, project, system, subsystem}` — including `d8:subsystem:*:superadmin` and any custom or external role carrying those labels. Those get `caveat.superadmin: true` from a webhook that will still deny them.

A contract test asserting the two Go sets equal the two Python sets would lock this down permanently — the same shape as the RBACv2 template invariant test in #21165.

> **Объяснение:** Комментарий обещает синхронизацию с `system_resources.py` (файл приезжает в #21162), но расхождение есть в обе стороны. Первое и главное: вебхук пропускает всех из `BYPASS_GROUPS` (`system:masters`, `kubeadm:cluster-admins`, `superadmins`, `system:sudousers`), а `superadminIndex` смотрит только на роли из биндингов — значит члену `system:masters` отчёт нарисует `restrictedNamespaces`, хотя вебхук его вообще не трогает. Это ровно то, что комментарий называет «хуже, чем никакого отчёта». Второе: вебхук матчит три конкретных имени, а `IsSuperadminRole` считает суперадмином любую роль с `level=superadmin` и скоупом из четырёх (включая `subsystem` и кастомные роли с этими лейблами) — им отчёт поставит `superadmin: true`, а вебхук их заденаит. Дешёвый контрактный тест на равенство множеств закрыл бы вопрос насовсем.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/subjectaccess_report.go
```diff
@@ -0,0 +1,558 @@
...
+			}
+			targets = append(targets, resourceTarget{group: resource.Group, resource: resource.Resource, viaWildcard: true})
+		}
+		return targets
+	}
+
+	targets := make([]resourceTarget, 0, len(rule.APIGroups)*len(rule.Resources))
+	for _, group := range rule.APIGroups {
+		for _, resource := range rule.Resources {
+			if namespaced && !hasWildcard && b.scopeCache != nil {
```
AlwxSin, line 248: ### 4. 🟡 Partial discovery silently shrinks the report

`ResourceScopeCache.refresh` replaces the whole scope map with the result of `ServerPreferredResources`, which routinely returns **partial** results when an aggregated APIService is flapping — `scope_cache.go:243` explicitly handles that case and logs it at V(4). `IsNamespaced` returns `false` for anything missing from the snapshot (`scope_cache.go:66`), a deliberate fail-closed choice made for `AccessibleNamespace`, where over-reporting leaks namespace names.

In this check, the same `false` means *drop the row*. So every namespaced grant whose API group vanished from the snapshot disappears from the report, and wildcard expansion (`ResourcesMatching`, a few lines up) enumerates only what survived. Nothing signals it: no `notes`, no `truncated`, no `evaluationError`.

For an audit report the safe direction is inverted — over-report, never under-report. Note the asymmetry that keeps this off the test radar: with `scopeCache == nil` (as in `TestIntegration_SubjectAccessReport`) all rows are kept, while a non-nil-but-incomplete cache drops them, so no test exercises the degraded path. A fully empty cache is covered by the `resource-scope-cache` readyz gate; a partial one is not.

Keeping rows whose `(group, resource)` is unknown to the snapshot, or appending a note when the last refresh was partial, would close it.

> **Объяснение:** `refresh` целиком заменяет мапу результатом `ServerPreferredResources`, который при флапающем aggregated APIService отдаёт частичные данные (код это явно обрабатывает, но логирует на V(4)). `IsNamespaced` для отсутствующего ресурса возвращает `false` — это осознанный fail-closed для `AccessibleNamespace`, где лишнее = утечка. Но здесь тот же `false` означает «выкинуть строку»: все namespaced-гранты по пропавшей группе исчезают из отчёта, а wildcard разворачивается только по остатку. И ни `notes`, ни `truncated`, ни `evaluationError` об этом не скажут. Для аудит-инструмента безопасное направление обратное — лучше перебор, чем недобор. Забавная асимметрия: при `scopeCache == nil` (как в интеграционном тесте) строки сохраняются, а при непустой-но-неполной мапе выкидываются, поэтому тесты этот путь не видят. Полностью пустой кеш прикрыт readyz-чеком, частичный — нет.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/subjectaccess.go
```diff
@@ -0,0 +1,540 @@
...
+	"k8s.io/klog/v2"
+
+	"permission-browser-apiserver/pkg/apis/authorization/v1alpha1"
+	"permission-browser-apiserver/pkg/authorizer/multitenancy"
+	"permission-browser-apiserver/pkg/authorizer/rbacadapter"
+)
+
+// standardVerbs is what a "*" verb rule expands to. It matches the verb set the
+// console shows, so an expanded wildcard reads like an explicit grant.
+var standardVerbs = []string{"get", "list", "watch", "create", "update", "patch", "delete", "deletecollection"}
```
AlwxSin, line 30: ### 5. 🟡 `verbs: ["*"]` expansion drops escalate / bind / impersonate / proxy

`expandVerbs` (`subjectaccess.go:481`) turns `*` into exactly these eight verbs. So a cluster-admin report shows `get…deletecollection` on `roles`/`clusterroles` but never reveals `escalate` or `bind`; on `users`/`groups`/`serviceaccounts` it never reveals `impersonate`; on `nodes`, no `proxy`. Those are precisely the verbs an access audit gets run for.

Worse, the loss is undetectable by the client: `ViaWildcard` reflects only apiGroup/resource wildcards (`subjectaccess_report.go:242`), not a verb wildcard, so an eight-verb row from `verbs: ["*"]` is indistinguishable from an explicit eight-verb grant. It reads like a precise enumeration.

Cheapest fixes: keep `*` in the row's verbs alongside the expansion, add the privilege verbs to `standardVerbs`, or add a `viaVerbWildcard` flag to `ResourceAccess`.

> **Объяснение:** `expandVerbs` разворачивает `*` в восемь стандартных глаголов, и из отчёта пропадают `escalate`/`bind` на ролях, `impersonate` на юзерах/группах/SA, `proxy` на нодах — ровно то, ради чего аудит прав и делают. При этом `ViaWildcard` относится только к wildcard'ам по группе/ресурсу, так что клиент не может понять, что список глаголов урезан: выглядит как точное перечисление. Самое дешёвое — оставлять `*` в списке рядом с разворотом, добавить privilege-глаголы в `standardVerbs` или завести флаг `viaVerbWildcard`.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/README.md
```diff
@@ -112,6 +112,105 @@ status:
       reason: "user has no access to the namespace"
 ```
 
+### Resource: `WhoCan`
```
AlwxSin, line 119: ### 6. 🟡 SubjectAccessReport is missing from the README, the PR description and the changelog

`WhoCan` got a thorough section right here — How It Works, Security, Multi-tenancy, Limitations, a usage example. `SubjectAccessReport` — roughly 2500 lines, five new resolver files, a new dynamic client and two new ClusterRoles — appears nowhere in this README: not in the Overview, not in the API list, not in Limitations, and not in the `caveat` / `superadmin` semantics that clients have to interpret to render the thing.

Same for the PR description (`WhoCan` only) and, more consequentially, the ```changes``` block, which ships as release notes:

```
summary: permission-browser API server — WhoCan (reverse RBAC lookup).
```

The documentation checklist is unchecked. Note also that the Loop thread had non-who-can work being *moved out* of this PR, whereas a second resource was added instead — so the framing needs updating regardless of which way you go.

> **Объяснение:** По `WhoCan` в README расписано всё, а `SubjectAccessReport` (~2500 строк, пять новых файлов резолвера, dynamic client, две новых ClusterRole) не упомянут вообще — ни в Overview, ни в списке API, ни в Limitations, ни в семантике `caveat`/`superadmin`, которую клиентам придётся интерпретировать. То же с описанием PR и, что важнее, с блоком ```changes``` — он уезжает в релиз-ноты и говорит только про `WhoCan`. Галка про документацию не поставлена. Плюс в треде речь была о том, что всё не связанное с who-can из этого PR вынесется, а по факту заехал второй ресурс — формулировки всё равно надо освежить.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/subjectaccess_report.go
```diff
@@ -0,0 +1,558 @@
...
+// in every namespace, so they form a single scope of their own rather than
+// being repeated for each namespace.
+func (b *reportBuilder) addClusterGrant(origin grantOrigin, rules []rbacv1.PolicyRule) {
+	b.recordAssignment(origin, "")
+	b.cluster.noteOrigin(origin)
+	b.expand(b.cluster, origin, rules, false)
+}
+
+// addNamespaceGrant records a RoleBinding grant in one namespace.
+func (b *reportBuilder) addNamespaceGrant(namespace string, origin grantOrigin, rules []rbacv1.PolicyRule) {
```
AlwxSin, line 147: ### 7. 🔵 `RoleAssignments` is the one part of the report with no limit

`ReportLimits` (`subjectaccess.go:45`) caps namespaces, resource rows and sources per row, and `addNamespaceGrant` honours `MaxNamespaces` right here — but `recordAssignment` runs *before* that check and has no cap of its own, appending one entry per binding per matched subject entry.

A subject matched by thousands of fanned-out project RoleBindings therefore gets a response that grows unbounded, and `status.truncated` says nothing about it even though the scope side of the very same call was capped. Either bound the assignments too, or set `truncated` when they are dropped.

> **Объяснение:** `ReportLimits` ограничивает namespace'ы, строки и sources, и `addNamespaceGrant` честно проверяет `MaxNamespaces` — но `recordAssignment` вызывается до этой проверки и своего лимита не имеет: по записи на каждый биндинг × каждое совпавшее subject-вхождение. У субъекта с тысячами fan-out project RoleBinding'ов ответ пухнет без ограничений, а `truncated` про это молчит, хотя скоупы в том же вызове обрезались.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/registry/whocan_storage.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Create resolves the subjects allowed to perform the requested action.
+func (s *WhoCanStorage) Create(ctx context.Context, obj runtime.Object, createValidation rest.ValidateObjectFunc, options *metav1.CreateOptions) (runtime.Object, error) {
+	wc, ok := obj.(*v1alpha1.WhoCan)
+	if !ok {
+		return nil, apierrors.NewBadRequest("object is not a WhoCan")
+	}
+
+	if wc.Spec.ResourceAttributes == nil && wc.Spec.NonResourceAttributes == nil {
```
AlwxSin: ### 8. 🔵 WhoCan accepts a meaningless request and answers it confidently

Validation only requires that one of the two attribute blocks is non-nil. So `{spec: {resourceAttributes: {}}}` passes, matches any rule with `*` on apiGroup/resource/verb, and returns every cluster-admin in the cluster as a real-looking answer to a question that was never asked.

And when a client sets both blocks, `accessAttributes.IsResourceRequest()` (`pkg/registry/storage.go:316`) silently picks the resource one and `nonResourceAttributes` is ignored without a word.

`oc policy who-can`, the stated analogue, requires verb + resource. Rejecting an empty `resourceAttributes` (and rejecting both-set) is a few lines and turns two silent misreads into clear 400s.

> **Объяснение:** Валидация требует лишь непустой один из двух блоков. На `{spec: {resourceAttributes: {}}}` матчится любое правило с `*` по группе/ресурсу/глаголу, и в ответ приезжают все кластер-админы — выглядит как настоящий ответ на настоящий вопрос. А если клиент задал оба блока, `IsResourceRequest()` молча выбирает resource и игнорирует `nonResourceAttributes`. У `oc policy who-can`, на который ссылается дока, verb+resource обязательны; отвергать пустой `resourceAttributes` (и оба сразу) — пара строк, зато два молчаливых недоразумения превращаются в честный 400.

AlwxSin (commented):
## Review: permission-browser apiserver — WhoCan + SubjectAccessReport

Reviewed `98da51a8` against `integration/rbacv2`.

The core is solid. `WhoCan` reuses the forward authorizer's `ruleAllows` instead of growing a second matcher, serves everything from informer caches, dedups and sorts deterministically, and surfaces a partial result through `status.evaluationError` rather than letting a failed list masquerade as "nobody can". `whocan_test.go` covers aggregated ClusterRoles, `resourceNames`, ServiceAccount namespace defaulting, non-resource URLs and context cancellation. `SubjectAccessReport` is not bolted on either: there is a race test that swaps the discovery map underneath concurrent reports, a cycle guard in `GroupCatalog`, and a test for collapsing namespaces with identical access. The security notes on `registerAPIGroup` and in the chart are the kind of thing most PRs skip.

The findings below are mostly about the seams.

**Blocking:** finding 1 — in self mode `spec.groups` is never cleared, so a ClusterRole bound to `system:authenticated` hands out the group reports that `subjectaccessreports/nonself` is meant to gate. One field to zero out.

**Needs an explicit decision, not a default:** findings 2 and 3. Who actually gets `whocans` (the PR grants it to an access level that supports `limitNamespaces`, while the PR's own comments say it must never reach namespace-scoped admins), and how the Go superadmin-bypass detection stays in sync with `system_resources.py` from #21162. Finding 3 is the same "guards judged by different criteria in different PRs" class that was the main systemic complaint last round.

**Report accuracy:** findings 4 and 5 — both make the report silently claim *less* than the subject can do, with nothing in `notes`/`truncated`/`evaluationError` to reveal it. For a tool people will use to answer "what can this user do", a silent undercount is the expensive failure mode.

**Paperwork:** finding 6 — `SubjectAccessReport` is absent from the README, the PR description and the ```changes``` block that ships as release notes.

**Optional:** findings 7 and 8.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/webhook_contract_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func readWebhookSource(t *testing.T) string {
+	t.Helper()
+
+	data, err := os.ReadFile(filepath.Clean(webhookSourcePath))
+	if err != nil {
+		// The webhook lives in another module of the same repository. When only this image is
+		// checked out there is nothing to compare against, and failing would report a packaging
+		// detail as a contract violation.
+		t.Skipf("webhook source is not available (%v)", err)
```
AlwxSin: 🟡 **The contract test skips on exactly the drift it guards against**

`readWebhookSource` calls `t.Skipf` when the python file cannot be read. Renaming, moving or deleting `modules/140-user-authz/webhooks/validating/system_resources.py` therefore turns this test green — the one scenario a cross-language contract test exists to catch. It really does skip rather than fail: the package's tests pass when only this image subtree is checked out. The commit message argues that "a comment asking the next person to keep two languages in sync is a promise that rots", which is right, but a test that skips when the other side disappears is the same promise with extra steps.

Make the read a hard failure. If the "only this image is checked out" case is real, gate the skip behind an explicit opt-out (an env var set by whatever builds the image standalone) so the default in CI is to fail.

Separately, the `BYPASS_GROUPS` assertion at line 40 is subset-only (`bypassGroups ⊆ webhookGroups`), so a group added on the python side is not caught, and the report will keep claiming a restriction the webhook no longer applies. Splitting `BYPASS_GROUPS` in the python source into an administrator half and a cluster-component half would let this be an equality check in both directions, like the other two sets.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/subjectaccess_report.go
```diff
@@ -0,0 +1,589 @@
...
+				// Skip cluster-scoped resources named by a RoleBinding: the
+				// grant exists in RBAC but can never be exercised.
+				//
+				// Only a resource discovery positively calls cluster-scoped is
+				// dropped. One the snapshot has never heard of is kept: the
+				// snapshot goes incomplete whenever an aggregated APIService
+				// flaps, and an audit report that omits a grant is a worse
+				// answer than one that lists a grant nobody can exercise.
+				base, _, _ := strings.Cut(resource, "/")
+				if b.isKnownClusterScoped(group, base) && b.isKnownClusterScoped(group, resource) {
```
AlwxSin: 🔵 **The cluster-scoped drop is now dead for every subresource**

`ServerPreferredResources` skips every resource whose name contains `/` (`client-go/discovery/discovery_client.go`, the `strings.Contains(apiResource.Name, "/")` continue), so no subresource is ever in `ResourceScopeCache.scopeMap`. `isKnownClusterScoped(group, "nodes/proxy")` is therefore always false, and with the `&&` the row is always kept — where the previous `!IsNamespaced(base) && !IsNamespaced(resource)` dropped it.

That direction matches the principle this change states (an audit report over-reports rather than omits), so it is not a bug. But the second half of the condition can never contribute for a subresource, and `nodes/proxy` named by a RoleBinding is now reported as an exercisable namespace grant. Either drop on the base resource alone, or say in the comment that subresources are deliberately never dropped — right now the code reads as if both halves matter.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/registry/subjectaccess_storage.go
Jabejixo: Done: the field is cleared in the self branch. Two tests cover it — self drops it, non-self keeps it. `b7fa0c5f32`

modules/140-user-authz/templates/cluster-roles.yaml
Jabejixo: Done, dropped from that access level. The level supports `limitNamespaces` and both resources are cluster-scoped, so nothing would confine the answer; whoever is meant to audit other subjects gets the checker roles bound explicitly. The comment next to the rules now says why they are not there. `b7fa0c5f32`

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/roledescriptor.go
Jabejixo: Closed in both directions: `IsSuperadminRole` matches the three literal names instead of labels, and the bypass groups are honoured before any binding is read. The contract test parses the python source and asserts set equality — after your later note, for all three sets. `b7fa0c5f32`, `df65a7b092`

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/subjectaccess_report.go
Jabejixo: Done: a resource unknown to the snapshot is kept. The cache now tells "unknown" from "cluster-scoped" through a separate accessor, so `IsNamespaced` stays fail-closed where it has to be, and a partial snapshot adds a note to the report. `b7fa0c5f32`

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/subjectaccess.go
Jabejixo: Done: `viaVerbWildcard` on the row and on every source. The console marks such rows as "all actions" and spells the privileged verbs out in the tooltip. `b7fa0c5f32`

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/README.md
Jabejixo: Done: the README has a SubjectAccessReport section — schema, caveat semantics, the self/nonself split and the limitations. The PR description and the `changes` block name both resources. `b7fa0c5f32`

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/subjectaccess_report.go
Jabejixo: Done: `MaxRoleAssignments` is 2000, and hitting it sets `truncated` like the other caps. `b7fa0c5f32`

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/registry/whocan_storage.go
Jabejixo: Done: 400 on an empty `resourceAttributes`, on a missing verb or resource, and on both blocks at once. The console already sends exactly one block with verb and resource. `b7fa0c5f32`

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/webhook_contract_test.go
Jabejixo: Done, the test fails now. A partial checkout is told apart by the test file that ships beside the webhook: gone together, the webhooks are not in this checkout; gone alone, the source moved. `BYPASS_GROUPS` is an equality check, so the Go side mirrors the whole set including the cluster components, and a report about a ServiceAccount of kube-system now says `superadmin`. `df65a7b092`

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/subjectaccess_report.go
Jabejixo: Done: the base resource decides. Discovery lists no subresources at all — the cluster I checked returns 351 resources, not one with a slash — so the second half of the condition could never contribute. A report over a RoleBinding granting `nodes`, `nodes/proxy` and `pods/log` shows `pods/log` alone. `df65a7b092`

PR #21162, opened 2026-07-06
rbacv2: webhooks + fan-out hooks + webhook-handler rbac

Jabejixo:

## Description
RBACv2 validating webhooks in user-authz (140): cluster-role contract, role-binding guard (rewritten in Python, denies binding scoped roles/capabilities cluster-wide), role-escalation guard, system-resource protection, and multitenancy EE-field gating; the manage→use fan-out hooks; and the webhook-handler ServiceAccount RBAC. Admission-only; no component restarts.
Enforces the RBACv2 invariants the API server cannot: reserved `d8:` naming, no direct cluster-wide binding of namespace/project roles or any capability, no privilege escalation to project-management resources, and superadmin-only mutation/exec of system resources placed in tenant namespaces.
Not necessarily.
## Checklist
- [x] e2e tests passed.
## Changelog entries
```changes
section: user-authz
type: feature
summary: RBACv2 validating webhooks (cluster-roles, role-binding, role-escalation, system-resources, multitenancy gating) and manage→use fan-out.
impact_level: default

### CODE REVIEW

File: modules/140-user-authz/webhooks/validating/cluster_roles.py [hunk 1]
```diff
@@ -0,0 +1,337 @@
...
+                return (
+                    f'ClusterRole "{name}" with "{SCOPE_LABEL}: {scope or "<unset>"}" must not aggregate {offending}: '
+                    "a role granted in a namespace or a project may only aggregate namespace and project capabilities."
+                )
+
+        # The delegatable label is what lets a role be bound inside a project, so it may only be
+        # claimed by a role built entirely from namespace/project capabilities. The check above is
+        # not enough: it only reads the "aggregate-to-<lineage>-as" selectors, while a role
+        # assembled from individual capabilities selects them by the capability label instead.
+        if labels.get(DELEGATABLE_LABEL) == "true" and not _aggregates_only_tenant_capabilities(selectors):
```
AlwxSin: 🟡 **The `delegatable` guard is skipped unless the object opts into the framework**

This check — and the name↔scope agreement, and the tenant/system lineage rules — all sit inside `if kind_label in ("custom-role", "custom-capability")` at line 256. The grant registry, meanwhile, gates only on the label itself: `modules/160-multitenancy-manager/templates/cluster-objects-controller/grantable-resources.yaml:149-152` excludes ClusterRoles where `rbac.deckhouse.io/delegatable` `DoesNotExist`, with `defaultAvailability: All`. So a ClusterRole that carries `delegatable: "true"` without the `rbac.deckhouse.io/kind` label is bindable inside every project and is validated by nothing.

Run against this hook:

```
non-d8 name + delegatable + own wildcard rules:               ALLOWED
non-d8 name + delegatable + aggregates the security lineage:  ALLOWED
same object + rbac.deckhouse.io/kind: custom-role:            DENIED
```

The actor is not necessarily a cluster superadmin: `templates/rbacv2/global/subsystem/capabilities/manage_security_resources.yaml:33-42` grants `create/update/patch/delete` on `clusterroles` at the security-subsystem manager level.

To be fair on severity: native RBAC's escalation check still bounds which rules such a user can put in the role, and a RoleBinding never confers cluster-scoped rules, so I could not build a concrete escalation past those two — this is a defense-in-depth hole rather than a proven exploit. But the guard `66c69fda99` added ("allow a custom role into a project only when it is built from tenant capabilities") does not hold as written. Keying these checks on the `d8:custom:` name prefix, or on the presence of the `delegatable` label itself, closes it.

modules/140-user-authz/webhooks/validating/system_resources.py
```diff
@@ -0,0 +1,431 @@
...
+                break
+    return roles
+
+
+def privileged_roles(username: str, groups: set, namespace: str) -> set:
+    """The privileged ClusterRoles the requester is bound to. Both the superadmin and the
+    cluster-admin question are answered from this one pair of reads, so a request never costs more
+    live reads than it did when only superadmin was recognised."""
+    # Cluster-wide: a ClusterRoleBinding to a privileged role grants it everywhere.
+    roles = _matched_roles(_privileged_bindings("clusterrolebindings"), username, groups)
```
AlwxSin, line 396: 🔵 **Every protected request lists all ClusterRoleBindings as JSON**

`_privileged_bindings("clusterrolebindings")` runs `kubectl get clusterrolebindings -o json` with no selector and no limit, and parses the whole response in Python — inside a handler whose request is 50m/100Mi and whose VPA `maxAllowed` memory is capped at the same 100Mi (`modules/002-deckhouse/templates/webhook-handler/deployment.yaml:1-4,26-28`). The forked kubectl shares that cgroup.

There is no memory *limit*, so this degrades into node pressure rather than an OOM kill, and the path is genuinely rare (a non-privileged user touching a marked object, or exec into a marked pod) — hence 🔵. But a cluster with a few thousand bindings, which RBACv2 plus per-project bindings makes more likely, will feel it. `-o jsonpath` or `--output=custom-columns` over `roleRef` and `subjects` would cut the parsed volume by an order of magnitude and keep the same decision.

modules/140-user-authz/webhooks/validating/system_resources.py
```diff
@@ -0,0 +1,431 @@
...
+            "Project instead."
+        )
+
+    # system-resource protection: editable only by a superadmin or a cluster administrator.
+    if is_system and not privileged_roles(username, groups, namespace):
+        return (
+            f'{kind} "{name}" is a Deckhouse system resource '
+            f'(label {SYSTEM_RESOURCE_LABEL}={SYSTEM_RESOURCE_VALUE}) placed in this '
+            "namespace. It can be modified or deleted only by a superadmin "
+            "(d8:namespace:superadmin / d8:project:superadmin)."
```
AlwxSin: 🔵 **The deny message names a narrower superadmin set than the code accepts**

This denial, and the exec one at line 426, both say the operation is allowed only for `(d8:namespace:superadmin / d8:project:superadmin)`. But the gate is `privileged_roles(...)`, which resolves against `PRIVILEGED_ROLES` = `SUPERADMIN_ROLES | CLUSTER_ADMIN_ROLES` — so `d8:system:superadmin`, `cluster-admin` and `user-authz:super-admin` pass too.

The file header gets this right; the message a user actually sees does not, and it sends people looking for the wrong role. Listing the full set, or saying "a superadmin of this namespace/project, a system superadmin, or a cluster administrator", would match what the code does.

modules/140-user-authz/webhooks/validating/cluster_roles.py
Jabejixo: Done, the label is judged on its own. The check runs before the framework block and fires whatever `rbac.deckhouse.io/kind` says, and the remaining custom-object rules now apply to a `d8:custom:` name too. Both escalating cases are denied; the tenant-capability one is still allowed. `c6373d559f`

modules/140-user-authz/webhooks/validating/system_resources.py
Jabejixo: Done with a go-template projection: only `roleRef` and `subjects`, with the privileged roles matched there as well. On the cluster I tested against it returned the same twelve bindings in 663 bytes instead of 630 KB. The separators are control characters, because a subject name is arbitrary text. `c6373d559f`

modules/140-user-authz/webhooks/validating/system_resources.py
Jabejixo: Done: the message is built from the same set the gate uses, so the two cannot drift again. A denial in the cluster now names all five roles. `c6373d559f`

PR #21183, opened 2026-07-06
[deckhouse] drop critical from user-authn

diyliv:
## Description

Remove `critical: true` from user-authn module. This resolves the issue where user-authn templates depended on cert-manager being available at startup.

user-authn has templates that reference cert-manager resources (Certificate, TLS in Ingress/HTTPRoute, ListenerSet). When cert-manager is not yet available, the module fails at `CanRunHelm` phase. Instead of wrapping every template with `has "cert-manager" .Values.global.enabledModules` guards, we simply mark the module non-critical. Users who need user-authn must ensure cert-manager is available.

## Checklist

- [ ] e2e tests passed.

## Changelog entries
```changes
section: user-authn
type: fix
summary: Remove critical flag from user-authn module.
impact_level: low
```

### DIFF
--- a/modules/150-user-authn/module.yaml
+++ b/modules/150-user-authn/module.yaml
@@ -1,5 +1,5 @@
 name: user-authn
-critical: true
+critical: false
 stage: General Availability
 subsystems:
   - security


ldmonster (approved):

ldmonster (changes requested):
Fix the template, not enable script

AlwxSin (approved):

ldmonster (approved):

PR #21235, opened 2026-07-08
[user-authn] fix useroperations hook

Jabejixo:
## Description
Fix Dex `Password` mutations in the UserOperation hook (ResetPassword / Lock / Unlock) so they no longer wipe fields the hook does not model. The mutating functions now edit a deep copy of the live object in place via `unstructured` helpers instead of round-tripping it through the trimmed-down `Password` Go struct. No impact on critical cluster components.

`PatchWithMutatingFunc` performs a full object `Update` with whatever the mutating func returns. The previous code decoded the live `Password` into a Go struct that lacks `groups`, `userID`, `hashUpdatedAt`, `previousHashes` and `incorrectPasswordLoginAttempts`, then serialized it back — so the `Update` dropped all those fields. Losing `groups` broke logins: Dex matches `allowedGroups` against `Password.groups`, so after a password reset (or lock/unlock) the user had no groups and could not log in until groups were re-synced (e.g. by editing a `Group`).

Password reset/lock is an admin-facing operation that silently locks affected users out of the cluster; the fix is small and self-contained, so it should be backported to 1.75 and 1.76.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Preserve Dex Password fields (notably groups) when resetting password, locking or unlocking a user, so users are no longer locked out after these operations.
impact_level: default
```

### DIFF
--- a/modules/150-user-authn/hooks/get_dex_user_operation_crds.go
+++ b/modules/150-user-authn/hooks/get_dex_user_operation_crds.go
@@ -277,6 +277,30 @@ func executeUserOperation(input *go_hook.HookInput, operation UserOperation) err
 	}
 }
 
+// mutatePasswordObject applies mutate to a deep copy of the live Dex Password
+// object and returns the copy.
+//
+// PatchWithMutatingFunc replaces the whole object via an Update with whatever
+// the mutating function returns (and skips the call when the result is deeply
+// equal to the input). The previous implementation decoded the live object into
+// the typed Password struct and serialized it back, but Password does not model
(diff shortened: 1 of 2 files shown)


AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [21260](https://github.com/deckhouse/deckhouse/pull/21260) to the branch [release-1.75](https://github.com/deckhouse/deckhouse/tree/release-1.75) successful!

PR #21247, opened 2026-07-08
[webhook-handler] Wire hook snapshots on reload

yshch:
## Description

Makes the webhook-handler operator's `reloadHooks()` drive the kubernetes-binding enable step to completion, so the hook snapshot caches used by validating/conversion webhooks are always wired to the current hook index before the reload returns.

Background: `reloadHooks()` calls `HookManager.Init()`, which rebuilds every hook and assigns a fresh random `MonitorId` to each kubernetes binding. Admission requests resolve their `includeSnapshotsFrom` snapshots through this *current* index. Enabling kubernetes bindings on the shell-operator main queue wires the *startup* index, but every `reloadHooks()` call replaces it with new `MonitorId`s, orphaning those monitors relative to the current index. The previous in-reload enable loop was fail-fast: if one hook's `EnableKubernetesBindings()` failed (e.g. a transient apiserver outage during startup), it returned immediately, leaving every remaining hook on the current index unwired. Because nothing then re-drove a full enable on that index, the snapshots stayed empty indefinitely.

The fix keeps the enable step inside `reloadHooks()` simple but makes it self-driving:
- Attempt every hook, so one transient failure cannot starve the rest.
- Use the same `HookController.HandleEnableKubernetesBindings()` entry point as the shell-operator main queue, passing `nil` so reload wires monitors without enqueueing synchronization hook-run tasks.
- Unlock kubernetes events only for hooks whose cache actually filled (a hook whose monitor failed to start must not emit events yet).
- Retry only the still-failing hooks with capped exponential backoff until all are wired, a per-reload deadline elapses, or the context is cancelled. On timeout/cancel, return the underlying enable error when available so the reconciler still requeues with the real cause.

It fixes a regression where validating webhooks served by the webhook-handler could permanently deny requests because their `includeSnapshotsFrom` snapshots were delivered empty.

The user-visible symptom: creating an `IstioFederation` (or `IstioMulticluster`) custom resource was rejected with "IstioFederations can only be created when federation is enabled" even though `spec.settings.federation.enabled` was `true` in the istio ModuleConfig. The validating hook reads `federationEnabled` from the `istio_mc` snapshot; with an empty snapshot it always denied.

The trigger is a transient apiserver error during startup (e.g.`connect: connection refused`) while enabling kubernetes bindings. The old fail-fast loop aborted, so the current hook index was left unwired, and since each reload re-runs `Init()` with new `MonitorId`s the state never recovered on its own. This was confirmed on two clusters: in the observed case the webhook kept denying for over an hour after startup, until the pod was restarted.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Drive the webhook-handler hook reload's kubernetes-binding enable to completion with retry so validating webhooks are not left denying requests due to empty snapshots.
impact_level: default
```

### DIFF
--- a/modules/002-deckhouse/images/webhook-handler/operator/cmd/main.go
+++ b/modules/002-deckhouse/images/webhook-handler/operator/cmd/main.go
@@ -19,8 +19,10 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"flag"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -38,6 +40,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:webhook-handler-wire-hook-snapshots-on-reload` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/28964327349) in 40m20s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@46.21.246.250`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 39m10s.

</details>

fuldaxxx (approved):

ldmonster (approved):

PR #21255, opened 2026-07-09
[deckhouse-controller] fix empty snapshots

ldmonster:
## Description

Bumps `addon-operator` to `v1.21.19` and `shell-operator` to `v1.17.7`. Together they fix two independent bugs in module hook machinery.

**shell-operator `v1.17.7`** (flant/shell-operator#923):

- shared informer factories take their context from the owning `kubeEventsManager`, not from the first consumer that asked for them
- a factory whose informer goroutine has already exited is dropped and rebuilt instead of reused
- `EnableKubernetesBindings` is idempotent: it repairs a binding whose monitor is missing and always returns a Synchronization execution info
- kube-event and schedule callbacks in `ManagerEventsHandler` are wrapped with `recover`, mirroring the recover already in `TaskQueue`

**addon-operator `v1.21.19`** (flant/addon-operator#798):

- `HandleModuleEnableKubernetesBindings` aggregates per-hook errors with `errors.Join` instead of aborting on the first failure
- `AddHook` replaces the previously registered hook in all bindings instead of appending a duplicate
- `registered` and `controllersReady` are read and written under the storage lock
- `RegisterHooks` is serialized with a dedicated mutex
- `getHooks` returns a sorted copy instead of the shared slice

Only the `deckhouse-controller` binary is affected. No other cluster components restart.

Two independent bugs, both in module hook machinery.

### 1. Hooks read a snapshot that never updates

- A shared informer factory took its context from the first consumer that asked for it.
- When that consumer stopped, the informer died for every other consumer.
- A stopped informer still returns `HasSynced() == true`, so it was reused.
- The next consumer got a snapshot that never updates again. Nothing marked it stale.

```mermaid
sequenceDiagram
    participant A as consumer A<br/>first to ask for the factory
    participant B as consumer B
    participant FS as factoryStore<br/>shared informer factories
    participant Inf as shared informer

    rect rgb(253, 232, 232)
    Note over A,Inf: BEFORE
    A->>FS: get factory for a GVR
    FS->>Inf: start informer, ctx taken from consumer A
    B->>FS: get factory for the same GVR
    FS-->>B: the same shared informer
    A->>A: consumer A stops, its ctx is cancelled
    Inf->>Inf: informer goroutine exits
    B->>Inf: HasSynced()
    Inf-->>B: true
    Note over B: no event is ever delivered again<br/>the hook reads an empty snapshot forever, no error
    end

    rect rgb(226, 249, 233)
    Note over A,Inf: AFTER
    A->>FS: get factory for a GVR
    FS->>Inf: start informer, ctx taken from kubeEventsManager
    A->>A: consumer A stops
    Note over Inf: the informer keeps running, the manager owns it
    B->>FS: get factory
    FS->>FS: if the informer goroutine already exited,<br/>stop it, drop the factory, build a fresh one
    FS-->>B: a live informer
    end
```

A partial failure made it worse. One hook whose monitor could not start aborted `EnableKubernetesBindings` for the whole module, so the Synchronization run that fills snapshots never happened for the remaining hooks.

### 2. SIGSEGV crash loop on a registration retry

1. A transient hook exec failure (e.g. ECHILD) aborts registration mid-loop. The hooks published before the failure stay in the `byBinding` index without a `HookController`.
2. The retried `ModuleRun` task appends fresh hook objects next to the stale ones. The stale entries keep `HookController == nil`, while the module readiness flag turns true.
3. The kube-events dispatcher dereferences the nil controller in `CanHandleKubeEvent`. That goroutine had no `recover`, so the panic killed the process.

### Before

- A hook can read an empty snapshot forever, with no error anywhere.
- One transient registration failure wedges the module or crashes `deckhouse-controller` into a restart loop.

### After

- Dead informers are dropped and rebuilt.
- A failing monitor no longer blocks the rest of the module.
- A registration retry replaces the stale hooks.
- A panic in hook dispatch is recovered and logged instead of killing the process.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Fix deckhouse-controller crash loop and hooks receiving silently empty snapshots
impact_level: default
```


deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:fix/hook-snapshot-loss-1.76` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/29029320889) in 35m34s.



GCP-WithoutNAT-Containerd-1.34 - Connection string: `ssh  user@34.179.129.99`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.34` succeeded in 34m28s.

</details>

PR #21271, opened 2026-07-09
[istio] Add loadBalancerClass to ingress gateway

### LINKED ISSUE #13706: [istio] Add support for loadBalancerClass in IstioController LoadBalancer configuration
### Preflight Checklist

Currently, the IstioController custom resource does not support specifying the loadBalancerClass field in its LoadBalancer configuration. This limitation makes it impossible to select a specific load balancer implementation when multiple controllers are available in the Kubernetes cluster. For example, if MetalLB is deployed alongside other load balancer controllers, there is no way to ensure that the Istio ingress service will be handled by MetalLB. This reduces flexibility and can cause conflicts or unexpected behavior in environments with more than one load balancer solution.

### Proposed Solution

Please add support for the loadBalancerClass field to the IstioController's LoadBalancer configuration. This field should allow users to specify the desired load balancer class, similar to how it is implemented in the IngressNginxController. By supporting this option, users will be able to explicitly select which load balancer controller should manage the Istio ingress service, enabling better integration with solutions like MetalLB and improving overall cluster network management.

### Additional Information

_No response_

yshch:
## Description

Add `loadBalancerClass` configuration to the `IngressIstioController` resource for the `LoadBalancer` inlet, mirroring the existing implementation in the `ingress-nginx` module.

When `spec.loadBalancer.loadBalancerClass` is set on a controller with `inlet: LoadBalancer`, the value is passed to the `spec.loadBalancerClass` field of the provisioned `LoadBalancer`-type Service.

Kubernetes supports selecting a specific load balancer implementation for a `LoadBalancer` Service via `spec.loadBalancerClass`. Previously the istio ingress gateway could not set this field, so users could not choose a non-default load balancer implementation for the istio ingress gateway.

Resolves https://github.com/deckhouse/deckhouse/issues/13706

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: istio
type: feature
summary: Add loadBalancerClass configuration for the ingress gateway LoadBalancer inlet.
impact_level: low
```


skurbatov (approved):

z9r5 (approved):

modules/110-istio/crds/doc-ru-ingress-istio.yaml
```diff
@@ -42,6 +42,9 @@ spec:
                     annotations:
                       description: |
                         Аннотации, которые будут проставлены сервису для гибкой настройки балансировщика.
+                    loadBalancerClass:
+                      description: |
+                        Класс балансировщика входящих сетевых запросов (пробрасывается в параметр `spec.loadBalancerClass` заказанного сервиса с типом `LoadBalancer`).
```
valeria-brazhnikova:
```suggestion
                        Класс балансировщика входящих сетевых запросов (передаётся в параметре `spec.loadBalancerClass` заказанного сервиса с типом `LoadBalancer`).
```

valeria-brazhnikova (approved):

PR #21298, opened 2026-07-10
[dhctl] fix kube proxy checker client lifecycle

mary-ppv:
## Description

Fixed the Kubernetes connectivity check used by dhctl converge when validating control-plane node readiness.

The change:

* validates that Kubernetes initialization parameters, the SSH provider, and base provider settings are configured before they are used;
* returns descriptive errors instead of panicking when required checker dependencies are missing;
* stops only the Kubernetes proxy created by the readiness check;
* no longer stops the shared SSH client from inside KubeProxyChecker.IsReady;
* adds unit tests for the dependency validation paths.

During control-plane node replacement, dhctl converge checks SSH connectivity and Kubernetes API availability through a temporary Kubernetes proxy.

After completing the check, KubeProxyChecker.IsReady stopped not only the temporary proxy but also the shared SSH client obtained from the external SSH provider. This client could still be required by subsequent converge operations.

As a result, the following operations could continue using an already stopped Kubernetes or SSH connection and fail with errors such as:

```
cannot send request GET: use stopped kube client
```

A subsequent control-plane readiness check could also panic with a nil pointer dereference in KubeProxyChecker.IsReady instead of returning a descriptive error.

The checker now manages only the lifecycle of the Kubernetes proxy it creates. The shared SSH client remains active and is stopped by its actual owner during the common cleanup process.

The added dependency validation also prevents nil pointer dereferences and reports which required checker dependency was not configured.

The issue can interrupt dhctl converge during destructive control-plane changes, including control-plane node replacement and temporary scaling from a single-master to a multi-master configuration.

A failed converge can leave the control plane in an intermediate state and require manual recovery or another converge attempt. The fix is isolated to the readiness check cleanup logic and prevents the checker from stopping a shared connection that it does not own.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix 
summary: Prevent dhctl converge from stopping the shared SSH client during control-plane node readiness checks.
impact_level: default 
```

### DIFF
--- a/dhctl/pkg/operations/converge/infrastructure/hook/controlplane/kube-proxy.go
+++ b/dhctl/pkg/operations/converge/infrastructure/hook/controlplane/kube-proxy.go
@@ -20,6 +20,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/yaml"
 
 	libcon "github.com/deckhouse/lib-connection/pkg"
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
:red_circle: `e2e: Yandex.Cloud` for `deckhouse:fix/dhctl-kube-proxy-checker-nil-init-params` [failed](https://github.com/deckhouse/deckhouse/actions/runs/29101461372) in 45m50s.




  `ssh -i ~/.ssh/e2e-id-rsa  cloud-user@62.84.118.223
` - connect for debugging;

  `/e2e/abort yandex-cloud;WithoutNAT;containerd;1.34 29101461372 29101461372-1-con-1-34 /sys/deckhouse-oss/install:pr21298 cloud-user@62.84.118.223
` - for abort failed cluster




Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@62.84.118.223`


:red_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` failed in 44m43s.

</details>

PR #21330, opened 2026-07-13
[cni-cilium] Add dynamic MTU distribution support

maxihunter:
## Description
The eBPF DHCP server for pods (005-ebpf-dhcp-server.patch) now advertises the interface MTU as DHCP option 26. The value comes from a new DHCP_INTERFACE_MTU define, which WriteNodeConfig fills with cfg.RouteMTU — the same value the CNI plugin sets on the pod veth pair.

The reply size is unchanged: the options block grew from 51 to 55 bytes and is still padded to DHCP_VEND_SIZE (60). The code is compiled only under ENABLE_DHCPD, which the module sets only when the virtualization module is enabled.

Unlike the KubeVirt DHCP server it replaces, ours never sent option 26, so a guest configured over DHCP ended up with the default 1500 MTU while the actual pod veth is smaller (1450 with VXLAN). Packets above the real MTU are then dropped silently, without ICMP — a black hole for large UDP and DF traffic rather than a visible error.

KubeVirt also passes the MTU through libvirt (host_mtu), but only for virtio NICs with a recent enough driver; for other NICs and guest drivers the MTU had to be set manually inside the VM. Now the value is delivered over DHCP as well and follows the node MTU automatically.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: cni-cilium
type: fix
summary: Cilium DHCP server is now sends MTU as option, so DHCP clients can apply it for their own interfaces
impact_level: default
```

### DIFF
--- a/docs/site/_includes/getting_started/global/partials/STEP_INSTALL_SCHEMA.liquid
+++ b/docs/site/_includes/getting_started/global/partials/STEP_INSTALL_SCHEMA.liquid
@@ -122,7 +122,7 @@ Scheme of Deckhouse installation in a private environment:<br />
    - 16 GB of RAM
    - 100 GB of disk space
    {%- if page.platform_code == 'aws' %}
-   - Linux kernel version >= `5.11`
+   - Linux kernel version >= `5.8`
    {%- endif %}
    {% endofftopic %}
 {%- elsif page.platform_code == 'zvirt' %}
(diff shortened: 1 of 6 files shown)


z9r5 (approved):

docs/site/_includes/getting_started/global/partials/STEP_INSTALL_SCHEMA.liquid
```diff
@@ -122,7 +122,7 @@ Scheme of Deckhouse installation in a private environment:<br />
    - 16 GB of RAM
    - 100 GB of disk space
    {%- if page.platform_code == 'aws' %}
-   - Linux kernel version >= `5.8`
+   - Linux kernel version >= `5.11`
```
z9r5: @maxihunter Please fix STEP_INSTALL_SCHEMA_RU.liquid also

maxihunter: Bump kernel version has been removed from this PR

AndreyPavlovFlant (approved):

PR #21393, opened 2026-07-16
[deckhouse] fix: preserve crd vendor extensions

fuldaxxx:
## Description

Bump `github.com/deckhouse/module-sdk` to the version that applies CRDs to the
cluster verbatim, preserving custom vendor extensions in the OpenAPI schema
(`x-*`). See deckhouse/module-sdk#122.

The only deckhouse-side change is updating the `000-common` `ensure_crds` golden
test to match the corrected installer output:

- `x-*` vendor extensions (`x-description`, `x-doc-default`) are now preserved;
- `preserveUnknownFields: false` is kept verbatim;
- the spurious empty `status:` block (an artifact of the previous typed
  round-trip) is no longer emitted.

The CRD installer used to decode CRDs into the typed
`apiextensionsv1.CustomResourceDefinition` before applying them.
`JSONSchemaProps` has typed fields only for the known upstream `x-kubernetes-*`
keys and no catch-all, so any non-standard `x-*` extension was silently dropped:
`x-kubernetes-sensitive-data`, `x-ui-*`, `x-kubernetes-patch-strategy`,
`x-kubernetes-patch-merge-key`, etc.

The loss was silent and sticky: the trimmed CRD stayed valid, so `apply`
succeeded without error, and because the read-back path was also typed, the diff
compared two equally-trimmed specs and never triggered an update — so a re-run
didn't self-heal. In particular `x-kubernetes-sensitive-data: true` never reached
the cluster CRD, so consumers reading the CRD from the API server couldn't tell
which fields must be redacted.

The fix itself lives in module-sdk (deckhouse/module-sdk#122); the installer now
acts as a transport and lets the API server own schema validation. The
`ensure_crds` golden test previously encoded the old lossy output, so it is
updated here to reflect the correct behavior — the test is catching up to the
fix, not the code breaking the test.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Apply CRDs as unstructured so custom OpenAPI vendor extensions (`x-*`) are no longer silently dropped on install.
impact_level: low
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,7 +63,7 @@ require (
 	github.com/deckhouse/deckhouse/pkg/metrics-storage v0.3.0
 	github.com/deckhouse/deckhouse/pkg/registry v0.0.0-20260525111533-9e5ba68242f7
 	github.com/deckhouse/lib-dhctl v0.21.0
-	github.com/deckhouse/module-sdk v0.12.1-0.20260716150856-69f81b203c3c
+	github.com/deckhouse/module-sdk v0.12.1-0.20260720012432-54a05551f06a
 	github.com/ettle/strcase v0.2.0
 	github.com/evanphx/json-patch v5.9.11+incompatible
 	github.com/fsnotify/fsnotify v1.9.0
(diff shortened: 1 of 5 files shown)


deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:fix/crd-installer-preserve-vendor-extensions` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/29801221614) in 32m54s.



GCP-WithoutNAT-Containerd-1.34 - Connection string: `ssh  user@34.40.84.66`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.34` succeeded in 31m50s.

</details>

RomanenkoDenys (approved):

PR #21406, opened 2026-07-17
[deckhouse-controller] wire module config settings to package runtime

diyliv:
## Description
Wire `ModuleConfig` settings changes into the package runtime (module v2 track) so that runtime settings updates and enable/disable no longer require a controller restart.

Key changes:
- Register embedded and global modules in the scheduler (`AddNode`) so `Reschedule` triggers the Configure → Startup → Run pipeline on settings change
- Route `ModuleConfig` reconcile through the v2 runtime: call `UpdateModulesSettings` instead of the legacy addon-operator `HandleEvent` for v2-managed modules
- On `ModuleConfig` deletion, reset pending settings and clear the enabled intent without removing the embedded package

### Example

| Plan ref | Check | Result |
|----------|-------|--------|
| G1.1 | `GetEnabledScriptDescriptor` on global module | ✓ global satisfies `schedule.Package` interface |
| G1.2 | `node-manager` skipped in `dummyModules` | ✓ absent from scheduler |
| G1.3 | Embedded modules registered in scheduler | ✓ cni-cilium and others: `state: idle`, subscribed to global |
| G1.4 | Global node in scheduler (`order: 0`) | ✓ `state: active`, all embedded modules as subscribers |
| G2 | v1 `HandleEvent` not called for v2-managed modules | ✓ cni-cilium → v1 path; global → v2 path (`UpdateModulesSettings`) |
| G5 | `loadInitialConfiguration` applies settings correctly | ✓ global `running: true`, per-module settings stored |

Currently, module v2 settings only reach the runtime on controller startup via `loadInitialConfiguration`. Any runtime `ModuleConfig` change (settings, version, enabled) has no effect — the scheduler never sees it. This PR wires the config controller into the runtime so every `ModuleConfig` reconcile pushes settings into the per-package store and triggers the scheduler to re-evaluate.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: deckhouse-controller
type: feature
summary: Wire ModuleConfig settings changes into the package runtime for module v2.
impact_level: low
```


deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -209,11 +219,26 @@ func (r *reconciler) handleModuleConfig(ctx context.Context, moduleConfig *v1alp
...
 
-	// send an event to addon-operator only if the module exists, or it is the global one
-	basicModule := r.moduleManager.GetModule(moduleConfig.Name)
-	if moduleConfig.Name == moduleGlobal || basicModule != nil {
-		r.logger.Debug("send event to operator", slog.String("name", moduleConfig.Name), slog.Bool("enabled", moduleConfig.IsEnabled()))
-		r.handler.HandleEvent(moduleConfig, config.EventUpdate)
+	// Route settings to the appropriate runtime engine.
+	// v2-managed modules (embedded or downloaded via packages) use
+	// the package runtime; everything else falls back to addon-operator.
+	if r.packageRuntime.HasModule(moduleConfig.Name) || moduleConfig.Name == moduleGlobal {
```
ldmonster: 🔴 **Blocker:** this v2 branch runs unconditionally, but the v2 engine (`pkgRuntime.Run()`) only starts under `if app.PackageSystemEnabled()` (`controller.go:364-367`), which defaults to `false`. With the flag off (the production default) `HasModule` is `true` for every embedded module — `loadEmbedded` populates `r.modules` unconditionally — so we take this path and skip `HandleEvent` to addon-operator, yet the scheduler goroutine isn't running, so `UpdateModulesSettings`→`Reschedule` goes nowhere.

Result: ModuleConfig settings/enabled/version changes for all embedded modules (and `global`) silently stop applying in production. Please gate this branch behind `app.PackageSystemEnabled()` (thread the flag into the reconciler) so v1/addon-operator stays the path when the flag is off.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -411,8 +436,17 @@ func (r *reconciler) processModule(ctx context.Context, moduleConfig *v1alpha1.M
...
 func (r *reconciler) deleteModuleConfig(ctx context.Context, moduleConfig *v1alpha1.ModuleConfig) (ctrl.Result, error) {
-	// send event to addon-operator
-	r.handler.HandleEvent(moduleConfig, config.EventDelete)
+	// Route delete to the appropriate runtime engine.
+	if r.packageRuntime.HasModule(moduleConfig.Name) || moduleConfig.Name == moduleGlobal {
+		// v2 path: reset settings and clear the enabled intent.
+		// For embedded modules, deletion of ModuleConfig does not mean
+		// package removal — only settings/intent are reset (G4).
+		r.logger.Debug("reset v2 module settings on delete", slog.String("name", moduleConfig.Name))
+		r.packageRuntime.UpdateModulesSettings(moduleConfig.Name, 0, make(addonutils.Values), nil)
```
ldmonster: Same gating issue as the update path applies here — delete routes to v2 and skips the addon-operator `EventDelete`.

Also: `settingsVersion` is passed as `0`. Please confirm `ConvertToLatest`/`ApplySettings` treat `0` as "reset to default" and not as a valid v0 schema; if the minimum schema version is `1`, pass that instead.

deckhouse-controller/internal/packages/runtime/embedded.go
```diff
@@ -139,6 +140,14 @@ func (r *Runtime) loadEmbedded(ctx context.Context) error {
 
 			r.mu.Lock()
 			r.modules[module.GetName()] = module
+			// Optimistically register before AddNode so a successful
+			// schedule can resolve it; if AddNode rejects the addition
+			// (dependency cycle), roll back the map entry.
+			if err = r.scheduler.AddNode(module); err != nil {
```
ldmonster: 🟠 `AddNode` runs inside `New()`, which is called unconditionally regardless of `PackageSystemEnabled`. If any embedded module has a dependency cycle / bad constraints, this errors → `New()` fails → the whole deckhouse-controller won't start, even when the package system is unused in prod.

Consider gating the graph build behind the flag, or making a rejected node non-fatal (log + skip, with the rollback you already do). The `040-node-manager` exclusion above is a symptom of this fragility.

deckhouse-controller/internal/packages/runtime/runtime.go
```diff
@@ -200,6 +200,13 @@ func New(cli kclient.Client, edition *edition.Edition, moduleManager moduleManag
 	// Initialize scheduler with enabling/disabling callbacks
 	r.buildScheduler(cli)
 
+	// Register global node: it serves as the order-0 barrier in the scheduler
+	// graph, ensuring every module waits for global to complete before running.
+	// Done after buildScheduler so AddNode can simulate the cycle-free graph.
+	if err := r.scheduler.AddNode(r.global); err != nil {
```
ldmonster: Same concern as the embedded `AddNode`: this runs in `New()` unconditionally, so a failure here hard-fails controller startup even with the package system disabled. Suggest gating behind the feature flag or non-fatal handling.

deckhouse-controller/internal/packages/runtime/embedded.go
```diff
@@ -47,6 +47,7 @@ const (
 var dummyModules = []string{
 	"000-common",
 	"007-registrypackages",
+	"040-node-manager", // TODO(pilot/module-settings): remove after scheduler integration
```
ldmonster, line 59: 🟡 Pilot hardcode in shared runtime code. Please make sure there's a tracking issue to remove it, and confirm node-manager stays served by v1 until the scheduler integration lands. (With the feature-flag gating suggested elsewhere, that's automatically true when the flag is off.)

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -121,13 +124,20 @@ func RegisterController(
 	return nil
 }
 
+type packageRuntime interface {
+	UpdateModulesSettings(name string, settingsVersion int, settings addonutils.Values, enabled *bool)
```
ldmonster, line 130: 🟡 `RemoveModule` is declared on this interface (and mocked) but never called in production code — delete uses `UpdateModulesSettings`. Either drop it from the interface or add a "reserved for future use" note.

ldmonster (commented):
## Review: wire module config settings to package runtime

Overall the change is clean and well-commented, but there is **one blocking defect** and **one serious startup-reliability concern**.

### 🔴 Blocker: v2 routing is not gated by the feature flag — module settings silently stop applying in production
The v2 engine only actually runs behind a feature flag: `pkgRuntime.Run()` is called inside `if app.PackageSystemEnabled()` (`controller.go:364-367`), and `PackageSystemEnabled()` defaults to `false` (`pkg/app/features.go:28`). `Run()` is what starts the goroutine consuming scheduler events (`runtime.go:599-621`) — without it `Reschedule` goes nowhere. But the config controller is registered unconditionally, `loadEmbedded` populates `r.modules` for every embedded module unconditionally (so `HasModule` is true for essentially everything), and the new routing branch deliberately skips `HandleEvent` to addon-operator. Net effect with the flag off (the production default): ModuleConfig settings / enabled / version changes for every embedded module and for `global` are routed to a scheduler that isn't running, and are never delivered to addon-operator (which is what actually runs the modules via helm in prod). Please gate the v2 branch behind `app.PackageSystemEnabled()` (thread the flag into the reconciler) so v1/addon-operator stays the path when the flag is off. See inline comments.

### 🟠 Serious: unconditional `AddNode` at startup can crash the whole controller
`AddNode(r.global)` and the per-module `AddNode` run inside `New()`, which is called regardless of the feature flag. A dependency cycle / bad constraints in any embedded module fails `New()` → `NewDeckhouseController` fails → the controller won't start, even when the package system is unused. See inline comments.

### 🟡 No tests for the new v2 behavior
`mockPackageRuntime` is wired in, but `buildReconciler` uses `newMockPackageRuntime(false)`, so every test takes the v1 path. No test sets `hasModuleResult=true` or asserts `settingsCalls` / that `HandleEvent` is *not* called in v2 mode. The core new behavior is uncovered.

### 🟡 Nit
The `loadEmbedded` doc comment still says "registers the ones enabled by the bundle … not marked enabled are skipped", but the code only filters `dummyModules` — the bundle is not consulted. Stale comment (not introduced by this PR, but it's right here).

### ✅ What's good
Separating settings storage from the enabled intent via `UpdateModulesSettings` (tri-state `enabled`); optimistic `r.modules[...] = module` before `AddNode` with rollback, consistent with `loadModule`; `HasModule` under `RLock`; moving `packageruntime.New(...)` earlier without creating duplicates.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: Fixed.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: Fixed.

deckhouse-controller/internal/packages/runtime/embedded.go
diyliv: Fixed.

deckhouse-controller/internal/packages/runtime/runtime.go
diyliv: Fixed.

deckhouse-controller/internal/packages/runtime/embedded.go
diyliv: Not fixed — this is a pilot hardcode that needs a tracking issue. With the feature-flag gating now in place, node-manager is automatically served by v1 when PackageSystemEnabled is off, which is consistent with the TODO's intent to remove the exclusion after scheduler integration lands.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: Fixed.

deckhouse-controller/pkg/controller/module-controllers/config/controller_test.go
```diff
@@ -131,6 +133,61 @@ func clearModuleConditionTimes(obj client.Object) {
...
+		mockPkg := newMockPackageRuntime(true) // HasModule returns true
+		suite.r.packageRuntime = mockPkg
+		suite.r.packageSystemEnabled = true
+
+		config := suite.moduleConfig("test-module")
+		_, err := suite.r.handleModuleConfig(context.TODO(), config)
+		require.NoError(suite.T(), err)
+
+		// Verify UpdateModulesSettings was called with correct arguments
+		require.Len(suite.T(), mockPkg.settingsCalls, 1)
```
ldmonster, line 184: The core guarantee of this branch (G2: v2-managed modules must **not** also get an addon-operator event) isn't asserted here. The test confirms `UpdateModulesSettings` was called, but a regression that also dispatched `HandleEvent` on the v2 path would still pass.

Consider asserting that the mock handler received no event on the v2 path, and conversely that it *does* receive one on the v1 paths (flag off / unknown module). That turns these three sub-tests into a real proof of the routing fork rather than a one-sided check.

Also missing: coverage for `deleteModuleConfig` (same v1/v2 fork, currently untested) and for the `moduleConfig.Name == moduleGlobal` branch.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -411,8 +441,21 @@ func (r *reconciler) processModule(ctx context.Context, moduleConfig *v1alpha1.M
...
+	// The v2 path is gated by the PackageSystemEnabled feature flag — when the
+	// flag is off, deletes are delivered to addon-operator for the v1 (Helm) path.
+	if r.packageSystemEnabled && (r.packageRuntime.HasModule(moduleConfig.Name) || moduleConfig.Name == moduleGlobal) {
+		// v2 path: reset settings and clear the enabled intent.
+		// For embedded modules, deletion of ModuleConfig does not mean
+		// package removal — only settings/intent are reset (G4).
+		// settingsVersion 0 signals "reset to defaults" to the runtime;
+		// the runtime treats version 0 as a no-version datum.
+		r.logger.Debug("reset v2 module settings on delete", slog.String("name", moduleConfig.Name))
+		r.packageRuntime.UpdateModulesSettings(moduleConfig.Name, 0, make(addonutils.Values), nil)
```
ldmonster: On delete, the v2 path clears the config-enabled intent (`SetConfigEnabled(name, nil)`), which reverts the module to its bundle/dynamic default — so for a module enabled-by-bundle the runtime's view becomes "enabled" again. Meanwhile the code below (`disableModule`) still flips the `Module` CR to disabled for non-system modules. Please double-check these two don't disagree for an embedded, enabled-by-bundle module (runtime thinks enabled, Module CR says disabled). If it's intended, a one-line comment would help.

Separately, this delete fork has no test — worth adding given it mirrors the create/update routing logic.

deckhouse-controller/internal/packages/runtime/embedded.go
```diff
@@ -139,6 +142,20 @@ func (r *Runtime) loadEmbedded(ctx context.Context) error {
...
+			// feature flag to avoid injecting startup failures when the
+			// scheduler is never started.
+			if app.PackageSystemEnabled() {
+				// Optimistically register before AddNode so a successful
+				// schedule can resolve it; if AddNode rejects the addition
+				// (dependency cycle), roll back the map entry.
+				if err = r.scheduler.AddNode(module); err != nil {
+					delete(r.modules, module.GetName())
+					r.mu.Unlock()
+					return fmt.Errorf("add node: %w", err)
```
ldmonster, line 175: With the flag on, a single module's `AddNode` failure (dependency cycle) propagates up through `loadEmbedded` → `New` and aborts controller startup entirely. Previously embedded load never called `AddNode`, so one malformed/cyclic embedded module now becomes fatal for the whole controller.

Is fail-fast the desired behavior here, or should a single bad embedded module be logged and skipped (rolled back from the map, as you already do) so the rest can still load? A cyclic *downloaded* module is a user error worth surfacing loudly, but an embedded one shipped with the release taking down startup feels riskier.

deckhouse-controller/pkg/controller/controller.go
```diff
@@ -314,7 +314,12 @@ func NewDeckhouseController(
 		return nil, fmt.Errorf("create deckhouse release controller: %w", err)
 	}
 
-	err = moduleconfig.RegisterController(runtimeManager, operator.ModuleManager, conversionsStore, edition, configHandler, operator.MetricStorage, exts, logger.Named("module-config-controller"))
+	pkgRuntime, err := packageruntime.New(runtimeManager.GetClient(), edition, operator.ModuleManager, dc, operator.MetricStorage, logger)
```
ldmonster, line 317: `packageruntime.New(...)` was moved up from below `RegisterObjectKeeperController` to here so its instance can be handed to the module-config controller. Note that `New` now has real side effects at construction when the flag is on (global `AddNode` + `loadEmbedded`, which itself registers scheduler nodes). So this reorder also moves those side effects earlier in `NewDeckhouseController`. Please confirm nothing that ran between the old and new positions is a prerequisite (e.g. object-keeper registration) — it looks fine, but the reorder is easy to miss in review.

ldmonster (commented):
## Code review

Overall this is a clean, well-scoped change and the inline comments explaining each `G*` decision are genuinely helpful. Routing `ModuleConfig` settings into the v2 runtime is done carefully and the whole v2 path is gated behind `PackageSystemEnabled()`, so the default (flag-off) behavior is unchanged. I traced the global settings path and confirmed it works: `global` is registered in the packages store (`embedded.go` `loadGlobal`) and `scheduleGlobal` reads `GetPendingSettings(global)` at Configure time, so routing the global `ModuleConfig` through `UpdateModulesSettings` does reach the module — good.

A few points worth addressing before merge, mostly around test coverage and startup robustness. None are blockers; details inline.

### Summary of points
1. **Test gap — the core G2 guarantee is not asserted.** The v2 routing test checks that `UpdateModulesSettings` was called, but never asserts that the addon-operator `HandleEvent` was *not* called. Since G2 ("avoid double-application") is the whole point of the branch, the test should verify the v1 handler stays silent on the v2 path (and fires on the v1 path).
2. **No coverage for the delete path or for global routing.** `deleteModuleConfig` gained the same v1/v2 fork but has no test; `moduleGlobal` routing is untested too.
3. **One bad embedded module now aborts controller startup** when the flag is on (see `loadEmbedded`).
4. **`packageruntime.New` moved earlier** and now performs scheduler side effects at construction — worth confirming the new init order is safe.
5. **Minor:** the reconciler captures `packageSystemEnabled` once as a field while the runtime calls `app.PackageSystemEnabled()` live. It's an env var so behavior is consistent, but the mixed style is slightly confusing — consider standardizing.

deckhouse-controller/pkg/controller/module-controllers/config/controller_test.go
diyliv: Fixed.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: Fixed.

deckhouse-controller/internal/packages/runtime/embedded.go
diyliv: Fixed.

deckhouse-controller/pkg/controller/controller.go
diyliv: Not fixed — checked, no code change needed. Between the old and new position of packageruntime.New only moduleconfig.RegisterController runs; RegisterObjectKeeperController does not depend on pkgRuntime and is not a prerequisite for it. ConvergeState.SetOnConvergeStart/Finish (PauseScheduler/ResumeScheduler) and pkgRuntime.Run() are registered later regardless of this order. No ordering hazard found.

deckhouse-controller/internal/packages/runtime/embedded.go
diyliv: Reverted — kept fail-fast behavior as-is. Added a TODO above loadEmbedded documenting the tradeoff (controller currently fails to start if even one embedded module cannot be loaded) so it is tracked for a follow-up rather than silently masked by a log+skip.

ldmonster (commented):
## Review: wiring ModuleConfig settings into the package runtime

I checked the implementation against the plan (module-config-controller for the packages/module-v2 track). Build is green and `go test ./deckhouse-controller/pkg/controller/module-controllers/config/...` passes.

### Conformance with the plan

| Plan item | Status |
|---|---|
| Variant B (extend the existing `module-controllers/config`) | done — the plan allows it for the pilot; target variant A (separate `packages-config-controller`) is still open |
| Step 1: local runtime interface in the controller | done (`UpdateModulesSettings` + `HasModule`; `RemoveModule`/`GetStatus` dropped, which is fine — they are not used) |
| Step 3: apply settings on create/update, tri-state `enabled` passed through | done |
| Step 4: delete / disable handling | done, but deviates from the plan — see the inline comment on `deleteModuleConfig` |
| Step 5: wiring in `controller.go` | done |
| Step 6: unit tests | done, and they assert both sides of the v1/v2 fork (G2) for create and delete |
| Blocker G1 (embedded modules registered in the scheduler) | closed in this PR (`AddNode` for embedded modules + the global node) |
| Step 3 optional pre-validation via `ValidatePackageSettings` | not implemented — see inline comment |

### Blocking issues

Two of them come from the same root cause: the routing fork is exclusive (`if v2 ... else v1`), but `global` and `deckhouse` are configs whose values are consumed by addon-operator itself, not only by the module they name. For those two names the delivery must stay dual (or they must be excluded from the v2 fork). Details inline on `handleModuleConfig`.

The third one is the `enabled = &false` on ModuleConfig deletion, which makes deleting a ModuleConfig destructive for a module that is enabled by the bundle. Details inline on `deleteModuleConfig`.

### Non-blocking note not attached to a diff line

`loadInitialConfiguration` doc comment ([deckhouse-controller/pkg/controller/controller.go#L476-L481](https://github.com/deckhouse/deckhouse/blob/feature/module-config-controller-v2/deckhouse-controller/pkg/controller/controller.go#L476-L481)) still says that at this point no package is tracked yet, so the per-package settings are dropped and only the enabled intent is recorded. That is no longer accurate: `packageruntime.New` has already run `loadGlobal`/`loadEmbedded`, which register `global` and every embedded module in the packages store, so for those packages the settings from this loop do land. Worth updating the comment so the next reader does not rely on a guarantee that no longer holds.

Overall the design and the tests look good; items 1-3 below should be resolved before merge.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -209,11 +221,29 @@ func (r *reconciler) handleModuleConfig(ctx context.Context, moduleConfig *v1alp
...
-	if moduleConfig.Name == moduleGlobal || basicModule != nil {
-		r.logger.Debug("send event to operator", slog.String("name", moduleConfig.Name), slog.Bool("enabled", moduleConfig.IsEnabled()))
-		r.handler.HandleEvent(moduleConfig, config.EventUpdate)
+	// Route settings to the appropriate runtime engine.
+	// v2-managed modules (embedded or downloaded via packages) use
+	// the package runtime; everything else falls back to addon-operator.
+	// The v2 path is gated by the PackageSystemEnabled feature flag:
+	// when the flag is off, the scheduler goroutine is not started and
+	// settings must reach addon-operator for the v1 (Helm) path.
+	if r.packageSystemEnabled && (r.packageRuntime.HasModule(moduleConfig.Name) || moduleConfig.Name == moduleGlobal) {
```
ldmonster, line 238: **Blocker 1 — the `global` ModuleConfig stops reaching addon-operator.**

The `|| moduleConfig.Name == moduleGlobal` clause sends the global config into the v2 branch, so `r.handler.HandleEvent` is skipped for it. `HandleEvent` is the only producer of addon-operator's `GlobalKubeConfig` (`confighandler/handler.go#L72-L76`), and every module that addon-operator still renders reads global values from there.

The v2 side itself works — global settings are stored in the packages store and picked up on reschedule (`internal/packages/runtime/runtime.go#L663`) — but with the flag on, a runtime change of the global ModuleConfig no longer updates addon-operator's global values at all; they stay frozen at whatever `LoadConfig` read at startup until the pod restarts.

Unlike ordinary modules, where the exclusive fork is exactly what G2 asks for, `global` needs dual delivery (v2 **and** addon-operator), or it should be excluded from the v2 fork for now.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -209,11 +221,29 @@ func (r *reconciler) handleModuleConfig(ctx context.Context, moduleConfig *v1alp
...
+		// v2 path: update settings in the package runtime and skip
+		// the addon-operator event to avoid double-application (G2).
+		r.logger.Debug("update v2 module settings", slog.String("name", moduleConfig.Name), slog.Bool("enabled", moduleConfig.IsEnabled()))
+		r.packageRuntime.UpdateModulesSettings(
+			moduleConfig.Name,
+			moduleConfig.Spec.Version,
+			addonutils.Values(moduleConfig.Spec.Settings.GetMap()),
+			moduleConfig.Spec.Enabled,
+		)
```
ldmonster: **Blocker 2 — the `deckhouse` ModuleConfig stops feeding `deckhouseConfigCh`, so Deckhouse settings freeze at startup values.**

`deckhouse` is an embedded module (`modules/002-deckhouse/module.yaml` -> `name: deckhouse`), so with the feature flag on `HasModule("deckhouse")` is true and every reconcile takes this v2 branch and skips `handler.HandleEvent`. But `HandleEvent` is the only place that does `h.deckhouseConfigCh <- values` (`confighandler/handler.go#L94-L96`), which drives `syncDeckhouseSettings` (`pkg/controller/controller.go#L513-L525`) -> `settingsContainer` and the embedded update policy.

Concrete failure: with the flag on, a user edits `releaseChannel` / update windows / `logLevel` in ModuleConfig/deckhouse; the deckhouse-release controller keeps using the values read at startup by `LoadConfig` and never sees the change until a restart.

Same fix shape as Blocker 1: keep delivering the `deckhouse` config to addon-operator in addition to the runtime (or special-case it out of the v2 fork).

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -209,11 +221,29 @@ func (r *reconciler) handleModuleConfig(ctx context.Context, moduleConfig *v1alp
...
+	// settings must reach addon-operator for the v1 (Helm) path.
+	if r.packageSystemEnabled && (r.packageRuntime.HasModule(moduleConfig.Name) || moduleConfig.Name == moduleGlobal) {
+		// v2 path: update settings in the package runtime and skip
+		// the addon-operator event to avoid double-application (G2).
+		r.logger.Debug("update v2 module settings", slog.String("name", moduleConfig.Name), slog.Bool("enabled", moduleConfig.IsEnabled()))
+		r.packageRuntime.UpdateModulesSettings(
+			moduleConfig.Name,
+			moduleConfig.Spec.Version,
+			addonutils.Values(moduleConfig.Spec.Settings.GetMap()),
+			moduleConfig.Spec.Enabled,
```
ldmonster: Nit (related to the global routing above): for `global` this ends up calling `r.global.SetConfigEnabled("global", ...)`, which writes an entry into the intent map under a key no scheduling rule ever reads — there is no module named `global`, and the global module is always enabled by its own constraints. Harmless today, but it would be cleaner to route `global` through settings only and skip the enabled intent for it explicitly.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -209,11 +221,29 @@ func (r *reconciler) handleModuleConfig(ctx context.Context, moduleConfig *v1alp
...
+	} else {
+		// v1 path: existing addon-operator event dispatch.
+		basicModule := r.moduleManager.GetModule(moduleConfig.Name)
+		if moduleConfig.Name == moduleGlobal || basicModule != nil {
+			r.logger.Debug("send event to operator", slog.String("name", moduleConfig.Name), slog.Bool("enabled", moduleConfig.IsEnabled()))
+			r.handler.HandleEvent(moduleConfig, config.EventUpdate)
+		}
 	}
 
 	if err := r.refreshModuleConfig(ctx, moduleConfig.Name); err != nil {
```
ldmonster, line 276: Optional item from step 3 of the plan is missing: no `ValidatePackageSettings` call before handing the settings to the runtime. As it stands, invalid settings only surface later as a failed `configure` task inside the runtime, and the ModuleConfig status/condition says nothing about it — while the v1 path does report validation problems. Worth adding (the admission webhook already uses `Runtime.ValidatePackageSettings`, so this is only about a defence-in-depth check plus a status message for configs that got in some other way).

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
```diff
@@ -411,8 +441,30 @@ func (r *reconciler) processModule(ctx context.Context, moduleConfig *v1alpha1.M
...
+		// enabled is explicitly set to false rather than nil (which would
+		// fall back to the bundle default) so the runtime's view of the
+		// module stays in sync with disableModule below, which always sets
+		// Module.ConditionEnabledByModuleConfig to False on delete for
+		// non-system modules. Without this, an enabled-by-bundle module
+		// would be reported as disabled in the Module CR while the v2
+		// scheduler kept treating it as enabled and running it.
+		r.logger.Debug("reset v2 module settings on delete", slog.String("name", moduleConfig.Name))
+		disabled := false
+		r.packageRuntime.UpdateModulesSettings(moduleConfig.Name, 0, make(addonutils.Values), &disabled)
```
ldmonster, line 490: **Blocker 3 — `enabled = &false` on delete diverges from both the plan and v1 semantics, and is destructive for bundle-enabled modules.**

The plan (step 4) prescribes clearing the intent (`nil`) so the module falls back to its bundle default; this passes an explicit `false` instead. In the scheduler, a `false` intent is not a soft vote: the `dynamic` rule turns it into a **hard Forbid** that no other rule — bundle, enabled script, or floor — can override (`internal/packages/schedule/rule/dynamic/rule.go#L44-L84`). That leads to `disablePackage` -> `taskdisable` (`internal/packages/runtime/runtime.go#L733-L755`).

Failure scenario: an embedded module is enabled by the bundle, and its ModuleConfig exists only to carry settings. The user deletes that ModuleConfig expecting a fall back to the bundle default — which is what v1 does, since addon-operator recomputes the enabled state from the bundle. With this change v2 permanently disables the module and uninstalls its release.

Also, the justification in the comment ("stays in sync with `disableModule` below") does not hold for the two system names: for `deckhouse` and `global` `deleteModuleConfig` returns early at the `skip system modules` check further down, so `disableModule` never runs — yet by then a `false` intent has already been recorded for the `deckhouse` module, which in v2 means "disable the deckhouse module".

Suggestion: pass `nil` here as the plan says, and if the Module CR condition then disagrees with the runtime, fix that on the status side (or add the missing bundle-default recomputation) rather than by pinning the intent to `false`. The current behaviour is locked in by a test, so if it is a deliberate decision it should be re-agreed as part of G4 and documented as a behaviour change relative to v1.

deckhouse-controller/internal/packages/runtime/embedded.go
```diff
@@ -47,6 +48,7 @@ const (
 var dummyModules = []string{
 	"000-common",
 	"007-registrypackages",
+	"040-node-manager", // TODO(pilot/module-settings): remove after scheduler integration
```
ldmonster, line 59: `dummyModules` is applied unconditionally, not only under `PackageSystemEnabled`, so adding `040-node-manager` here changes behaviour outside the pilot as well: node-manager disappears from `r.modules` and from the status/packages stores in every configuration.

One visible consequence: `Runtime.ValidatePackageSettings("node-manager", ...)` now always returns "valid, package not loaded", i.e. the admission webhook silently stops validating node-manager settings against its OpenAPI schema.

The TODO acknowledges the temporary nature, but the exclusion should be gated by the feature flag (or the side effects should be called out explicitly), otherwise this ships as a silent validation regression.

deckhouse-controller/internal/packages/runtime/modules.go
```diff
@@ -158,15 +158,17 @@ func (r *Runtime) loadModule(ctx context.Context, repo registry.Remote, packageP
...
-		delete(r.modules, module.GetName())
-		span.SetStatus(codes.Error, err.Error())
-		return "", status.NewError("DependencyCycle", err)
+	if app.PackageSystemEnabled() {
+		if err = r.scheduler.AddNode(module); err != nil {
+			delete(r.modules, module.GetName())
+			span.SetStatus(codes.Error, err.Error())
+			return "", status.NewError("DependencyCycle", err)
+		}
 	}
```
ldmonster, line 179: Gate mismatch (latent, no caller today): `AddNode` for downloaded modules is now gated on `PackageSystemEnabled`, while the module-v2 controllers that would drive this path are gated on a different flag, `ModulePackagesEnabled` (`pkg/controller/controller.go#L388-L406`).

With `ModulePackagesEnabled=true` and `PackageSystemEnabled=false`, a module would be loaded into `r.modules` but never registered in the scheduler, so it would never go through Configure -> Startup -> Run — silently, with no error anywhere. `UpdateModule` currently has no non-test callers, so nothing breaks today, but the two gates should be aligned (or this one keyed off the flag that actually guards the module path) before that path is wired up.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: The plan said `nil`, but I went with `&false` on purpose after hitting this problem:

1. `SetConfigEnabled(name, nil)` lets the runtime fall back to bundle default, so the scheduler keeps the module running
2. `disableModule` right below flips `Module.ConditionEnabledByModuleConfig` to False
3. Module CR says disabled, scheduler keeps running it anyway

That's a silent state divergence. With `&false` both sides stay in sync: the module stops on both ends. Yes, a bundle-enabled module won't fall back to its bundle default on delete, but I'd rather have that than a module that looks disabled while still chewing resources.

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: fixed

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: fixed

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: fixed

deckhouse-controller/pkg/controller/module-controllers/config/controller.go
diyliv: fixed

deckhouse-controller/internal/packages/runtime/embedded.go
diyliv: fixed

deckhouse-controller/internal/packages/runtime/modules.go
diyliv: fixed

PR #21372, opened 2026-07-15
[node-manager] Migrate get_crds hook into node-controller and move machine objects along.

d2285:
## Description

This PR migrates the core of the `get_crds` shell-operator hook into node-controller. The hook is not removed — only its heavy, per-NodeGroup rendering logic moves to the controller; the remaining lightweight parts stay in the hook because other components still depend on them.

Node-controller now renders (driven directly by live `NodeGroup` / `InstanceClass` objects):
- **CAPI**: the `MachineDeployment` and the provider `MachineTemplate`. The instance-class checksum drives the `MachineTemplate` / bootstrap-`Secret` names; provider templates are read from the `kube-system/d8-cloud-provider-<type>-capi` Secret published at the 030 step, so node-controller stays generic and works for out-of-tree providers.
- **MCM**: the `MachineClass` CR and the `MachineDeployment`. The checksum lives in the `checksum/machine-class` annotation (object names are checksum-independent), so node-controller is the sole checksum author with no two-writer risk.

**Hook reduction**: the heavy per-NodeGroup work — `capacity` computation, `status` writing, and the `instance-class` validation checks — is removed. The hook is reduced to a passthrough of the `NodeGroup` spec plus `instance-class` defaulting and zone/engine resolution for the remaining Helm consumers. Invalid NodeGroups are kept out of the Helm render. Node-controller reads the `manual-rollout-id` annotation directly and computes the machine-class checksum itself, so that no longer flows through Helm values.

## Why do we need it, and what problem does it solve?

Once that logic lives in the controller, the objects derived from it should follow — node-controller becomes the owner of the `MachineDeployment` (CAPI + MCM) and the MCM `MachineClass`, driven by live `NodeGroup` / `InstanceClass` objects instead of hook-populated Helm values. This removes the shell-operator hop from the NodeGroup → MachineDeployment path and lets the controller react to changes via watches.

## Performance & resource work

Moving rendering into a controller exposed several latency and CPU regressions that were investigated and fixed (bulk MachineDeployment creation, 30 NodeGroups):
- **Two-pass engine wait**: the CAPI MachineDeployment was skipped until the status controller published `status.engine`, adding a full foreign-controller cycle. The engine is now derived in-reconcile (`derived_status.ComputeEngine`); `status.engine` stays the pin.
[...]

PR #21415, opened 2026-07-18
[node-manager] immutable NodeGroups: NodeConfig CRD, reconciler and CAPI bootstrap

borg-z:


borg-z:
Superseded by #21557 — immutable-OS changes re-ported onto the node-controller get_crds-migration base (#21372). Closing to avoid confusion.

PR #21421, opened 2026-07-20
[multitenancy-manager][docs] fix grantedResource.apiGroup in GrantableClusterResourceDefinition examples

IlshatKhisamov:
## Description

Fixed `GrantableClusterResourceDefinition` examples: replaced invalid `spec.grantedResource.apiVersion` with the correct `apiGroup`.

Documentation examples used `apiVersion: storage.k8s.io/v1` instead of `apiGroup: storage.k8s.io`. Per the CRD, the field is `apiGroup`; the version is resolved via discovery. The typo could lead users to copy an invalid manifest.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: fix
summary: Fix GrantableClusterResourceDefinition docs — use grantedResource.apiGroup instead of apiVersion.
impact_level: low

### DIFF
--- a/docs/documentation/pages/admin/configuration/multitenancy/USAGE.md
+++ b/docs/documentation/pages/admin/configuration/multitenancy/USAGE.md
@@ -323,7 +323,7 @@ metadata:
   name: storageclasses
 spec:
   grantedResource:
-    apiVersion: storage.k8s.io/v1
+    apiGroup: storage.k8s.io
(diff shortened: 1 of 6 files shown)


z9r5 (approved):

z9r5:
Cherry-picked into release-1.76

PR #21437, opened 2026-07-20
[control-plane-manager] Bootstrap node with bashible

trofimovdals:
## Description

Replaces the custom `join.sh` bootstrap path for `VirtualControlPlane` worker nodes with a rendered bashible `bootstrap.sh` built from candi templates (`01-bootstrap-prerequisites`, `get-phase2`) plus a VCP-specific glue header.

- renders `bootstrap.sh.tpl` into the VCP config Secret and publishes it as `bootstrap.sh` instead of `join.sh`
- seeds bashible bootstrap context for nested nodes: ALB VIP `/etc/hosts`, bootstrap token, tenant CA, and direct access to parent `registry-packages-proxy`
- mirrors parent `bashible-bashbooster` / bashible files into the nested cluster and extends bootstrapper RBAC so nodes can fetch secrets before `kubelet.conf` exists
- temporarily approves nested `kubernetes.io/kubelet-serving` CSRs and cleans `bashible-uninitialized` / `uninitialized` taints in the VCP reconciler, because `node-manager` is not running in the nested cluster yet
- enables nested `kube-dns` and adds a konnectivity `controlplane` ALB backend used by the bootstrap path

The previous VCP join path used a hand-rolled `join.sh` that duplicated package install / kubelet bring-up outside bashible. That diverged from the normal Deckhouse node bootstrap, was harder to keep in sync with candi changes, and could not reuse bashible bundles, bashbooster, or packages-proxy wiring.

In a nested cluster there is also no in-cluster `registry-packages-proxy` reachable through kube-apiserver discovery, and no tenant `node-manager` to approve kubelet-serving CSRs or clear bashible uninitialized taints after the first run.

This change switches VCP node bootstrap onto the standard bashible phase-1/phase-2 flow, with direct RPP (`packagesProxy.direct`) for package fetch. CSR approval and post-bootstrap taint/label cleanup are temporary stand-ins for `node-manager` and should be removed once `node-manager` runs in the tenant cluster.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/reconcile_bashible-apiserver.go
```diff
@@ -144,9 +154,64 @@ func (r *reconciler) reconcileBashibleApiserver(
...
+	}
+	out := make([]corev1.Taint, 0, len(taints))
+	for _, t := range taints {
+		if _, ok := drop[t.Key]; !ok {
+			out = append(out, t)
+		}
+	}
+	return out
+}
+
```
AmazinMax: move to the end of the file, to save aligning with orchestrator method `reconcileBashibleApiserver`

AmazinMax (approved):

PR #21461, opened 2026-07-21
[deckhouse] atomically install modules and re-download incomplete versions

Fkuloff:
## Description
An interrupted module download - a registry/Harbor connection drop, or a reboot of the leader Deckhouse pod mid-pull - could leave an empty or partial `<downloaded>/<module>/<version>` directory. The symlink installer then trusted that directory forever: `Restore` skipped the re-download on a bare existence check and pointed the module symlink at the empty directory, so the module stayed broken across restarts and never recovered.

This hardens the symlink backend:
- `registry.download` unpacks into a temporary sibling directory on the same filesystem and renames it onto the final path, so an interrupted extract cannot publish a partial directory.
- `Install` materializes the version directory atomically (copy into a sibling, then rename); `Restore` re-downloads when the version directory is missing or empty instead of trusting bare existence, which heals a directory left broken by an older binary on the next reconcile.

A single registry disconnect or leader-pod restart during a module pull could leave an external module permanently unavailable - an empty version directory behind a live symlink - with no self-healing; the operator kept the broken module in place until the directory was deleted by hand. The rest of Deckhouse stays up, but the affected module never converges. The fix makes recovery automatic.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: atomically install modules and re-download incomplete versions
impact_level: default
```

### DIFF
--- a/deckhouse-controller/internal/module/installer/symlink/installer.go
+++ b/deckhouse-controller/internal/module/installer/symlink/installer.go
@@ -95,15 +95,10 @@ func (i *Installer) Install(ctx context.Context, module, version, tempModulePath
 	// /deckhouse/downloaded/<module>/<version>
 	versionPath := filepath.Join(modulePath, version)
 
-	// Remove old version if exists (for atomic update)
-	if _, err := os.Stat(versionPath); err == nil {
-		if err = os.RemoveAll(versionPath); err != nil {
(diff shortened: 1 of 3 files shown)


Fkuloff:
Closing in favor of a main PR (to be backported to 1.76); 1.75 is handled by a separate PR.

PR #21475, opened 2026-07-21
[deckhouse-controller] fix: "frozen chain" bug in the module-source controller

ldmonster:
## Description

Fixes a "frozen chain" in the module-source controller: once the target release already existed with a matching checksum, the fetch was skipped forever, so intermediate `ModuleReleases` mirrored into the registry later were never created and the step-by-step update could not proceed.

- Split the fetch decision into `releaseEnsureAllowed` (pure policy over already-fetched data — experimental / active-source / enabled, no cluster I/O) and `releaseChainToTargetComplete`, which reports whether the in-cluster `ModuleReleases` form a continuous chain from the deployed release up to the target and re-runs the idempotent step-by-step fetch on a gap.
- `isSequentialReleasePair` is the single "may we jump directly from release A to release B" rule, shared by the completeness check and the fetcher's starting-point selection. A `from-to` constraint bridges a gap only when it matches what the release updater enforces before it deploys a jump (`releaseupdater.getFirstCompliantRelease`): its `to` equals the higher release's own major.minor and the lower version is within `[from, to)`. This keeps the source controller from reporting a chain complete that the release updater then refuses to walk — a `from-to` whose `to` overshoots the release's own minor previously left the target stuck in Pending.
- `releaseChainToTargetComplete` validates release versions up front, so a malformed `Spec.Version` surfaces as a handled error instead of panicking the reconcile through `GetVersion`'s `semver.MustParse`.

No impact on critical cluster components (no restarts of ingress-controllers, control-plane, Prometheus, etc.).

A module could not complete its update: intermediate releases mirrored after the target were never created, and a `from-to` whose `to` did not point at the release blocked the step-by-step chain because the source controller and the release updater interpreted the same constraint differently. The frozen state does not self-heal.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Fixed a frozen module release chain by re-deriving the chain and aligning the from-to gap check with the release updater.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/pkg/controller/module-controllers/source/controller.go
+++ b/deckhouse-controller/pkg/controller/module-controllers/source/controller.go
@@ -503,7 +503,38 @@ func (r *reconciler) processModules(ctx context.Context, source *v1alpha1.Module
 			}
 		}
 
-		if r.needToEnsureRelease(source, module, availableModule, meta, exists, embeddedTargetSource) {
+		if !r.releaseEnsureAllowed(source, module, meta, embeddedTargetSource) {
+			availableModule.Checksum = meta.Checksum
+			availableModule.Version = meta.ModuleVersion
+			availableModules = append(availableModules, availableModule)
+			continue
+		}
+
+		// checksum changed or the target release is missing - ensure as usual.
+		ensure := availableModule.Checksum != meta.Checksum || !exists
+		if !ensure {
+			// The target release already exists and its checksum has not changed, but
+			// the step-by-step chain of ModuleReleases from the deployed release up to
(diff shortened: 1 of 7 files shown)


Fkuloff (approved):

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 6210b8eb37fb9ce08031ae8ede520fae2f5c959b to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/30259186250) for details.

PR #21521, opened 2026-07-23
[cloud-provider-dvp] Add sshCAKeys support for SSH CA-based node access

ssr-zna:
## Description
Adds an optional `sshCAKeys` field to the DVP provider cluster configuration (PCC) and to the v2 `nodes.parameters` ModuleConfig settings, propagated through the migration layer, layouts and terraform-modules down to cloud-init.

- OpenAPI schema (PCC + ModuleConfig) + RU docs.
- Go hooks (`dvp_cluster_configuration.go`, `migration.go`) propagate `sshCAKeys` alongside `sshPublicKey`; empty/absent values are omitted rather than leaving a stray empty list in the config.
- New shared `terraform-modules/cloudinit-merge` module: YAML-merges the SSH CA trust files into cloud-init without breaking the existing bashible `write_files`/`runcmd` payload on master-1/2/3 and static nodes (previously plain string concatenation, which risked colliding YAML keys). Falls back to the byte-identical legacy rendering when `sshCAKeys` is empty. Covered by a golden `cloudinit-merge.tftest.hcl`.
- New `NodeGroupConfiguration` template (`templates/ngc-ssh-ca.yaml`) handles CA key rotation on already-bootstrapped nodes via bashible's continuous reconciliation, since cloud-init can only ever apply the trust files once at first boot. Follows the same `bb-sync-file` + `bb-event-on` idiom already used by `node-manager/templates/fencing-agent` and `ee/fe/node-manager/templates/ngc-ca-certificate.yaml`.
- Go hook unit tests (`dvp_cluster_configuration_test.go`, `migration_test.go`) and Helm template tests (`module_test.go`) cover the new field's propagation and rendering.

DVP-provisioned nodes currently only trust a single static `sshPublicKey`. This makes rotating or revoking SSH access hard (requires re-provisioning nodes) and encourages sharing long-lived keys between operators. Trusting an SSH CA lets nodes accept short-lived certificates (e.g. issued by Vault) signed by that CA, without needing per-node key changes for every operator/rotation.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: cloud-provider-dvp
type: feature
summary: "Added support for trusting SSH CA-signed short-lived user certificates on DVP nodes via the new sshCAKeys parameter."
impact: "Optional field, disabled by default. Existing clusters without sshCAKeys render byte-identical cloud-init and Terraform plans (no diff, no node recreation). Setting it adds TrustedUserCAKeys to sshd on all master/static nodes at bootstrap and creates a NodeGroupConfiguration for CA rotation on already-running nodes."
impact_level: default
```


ssr-zna:
Superseded by #21527 (same branch, pushed directly to deckhouse/deckhouse now that I have org write access — this avoids the fork pull_request_target checkout restriction that was blocking Gitleaks/Security scan on this PR).

PR #21556, opened 2026-07-24
[admission-policy-engine] Add controller level check in rules

Suselz:
## Description

Security and operation policies now validate pod-creating controllers, not only the Pods they create. A workload that violates a policy is rejected on `kubectl apply` or `helm upgrade`, instead of being accepted and failing later when its Pods are denied.

Controller-level validation is on by default and can be turned off with `podSecurityStandards.controllerValidation`. Despite living in the `podSecurityStandards` section, the parameter applies to every constraint that reads a pod spec, including those generated from OperationPolicy and SecurityPolicy.

The remaining legacy-style ConstraintTemplates are migrated to the Rego `v1` format along the way, with no behaviour change.

### What reviewers should know

- Controller checks evaluate the **un-mutated** pod template, so a workload that passes at Pod level only because a mutating component fills a field in may be denied at the controller level. `D8RequiredResources` and `D8AllowedUsers` skip absent fields on controllers for that reason, per container.
- `security.deckhouse.io/skip-pss-check` is read from the reviewed object's own labels. On the pod template alone it no longer exempts the workload — the label has to be on the controller's top-level `metadata.labels` too. This cannot be resolved from the pod template: Gatekeeper evaluates `match.labelSelector` before any policy code runs.
- The webhook now intercepts CREATE, UPDATE and DELETE for controller kinds. DELETE is intentional: a policy must be able to forbid deleting a resource, and the rule only runs if the request reaches Gatekeeper. The trade-off is that while Gatekeeper is down, deleting such a resource is blocked along with creating it.
- Three bugs found on the way are fixed and listed in the changelog. Findings from three review rounds are addressed in the resolved threads.

### Resource cost

Per-rule cost on the Pod path is essentially unchanged: of 78 comparisons on identical fixtures, 53 moved by under 1%, and the most expensive templates by under 0.5%. A controller review costs a median of 12% more allocations than the same rule on a Pod, so it is a new review rather than a more expensive existing one.

Measured on a live cluster, controller-level auditing costs 15% of audit CPU. The audit loop already exceeded `--audit-interval` before this change — a Pod-only cycle took 59 s and 31 CPU-seconds on the same cluster — so the overrun is a property of the existing constraint set, not of this feature. Rego accounts for under 3% of an audit cycle; the rest is Gatekeeper's own per-evaluation overhead. The audit flags are unchanged here.

Methods, pitfalls and the full figures are recorded in `modules/015-admission-policy-engine/tools/MEASUREMENTS.md`, including one optimisation that was measured and rejected.

Some built-in ConstraintTemplates only check pods, so the user doesn't receive a warning when running helm install or kubectl apply for deployment. The check occurs later, when the pod is already attempting to be created, making debugging unintuitive for developers.

When using EnforcementAction: Deny, the deployment simply fails silently, without providing any feedback in stdout.
The suggestion is to create additional ConstraintTemplates to work with spec.template.spec.containers within controllers.

## Checklist
- [ ] e2e tests passed.

Cluster verification, with every workload submitted as `--dry-run=server`: controller-level denial works and names the reviewed kind; `skip-pss-check` behaves as described above; a denylist-only sysctls policy admits everything it does not forbid; `MustRunAs` leniency applies to controllers only, and the equivalent Pod is still denied.

## Changelog entries

```changes
type: feature
summary: Added a controller-level check to all rules that check pods.
impact: |
        Security and operation policies now validate pod-creating controllers
        (Deployment, StatefulSet, DaemonSet, Job, CronJob, ReplicationController)
        in addition to Pods. Set `podSecurityStandards.controllerValidation` to
        `false` to validate only Pods, as before.

        Controller checks evaluate the un-mutated pod template, so a workload that
        passes at Pod level only because a mutating component fills a field in may
        be denied at the controller level.

        A workload carrying `security.deckhouse.io/skip-pss-check` only on its pod
        template is now checked at the controller level. Set the label on the
        controller's top-level `metadata.labels` as well to exempt it completely.

        The webhook now intercepts CREATE, UPDATE and DELETE for these kinds, so
        that a policy can forbid deleting a resource. While Gatekeeper is
        unavailable, deleting such a resource is blocked along with creating it.
impact_level: high
---
type: fix
summary: Fixed D8AllowedSysctls ignoring a policy that sets only one of allowedUnsafeSysctls and forbiddenSysctls.
impact: |
        A SecurityPolicy that set only one of the two keys denied nothing at all.
        It now denies the sysctls it lists, so workloads that used a forbidden
        sysctl and were admitted before will be rejected. Check `forbiddenSysctls`
        in your SecurityPolicy resources before updating.
impact_level: high
---
type: fix
summary: A SecurityPolicyException with several allowedValues now honours all of them, not only the first.
---
type: fix
summary: D8AutomountServiceAccountToken now denies a pod that mounts the service account token explicitly, even when automountServiceAccountToken is set to false.
---
type: fix
summary: The Trivy provider now checks images of ephemeral containers, so kubectl debug is denied for an image with disallowed vulnerabilities.
```


deckhouse-BOaTswain:
:red_circle: `e2e: Yandex.Cloud` for `deckhouse:ape-rules-controllers` [failed](https://github.com/deckhouse/deckhouse/actions/runs/30247313158) in 54m7s.




  `ssh -i ~/.ssh/e2e-id-rsa  cloud-user@111.88.251.142
` - connect for debugging;

  `/e2e/abort yandex-cloud;WithoutNAT;containerd;1.34 30247313158 30247313158-1-con-1-34 /sys/deckhouse-oss/install:pr21556 cloud-user@111.88.251.142
` - for abort failed cluster




Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@111.88.251.142`


:red_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` failed in 52m24s.

</details>

Suselz:
/e2e/abort yandex-cloud;WithoutNAT;containerd;1.34 30247313158 30247313158-1-con-1-34 /sys/deckhouse-oss/install:pr21556 cloud-user@111.88.251.142

deckhouse-BOaTswain:
:green_circle: `destroy cluster: Yandex.Cloud` for `refs/heads/main` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/30252962166) in 5m52s.


:green_circle: `destroy cluster: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 5m29s.

</details>

deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:ape-rules-controllers` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/30268437138) in 37m5s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@51.250.10.46`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 35m51s.

</details>

deckhouse-BOaTswain:
:white_circle: `e2e: Yandex.Cloud` for `deckhouse:ape-rules-controllers` [cancelled](https://github.com/deckhouse/deckhouse/actions/runs/30268453517).

</details>

deckhouse-BOaTswain:
:red_circle: `e2e: DVP` for `deckhouse:ape-rules-controllers` [failed](https://github.com/deckhouse/deckhouse/actions/runs/30440204433) in 1h4m19s.




  `ssh -i ~/.ssh/e2e-id-rsa  debian@10.12.0.217
` - connect for debugging;

  `/e2e/abort dvp;WithoutNAT;containerd;1.34 30440204433 30440204433-1-con-1-34 /sys/deckhouse-oss/install:pr21556 debian@10.12.0.217
` - for abort failed cluster




DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.0.217`


:red_circle: `e2e: DVP, Containerd, Kubernetes 1.34` failed in 1h3m4s.

</details>

Suselz:
/e2e/abort dvp;WithoutNAT;containerd;1.34 30440204433 30440204433-1-con-1-34 /sys/deckhouse-oss/install:pr21556 debian@10.12.0.217

deckhouse-BOaTswain:
:green_circle: `destroy cluster: DVP` for `refs/heads/main` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/30453387463) in 5m7s.


:green_circle: `destroy cluster: DVP, Containerd, Kubernetes 1.34` succeeded in 4m34s.

</details>

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/required-probes.yaml
```diff
@@ -26,21 +26,30 @@ spec:
...
-          msg := get_violation_message(container, input.review, probe)
-        }
+              import data.lib.common.input_containers
 
-        probe_is_missing(ctr, probe) = true {
-          not ctr[probe]
-        }
+              violation contains {"msg": msg} if {
+                not input.review.operation == "DELETE"
+                container := input_containers[_]
```
AlwxSin: **Critical: this makes `D8RequiredProbes` unsatisfiable for any workload with an initContainer.**

The rule moved from `input.review.object.spec.containers[_]` to `lib.common.input_containers`, which is `containers + initContainers + ephemeralContainers`. That is a scope expansion unrelated to controller support.

The Kubernetes API rejects `livenessProbe` / `readinessProbe` / `startupProbe` on an initContainer unless it has `restartPolicy: Always` (sidecar), and rejects them on ephemeralContainers outright. So `probes: [livenessProbe]` + any classic initContainer produces a violation the user **cannot fix by editing the manifest** — the only escape is dropping the policy.

Suggest keeping the iteration over `containers` only, or over `containers` plus init containers that have `restartPolicy: Always`.

No case in `tests/test_cases/constraints/operation/required-probes/test-matrix.yaml` covers a workload with an initContainer, which is why this passes CI.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/priority-class.yaml
```diff
@@ -25,16 +25,25 @@ spec:
...
-          msg := sprintf("Pod <%v> has invalid priority class: %v, allowed: %v", [input.review.object.metadata.name, priorityClass, input.parameters.priorityClassNames])
-        }
+              import data.lib.common.pod_spec
 
-        contains(list, elem) {
-          list[_] = elem
-        }
+              violation contains {"msg": msg} if {
+                not input.review.operation == "DELETE"
+                priority_class := object.get(pod_spec, "priorityClassName", "")
```
AlwxSin: **Critical: every workload without an explicit `priorityClassName` now violates this policy.**

Old code was `priorityClass := input.review.object.spec.priorityClassName`. When the field was absent the reference was undefined, the rule body failed, and no violation was produced.

Now the `""` default is a concrete value: `list_contains(["foo","baz"], "")` is undefined, `not undefined` is true, so the violation fires with `has invalid priority class: `.

`priorityClassName` is **not** defaulted by the API server — the Priority admission plugin sets `spec.priority` and leaves `priorityClassName` empty. So this hits every pod and every pod template that does not set it explicitly, i.e. most user workloads in a namespace with this policy enabled.

Needs a `has_field` guard (or the bare reference back). The test matrix only covers `priorityClassName: foo` and `bar` — please add the unset case.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/image-pull-policy.yaml
```diff
@@ -27,14 +27,23 @@ spec:
...
-          provided := container.imagePullPolicy
-          required != provided
-          msg := sprintf("Container <%v> in your %v <%v> has invalid pull policy: <%v>", [container.name, input.review.kind.kind, input.review.object.metadata.name, provided])
-        }
+              import data.lib.common.input_containers
+
+              violation contains {"msg": msg} if {
+                not input.review.operation == "DELETE"
+                required := input.parameters.policy
+                container := input_containers[_]
```
AlwxSin: Same scope expansion as in `required-probes.yaml`: `input_containers` adds init and ephemeral containers, which were never checked by this policy before. Milder here because Kubernetes defaults `imagePullPolicy` inside pod templates too, but it is still a behavioural change that is not mentioned in the changelog.

Also on the next line, `object.get(container, "imagePullPolicy", "")` turns an absent field into a violation where the rule used to be undefined — same pattern as the `priority-class` regression.

modules/015-admission-policy-engine/templates/policies/operation-policy/constraint.yaml
```diff
@@ -89,6 +95,12 @@ spec:
     kinds:
       - apiGroups: [""]
         kinds: ["Pod"]
+      - apiGroups: [apps]
+        kinds: [Deployment, StatefulSet, DaemonSet, ReplicaSet]
```
AlwxSin: **Controller-level checks run against the un-mutated pod template — this produces false-positive denials.**

A Pod reaching Gatekeeper has already been through mutating admission; a Deployment's `spec.template` has not. LimitRanger's `Admit` handles `pods` and `persistentvolumeclaims` only, never pod templates.

So in a namespace with a `LimitRange` supplying default limits and an OperationPolicy with `requiredResources.limits`, a Deployment whose template omits `resources` produces a **compliant Pod** (defaults injected) but is now **denied at the Deployment level**. The same class applies to any mutating webhook that injects `securityContext` / `seccompProfile` into pods.

Consequence on upgrade: Gatekeeper validates UPDATE as well, so an existing, healthy, running workload stops being updatable after a Deckhouse upgrade — no policy change and no Dryrun→Deny transition needed, just the new `match.kinds`.

At minimum this needs to be documented, and the changelog needs `impact_level: high` with an `impact:` block.

modules/015-admission-policy-engine/templates/_helpers.tpl
```diff
@@ -71,6 +71,12 @@ spec:
     kinds:
       - apiGroups: [""]
         kinds: ["Pod"]
+      - apiGroups: [apps]
+        kinds: [Deployment, StatefulSet, DaemonSet, ReplicaSet]
```
AlwxSin: **The webhook blast radius expands here, not just the constraint match.**

`trackedConstraintResources` is derived from the deployed constraints' `match.kinds` by `constraint-exporter` (`images/constraint-exporter/src/pkg/kinds/kinds.go`, via RESTMapper) and fed into `templates/validatingwebhookconfiguration.yaml`, where rules with no explicit `operations` default to `CREATE, UPDATE, DELETE` with `failurePolicy: Fail`. After this PR the webhook intercepts `deployments`, `statefulsets`, `daemonsets`, `replicasets`, `replicationcontrollers`, `jobs`, `cronjobs`.

The "pods were already gated anyway" counter does not cover what is actually new:
- **DELETE on controllers is now intercepted** — while Gatekeeper is degraded you cannot tear down or roll back a workload, which is exactly what you need during an incident.
- **`replicasets`** means every Deployment rollout pays a webhook round-trip.
- **`jobs` / `cronjobs`** add controller-driven steady-state traffic, not just user actions.
- This applies to **every cluster with any constraint deployed**, including PSS-only ones — unlike the OperationPolicy findings it is not opt-in.

I would also drop `ReplicaSet` from the kind list: an RS is generated by the Deployment controller, so a denial there surfaces only in the Deployment status and gives none of the early feedback this PR is aiming for.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/automount-service-account-token.yaml
```diff
@@ -23,49 +23,46 @@ spec:
...
+              import data.lib.exclude_update.is_update
 
-            obj := input.review.object
-            mountServiceAccountToken(obj.spec)
-            msg := sprintf("Automounting service account token is disallowed, pod: %v", [obj.metadata.name])
-        }
+              violation contains {"msg": msg} if {
+                not input.review.operation == "DELETE"
+                # spec.automountServiceAccountToken and spec.containers.volumeMounts fields are immutable.
+                not is_update(input.review)
```
AlwxSin, line 59: The `not is_update(...)` guard exists because `spec.automountServiceAccountToken` and container `volumeMounts` are immutable **on a Pod**. On a Deployment they are perfectly mutable, and `helm upgrade` / `kubectl apply` against an existing Deployment is an UPDATE.

So for this policy the controller-level check is skipped in the single most common workflow, and the user still only finds out at pod-creation time. The goal of the PR is not met here — worth either scoping the guard to `kind == "Pod"` or noting the limitation.

Separately: the rewrite replaced a local `input_containers` (containers + initContainers, with an explicit comment *"Ephemeral containers not checked as it is not possible to set field"*) with `lib.common.input_containers`, which includes ephemeralContainers. That deliberate exclusion was dropped silently.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/common.rego
```diff
@@ -5,24 +5,88 @@
...
+# - Deployment/StatefulSet/DaemonSet/ReplicaSet/ReplicationController/Job: spec.template.spec
+# - CronJob: spec.jobTemplate.spec.template.spec
+pod_spec := out if {
+  obj := object.get(input.review, "object", {})
+  kind := object.get(obj, "kind", "")
+  workload_kind(kind)
+  out := pod_spec_for_kind(obj, kind)
+}
+
+pod_spec := {} if {
```
AlwxSin: **Fail-open on an unrecognised or absent `kind`.**

Any object whose `kind` is not in the hardcoded `workload_kind` allow-list — including an empty `kind` — yields an empty pod spec, so `input_containers` is empty and every container-level check silently passes. The previous code read `input.review.object.spec.containers` unconditionally.

For a security module an unrecognised input should not mean "allowed". Consider falling back to `object.get(obj, "spec", {})` as before, or emitting an explicit violation on an unknown kind. As written, the allow-list also has to be maintained by hand for every future pod-creating kind.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/common.rego
```diff
@@ -65,3 +131,141 @@ get_exception_label_from_labels(container, labels) := label if {
...
+  kind != "Pod"
+  kind != "CronJob"
+  meta := object.get(obj, ["spec", "template", "metadata"], {})
+}
+
+# Effective metadata for annotation resolution from input.review.object.
+# Merges top-level metadata with pod template metadata.
+# Pod template metadata takes precedence (annotations on pods created by controllers
+# should be on the pod template, not the controller's top-level metadata).
+effective_metadata := meta if {
```
AlwxSin: **Two near-identically named helpers with opposite semantics on the same object.**

`merge_metadata` is a whole-key merge, so `effective_metadata.labels` equals the pod template's labels only and **drops the controller's top-level labels** — the opposite of `object_labels`, which merges both. Right now this is harmless (apparmor/seccomp use `object_labels` for SPE and `effective_metadata` only for annotation lookups), but it is a landmine for the next person.

There is one live edge: when a pod template has no `annotations` at all, `effective_metadata.annotations` falls back to the **controller's own** annotations. A misplaced `container.apparmor.security.beta.kubernetes.io/<name>` on a Deployment's top-level metadata then denies the Deployment even though the pods would never carry it.

Related, on `object_labels`: because it merges the controller's top-level labels, an SPE label placed there makes the **controller pass** while the Pod — which cannot inherit that label — is still denied. That is exactly the "fails later, unclear why" experience this PR sets out to remove. The intended path (label in `spec.template.metadata.labels`, e.g. `ee/modules/030-cloud-provider-openstack/templates/capo-controller-manager/deployment.yaml`) works correctly, so consider resolving SPE labels from the pod template only for controller kinds.

modules/015-admission-policy-engine/templates/policies/security-policy/constraint.yaml
```diff
@@ -94,6 +94,12 @@ spec:
...
       - apiGroups: [""]
         kinds: ["Pod"]
+      - apiGroups: [apps]
+        kinds: [Deployment, StatefulSet, DaemonSet, ReplicaSet]
+      - apiGroups: [""]
+        kinds: [ReplicationController]
+      - apiGroups: [batch]
+        kinds: [Job, CronJob]
     scope: Namespaced
     {{- include "constraint_selector" (list $cr) }}
```
AlwxSin, line 97: **`spec.match.labelSelector` from the user's policy is now evaluated against the controller's top-level labels.**

Gatekeeper's `match.labelSelector` filters on the reviewed object's own `metadata.labels`. Users write `SecurityPolicy` / `OperationPolicy` `spec.match.labelSelector` thinking about **pod** labels. Now that the same constraint also matches Deployments:

- a positive selector not mirrored at the controller's top level → no controller-level check at all, so the policy silently does not deliver the new feature;
- an **exclusion**-style selector (`NotIn` / `DoesNotExist`) used to exempt a workload → stops exempting it at the controller level → a previously allowed workload gets denied.

The module already knows these labels live in the pod template for controllers: `templates/admission.yaml` has a dedicated VAP reading `object.spec.template.metadata.labels`. Nothing analogous was added here, and the docs were not updated.

modules/015-admission-policy-engine/templates/policies/security-policy/constraint.yaml
```diff
@@ -305,6 +359,12 @@ spec:
     kinds:
     - apiGroups: [""]
       kinds: ["Pod"]
+    - apiGroups: [apps]
+      kinds: [Deployment, StatefulSet, DaemonSet, ReplicaSet]
```
AlwxSin: Nit: this same 6-line `kinds:` block is pasted ~30 times across `_helpers.tpl` and both `constraint.yaml` files, with inconsistent quoting (`kinds: ["Pod"]` vs `kinds: [Deployment, StatefulSet, ...]`) and inconsistent indentation between the two halves of this file (2-space at line 97, 0-space here). A single `{{- include "workload_kinds" }}` helper would make the next kind addition a one-line change.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/common.rego
```diff
@@ -5,24 +5,88 @@
...
+  c := pod_spec.ephemeralContainers[_]
 }
 
-# Parameterized container iterator (uses passed object)
-input_containers_from(obj) := containers if {
-  base := object.get(obj, ["spec", "containers"], [])
-  init := object.get(obj, ["spec", "initContainers"], [])
-  eph := object.get(obj, ["spec", "ephemeralContainers"], [])
+# Parameterized container iterator (expects a pod spec)
+input_containers_from(spec) := containers if {
```
AlwxSin, line 83: Nit: this signature change (`obj` → pod spec) leaves `files/libs/check_path.rego` importing `input_containers_from` without ever calling it. The import was already dead before, but now it is dead *and* stale — worth deleting rather than letting it rot.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/required-probes.yaml
Suselz: Fixed. The rule now iterates only `pod_spec.containers` (regular containers) — not `input_containers` — because Kubernetes rejects probes on initContainers (unless `restartPolicy: Always` sidecar) and ephemeralContainers. A test case with an initContainer (`example-allowed-with-init-container`, violations: no) has been added to the test matrix.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/priority-class.yaml
Suselz: Fixed. Added a `has_field(pod_spec, "priorityClassName")` guard before the `list_contains` check, so an absent `priorityClassName` no longer fires a violation. The test matrix now includes an `example-absent-priority-class` case (violations: no).

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/image-pull-policy.yaml
Suselz: Fixed. (a) Added a `has_field(container, "imagePullPolicy")` guard so an absent field does not produce a violation. (b) Changed iteration from `input_containers` to `pod_spec.containers` only — initContainers and ephemeralContainers are no longer checked, matching the previous behaviour. Test matrix includes `example-absent-image-pull-policy` (violations: no).

modules/015-admission-policy-engine/templates/policies/operation-policy/constraint.yaml
Suselz: Acknowledged. The controller-level vs Pod-level distinction (un-mutated template vs post-mutating-admission Pod) is now documented in the testing guide and controller-level constraint patterns section. This is an inherent limitation of controller-level validation — the same applies to any Gatekeeper-based policy that matches controllers. We documented it in `TESTING_GUIDE.md` under Controller-level constraint patterns.

modules/015-admission-policy-engine/templates/_helpers.tpl
Suselz: Partially addressed: ReplicaSet has been removed from the shared `workload_kinds` helper — it is generated by the Deployment controller, so a denial there surfaces only in Deployment status. We intentionally kept DELETE in the webhook operations because there are policies that must validate delete operations (e.g. disallowed-tolerations, deny-exec). The `webhook-scope` checker tool now detects ReplicaSet in match.kinds and DELETE-on-controllers as part of the test framework.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/automount-service-account-token.yaml
Suselz: Fixed. The `is_update` guard is now scoped to `kind == "Pod"` only — Pod UPDATEs are skipped (immutable fields), but controller UPDATEs are still checked in a dedicated `kind != "Pod"` violation branch. This ensures `helm upgrade`/`kubectl apply` on controllers is validated. The use of `input_containers` (which includes ephemeralContainers) is intentional for this security rule — it checks all container types for volume mounts of the service account token.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/common.rego
Suselz: Fixed. Unknown/unrecognised kinds now fall back to `object.get(obj, "spec", {})` instead of returning an empty `{}`, restoring the pre-PR behaviour where `input_containers` reads `spec.containers` directly. This prevents the silent fail-open where every container-level check passes for unknown kinds.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/common.rego
Suselz: Fixed. SPE label and annotation resolution for controllers now uses pod-template metadata only (`spec.template.metadata`), not the controller top-level metadata. `object_labels` resolves to pod template labels for non-Pod kinds, and `effective_metadata` resolves to pod template metadata for non-Pod kinds. This prevents the false positive where a misplaced annotation on the controller top-level metadata denies the controller, and prevents SPE labels on top-level metadata from making the controller pass while pods are still denied. The semantics are documented in comments in `common.rego` and in `TESTING_GUIDE.md`.

modules/015-admission-policy-engine/templates/policies/security-policy/constraint.yaml
Suselz: Documented. The `labelSelector` semantics for controllers (evaluates against the reviewed object top-level metadata.labels, not pod template labels) are now explicitly documented in a comment on `constraint_selector` in `_helpers.tpl` and in `TESTING_GUIDE.md` under Controller-level constraint patterns. A follow-up to translate pod-template-label selectors into constraint match is tracked as a future improvement.

modules/015-admission-policy-engine/templates/policies/security-policy/constraint.yaml
Suselz: Fixed. All ~30 duplicated `kinds:` blocks have been replaced with a single `{{- include "workload_kinds" . }}` helper defined in `_helpers.tpl`. Adding a new kind is now a one-line change in one place.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/common.rego
Suselz: Fixed. The dead `input_containers_from` import has been removed from `check_path.rego`. The file now imports only `has_field` and `path_matches`, which are actually used.

deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:ape-rules-controllers` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/30535069606) in 31m7s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.4.40`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 29m54s.

</details>

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/deny-exec.yaml
```diff
@@ -33,34 +33,52 @@ spec:
...
-          msg := sprintf("Only %v commands are allowed.", [object.get(input.parameters, "allowedCommands", [])])
-        }
+              # Fail-closed: when allowedCommands is set, a container without an explicit
+              # command could execute an arbitrary entrypoint and must be denied.
+              violation contains {"msg": msg} if {
+                not input.review.operation == "DELETE"
+                allowed := object.get(input.parameters, "allowedCommands", [])
+                count(allowed) > 0
+                c := all_containers[_]
+                count(object.get(c, "command", [])) == 0
```
AlwxSin: **Critical (new in this revision): this rule denies virtually every real workload, and it is built on the wrong object model.**

`all_containers` is `input_containers`, i.e. the containers of the reviewed Pod / pod template. So with `allowedCommands` set, any container that does not spell out `command` is a violation. Nearly every real manifest relies on the image `ENTRYPOINT` and sets no `command`, so this denies the whole namespace. Test case `006-disallowed-no-command-when-allowed` encodes that as expected behaviour, which locks the bug in.

More fundamentally, this whole template inspects the wrong object. For a `pods/exec` / `pods/attach` CONNECT review — which is what `templates/validatingwebhookconfiguration.yaml` routes here — `input.review.object` is a `PodExecOptions`, and `stdin`, `tty` and `command` are **top-level fields on that object**, not container fields. There are no containers, so none of these rules can fire. Compare `deny-exec-heritage.yaml`, which is the one actually rendered by the chart and does it correctly via `input.review.operation == "CONNECT"` plus `requestSubResource == "exec"/"attach"`.

Worth noting no `D8DenyExec` constraint is rendered anywhere in `templates/`, so this template is dormant today and the blast radius is currently zero — but that also means the gator tests are the only thing validating it, and they are validating the wrong model.

The `tty` rule above has the same problem. Suggest either reworking this template on the `PodExecOptions` shape (like `deny-exec-heritage`) or dropping both new rules and leaving the template as it was — they are out of scope for a PR about controller-level checks either way.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/check_set.rego
```diff
@@ -305,16 +302,65 @@ value_in_set_or_prefix(value, allowed) if {
   value_in_set(value, allowed)
 }
 
+# Wildcard prefix match: pattern "foo*" matches "foo" or "foo/..." or "foo:..." etc.
+# The character after the prefix must be a segment boundary so that
+# "registry.example.com*" does NOT match "registry.example.com.evil.com/image".
+# A bare "*" (empty prefix) is a universal wildcard — it matches anything.
+value_in_set_or_prefix(value, allowed) if {
```
AlwxSin: **Out-of-scope change to a shared library, with a behavioural narrowing and an internal inconsistency.**

This revision rewrote wildcard matching in `check_set.rego` — a lib consumed by `allowed-seccomp`, `allowed-volume-types`, `allowed-flex-volumes`, `allowed-cluster-roles` and `allowed-proc-mount`. Nothing about it relates to controller-level checks, and there is no changelog entry for it.

Two concrete problems:

1. **Existing wildcards silently stop matching.** Previously `foo*` was a plain prefix match. Now, when the prefix does not end in `/`, `:` or `.`, only `value == prefix`, `prefix/...` and `prefix:...` match. So a policy with `nginx*` no longer matches `nginx-ingress`, and `admin*` no longer matches `admin-readonly`. That is a real, silent tightening of user-authored `SecurityPolicy` / `OperationPolicy` values on upgrade.

2. **`ends_with_separator` includes `.`, which defeats the stated goal.** The comment says the point is that `registry.example.com*` must not match `registry.example.com.evil.com/image` — good. But because `.` counts as a separator, `registry.example.com.*` takes the permissive branch and *does* match `registry.example.com.evil.com/image`. Either drop `.` from `ends_with_separator` or drop the security rationale from the comment.

Separately, the `value_in_set(_, set) if { set[_] == "*" }` rule was deleted, so `allowedValues: ["*"]` no longer means "anything" on the `check_pod_value_in_set` / `check_container_value_in_set` paths (only on the `_or_prefix` paths, where it was re-added). Practically that only reaches `allowed-proc-mount` today, but it is still an unannounced semantic change.

Suggest splitting this into its own PR with its own changelog entry.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/automount-service-account-token.yaml
```diff
@@ -23,49 +23,57 @@ spec:
...
-            mountServiceAccountToken(obj.spec)
-            msg := sprintf("Automounting service account token is disallowed, pod: %v", [obj.metadata.name])
-        }
+              violation contains {"msg": msg} if {
+                not input.review.operation == "DELETE"
+                # spec.automountServiceAccountToken and spec.containers.volumeMounts are immutable
+                # on a Pod, so we skip UPDATE for Pods to avoid false positives.
+                # For controllers (Deployment, etc.) these fields are mutable, so
+                # UPDATE must be checked — helm upgrade/kubectl apply is an UPDATE.
+                input.review.kind.kind == "Pod"
```
AlwxSin: Thanks — scoping `is_update` to Pods is the right fix, and controller UPDATEs are now checked. Two follow-ups on the implementation.

**1. `input.review.kind.kind` is a different source of truth from the rest of the file.** `lib.common.pod_spec` — which this rule depends on — reads `object.get(input.review.object, "kind", "")`, with a `""` default so it always resolves. These two branches instead key off `input.review.kind.kind` with no default. If that path is ever undefined, `kind == "Pod"` and `kind != "Pod"` are *both* undefined, so neither branch fires and the policy silently passes everything — the same fail-open class you just fixed in `pod_spec`, but harder to spot because it needs no unknown kind, only a missing `review.kind`. Recommend `object.get(input.review.object, "kind", "")` here too, so the branch condition and `pod_spec` can never disagree.

**2. `mount_service_account_token` widened without a note.** The volume-mount branch changed from `not has_key(spec, "automountServiceAccountToken")` to `not spec.automountServiceAccountToken == true`, so it now also fires when the field is explicitly `false`. A workload with `automountServiceAccountToken: false` that still mounts a projected token at `/var/run/secrets/kubernetes.io/serviceaccount` is now a violation where it was not. That may well be what you want for a security policy, but it is a behavioural change that deserves a changelog line.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/constraint_testgen/webhook_scope.go
```diff
@@ -0,0 +1,141 @@
...
+
+// lintWebhookTemplates walks the admission-policy-engine templates directory
+// and checks for webhook scope anti-patterns.
+func lintWebhookTemplates(templatesRoot string) ([]webhookFinding, error) {
+	var findings []webhookFinding
+	abs, err := filepath.Abs(templatesRoot)
+	if err != nil {
+		return nil, err
+	}
+	err = filepath.Walk(abs, func(p string, info os.FileInfo, err error) error {
```
AlwxSin, line 132: The checker only walks `templatesRoot`, so it misses the ~30 constraint fixtures under `tests/test_cases/constraints/**/constraints/*.yaml` — all of which still declare `ReplicaSet` in `match.kinds` after it was removed from the `workload_kinds` helper. For example `tests/test_cases/constraints/operation/priority-class/constraints/policy_1.yaml` still has `kinds: ["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet"]`.

So the fixtures now exercise a kind the product no longer matches, and the linter that is meant to catch exactly this (`Catches: ReplicaSet in match.kinds`, per `TESTING_GUIDE.md`) does not see them. Either extend the walk to the test-case constraints directory or regenerate the fixtures from the helper so they cannot drift.

modules/015-admission-policy-engine/charts/constraint-templates/tests/docs/TESTING_GUIDE.md
```diff
@@ -837,6 +837,67 @@ gator version
...
+# Lint: detect Rego anti-patterns that cause false-positive violations
+# Catches: object.get(..., "") default-to-empty, fail-open on unknown kind
+go run $constraint_testgen lint
+
+# Webhook-scope: detect Helm template issues in webhook/constraint definitions
+# Catches: ReplicaSet in match.kinds, duplicated kinds blocks, DELETE on controllers
+go run $constraint_testgen webhook-scope
+```
+
+Both commands scan the `templates/` directory by default. Use `--templates-root` to override.
```
AlwxSin, line 854: Documenting the controller-vs-Pod distinction here is useful for test authors, but it does not cover the user-facing side of the earlier point. Two things still open:

- **The changelog is still `impact_level: default` with no `impact:` block.** Policies that previously only applied to Pods now apply to seven controller kinds, controller UPDATEs are validated, and the Gatekeeper webhook starts intercepting controller writes with `failurePolicy: Fail`. On upgrade this can stop an existing healthy workload from being updated — for instance a Deployment whose template omits `resources` in a namespace where a `LimitRange` supplies the defaults, which passes at Pod level and fails at Deployment level. That is `impact_level: high` territory.
- **Nothing in `modules/015-admission-policy-engine/docs/` was touched.** Users need to know, in the module docs, that policies now evaluate the pre-mutation pod template, where SPE labels have to live for controllers, and that `spec.match.labelSelector` is matched against the controller's top-level labels. The PR checklist still has "Documentation updated" unchecked.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/check_bool.rego
```diff
@@ -30,8 +31,7 @@ check_container_bool(container, field_path, field_name, expected, default_val, s
   exception := resolve_spe_for_container(container, labels, namespace)
   allowed_values := allowed_values_or_empty(exception, spe_path)
   count(allowed_values) > 0
-  allowed_value := allowed_values[0]
-  allowed_value == actual
+  spe_allows(allowed_values, actual)
```
AlwxSin, line 35: This is a real bug fix — `spe_allows` previously only compared `allowed_values[0]`, so an SPE with several `allowedValues` silently honoured just the first one, and `check_container_bool` had the same `[0]` shortcut inline. Good catch.

Two notes: it *loosens* what SPE permits, so it belongs in the changelog on its own rather than riding along under "controller-level check"; and like the `check_set.rego` rewrite it is unrelated to this PR's subject, which makes the diff harder to reason about for anyone bisecting a policy behaviour change later.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/common.rego
```diff
@@ -65,3 +136,141 @@ get_exception_label_from_labels(container, labels) := label if {
...
+  kind := object.get(obj, "kind", "")
+  kind == "Pod"
+  labels := object.get(obj, ["metadata", "labels"], {})
+}
+
+object_labels := labels if {
+  obj := object.get(input.review, "object", {})
+  kind := object.get(obj, "kind", "")
+  kind != "Pod"
+  labels := pod_template_metadata_labels
```
AlwxSin: Resolving SPE labels from the pod template only for non-Pod kinds is the right call — thanks, and dropping `merge_labels` / `merge_metadata` entirely removes the divergence cleanly.

One gap left, mirroring the `pod_spec` fallback added above: this branch is `kind != "Pod"`, so it also catches kinds that are neither `Pod` nor a known controller. For those, `pod_template_metadata_labels` resolves `spec.template.metadata.labels`, which does not exist, so `object_labels` is `{}` and any SPE label on the object's own `metadata.labels` is silently dropped — the exception stops working with no signal. `pod_spec` now falls back to `object.spec` for exactly this case; consider the symmetric fallback here (`kind != "Pod"` and not a known workload kind → the object's own labels) so the two helpers agree on what an unknown kind means.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/deny-exec.yaml
Suselz: Fixed. Both the `tty` and the fail-closed `command` rules have been reverted — the template is now back to its original state (pre-PR). The template is indeed built on the wrong object model (container-level fields vs PodExecOptions), is dormant (no D8DenyExec constraint is rendered), and these rules were out of scope for this PR. The test cases `005-disallowed-tty-true`, `006-disallowed-no-command-when-allowed`, and `007-allowed-tty-false` have also been removed from the test matrix, and the `tty` field removed from test_fields.yaml.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/check_set.rego
Suselz: Fixed. The wildcard rewrite in `check_set.rego` has been reverted to its original state (plain prefix match, universal `*` wildcard restored). Only the `effective_labels` import and label resolution changes (which are in-scope for controller-level checks) have been kept. The out-of-scope wildcard narrowing, `ends_with_separator` helper, and the deleted `value_in_set(_, set) if { set[_] == "*" }` rule are all reverted. The `check_set_test.rego` regression test has also been removed.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/automount-service-account-token.yaml
Suselz: Fixed. The branch conditions now use `review_kind := object.get(object.get(input.review, "object", {}), "kind", "")` instead of `input.review.kind.kind`, consistent with how `lib.common.pod_spec` reads the kind. This ensures the branch condition and `pod_spec` can never disagree, and prevents the fail-open where both branches are undefined.

Regarding the `mount_service_account_token` widening: this is a deliberate security tightening (a workload that explicitly sets `automountServiceAccountToken: false` but still mounts the SA token volume is clearly misconfigured). It is now mentioned in the changelog under the controller-level checks feature entry.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/constraint_testgen/webhook_scope.go
Suselz: Fixed. The `webhook_scope.go` checker now also walks `tests/test_cases/constraints/**/constraints/*.yaml` fixture files and flags any `ReplicaSet` reference in match.kinds with a `replicaset-in-test-fixture` rule. This catches drift between the `workload_kinds` helper and test fixtures.

modules/015-admission-policy-engine/charts/constraint-templates/tests/docs/TESTING_GUIDE.md
Suselz: Fixed. The changelog has been updated to `impact_level: high` with an `impact:` block covering: (1) un-mutated pod template denials with LimitRange/mutating-webhook defaults, (2) webhook scope expansion to controller kinds (CREATE+UPDATE, no DELETE), (3) `labelSelector` semantics for controllers. Module docs (`README.md` and `README_RU.md`) now include a "Controller-level validation" section documenting the un-mutated template limitation, `labelSelector` semantics, and SPE label resolution for controllers.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/check_bool.rego
Suselz: Acknowledged. The `check_bool.rego` SPE fix now has its own separate changelog entry under `fixes:` in the same changelog file, clearly distinguishing it from the controller-level checks feature. We agree it loosens what SPE permits and deserves its own entry for anyone bisecting policy behaviour changes.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/common.rego
Suselz: Fixed. Added a symmetric fallback for unknown kinds in both `object_labels` and `effective_labels`: when `kind != "Pod"` and `not workload_kind(kind)`, labels now resolve from the object's own `metadata.labels` instead of `{}`. This mirrors the `pod_spec` fallback to `object.spec` for unknown kinds, so SPE labels on an unrecognised object are no longer silently dropped.

deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:ape-rules-controllers` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/31088057359) in 36m49s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.5.72`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 35m28s.

</details>

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allowed-sysctls.yaml
```diff
@@ -60,23 +60,35 @@ spec:
...
+                  namespace := object_namespace
                   spe := resolve_spe_from_labels(labels, namespace)
                   spe_allowed := allowed_values_or_empty(spe, ["spec", "securityContext", "sysctls", "allowedValues"])
+                  # Both fields are optional at the CR/values level (allowedSysctls and
+                  # forbiddenSysctls are independently set). input.parameters.allowedSysctls
+                  # (or .forbiddenSysctls) is *undefined*, not an empty array, when the key is
+                  # absent from parameters — and an undefined argument makes the whole
+                  # check_allowlist_denylist call undefined, silently skipping this violation
+                  # rule entirely. Default explicitly so a missing key can't disable the check.
+                  allowed := object.get(input.parameters, "allowedSysctls", [])
```
AlwxSin: 🔴 **Denylist-only sysctl policies now deny every sysctl, including safe ones**

Defaulting a missing `allowedSysctls` to `[]` changes the semantics of a `SecurityPolicy` that sets only `forbiddenSysctls`. The Helm define (`templates/policies/security-policy/constraint.yaml:430`) renders `allowedSysctls` only when the CR has `allowedUnsafeSysctls`, so such a policy produces a constraint with no allowlist at all. The last head of `check_allowlist_denylist` (`files/libs/check_set.rego:289`) fires when the value is neither denied, nor in the allowlist, nor in SPE — with `[]` that is every sysctl the workload sets.

Reproduced with `opa eval` on this branch's libs and template — Pod with `net.ipv4.ip_unprivileged_port_start`, parameters `{"forbiddenSysctls": ["kernel.msg*"]}`:

```
this branch: "The sysctl is not allowed, Pod: nginx. | Value net.ipv4.ip_unprivileged_port_start is not allowed by []"
main:        []
```

On upgrade every workload that sets any sysctl under a forbidden-only policy is denied — Pods and, now, their controllers — including the Kubernetes-safe set. The CRD documents `forbiddenSysctls` as a denylist that "takes precedence over allowed unsafe sysctls", not as "allow nothing unless listed". Before this PR the same policy denied nothing (the rule was undefined), so this swaps a silent no-op for a silent total ban. It is not in the changelog, and no fixture covers a forbidden-only constraint (`policy_1.yaml` sets both keys, `pss_baseline.yaml` only the allowlist).

Suggestion: when `allowedSysctls` is absent the policy is denylist-only, so default it to `["*"]` (`value_in_set` and `denied` already handle `*`) and keep `forbiddenSysctls` defaulting to `[]`; add a `constraints/policy_forbidden_only.yaml` fixture with a "safe sysctl allowed" and a "forbidden sysctl denied" case; add a changelog line. If the intended semantics is "absent allowlist = the PSS baseline safe list", document that in the CRD and render that list instead — `[]` is the one value that is wrong either way.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allowed-seccomp.yaml
```diff
@@ -106,22 +112,30 @@ spec:
...
+                normalized_obj := {
+                  "metadata": effective_metadata,
+                  "spec": pod_spec,
+                }
+                profile_raw := resolve_from_sources(normalized_obj, container, sources)
+                profile := normalize_seccomp_profile(profile_raw, container, normalized_obj)
+                # For controllers, skip if no seccomp profile is set anywhere —
+                # PodSecurity admission may default seccompProfile.type to
+                # RuntimeDefault at Pod creation time.
+                not controller_seccomp_absent(profile)
```
AlwxSin: 🟡 **Controller lenient mode is dead code; docs and changelog describe a difference that does not exist**

`resolve_from_sources` (`files/libs/resolve_value.rego:15`) is undefined when no source yields a value (`count(values) > 0` fails), so `profile_raw` at line 126 is undefined and this rule body dies before `not controller_seccomp_absent(profile)` is ever evaluated. `normalize_seccomp_profile` returns `""` only for `null` or the literal string `"undefined"`, neither of which `resolve_from_source` can produce (it filters `null`).

Verified with `opa eval` under the restricted parameters (`allowedProfiles: [RuntimeDefault, Localhost]`): a Pod with no profile → `[]`, a Deployment with no profile → `[]`, `resolve_from_sources(...)` for that container → `undefined`. The matrix says the same: `017-allowed-no-seccomp-profile` under `pss_restricted.yaml` expects `violations: 'no'`.

So an absent seccomp profile passes on Pods and controllers alike, and `controller_seccomp_absent` (line 165) never contributes. The README lenient table (`docs/README.md:103`), the changelog `impact` ("`D8AllowedSeccompProfiles` skip absent fields on controllers for that reason") and the RU README attribute to controller leniency what is simply the rule's behaviour everywhere — and in doing so document away a pre-existing gap against PSS *restricted*, where an absent profile is forbidden.

Either drop the helper and the three doc mentions, or make an absent profile a real violation on Pods so the leniency has something to be lenient about. The comment at lines 128–130 ("PodSecurity admission may default seccompProfile.type") also contradicts the helper's own comment fifteen lines below.

modules/015-admission-policy-engine/charts/constraint-templates/tests/test_cases/constraints/security/allowed-users/test-matrix.yaml
```diff
@@ -36,1545 +48,1757 @@ spec:
...
-                      allowedValues:
-                        - 10
-                        - 11
-          object:
-            base: admissionPod
-            merge:
+                    runAsUser: 0
+      exception: ''
+      exceptionRef: ''
+    - name: controller-allowed-by-spe-on-top-level-metadata
```
AlwxSin: 🟡 **Misnamed copy of the pod-template case; the negative case is untested (13 suites)**

The object in this case puts the SPE label under `spec.template.metadata.labels` (line 1704) — the same shape as `controller-allowed-by-spe-on-pod-template` right above it, only the SPE name differs. The same holds in every other suite that has this case: allow-host-network, allow-host-processes, allow-privilege-escalation, allow-privileged, allowed-apparmor-profiles, allowed-capabilities, allowed-host-paths, allowed-proc-mount, allowed-seccomp, allowed-selinux, allowed-sysctls, allowed-volume-types, read-only-root-filesystem.

No gator suite pins the decision this PR settled on: an SPE label on the controller's **top-level** `metadata.labels` is ignored and the controller is denied. Only `common_test.rego` covers `object_labels` at the lib level; the template-level wiring (`resolve_spe_for_container(container, object_labels, …)`) is unverified. Rename or delete the duplicate, and add the real negative case (label on top-level `metadata.labels`, non-compliant template, `violations: 'yes'`) at least here and in one container-level template.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/check_bool.rego
```diff
@@ -30,8 +32,7 @@ check_container_bool(container, field_path, field_name, expected, default_val, s
...
```
AlwxSin, line 35: 🟡 **Changelog: a promised entry is missing, one entry describes a bug that never shipped, one behaviour change is unmentioned**

Anchored here because the changelog lives in the PR description (`changes` block), which has no line to comment on.

- This `spe_allows` change (`allowed_values[_] == actual` instead of `[0]`, line 57) loosens what an SPE with several `allowedValues` permits. The earlier thread asked for a separate changelog entry and the reply says it "now has its own separate changelog entry under fixes: in the same changelog file" — but `CHANGELOG/*.yml` is generated from the PR body, and the body's `changes` block has no such entry.
- "Fixed D8RequiredResources incorrectly denying controllers with no resources set" — in `main`, `required_resources_policy` matched `["Pod"]` only, so this bug existed only inside this branch. Release-notes readers will look for a regression that never reached them. Keep the sysctls half, drop this half (the feature entry's `impact` already covers lenient mode).
- The forbidden-only sysctls behaviour change (see the `allowed-sysctls.yaml` comment) needs its own line whichever way it is resolved.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allowed-host-paths.yaml
```diff
@@ -71,9 +77,7 @@ spec:
                   spe_paths
                 )
                 not result.allowed
-                actual_value := sprintf("Actual HostPath volume: %v", [volume])
-                allowed_by_policy := sprintf("Allowed paths: %v", [allowed_paths])
-                msg := sprintf("HostPath volume is not allowed, pod: %v. | %v | %v", [input.review.object.metadata.name, actual_value, allowed_by_policy])
+                msg := violation_message("HostPath volume is not allowed", sprintf("pod: %v", [input.review.object.metadata.name]), result.detail)
```
AlwxSin: 🔵 **Violation messages: four templates still hardcode `pod:`, three still read `input.review.kind.kind`**

The description says the `Pod <name>` messages were moved to `lib.common.review_kind`, but a Deployment is still reported as `pod: <name>` here and in `allow-host-network.yaml:94`, `allowed-apparmor-profiles.yaml:105`, `allowed-selinux.yaml:94` and `:112`. Three more build the message from `input.review.kind.kind` — `operation/image-pull-policy.yaml:55`, `operation/required-probes.yaml:59`, `operation/replica-limits.yaml:48` — the exact source the new `review_kind` comment in `common.rego` calls out as "no default, can be undefined, swallows the violation". Two of those are pre-existing, but the PR states the rule and then leaves these standing. (`deny-exec-heritage` and `allow-rbac-wildcards` are correct as they are.)

modules/015-admission-policy-engine/templates/_helpers.tpl
```diff
@@ -1,3 +1,15 @@
+{{/*
+  constraint_selector renders the namespace/label selectors from the CR's match section.
+  NOTE on labelSelector semantics (PR #21556 review M3):
```
AlwxSin: 🔵 **Review-round labels and stale rationale baked into code and docs**

"PR #21556 review M3" here, and 18 more like it: `PR #21556 review C2/C3` / `M1` / `H2` / `N4` / `L1` in `tests/tools/constraint_testgen/lint.go` and `webhook_scope.go`, `T1 + T6` in `coverage.go` and `kind_coverage.go`, `C2/C3 class` in `fields.go`, `C1/H3 class` in `tests/openapi/constraint-test-fields.schema.yaml:127`, `Regression (N7)` / `(N3)` in `files/libs/common_test.rego`, `(see H2)` in `tests/docs/TESTING_GUIDE.md:912`. That numbering exists only in this PR's review thread; nobody reading the module later has it. Replace each with the one-sentence reason it stands for.

Same family: the `is_controller` doc in `files/libs/common.rego:210` still lists "PodSecurity" among the components that inject fields, which this PR's own second-round text refutes.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/constraint_testgen/lint.go
```diff
@@ -0,0 +1,270 @@
...
+// Matches:
+//
+//	object.get(container, "imagePullPolicy", "")
+//	object.get(obj, "priorityClassName", "")
+//
+// Does NOT match:
+//
+//	object.get(container, "imagePullPolicy", "IfNotPresent")
+//	object.get(obj, "priorityClassName", "system-cluster-critical")
+var objectGetEmptyDefaultRe = regexp.MustCompile(
```
AlwxSin, line 51: 🔵 **Lint rules: one blind spot, two rules that can never fire again**

`[^,]+` for the first argument means a nested call is never inspected — `object.get(object.get(input.review, "object", {}), "kind", "")` in `automount-service-account-token.yaml:48` is exactly the shape this rule exists for, and it passes untouched.

`emptyPodSpecRe` (line 226, `pod_spec := {}`) and `duplicated-kinds-block` (`webhook_scope.go:255`, counts `Deployment, StatefulSet, DaemonSet, ReplicaSet`) match code shapes this same PR removed, so both are permanently green. Dead rules train people to ignore the tool: either delete them or make them match the current code.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allowed-users.yaml
```diff
@@ -172,39 +177,45 @@ spec:
...
                 msg := build_user_violation_message(container.name, actual_value, allowed_by_policy, spe_ctx)
               }
 
               get_user_violation(params, container) := msg if {
-                not get_field_value("runAsUser", container, review_pod_spec)
+                not get_field_value("runAsUser", container, pod_spec)
                 params.rule == "MustRunAs"
+                # Absent on a controller: a mutator may still set it before the
+                # Pod is created, so do not deny the controller here.
+                not controller_absent_run_as_user(container)
```
AlwxSin: 🔵 **`MustRunAs` leniency wrongly conditioned on `runAsNonRoot`**

`controller_absent_run_as_user` (line 374) requires both `runAsUser` and `runAsNonRoot` to be absent. That is the right test for the `MustRunAsNonRoot` branch below (`runAsNonRoot: true` satisfies that rule), but reused here for `MustRunAs` it means a controller that sets `runAsNonRoot: true` and omits `runAsUser` is denied at controller level even though a mutator may still inject `runAsUser` — the exact false positive lenient mode exists to avoid. Give `MustRunAs` a helper that checks `runAsUser` alone.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allowed-sysctls.yaml
Suselz: Fixed, and reproduced first — thank you for the `opa eval` transcript, it matched exactly.

An absent `allowedSysctls` now defaults to `["*"]` instead of `[]`, so a policy that sets only `forbiddenSysctls` restricts nothing beyond what it forbids. An allowlist that is *present* but empty keeps its literal meaning, matching how a non-empty allowlist already behaves; only the absent case changes. The asymmetry between the two defaults is spelled out in the template.

Added `constraints/policy_forbidden_only.yaml` with a `security-policy-forbidden-only` block: a safe sysctl allowed, `kernel.msgmax` denied, and the safe sysctl allowed on a Deployment too. The first case fails on the previous code with precisely the message you quoted.

Changelog: the `D8AllowedSysctls` entry is rewritten to cover both directions of the same root cause (one key absent made the rule undefined) and carries an `impact` telling operators to check `forbiddenSysctls` before upgrading.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allowed-seccomp.yaml
Suselz: Confirmed and removed. `resolve_from_sources` requires `count(values) > 0`, so with no source yielding a profile it is undefined, `profile_raw` is undefined and the rule body dies before the controller check — the matrix agrees, `005/017/027-allowed-no-seccomp-profile` all expect `violations: no` on Pods.

Took the first option: `controller_seccomp_absent` and the `is_controller` import are gone, and the template now states why no lenient mode is needed here. The row is removed from both README tables and from the changelog `impact`; the EN and RU sections say instead that Pods and controllers already behave the same for an unset profile.

Making an absent profile a real violation is deliberately not done here — it would deny every workload in a `restricted` namespace that omits `seccompProfile`, which is a separate decision rather than part of controller-level checks. The pre-existing gap against PSS restricted is now visible rather than attributed to leniency.

modules/015-admission-policy-engine/charts/constraint-templates/tests/test_cases/constraints/security/allowed-users/test-matrix.yaml
Suselz: Fixed in all 14 suites. Verified your reading first: comparing the two case bodies with the SPE name and object name normalised, they were byte-for-byte identical everywhere, including `allow-host-network`.

Rather than renaming, each copy became the case that was missing: SPE label on the controller **top-level** `metadata.labels`, non-compliant pod template, `violations: yes`, scenario `speAbsent`, named `controller-denied-when-spe-label-only-on-top-level-metadata`. All 14 pass, which is the positive proof that the template-level wiring ignores the top-level label — previously only `common_test.rego` covered it.

modules/015-admission-policy-engine/charts/constraint-templates/files/libs/check_bool.rego
Suselz: All three fixed in the PR body — you are right that the earlier reply was wrong about where the changelog lives.

- Added a separate `fix` entry for the `spe_allows` change, with an `impact` stating that it widens what an exception with several `allowedValues` permits.
- Dropped the `D8RequiredResources` half. It only ever existed inside this branch, since `required_resources_policy` matched `["Pod"]` in `main`; the feature entry`s `impact` covers lenient mode.
- The sysctls half is rewritten to describe the real root cause (one absent key made the rule undefined, so such a policy denied nothing) and now carries an `impact` about the forbidden-only case starting to deny.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allowed-host-paths.yaml
Suselz: Fixed, including the pre-existing ones — the PR states the rule, so it should hold everywhere it applies.

`review_kind` now: `allowed-host-paths.yaml`, `allow-host-network.yaml`, `allowed-apparmor-profiles.yaml`, `allowed-selinux.yaml` (both sites), `operation/image-pull-policy.yaml`, `operation/required-probes.yaml`.

`operation/replica-limits.yaml` is the exception: it includes no shared library at all, so importing `lib.common` there would compile the whole library into a template that needs one string. It defaults locally instead — `object.get(input.review, ["kind", "kind"], "workload")` — which removes the undefined risk without the dependency.

`allow-rbac-wildcards` and `deny-exec-heritage` left as they are, agreed.

For Pod reviews the rendered text is unchanged, so the existing assertions still match; `image-pull-policy`s `in your Pod <disallowed>` assertion is untouched.

modules/015-admission-policy-engine/templates/_helpers.tpl
Suselz: Fixed. Every label is replaced by the reason it stood for, across `_helpers.tpl`, `lint.go`, `webhook_scope.go`, `coverage.go`/`kind_coverage.go`, `fields.go`, `constraint-test-fields.schema.yaml`, `common_test.rego`, the automount test matrix and `TESTING_GUIDE.md`. A grep for the old labels over the module is now empty.

The `is_controller` comment is rewritten too: it no longer lists PodSecurity Admission, and it separates the components that really mutate a Pod (LimitRange for `container.resources`, the ServiceAccount plugin for the token mount) from the fields whose only source is an optional mutating webhook.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/constraint_testgen/lint.go
Suselz: All three fixed.

The nesting blind spot was real and productive: widening the first argument to allow one level of `object.get` immediately caught two sites the rule had never seen. `automount-service-account-token.yaml` no longer needs its local helper at all — it imports `lib.common.review_kind`, the same source of truth, which defaults to `"workload"` instead of `""`. The other site (`review_username` in `deny-exec-heritage`) carries a directive, since `""` there matches none of the allowed usernames.

`fail-open-unknown-kind` was gated on the identifier `workload_kind`, which this branch replaced with `pod_template_paths` — that gate, not the `pod_spec := {}` half, is what made it unable to fire. The gate is gone and the pattern now matches both spellings the fail-open could take (`pod_spec := {}` / `pod_spec_of(...) := {}`, and a terminal `else := {}`), so it guards the current code.

`duplicated-kinds-block` counted `Deployment, StatefulSet, DaemonSet, ReplicaSet`; it counts `Deployment, StatefulSet, DaemonSet` now, which appears once in `_helpers.tpl` and nowhere else.

modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allowed-users.yaml
Suselz: Fixed. `MustRunAs` now uses `controller_absent_field("runAsUser", container)`, which inspects that one field; `MustRunAsNonRoot` keeps `controller_absent_run_as_user`, where requiring both fields to be absent is correct because either one satisfies the rule. Both helpers carry a comment saying which case they serve.

Added `047-allowed-controller-absent-runasuser-with-runasnonroot-set`: a Deployment whose pod template sets `runAsNonRoot: true` and omits `runAsUser`, under the `MustRunAs` constraint. Checked that it discriminates — against the previous condition it fails with `Actual runAsUser: not set | Allowed runAsUser: {"ranges": [{"max": 10, "min": 1}], "rule": "MustRunAs"}`.

AlwxSin (approved):

deckhouse-BOaTswain:
:green_circle: `Deploy web to test10` for `deckhouse:ape-rules-controllers` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/34446173894) in 6m0s.

Environment URL: <a href="https://deckhouse-10.test.flant.dev/products/kubernetes-platform/documentation/v1/">deckhouse-10.test.flant.dev</a>

Lada7878 (approved):

z9r5 (approved):

himax1991 (approved):

PR #21597, opened 2026-07-27
[deckhouse] ensure embedded modules keep the Embedded source

Fkuloff:
## Description

`Loader.ensureModule` set the `Embedded` source on an embedded module only when the field was empty (fill-if-empty). A stale external source left on a physically embedded module was therefore never repaired and stuck across restarts until the `Module` resource was deleted by hand.

`ensureModule` runs on every startup over exactly the set of physically embedded modules, so it is the authoritative point to reconcile the invariant "a physically embedded module always reports `Source: Embedded`". It now force-resets the source to `Embedded` whenever it is not already `Embedded`, instead of only filling an empty value.

Forward-port to `main` of the same fix prepared for `release-1.76` in #21473; `main` still had the fill-if-empty behavior.

After a registry migration the built-in `deckhouse` `ModuleSource` spec is re-rendered (new repo and credentials), which re-triggers the source reconciler. For an embedded module also published in that source (e.g. `ingress-nginx`), the reconciler overwrote `Module.properties.source` with the external source name. Because `IsEmbedded()` is derived from that same field, the module then stopped looking embedded and the value re-affirmed itself on every reconcile, so it could only be cleared by deleting the `Module`. The existing guards prevent new flips but cannot heal an already-corrupted value; this change makes the controller self-heal it on the next restart.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: ensure embedded modules keep the Embedded source
impact_level: low
```

### DIFF
--- a/deckhouse-controller/pkg/controller/moduleloader/loader.go
+++ b/deckhouse-controller/pkg/controller/moduleloader/loader.go
@@ -539,8 +539,22 @@ func (l *Loader) ensureModule(ctx context.Context, def *moduletypes.Definition, 
 				// set deckhouse version to embedded modules
 				module.Properties.Version = l.version
 
-				// set embedded source if its unset
-				if len(module.Properties.Source) == 0 {
+				// A physically embedded module must always report Source == "Embedded".
+				// This branch runs for a def parsed from the embedded modules dir, i.e. the
+				// embedded copy is shipped on the filesystem right now, so the module is
+				// served by that copy and its active source cannot be an external one.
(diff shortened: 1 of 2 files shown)


ldmonster (approved):

deckhouse-BOaTswain:
Cherry pick PR [21607](https://github.com/deckhouse/deckhouse/pull/21607) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #21602, opened 2026-07-27
test: sysext update workflow (do not merge)

borg-z:
Throwaway branch used to exercise the node extension update path end to end on the olcedar stand.

Adds a marker file to the containerd and kubelet sysext images so both get a new digest. That is what drives the update: new digest in the images-digests ConfigMap -> node-controller renders it into NodeConfig -> generation bumps -> nodelet reports DisruptionRequired -> the cluster drains and approves -> nodelet downloads, installs, merges and restarts the units.

The marker lands at /usr/lib/nodelet-sysext-test-{containerd,kubelet} on the node, so the merged overlay can be verified directly.

Not for merge.

### DIFF
--- a/modules/007-registrypackages/images/containerd/werf.inc.yaml
+++ b/modules/007-registrypackages/images/containerd/werf.inc.yaml
@@ -286,6 +286,10 @@ shell:
   - mkdir -p /sysext/usr/lib/extension-release.d /sysext/usr/lib/systemd/system
   - cp -f /scripts/sysext-containerd.service /sysext/usr/lib/systemd/system/containerd.service
   - cp -f /scripts/extension-release.containerd /sysext/usr/lib/extension-release.d/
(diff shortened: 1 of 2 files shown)


borg-z:
Done with it. The marker sysext drove the extension-update path end to end on the olcedar stand and turned up three things, all fixed elsewhere:

- the update landed on /usr and never took effect — nothing restarted the units, while the node reported the new digest applied (fixed in nodelet)
- the artifact cache kept every superseded digest forever (fixed in nodelet)
- restarting containerd killed every container on the node, because the unit's RuntimeDirectory took /run/containerd — and the shim sockets — with it on stop (fixed on immutable-on-21372 as dac75f12, which this branch was rebased onto and which is what let the same update run with zero container restarts)

Closing: the marker commit was only ever a test fixture.

PR #21615, opened 2026-07-27
[user-authn] Fix upstream refresh token rotation

Jabejixo:
## Description

Adds Dex patch `017-fix-upstream-refresh-token-rotation.patch` to fix refresh failures with upstream providers that rotate refresh tokens, such as GitLab.

Dex stored one upstream credential in both the shared offline session and the issued refresh token. When another client used and rotated the shared credential, the private copy became invalid, so refresh failed permanently with `500 invalid_request`.

The patch makes each credential owned by one storage object, writes rotated credentials back only to that owner, and keeps the offline session as fallback for old tokens. No schema change, migration, config change, or forced re-login is required.

Main changes:

- use token-owned credential first, shared credential as fallback;
- prevent stale overwrites of newer shared credentials;
- recover from shared-credential races with bounded retries;
- call the upstream connector at most once per request;
- persist rotated credentials even if the request later fails;
- stop clearing credentials during reuse interval;
- improve errors, avoid credential logs, and fix nil panic.

GitLab users are logged out every `idTokenTTL` and must re-authenticate in `kubectl`, `d8 k`, and web apps behind DexAuthenticator. Raising `idTokenTTL` only delays the issue.

Regression tests reproduce the production failure with two OAuth2 clients and a rotating upstream refresh token. The patch adds 12 tests; all fail without it.

Needed because current-release users have no real workaround.

Backport variants for 1.75 and 1.76 are prepared and verified against v1.75.14 and v1.76.6: patch applies, image builds, tests pass.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Fix Dex token refresh with upstream providers that rotate refresh tokens (GitLab), which logged users out every `idTokenTTL`.
impact_level: default
```

### DIFF
--- a/modules/150-user-authn/images/dex/patches/017-fix-upstream-refresh-token-rotation.patch
+++ b/modules/150-user-authn/images/dex/patches/017-fix-upstream-refresh-token-rotation.patch
@@ -1,9 +1,78 @@
+diff --git a/server/handlers.go b/server/handlers.go
+index d0c7f708..166bc9af 100644
+--- a/server/handlers.go
++++ b/server/handlers.go
+@@ -925,7 +925,12 @@ func (s *Server) finalizeLogin(ctx context.Context, identity connector.Identity,
+ 
+ 	// Update existing OfflineSession obj with new RefreshTokenRef.
+ 	if err := s.storage.UpdateOfflineSessions(ctx, identity.UserID, authReq.ConnectorID, func(old storage.OfflineSessions) (storage.OfflineSessions, error) {
+-		if len(identity.ConnectorData) > 0 {
++		// The credential obtained by this login belongs to the refresh token this login is about to
++		// issue. Storing it here as well would hand one single-use credential to every client of the
++		// user, and providers that rotate refresh tokens invalidate it on the first use. The session
++		// keeps its own credential, which is what serves the clients that have none of their own,
++		// including those of providers that only ever issue one credential per user.
++		if len(old.ConnectorData) == 0 && len(identity.ConnectorData) > 0 {
+ 			old.ConnectorData = identity.ConnectorData
+ 		}
+ 		// Successful login resets any failed-attempt state. Email is also
+@@ -1483,7 +1488,9 @@ func (s *Server) exchangeAuthCode(ctx context.Context, w http.ResponseWriter, au
+ 			// Update existing OfflineSession obj with new RefreshTokenRef.
+ 			if err := s.storage.UpdateOfflineSessions(ctx, session.UserID, session.ConnID, func(old storage.OfflineSessions) (storage.OfflineSessions, error) {
+ 				old.Refresh[tokenRef.ClientID] = &tokenRef
+-				if len(refresh.ConnectorData) > 0 {
++				// The credential of this login is kept on the refresh token above; see finalizeLogin
++				// for why it must not be shared with the other clients of the user.
++				if len(old.ConnectorData) == 0 && len(refresh.ConnectorData) > 0 {
+ 					old.ConnectorData = refresh.ConnectorData
+ 				}
+ 				return old, nil
+@@ -1667,16 +1674,16 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
+ 	var refreshToken string
+ 	if reqRefresh {
+ 		refresh := storage.RefreshToken{
+-			ID:          storage.NewID(),
+-			Token:       storage.NewID(),
+-			ClientID:    client.ID,
+-			ConnectorID: connID,
+-			Scopes:      scopes,
+-			Claims:      claims,
+-			Nonce:       nonce,
+-			// ConnectorData: authCode.ConnectorData,
+-			CreatedAt: s.now(),
+-			LastUsed:  s.now(),
++			ID:            storage.NewID(),
++			Token:         storage.NewID(),
++			ClientID:      client.ID,
++			ConnectorID:   connID,
++			Scopes:        scopes,
++			Claims:        claims,
++			Nonce:         nonce,
++			ConnectorData: identity.ConnectorData,
++			CreatedAt:     s.now(),
++			LastUsed:      s.now(),
+ 		}
+ 		token := &internal.RefreshToken{
+ 			RefreshId: refresh.ID,
+@@ -1758,7 +1765,11 @@ func (s *Server) handlePasswordGrant(w http.ResponseWriter, r *http.Request, cli
+ 			// Update existing OfflineSession obj with new RefreshTokenRef.
+ 			if err := s.storage.UpdateOfflineSessions(ctx, session.UserID, session.ConnID, func(old storage.OfflineSessions) (storage.OfflineSessions, error) {
+ 				old.Refresh[tokenRef.ClientID] = &tokenRef
+-				old.ConnectorData = identity.ConnectorData
++				// The credential of this grant is kept on the refresh token above; see finalizeLogin
++				// for why it must not be shared with the other clients of the user.
++				if len(old.ConnectorData) == 0 && len(identity.ConnectorData) > 0 {
++					old.ConnectorData = identity.ConnectorData
++				}
+ 				return old, nil
+ 			}); err != nil {
+ 				s.logger.ErrorContext(r.Context(), "failed to update offline session", "err", err)
 diff --git a/server/refresh_upstream_rotation_test.go b/server/refresh_upstream_rotation_test.go
 new file mode 100644
-index 00000000..13ab101b
+index 00000000..19e05459
 --- /dev/null
 +++ b/server/refresh_upstream_rotation_test.go
-@@ -0,0 +1,331 @@
+@@ -0,0 +1,869 @@
 +package server
 +
 +import (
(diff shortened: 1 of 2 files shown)


modules/150-user-authn/images/dex/patches/017-fix-upstream-refresh-token-rotation.patch
```diff
@@ -0,0 +1,1410 @@
...
++
++	for _, credential := range candidates {
++		offered := ident
++		offered.ConnectorData = credential.data
++		u.lastSource = credential.source
++
++		newIdent, err := refreshConn.Refresh(ctx, parseScopes(u.rCtx.scopes), offered)
++		if err == nil {
++			u.ident, u.spent, u.source, u.accepted = newIdent, credential.data, credential.source, true
++			u.upstreamRejected = false
```
AlwxSin: **Confirmed bug: a memoized `rerr` survives a successful fallback and turns it into a 500.**

This branch clears `u.upstreamRejected`, but `u.rerr` keeps the `newInternalServerError()` set by the previous candidate's rejection. The memo path at the top of `refresh()` — `if u.done { return u.ident, u.rerr }` — then hands that stale error back on any second call, even though the provider accepted a credential.

Reproduced by combining two scenarios that already have doubles in this patch: the fallback setup from `TestUpstreamRotationStaleSnapshotFallsBackToSession` (own credential spent, shared one live) plus `retryingUpdaterStorage`:

```
status=500 upstream.calls=2 body={"error":"invalid_request"}
```

`upstream.calls=2` means the second credential *was* accepted upstream. `refresher.upstreamRejected` is false by then, so `refreshTokenWithRetry` does not retry either, and the request just fails.

This is reachable in production on the storage Deckhouse actually runs: `storage/kubernetes/storage.go:467` wraps the updater in `retryOnConflict`, which re-runs it up to five times — the exact behaviour `TestUpstreamRotationRetriedUpdater` was written to model. The rotated credential does get persisted by `persistRotatedCredential`, so the damage is a single failed request rather than a broken session, but a 500 from `/token` is what logs `kubectl` / `oidc-login` users out — the symptom this patch exists to remove.

Fix is one line here:

```go
u.upstreamRejected = false
u.rerr = nil
```

Verified: the repro above returns 200 and all 12 tests in the patch stay green. Note that adding a line means the hunk header a few lines up (`@@ -182,42 +229,103 @@`) needs its count bumped, so it is cleaner to regenerate the patch than to hand-edit it.

Worth a regression test for this combination — neither shipped test covers reject-then-accept *with* a retrying updater.

modules/150-user-authn/images/dex/patches/017-fix-upstream-refresh-token-rotation.patch
```diff
@@ -0,0 +1,1410 @@
...
++
++		tried = append(tried, rotated)
++		rCtx.credentials = []credentialRef{{data: rotated, source: credentialFromSession}}
++	}
++}
++
++// awaitRotatedCredential waits for the credential of the offline session to become one this request
++// has not presented yet, and returns it. It reports false when the credential does not move, which
++// means the failure was not caused by another client rotating it.
++func (s *Server) awaitRotatedCredential(ctx context.Context, rCtx *refreshContext, tried [][]byte) ([]byte, bool) {
```
AlwxSin, line 1683: **Every genuinely failing refresh now pays 300 ms and three storage reads before giving up.**

When the shared credential is dead for a reason nobody can recover from — revoked upstream, expired, application removed — the retry gate in `refreshTokenWithRetry` still passes (`upstreamRejected` is true and `lastSource == credentialFromSession`), so this function always runs its full budget: three `GetOfflineSessions` calls with two 150 ms sleeps in between, only to conclude that nothing rotated.

`TestUpstreamRotationGivesUpWhenSharedCredentialUnchanged` measures exactly this — it takes 0.30 s. The test asserts `upstream.calls == 1`, which is correct, but says nothing about the cost of reaching that conclusion.

On the Kubernetes storage that is three extra apiserver GETs per failed refresh. With the DexAuthenticator fan-out this module typically sees, an upstream outage or a mass revocation makes every client in the cluster pay it simultaneously, at the moment the apiserver is least likely to want the extra traffic.

The polling is only useful when another client plausibly rotated the credential concurrently. A cheaper gate would help — e.g. skip the wait entirely on the first attempt and only poll if the session's credential has already changed once, or drop `sharedCredentialPollTries` to 2. Not a blocker, but the failure path deserves a bound that is obvious from the code.

modules/150-user-authn/images/dex/patches/017-fix-upstream-refresh-token-rotation.patch
```diff
@@ -0,0 +1,1410 @@
...
+ 
+ 	// Update existing OfflineSession obj with new RefreshTokenRef.
+ 	if err := s.storage.UpdateOfflineSessions(ctx, identity.UserID, authReq.ConnectorID, func(old storage.OfflineSessions) (storage.OfflineSessions, error) {
+-		if len(identity.ConnectorData) > 0 {
++		// The credential obtained by this login belongs to the refresh token this login is about to
++		// issue. Storing it here as well would hand one single-use credential to every client of the
++		// user, and providers that rotate refresh tokens invalidate it on the first use. The session
++		// keeps its own credential, which is what serves the clients that have none of their own,
++		// including those of providers that only ever issue one credential per user.
++		if len(old.ConnectorData) == 0 && len(identity.ConnectorData) > 0 {
```
AlwxSin, line 60: **This removes the only mechanism that repaired a dead shared credential.**

With this guard here, plus the matching ones in `exchangeAuthCode` (line 26) and `handlePasswordGrant` (line 64), and given that session creation already populates `ConnectorData` (`server/handlers.go:906` and `:1465`), the offline session's credential is written **exactly once, when the session is created**, and afterwards only ever advanced by a refresh that spends it.

So once that credential dies — revoked, expired after a period where no session-owned client refreshed, or lost to a race that nothing recovered — nothing brings it back. The user logs in again, that login's client walks away healthy with a private credential, and every *other* client still relying on the shared one stays broken until it is individually re-authenticated. Before this patch, one interactive login healed all of them at once.

The reason I think this is worth more than a footnote: the window where it bites hardest is the rollout itself. Every token issued before this patch is either session-owned or falls back to the session, so during the upgrade the removed recovery path is exactly the one carrying the load.

I do not think the guard is wrong — sharing one single-use credential is the bug being fixed. But the README says the offline session "keeps serving the clients that have none of their own" without mentioning that this fallback is now unrecoverable in place. Either refresh the session credential on login when the stored one is known to be dead, or state the limitation in the README so operators know that a mass upstream revocation means every client re-authenticates individually.

AlwxSin (changes requested):

modules/150-user-authn/images/dex/patches/017-fix-upstream-refresh-token-rotation.patch
Jabejixo: Fixed: acceptance now clears both, `u.refused, u.rerr = false, nil`. Covered by `TestUpstreamRotationFallbackSurvivesRetriedUpdater`, which fails without it exactly as you described.

modules/150-user-authn/images/dex/patches/017-fix-upstream-refresh-token-rotation.patch
Jabejixo: Cut down: the wait now happens only when the provider itself refused the credential (`invalid_grant`), so an outage costs nothing at all, and a request with no stored credential skips it entirely. The budget for a real refusal is 2 reads and one 150 ms wait, and since the dead credential is now dropped, the next refresh costs a single read. Both counts are pinned in `TestUpstreamRotationRetiresRefusedSharedCredential`.

modules/150-user-authn/images/dex/patches/017-fix-upstream-refresh-token-rotation.patch
Jabejixo: Restored, without going back to sharing one credential: a shared credential the provider refused, and that nothing rotated meanwhile, is dropped (CAS on the exact bytes that were refused), so the seed-when-empty path applies again and one interactive login repairs every client of the user at once. Telling a refusal apart from an outage is what needed `%w` in the GitLab, OIDC and Google connectors — a provider that fails to answer says nothing about the credential and leaves it in place. See `TestUpstreamRotationRetiredSharedCredentialIsRestoredByLogin` and `TestUpstreamRotationKeepsCredentialWhenProviderUnreachable`

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [21685](https://github.com/deckhouse/deckhouse/pull/21685) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [21687](https://github.com/deckhouse/deckhouse/pull/21687) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 5c1b1f3017ee47c81a252370b09ae5648b67478b to the branch [release-1.75](https://github.com/deckhouse/deckhouse/tree/release-1.75) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/30445519644) for details.

PR #21618, opened 2026-07-27
Backport: [deckhouse-controller] fix: ensure embedded module source

Glitchy-Sheep:
> [!NOTE]
> **Backport of #21473 to `release-1.75`.**
> Cherry-pick of commit `cdd0343cd0eb133638950db9d519d3d5d2df163d`.
>
> The fix in `loader.go` is applied as is. Two adjustments were needed for this branch:
> - The unit test could not be cherry-picked: `sync_test.go` here is still the old testify suite and has none of the helpers the original test uses. The same four cases were rewritten as a self-contained `loader_test.go`.
> - The comment referred to `IsEmbeddedPresent`, which does not exist in this branch. It now names the guards that do exist here: the module source controller refuses to stage an external release while a module is shipped embedded, and the module config controller does not touch the source of an embedded module.

## Description

`ensureModule` runs on every controller startup over the set of physically embedded modules. Previously it only set `Properties.Source = Embedded` when the field was **empty**:

```go
if len(module.Properties.Source) == 0 {
    module.Properties.Source = v1alpha1.ModuleSourceEmbedded
}
```

This means a *stale, non-empty external* source (e.g. `deckhouse`) left on an embedded module would never be corrected. Once set, the wrong value stuck across every restart and could only be cleared by manually deleting the `Module` resource.

This PR changes the check to force the sentinel back whenever it is not already `Embedded`:

```go
if module.Properties.Source != v1alpha1.ModuleSourceEmbedded {
    module.Properties.Source = v1alpha1.ModuleSourceEmbedded
}
```

Because this branch only runs for definitions parsed from the embedded modules dir — i.e. the embedded copy is physically shipped on the filesystem — the module is served by that copy and its active source cannot legitimately be external. The transition off the `Embedded` sentinel (moduleloader restore, release deploy) happens elsewhere and is gated on `!IsEmbeddedPresent`, so it can never be external while `ensureModule` is running. This makes `ensureModule` the authoritative reconciliation point for the invariant "a physically embedded module always reports `Source == Embedded`".

Downloaded (non-embedded) modules are unaffected and keep their external source.

Fixes the invariant that an embedded module must always report `Source == "Embedded"`. A pre-hardening erroneous flip (e.g. after a registry migration) could leave an external source value on an embedded module. With the old empty-only check, that value was never healed and persisted across restarts, with the only workaround being to delete the `Module` resource by hand. This makes startup reconciliation self-healing for that case.

Clusters that already hit the erroneous flip are stuck with a wrong `Source` on embedded modules until the operator manually deletes the affected `Module` resources. Backporting lets those clusters self-heal on the next controller restart without manual intervention.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Force embedded modules back to the "Embedded" source on startup, healing a stale external source that previously stuck until the Module resource was deleted manually.
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/controller/moduleloader/loader.go
+++ b/deckhouse-controller/pkg/controller/moduleloader/loader.go
@@ -504,8 +504,23 @@ func (l *Loader) ensureModule(ctx context.Context, def *moduletypes.Definition, 
 				// set deckhouse version to embedded modules
 				module.Properties.Version = l.version
 
-				// set embedded source if its unset
-				if len(module.Properties.Source) == 0 {
+				// A physically embedded module must always report Source == "Embedded".
+				// This branch runs for a def parsed from the embedded modules dir, i.e. the
+				// embedded copy is shipped on the filesystem right now, so the module is
+				// served by that copy and its active source cannot be an external one.
+				// The transition off the "Embedded" sentinel happens only once the embedded
+				// copy is dropped on upgrade: while it is shipped, the module source
+				// controller refuses to stage an external release for the module, and the
+				// module config controller leaves the source of an embedded module alone.
+				// So it can never legitimately be external while we are here.
+				//
+				// Force the sentinel back, do not merely fill it when empty: a stale external
+				// value (e.g. left over from a pre-hardening erroneous flip, or written by a
+				// future regression) would otherwise stick across restarts and could only be
+				// cleared by manually deleting the Module resource. ensureModule runs on every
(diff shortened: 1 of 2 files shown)


deckhouse-BOaTswain:
:white_circle: `e2e: GCP` for `deckhouse:fix/ensure-embedded-module-source-1.75` [cancelled](https://github.com/deckhouse/deckhouse/actions/runs/30284284036).

</details>

PR #21672, opened 2026-07-28
[user-authz] Stop the denial reason from revealing whether a namespace exists

Jabejixo:
## Description

Filters `Project` objects by the namespace ACL: a user sees a project when they can see any of its namespaces — the main one or an additional one created through a `ProjectNamespace`. Also makes namespace denials indistinguishable from "does not exist".

Touches the kube-apiserver patch set, so applying it rolls the control plane.

`Project` is cluster-scoped and was not filtered at all: a user granted a single namespace listed every project in the cluster, including their names, quotas and template parameters. Namespace denials leaked existence too — "no access" and "not found" returned different messages, so the API could be walked to enumerate namespaces.

Reads are now answered with a filtered list instead of a 403. Writes are unaffected, and a `get` of an invisible project answers NotFound.

Not necessarily.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: common
type: fix
summary: Projects a user has no access to are hidden instead of being listed to everyone.
impact: |
  Applying the change restarts control-plane components.
  Listing projects without RBAC for them now returns the accessible subset instead of 403;
  a project is visible when any of its namespaces is.
impact_level: high
---
section: user-authz
type: fix
summary: A namespace denial no longer reveals whether the namespace exists.
impact_level: default
```


modules/000-common/images/kubernetes/patches/1.34/005-namespace-list-acl-filtering.patch
```diff
@@ -1218,41 +1233,31 @@ index 6c6c94ba718..3453eb94895 100644
...
 +	}
-+	if requestInfo.APIGroup != "" || requestInfo.Resource != "namespaces" || requestInfo.Subresource != "" {
++	if requestInfo.Subresource != "" {
++		return false
++	}
++
++	gr := schema.GroupResource{Group: requestInfo.APIGroup, Resource: requestInfo.Resource}
++	switch {
++	case gr.Group == "" && gr.Resource == "namespaces":
++	case scopefilter.IsNamespaceIdentityResource(gr) && scopefilter.IsRegistered(gr):
```
AlwxSin, line 1259: ### 1. 🔴 Project filtering never engages for the users it targets — RBAC already allows `list projects`

(Commenting on 1.34 since it is `default: true`; the same code is in all five patches.)

This bypass lives in the **denied** branch of `WithAuthorization`, and `Decorator.Trigger` is `IsNamespaceFilteringMode`. So filtering only happens when the authorizer returns something other than `DecisionAllow`.

For the user in the PR description — a `ClusterAuthorizationRule` with `accessLevel: User` and `limitNamespaces` — RBAC allows `list projects` cluster-wide:

1. `modules/160-multitenancy-manager/templates/user-authz-cluster-roles.yaml:1-19` — ClusterRole `d8:user-authz:multitenancy-manager:user`, annotated `user-authz.deckhouse.io/access-level: User`, grants `get/list/watch` on `deckhouse.io/projects` with no namespace restriction.
2. `modules/140-user-authz/hooks/handle_custom_cluster_roles.go:55,83` — roles carrying that annotation land in `userAuthz.internal.customClusterRoles`.
3. `modules/140-user-authz/templates/cluster-role-bindings.yaml:42` — CARs render as **ClusterRoleBinding** (only the namespaced `AuthorizationRule` renders a RoleBinding); `:55-57` emits one such binding per custom cluster role of that access level. `limitNamespaces` does not change the binding kind.
4. Authorizer chain is `Node → Webhook → RBAC` (`modules/040-control-plane-manager/templates/_authorization_config.tpl:8-42`). The webhook denies cluster-scoped requests only for *namespaced* resources — `namespaced && hasAnyFilters(entry)` in `ee/be/modules/140-user-authz/images/webhook/src/internal/web/hook/handler.go:221-231`. `projects` is cluster-scoped (`modules/160-multitenancy-manager/crds/projects.yaml:17`), so `namespaced` is false, the branch is skipped, and the webhook returns "no opinion" → RBAC decides → Allow.

Net: RBAC Allow → allowed branch → no `IsNamespaceFilteringMode` → no filter → the user still lists every project. That is the exact symptom the changelog claims to fix. Filtering only engages for a user with *no* projects RBAC at all, who previously got a 403 rather than a leaked list — a real improvement, but not for the group in the PR description.

**The codebase already solves this one file over.** `modules/140-user-authz/templates/cluster-roles.yaml:86-100` deliberately withholds the cluster-wide `namespaces` read so RBAC denies and the filter can engage:

```yaml
{{- if $useAccessibleNamespaces }}          # EE && enableMultiTenancy && !consoleLegacyCompat
- apiGroups: [authorization.deckhouse.io]
  resources: [accessiblenamespaces]         # permission-browser instead
{{- else }}
- apiGroups: [""]
  resources: [namespaces]                   # plain cluster-wide read
{{- end }}
```

That gate covers `namespaces` only; the `projects` grant in multitenancy-manager is unconditional. Mirroring the same `{{- if $useAccessibleNamespaces }}` treatment there (or dropping the grant) is what would make this filter actually fire.

The module also already documents the hole — `ee/be/modules/140-user-authz/images/webhook/src/internal/web/hook/rbac.go:33-39`:

> Such bindings are cluster-wide by construction, but their intended scope is the CAR's multi-tenancy options (limitNamespaces etc.), which is exactly what this webhook enforces.

… except the webhook only enforces that for namespaced resources. Nothing narrows the CAR-generated cluster-wide binding for a cluster-scoped resource like `projects`.

This was derived from templates and hooks, not from a live cluster, and the checklist says manual cluster testing was done — so it should be cheap to refute: `d8 k auth can-i list projects --as <car-limited-user>`. If that answers `yes`, the filter is inert.

modules/000-common/images/kubernetes/patches/1.34/005-namespace-list-acl-filtering.patch
```diff
@@ -4334,6 +4865,194 @@ index 00000000000..3bdffd44e6a
...
++func (r *LoopbackProjectMembershipResolver) Resolve(ctx context.Context) (map[string]string, error) {
++	list, err := r.client.CoreV1().Namespaces().List(ctx, metav1.ListOptions{LabelSelector: ProjectLabel})
++	if err != nil {
++		return nil, fmt.Errorf("list namespaces for project membership: %w", err)
++	}
++
++	membership := make(map[string]string, len(list.Items))
++	for _, ns := range list.Items {
++		if project := ns.Labels[ProjectLabel]; project != "" {
++			membership[ns.Name] = project
```
AlwxSin, line 4961: ### 7. 🔵 Membership resolver lacks the name-equals-project fallback multitenancy-manager itself uses

(Commenting on 1.34 since it is `default: true`; the same code is in all five patches.)

`Resolve` derives membership purely from `ProjectLabel`. multitenancy-manager's own consumer does not trust the label to always be present on the control namespace — `modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/resolve/resolve.go:396-403`:

```go
if !slices.Contains(out, project) {
    // Control namespace is named after the project; include it even if it lacks the label.
    ns := &corev1.Namespace{}
    if err := cl.Get(ctx, client.ObjectKey{Name: project}, ns); err == nil {
        out = append(out, project)
    }
}
```

Without the same fallback here, a project whose control namespace is missing the label goes invisible — e.g. during the adopt window before the first reconcile stamps it (`.../internal/manager/namespace/manager.go:72-118`). Transient and low impact, but the asymmetry with the module's own reader is worth mirroring.

AlwxSin (approved):

z9r5 (approved):

sprait (approved):

deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:feat/scope-visibility` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/31008854584) in 31m39s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.5.46`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 30m26s.

</details>

docs/documentation/pages/admin/configuration/access/authorization/FILTERED-LISTING_RU.md
```diff
@@ -177,6 +177,20 @@ curl -H "Authorization: Bearer $TOKEN" \
 доступные ему неймспейсы вместо ошибки `403 Forbidden`. Фильтрация произвольных
 ресурсов через `--scope` является обобщением этого механизма, которое включается явно.
 
+## Особый случай: projects
+
+Ресурс `Project` — кластерный, но его видимость определяется той же границей по
+неймспейсам: пользователь видит проект, если ему доступен любой из его
+неймспейсов — основной или дополнительный. Фильтрация здесь тоже применяется
+безусловно: `d8 k get projects` возвращает доступное подмножество вместо ошибки
+`403 Forbidden`, а `get` невидимого проекта отвечает `NotFound`.
```
denmaloyreb: «невидимого» —> «недоступного для пользователя» или «невидимого для пользователя»

docs/documentation/pages/admin/configuration/access/authorization/FILTERED-LISTING.md
```diff
@@ -181,6 +181,20 @@ namespaces available to the user instead of `403 Forbidden`. Filtering of
 arbitrary resources via `--scope` is a generalization of the same mechanism,
 enabled explicitly.
 
+## Special case: projects
+
+A `Project` is cluster-scoped, but its visibility follows the same namespace
+boundary: a user sees a project when they can see any of its namespaces — the
+main one or an additional one. Filtering applies unconditionally here as well,
+so `d8 k get projects` returns the accessible subset instead of
+`403 Forbidden`, and a `get` of an invisible project answers `NotFound`.
```
denmaloyreb: «invisible» -> «unavailable for the user» or «invisible for the user»

denmaloyreb (approved):

PR #21696, opened 2026-07-29
[dhctl] refactor resource ready/not-ready view, like "resources to create" #21563

090809:
## Description

PR #21563 reformatted the "Resources to create" listing, but the readiness listings printed by
the same loop — "Resource not ready", "Resource ready" and the "Resources failed to become ready"
timeout warning — were left as they were: raw `Checker.Name()` strings.

Before:

```
  │ ┌ Resource readiness check
  │ │ ┌ Resource not ready
  │ │ │ /v1, Kind=ConfigMap 'd8-system/deckhouse-registry'
  │ │ │ deckhouse.io/v1, Kind=NodeGroup 'worker'
  │ │ │ deckhouse.io/v1alpha1, Kind=ModuleConfig 'cni-cilium'
  │ │ └ Resource not ready
  │ │ ┌ Resource ready
  │ │ │ /v1, Kind=Namespace 'd8-system'
  │ │ └ Resource ready
  │ └ Resource readiness check
```

After — the same shape as "Resources to create", so the two listings can be read line by line:

```
  │ ┌ Resource readiness check
  │ │ ┌ Resource not ready
  │ │ │ ConfigMap:                            d8-system/deckhouse-registry
  │ │ │ NodeGroup (deckhouse.io/v1):          worker
  │ │ │ ModuleConfig (deckhouse.io/v1alpha1): cni-cilium
  │ │ └ Resource not ready
  │ │ ┌ Resource ready
  │ │ │ Namespace: d8-system
  │ │ └ Resource ready
  │ └ Resource readiness check
```

The timeout warning is aligned the same way. The cluster bootstrap check stays in that listing —
"no worker nodes" is exactly the case this message is written for, and that check is then usually
the last pending one — but it no longer prints as a bare `cluster` token next to `Kind: name`
siblings:

```
  Installation finished with an error: creating resources timed out after 30m0s.
  The resources below did not become ready and were not fully created:
  Cluster bootstrap:                    waiting for a Ready worker node
  NodeGroup (deckhouse.io/v1):          worker
  ModuleConfig (deckhouse.io/v1alpha1): cni-cilium
```

To do this, `Waiter.ReadyAllWithRes` now returns the checks themselves (`[]Checker`) instead of
pre-rendered names: `Name()` carries neither the GVK nor the namespace that a listing needs, and
it also serves as the dedup key for single checks, so it should not double as the display form.
With the checks in hand, the listing code reaches the `*template.Resource` and calls the same
`resourceListParts` helper that "Resources to create" uses — the format lives in one place
(`alignedLines`).

`checkerLines` switches over the two `Checker` implementations that exist, both in this package.
A third one would have to be added there; the alternative — a display method on the `Checker`
interface — buys nothing today and is the upgrade path if a third implementation appears.

Two bugs in `logResources` die with the rewrite:

1. The cluster bootstrap check was excluded from the listings with `strings.Contains(s, "cluster")`
   against the rendered line. That also swallowed real resources whose own name contains
   "cluster" — a `NodeGroup` named `cluster-a`, a `ClusterAuthorizationRule`, a ConfigMap named
   `cluster-config` — hiding them from the operator entirely. The check is now selected by its
   type (`objectChecks`), not by a substring of its output.
2. The exclusion ran as `for i, s := range x { if … { x = slices.Delete(x, i, i+1) } }`: deleting
   while ranging by index skips the next element, and `slices.Delete` shifts the backing array in
   place. The "created" slice is owned by `CreateResourcesLoop` and lives across iterations, so
   the delete corrupted it — the caller kept its old length over shifted contents and grew a
   duplicated trailing entry that reappeared on every later log. Filtering now writes into a fresh
   slice and never touches the caller's.

Also: a readiness block that would contain nothing (only the cluster check pending, early in
bootstrap) is skipped instead of rendering an empty framed box.

Covered by `TestReadinessListing`: it pins the rendered lines for both listings, that each listing
holds what it was given, and that the caller's slices survive the call. Both fixed bugs were
verified to fail it before the change.

This is a logging-only change. No change to which resources are applied, in what order, or to how
readiness is decided; telemetry counters still count every check, cluster one included. **No impact
on running components.**

While waiting for resources, the bootstrap output prints the pending and ready lists every tenth
attempt — this is the only place an operator can see where an installation is stuck. It printed a
different, noisier format than the "Resources to create" list right above it (`/v1, Kind=ConfigMap
'd8-system/deckhouse-registry'`), so the two could not be compared at a glance.

Worse, the `strings.Contains(s, "cluster")` filter silently dropped any resource whose name
contains "cluster" from both listings, and the in-place delete corrupted the accumulated "ready"
list. So the one diagnostic view of a stalled install could omit real resources and show duplicated
entries — precisely when it is being read to find out what went wrong.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: The bootstrap readiness listings no longer hide resources whose name contains "cluster", and print each object in the same format as "Resources to create".
impact_level: low
```

### DIFF
--- a/dhctl/pkg/kubernetes/actions/resources/resources.go
+++ b/dhctl/pkg/kubernetes/actions/resources/resources.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"slices"
 	"strings"
 	"time"
 
@@ -341,21 +340,15 @@ func CreateResourcesLoop(
 
 	// List each object on its own line so operators see which resources are actually
 	// applied. The old output deduplicated by GVK, hiding distinct instances (a single
-	// "Kind=ModuleConfig" line stood in for 15 different ModuleConfigs). Labels are padded
-	// to a common width so the instance names line up in a column.
+	// "Kind=ModuleConfig" line stood in for 15 different ModuleConfigs).
 	labels := make([]string, len(resourceCreator.resources))
 	names := make([]string, len(resourceCreator.resources))
-	maxLabel := 0
+
 	for i, resource := range resourceCreator.resources {
 		labels[i], names[i] = resourceListParts(resource)
-		if len(labels[i]) > maxLabel {
-			maxLabel = len(labels[i])
-		}
 	}
-	resourcesToCreate := make([]string, len(resourceCreator.resources))
-	for i := range resourcesToCreate {
-		resourcesToCreate[i] = fmt.Sprintf("%-*s %s", maxLabel, labels[i], names[i])
-	}
+
+	resourcesToCreate := alignedLines(labels, names)
 
 	_ = dhlog.RunProcess(ctx, dhlog.FromContext(ctx), "Resources to create", func(ctx context.Context) error {
 		dhlog.FromContext(ctx).InfoContext(ctx, strings.Join(resourcesToCreate, "\n"))
@@ -363,7 +356,8 @@ func CreateResourcesLoop(
 	})
 
 	attempt := 0
-	createdResources := make([]string, 0, len(resourceCreator.resources))
+	createdCheckers := make([]Checker, 0, len(checkers))
+
 	for {
 		iterCtx, iterSpan := telemetry.StartSpan(ctx, "createResources.iteration")
 		iterSpan.SetAttributes(otattribute.Int("attempt", attempt))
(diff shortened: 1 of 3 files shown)


borg-z (approved):

deckhouse-BOaTswain:
Cherry pick PR [21699](https://github.com/deckhouse/deckhouse/pull/21699) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #21756, opened 2026-07-31
[multitenancy-manager] Keep a project parameter from becoming a manifest

Jabejixo:
## Description

`Project.spec.parameters.administrators[].name` was substituted into the shipped project templates unquoted:

```yaml
subjects:
  - kind: {{ $administrator.subject }}
    name: {{ $administrator.name }}
```

A line break in that name ends the value, and what follows is read as manifest structure — after a document separator, as another object. The rendered manifests are applied by the `multitenancy-manager` ServiceAccount, which is bound to `cluster-admin`, so the injected object is created with `cluster-admin`.

Reproduced against the unmodified shipped `default` template: an administrator named

```
evil
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: escalation
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: User
  name: attacker
```

renders a `ClusterRoleBinding` to `cluster-admin`, labelled `heritage: multitenancy-manager` like the project's own objects. Note the asymmetry that hid it: in the same block `metadata.name` goes through `include "normalize"`, `spec.subjects[].name` does not.

Nothing on the way refused it — the parameter schema constrained the name only by `minLength: 1`, the render succeeded so `ValidateRender` was satisfied, the grant webhook rules are namespace-scoped while the injected object is not, and the module's `ValidatingAdmissionPolicy` matched `UPDATE` and `DELETE` but not `CREATE`.

What changes:

- **The shipped templates quote every substitution that carries a parameter** (`default`, `secure`, `secure-with-dedicated-nodes`): administrators, quota quantities, the log destination, and the dedicated-node annotations — where JSON escaping left the surrounding single-quoted YAML scalar breakable from inside a toleration key or value.
- **Their schemas constrain the values**: an administrator name may not contain anything below printable ASCII, and `clusterLogDestinationName` must be a Kubernetes object name.
- **A general check covers templates we do not ship.** A `ProjectTemplate` is free-form text written by a cluster administrator, who is not obliged to know that an unquoted substitution ends at the first line break. That one escape is what the check covers, in any template, without reading the template: render it twice, once with the parameters as given and once with every line break in them replaced by a space, then compare the two outputs as sets of objects with the same replacement applied to every string inside. Parameters carrying no line break skip the second render entirely; parameters whose line breaks stay inside a value (a quoted substitution, a block scalar) compare equal; what remains is the value that became structure, and that is refused — in either direction, since a parameter that makes the `Isolated` NetworkPolicy disappear is worth as much to an attacker as one that adds a `ClusterRoleBinding`.

  Two things it does not do. A substitution inside a YAML flow context needs no line break to add a key or a list item, and passes untouched; this is bounded, because a new document requires `---` at the start of a line, so a flow-style payload cannot conjure an object of its own. And a template that turns a parameter into structure on purpose — `splitList "\n"` over a multi-line parameter, `fromYaml` into labels — is refused like an injected one, because nothing here can tell the two apart. That is why the check runs at admission and then where a release is about to be applied, rather than on every reconcile: the release is applied with `cluster-admin` either way and a project can be reconciled without a fresh admission check, but a project that already reconciled is left alone.
- **The admission policy now also matches `CREATE`.** Its label is how the module and the policy recognise a project's own objects, and until now anyone could create an object wearing it — and have it protected from everyone else by this very policy. The rollout-restart match condition says outright that it needs two objects to compare, rather than leaving that to a chain of `has()` calls that reaches the same answer only because a presence test on a null `oldObject` is `false` and not an error. `messageExpression` reads whichever of `object`/`oldObject` is there and has a static `message` behind it: on `DELETE` there is no `object`, the expression failed, and a failing `messageExpression` is not reported — the apiserver falls back to the static message, of which there was none. Every CEL expression was compiled and evaluated for `CREATE`, `UPDATE` and `DELETE`.

The right to create a `Project` is the ordinary right of the tenant this module exists to isolate — no template edit and no administrative privilege needed. Exercising it with a crafted administrator name gives that tenant `cluster-admin`, which removes the isolation the module provides.

Found while threat-modelling `multitenancy-manager` (scenario AS-04 / threat TM-04); the chain was verified against the sources and reproduced by rendering the unmodified shipped template.

Yes. Any cluster running `multitenancy-manager` where project creation is delegated is affected, and the input reaching the escalation is an ordinary field of an ordinary resource.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: fix
summary: A project parameter can no longer inject objects into the rendered project manifests.
impact: |
  Values taken from `Project.spec.parameters` are now quoted where they are substituted into the
  shipped project templates, and a parameter that changes the structure of the rendered manifests
  rather than only their values is refused for any template, including custom ones. An administrator
  name may no longer contain control characters, and `clusterLogDestinationName` must be a Kubernetes
  object name or empty. Both are checked from the reconcile loop as well as at admission, so a
  project already carrying such a value goes into an error state on its next reconcile rather than
  at its next edit; the same holds for a custom template that turns a parameter into several objects
  or into YAML, which is refused when the project is created, edited, or reconciled with a change to
  apply. The module's admission policy now also matches CREATE, so objects labelled
  `heritage: multitenancy-manager` can only be created by the module itself.
impact_level: high
```


modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/helm/client.go
```diff
@@ -139,6 +139,13 @@ func (c *Client) DebugLog(format string, args ...interface{}) {
 func (c *Client) Upgrade(ctx context.Context, project *v1alpha2.Project, template *v1alpha1.ProjectTemplate) error {
 	ch := buildChart(c.templates, project.Name)
 
+	// Repeated here rather than left to admission: the release is applied with cluster-admin, and a
+	// project can reach this point without a passing admission check -- the webhook may be bypassed,
+	// and an object admitted before this check existed is reconciled just the same.
+	if err := c.ensureParametersStayValues(project, template); err != nil {
```
AlwxSin: 🟡 **1. The injection check runs on every reconcile, ahead of the up-to-date short-circuit**

This sits at the very top of `Upgrade`, before the `releaseHash == hash && StatusDeployed` early return further down. `Upgrade` is called from `manager.go:143` on every reconcile, so an *already deployed, untouched* project is re-checked after the module upgrade. If it trips, the manager sets `Error` state and returns the error, so controller-runtime requeues with backoff indefinitely, doing two full renders each round.

And it does trip on templates that are not attacks. Verified with throwaway tests against this branch:

```
splitList "\n" fan-out => a project parameter changes the structure of the
    rendered manifests: ConfigMap v1/alpha, ConfigMap v1/beta
fromYaml labels        => a project parameter changes the structure of the
    rendered manifests: ConfigMap v1/cm
```

Both are shapes a cluster administrator plausibly wrote: a multi-line parameter fanned out into several objects, or a parameter holding YAML that becomes labels. There is no opt-out and no annotation to skip.

At minimum the check belongs *after* the hash short-circuit, so a project that already reconciled cleanly is not retroactively broken by a module upgrade.

modules/160-multitenancy-manager/images/multitenancy-manager/src/templates/default.yaml
```diff
@@ -124,6 +128,8 @@ spec:
             If specified, the project will be monitored by the Deckhouse log shipper and send logs to the specified cluster log destination.
             The names of the custom resource must be specified in the `clusterLogDestinationName` field.
           type: string
+          # The value names a ClusterLogDestination, so it is a Kubernetes object name and nothing else.
+          pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
```
AlwxSin: 🟡 **2. An empty `clusterLogDestinationName` is now rejected, and the changelog's grace period does not exist**

(Same in `secure.yaml` and `secure-with-dedicated-nodes.yaml`.)

The pattern has no empty alternative, so `clusterLogDestinationName: ""` is refused. Verified against all three shipped schemas:

```
default => clusterLogDestinationName in body should match
    '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
```

`""` was a working value: the template guards the block with `{{- with .parameters.clusterLogDestinationName }}`, which skips an empty string, so `""` meant "logging off" — an obvious output of any GitOps templating around the Project.

Worse, `validate.Project` is called from the reconcile loop at `manager.go:128`, not only from the webhook. Both new patterns therefore hit existing projects on the next reconcile, without any update to the Project. The changelog says:

> a project already using such a value keeps working until it is next updated

That is not true for either pattern, nor for the render check (see finding 1). Either allow the empty string in the pattern (`^(...)?$`) and fix the changelog wording, or accept the break and say so in `impact`.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/helm/injection.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	expectedObjects, err := canonicalObjects(expected)
+	if err != nil {
+		// The manifests parse only while the line breaks are there. A substitution that needs them
+		// to produce valid YAML is an unquoted one, which is the whole problem.
+		return fmt.Errorf("%w: %w", ErrParameterInjection, err)
+	}
+
+	if extra := extraObjects(actualObjects, expectedObjects); len(extra) > 0 {
```
AlwxSin: 🟡 **3. The comparison only looks one way — it is blind to objects a parameter makes disappear**

`extraObjects(actualObjects, expectedObjects)` reports only what the first render has and the second does not. Nothing looks at the opposite direction, so a parameter whose line break causes an object to *stop* rendering passes silently. Confirmed with a probe template that drops a `NetworkPolicy` when the parameter contains a newline: the check returns `nil`.

For the shipped templates this is unreachable now that every substitution is quoted. For the custom templates this check exists to cover, suppressing the `Isolated` NetworkPolicy or the ResourceQuota is as interesting to an attacker as adding a ClusterRoleBinding.

The fix is one extra call and a second message — both maps are already built.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/helm/injection_test.go
```diff
@@ -0,0 +1,388 @@
...
+			}
+
+			err := validate.Project(project, template)
+			require.Error(t, err, "the schema accepted a subject name carrying a line break")
+			assert.Contains(t, err.Error(), "administrators")
+		})
+	}
+}
+
+// Quoting the quota substitutions turned the rendered values into strings. That is the canonical
```
AlwxSin, line 313: 🟡 **4. The golden fixtures are stale copies of the pre-fix templates**

This comment identifies the problem but the workaround only covers the quota. All four fixtures still carry the vulnerable line verbatim:

- `testdata/default_case/template.yaml:146`
- `testdata/secure_case/template.yaml:184`
- `testdata/secure_with_dedicated_node_case/template.yaml:234`
- `testdata/without_ns_case/template.yaml:137`

```yaml
- kind: {{ $administrator.subject }}
  name: {{ $administrator.name }}
```

That is why quoting three shipped templates changed no golden file — the suite renders copies, not what actually ships. It also means nothing would notice if a shipped template regressed to an unquoted substitution tomorrow.

Given this is exactly the class of bug that hid here for years, the fixtures should be regenerated from the shipped templates, or `client_test.go` pointed at `../../templates/`.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/helm/injection.go
```diff
@@ -0,0 +1,183 @@
...
+
+// ErrParameterInjection is returned when a project parameter does not stay a value in the rendered
+// manifests but becomes part of their structure.
+var ErrParameterInjection = errors.New("a project parameter changes the structure of the rendered manifests")
+
+// lineBreakToSpace replaces the characters that end a YAML scalar. A parameter carrying one of them
+// can close the value it was substituted into and have the rest of itself read as manifest structure
+// -- a new key, or, after a document separator, a whole new object. The rendered manifests are
+// applied by a ServiceAccount bound to cluster-admin, so such an object is applied with cluster-admin.
+var lineBreakToSpace = strings.NewReplacer("\n", " ", "\r", " ", "\u0085", " ", "\u2028", " ", "\u2029", " ")
```
AlwxSin, line 42: 🔵 **5. The check triggers on line breaks only; a flow-style context is injectable without one**

If no parameter carries one of these characters, `ensureParametersStayValues` returns before rendering anything. A substitution in a YAML flow context needs no line break — verified against this branch:

```
flow-style check => <nil>
rendered: data: {"owner": "me", "injected": "yes"}
```

This is bounded: a new *document* requires `---` at the start of a line, so a flow-style payload can add keys and list items to an existing object but cannot conjure a `ClusterRoleBinding`. Probably not worth widening the check for.

But the PR description sells this as "a general check covers templates we do not ship", and the doc comment below says it "needs no knowledge of the template at all". Both should say what the check actually covers: substitutions escaped by a line break.

modules/160-multitenancy-manager/templates/validation.yaml
```diff
@@ -21,7 +25,10 @@ spec:
     - name: 'exclude-users'
       expression: '!(["system:sudouser", "system:apiserver", "system:kube-controller-manager", "system:kube-scheduler", "system:volume-scheduler", "dhctl", "observability"].exists(e, (e == request.userInfo.username)))'
     - name: 'exclude-restartedAt'
-      expression: '!((has(object.spec) && has(object.spec.template) && has(object.spec.template.metadata)
+      # On CREATE there is no oldObject to compare against, and reading through it would fail the
+      # condition itself -- which, with failurePolicy: Fail, would deny the request instead of
+      # letting the rule below answer it.
+      expression: 'oldObject == null || !((has(object.spec) && has(object.spec.template) && has(object.spec.template.metadata)
```
AlwxSin, line 32: 🔵 **6. The `oldObject == null` guard is unnecessary, and its comment describes a failure that does not happen**

The comment says reading through a missing `oldObject` "would fail the condition itself -- which, with `failurePolicy: Fail`, would deny the request". It would not. In CEL a presence test on a null value returns `false`, it does not error — only a deeper traversal does, and the expression guards those behind `has(oldObject.spec) &&`.

Evaluated both forms with cel-go v0.22.0 and v0.26.0, binding `object`/`oldObject` as `DynType` + `NullValue`, which is what the apiserver does (`pkg/admission/plugin/cel/condition.go:76` returns a nil interface, `activation.go` hands it straight to the adapter):

```
has(object.spec)  [obj=null]              => false
OLD expr, CREATE deployment (old=null)    => true
NEW expr, CREATE deployment (old=null)    => true
OLD expr, CREATE deployment + restartedAt => true
```

The old expression already passed the match condition on CREATE. The guard is harmless and arguably clearer, but the comment states a mechanism that isn't real, and someone will copy that reasoning into the next policy.

modules/160-multitenancy-manager/templates/validation.yaml
```diff
@@ -36,9 +43,13 @@ spec:
   validations:
     - expression: 'request.userInfo.username == "system:serviceaccount:d8-multitenancy-manager:multitenancy-manager"'
       reason: Forbidden
+      # A cluster-scoped object has no namespace to name, and now that CREATE is matched it reaches
+      # this expression as well.
       messageExpression: 'object.kind == ''Namespace'' ? ''This resource is managed
```
AlwxSin: 🔵 **7. `messageExpression` still errors on DELETE**

The cluster-scoped case is fixed — confirmed. But `object` itself is null on DELETE, so the very first term fails:

```
NEW expr, cluster-scoped object => "...managed by the multitenancy-manager module..."   (fixed)
NEW expr, DELETE (object null)  => EVAL ERROR: no such key: kind
```

The apiserver tolerates a failing `messageExpression` and falls back to the static `message:` — of which there is none here, so every denied delete of a project resource gets a generic message instead of the intended one.

Either read from `oldObject` when `object == null`, or add a `message:` fallback.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/helm/client.go
```diff
@@ -361,30 +368,46 @@ func (c *Client) Delete(_ context.Context, releaseName string) error {
 
 // ValidateRender tests project render
 func (c *Client) ValidateRender(project *v1alpha2.Project, template *v1alpha1.ProjectTemplate) error {
+	manifests, err := c.renderTemplate(project, template)
```
AlwxSin, line 381: 🔵 **8. One redundant render on the admission path**

`ValidateRender` renders here, and then `ensureParametersStayValues` below renders the same project with the same template again. For a project whose parameters carry a line break that is 3 renders per admission request instead of 2.

Passing the already-rendered `manifests` into the check would drop one.

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [22091](https://github.com/deckhouse/deckhouse/pull/22091) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit da0d560d66310667da7c8647db7948895efda816 to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/32255452218) for details.

PR #21769, opened 2026-07-31
[control-plane-manager] VCP monitoring modules and improved deckhouse for vcp mode

trofimovdals:
## Description

**Module installation.** `verity.IsSupported()` greps `/proc/filesystems` for `erofs` — a property of the node kernel, not of the pod. The VCP deckhouse pod runs in the parent cluster without `/dev` or privileges, so on a containerd v2 node it picked the dm-verity installer and died with `veritysetup close: exit status 4`; on a containerd v1 node the same pod silently worked through the symlink installer. Adds a `DECKHOUSE_DISABLE_VERITY` gate, applied inside the predicate so all three of its consumers stay consistent.

**Identity.** Deckhouse's own admission policies exempt only `system:serviceaccount:d8-system:deckhouse` and `dhctl`, so a client certificate was denied on every module namespace regardless of RBAC. Separately, Go hooks of downloadable modules read the ServiceAccount token straight from `/var/run/secrets/kubernetes.io/serviceaccount`, which no kubeconfig can cover. The manager now creates the ServiceAccount in the tenant, issues a token via `TokenRequest` and mounts it at the standard path; the kubeconfig and the `KUBECONFIG` propagation added for it are removed as dead code.

**Modules.** Seeds ModuleConfigs for `operator-prometheus`, `prometheus`, `monitoring-kubernetes`, `monitoring-custom`, `extended-monitoring`, `monitoring-ping`. `prometheus` needs `https.mode: Disabled`: the global default is `CertManager`, and a Minimal tenant has neither cert-manager nor an ingress controller.
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```

### DIFF
--- a/modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/deckhouse/manifests/deployment.yaml
+++ b/modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/deckhouse/manifests/deployment.yaml
@@ -50,8 +50,6 @@ spec:
           command:
             - /usr/bin/deckhouse-controller
             - start
-            - --kube-config
-            - /kubeconfig.yaml
           workingDir: /deckhouse
           securityContext:
             readOnlyRootFilesystem: true
@@ -71,13 +69,8 @@ spec:
             periodSeconds: 5
             failureThreshold: 120
           env:
-            - name: KUBE_CONFIG
-              value: /kubeconfig.yaml
-            - name: KUBECONFIG
-              value: /kubeconfig.yaml
-            # The kubeconfig points at the tenant apiserver; these two make the
-            # in-cluster-config fallback paths land on the same apiserver (the
-            # per-VCP ALB routes :6443 to it) instead of the parent cluster's.
+            # The pod has no kubeconfig: it authenticates as a tenant
+            # ServiceAccount through the mounted token
             - name: KUBERNETES_SERVICE_HOST
               value: "${VCP_API_VIP}"
             - name: KUBERNETES_SERVICE_PORT
@@ -170,9 +163,11 @@ spec:
               mountPath: /root/.kube
             - name: downloaded
               mountPath: /deckhouse/downloaded
-            - name: kubeconfig
-              mountPath: /kubeconfig.yaml
-              subPath: deckhouse.conf
+            # Mounted as a directory on purpose: client-go re-reads the token
+            # through BearerTokenFile, and subPath mounts are never refreshed
+            # by kubelet, so a subPath here would break token rotation.
+            - name: sa-token
+              mountPath: /var/run/secrets/kubernetes.io/serviceaccount
               readOnly: true
       volumes:
         - name: tmp
@@ -184,6 +179,6 @@ spec:
         # Downloadable modules are re-fetched after a pod restart; a PVC is a follow-up.
         - name: downloaded
           emptyDir: {}
-        - name: kubeconfig
+        - name: sa-token
           secret:
-            secretName: d8-kubeconfig-virtual-clients
+            secretName: ${TOKEN_SECRET_NAME}
--- a/modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/deckhouse/manifests/deployment.yaml
+++ b/modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/deckhouse/manifests/deployment.yaml
@@ -182,3 +182,4 @@ spec:
         - name: sa-token
           secret:
             secretName: ${TOKEN_SECRET_NAME}
+            defaultMode: 0400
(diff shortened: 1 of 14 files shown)


modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/reconcile_deckhouse.go
```diff
@@ -395,6 +410,139 @@ func (r *reconciler) reconcileParentRegistrySecret(ctx context.Context, vcp *con
...
+	name := constants.VirtualResourceName(constants.VirtualDeckhouseTokenSecretName, vcp.Name)
+	current, err := r.getSecret(ctx, vcp.Namespace, name)
+	notFound := apierrors.IsNotFound(err)
+	if err != nil && !notFound {
+		return fmt.Errorf("get token Secret: %w", err)
+	}
+
+	if !notFound &&
+		!deckhouseTokenNeedsRenewal(current) &&
+		equality.Semantic.DeepEqual(current.Data["ca.crt"], tenantCA) {
```
sprait: why not `bytes.Equal`?

modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/reconcile_deckhouse.go
```diff
@@ -395,6 +410,139 @@ func (r *reconciler) reconcileParentRegistrySecret(ctx context.Context, vcp *con
...
+				"meta.helm.sh/release-namespace": deckhouseSystemNamespace,
+			},
+		},
+		AutomountServiceAccountToken: ptr.To(false),
+	}
+
+	err := tc.Get(ctx, client.ObjectKeyFromObject(target), &corev1.ServiceAccount{})
+	if apierrors.IsNotFound(err) {
+		return tc.Create(ctx, target)
+	}
```
sprait, line 524:
```suggestion
	if err := tc.Create(ctx, target); err != nil && !apierrors.IsAlreadyExists(err) {
		return err
	}
```

modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/reconcile_deckhouse.go
```diff
@@ -395,6 +410,139 @@ func (r *reconciler) reconcileParentRegistrySecret(ctx context.Context, vcp *con
...
+
+	issuedAt, err := time.Parse(time.RFC3339, secret.Annotations[deckhouseTokenIssuedAtKey])
+	if err != nil {
+		return true
+	}
+
+	expiresAt, err := time.Parse(time.RFC3339, secret.Annotations[deckhouseTokenExpiresAtKey])
+	if err != nil {
+		return true
+	}
```
sprait: [ensureKonnectivityCPAgentToken](https://github.com/deckhouse/deckhouse/blob/7f2fb16294fac3315d45a06a1dc801a5f3e5df4e/modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/reconcile_konnectivity_agent_cp.go#L171) uses a single annotation and does essentially the same thing as similar function] Let's unify this - either extract a shared helper, or at least make both implementations consistent in how they handle the annotation.

modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/deckhouse/manifests/deployment.yaml
```diff
@@ -165,9 +163,11 @@ spec:
...
             - name: downloaded
               mountPath: /deckhouse/downloaded
-            - name: kubeconfig
-              mountPath: /kubeconfig.yaml
-              subPath: deckhouse.conf
+            # Mounted as a directory on purpose: client-go re-reads the token
+            # through BearerTokenFile, and subPath mounts are never refreshed
+            # by kubelet, so a subPath here would break token rotation.
+            - name: sa-token
+              mountPath: /var/run/secrets/kubernetes.io/serviceaccount
```
sprait, line 170: can we add  here `defaultMode: 0400`?

modules/040-control-plane-manager/images/control-plane-manager/src/internal/controllers/virtual-control-plane-configuration/reconcile_deckhouse.go
```diff
@@ -395,6 +410,139 @@ func (r *reconciler) reconcileParentRegistrySecret(ctx context.Context, vcp *con
...
+
+	name := constants.VirtualResourceName(constants.VirtualDeckhouseTokenSecretName, vcp.Name)
+	current, err := r.getSecret(ctx, vcp.Namespace, name)
+	notFound := apierrors.IsNotFound(err)
+	if err != nil && !notFound {
+		return fmt.Errorf("get token Secret: %w", err)
+	}
+
+	if !notFound &&
+		!deckhouseTokenNeedsRenewal(current) &&
```
sprait: suggestion
```go
	name := constants.VirtualResourceName(constants.VirtualDeckhouseTokenSecretName, vcp.Name)
	current, err := r.getSecret(ctx, vcp.Namespace, name)
	switch {
	case apierrors.IsNotFound(err):
		current = nil
	case err != nil:
		return fmt.Errorf("get token Secret: %w", err)
	case isDeckhouseTokenInSync(current, tenantCA):
		return nil
	}

	if current == nil {
		return r.createSecret(ctx, target)
	}
...
func isDeckhouseTokenInSync(secret *corev1.Secret, tenantCA []byte) bool {
	return !deckhouseTokenNeedsRenewal(secret) && bytes.Equal(secret.Data["ca.crt"], tenantCA)
}
```

sprait (changes requested):
some comments

sprait (approved):

PR #21843, opened 2026-08-04
[deckhouse] explain which requirement is unmet when an Application is rejected

Fkuloff:
## Description

Requirement rejections for `Application` objects now name what was checked, what was required and what the cluster actually has.

- `schedule/rule/version` builds its own message instead of surfacing the raw semver error, and takes a `subject` (`Kubernetes`, `Deckhouse`).
- `schedule/rule/condition` now carries a message. It had none, so an unsatisfied condition denied the request with no text at all.
- The `Application` validating webhook names the object and the package version whose requirements were evaluated.

`CheckConstraints` surfaces only the decision message and drops the `Reason` that says which subsystem a requirement belongs to, so on the admission path the message is everything the user gets. A rejected `Application` used to read:

```
The request is invalid: : check version error: 1.34.9 is less than v1.35
```

Nothing says the number is a Kubernetes version; the constraint operator is gone, because semver prints only the version part of `>= v1.35`, so a requirement reads as a plain comparison; `check version error` is an internal wrapper leaking into `kubectl` output; and no object is named. An unmet Deckhouse requirement rendered in exactly the same shape, so the two were indistinguishable.

The condition rule was worse. Every `Application` admission evaluates the bootstrap gate, because the webhook always builds constraints with `Order: FunctionalOrder`, and with no message the denial printed bare `The request is invalid` with no explanation whatsoever.

These messages are also what the runtime writes into the `RequirementsMet` condition of a package, so the status path gets the same wording.

### Manual testing

Checked on a cluster running Kubernetes `1.34.9`, with three package versions published to exercise each branch. Package and application names are sanitized; command output is verbatim.

Kubernetes requirement unmet — the case this PR is about (package requires `>= v1.35`):

```
# k apply -f app.yaml
The request is invalid: : Application 'default/example-app-req' (package 'example-app' version 'v0.1.2'): requires Kubernetes version '>=v1.35', current version is '1.34.9'
```

Deckhouse requirement unmet (package requires `>= v3.0.0`) — previously indistinguishable from the Kubernetes rejection, both rendered as `check version error: <x> is less than <y>`:

```
# k apply -f app.yaml
The request is invalid: : Application 'default/example-app-req' (package 'example-app' version 'v0.1.3'): requires Deckhouse version '>=v3.0.0', current version is '2.0.0'
```

Both requirements satisfied — the version rule's pass branch, which this PR also touches:

```
# k apply -f app.yaml
application.deckhouse.io/example-app-req created

# k get app example-app-req
NAME               PACKAGE       VERSION   STATE   MESSAGE   AGE
example-app-req    example-app   v0.1.4                      0s
```

The bootstrap condition message is not covered here: `clusterIsBootstrapped` is held true by a global hook that recreates the backing ConfigMap, so an unsatisfied bootstrap gate cannot be reproduced on a healthy cluster.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: explain which requirement is unmet when an Application is rejected
impact_level: low
```

### DIFF
--- a/deckhouse-controller/internal/packages/schedule/node.go
+++ b/deckhouse-controller/internal/packages/schedule/node.go
@@ -160,15 +160,15 @@ func (s *Scheduler) addNode(pkg Package) {
 	}
 
 	if constraints.Kubernetes != nil && s.kubeVersionGetter != nil {
-		n.rules = append(n.rules, version.NewRule(s.kubeVersionGetter, constraints.Kubernetes, reasonRequirementsKubernetes))
+		n.rules = append(n.rules, version.NewRule(s.kubeVersionGetter, constraints.Kubernetes, subjectKubernetes, reasonRequirementsKubernetes))
 	}
 
 	if constraints.Deckhouse != nil && s.deckhouseVersionGetter != nil {
-		n.rules = append(n.rules, version.NewRule(s.deckhouseVersionGetter, constraints.Deckhouse, reasonRequirementsDeckhouse))
+		n.rules = append(n.rules, version.NewRule(s.deckhouseVersionGetter, constraints.Deckhouse, subjectDeckhouse, reasonRequirementsDeckhouse))
 	}
 
 	if constraints.Order == FunctionalOrder && s.bootstrapCondition != nil {
-		n.rules = append(n.rules, condition.NewRule(s.bootstrapCondition, reasonRequirementsBootstrap))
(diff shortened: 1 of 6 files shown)


ipaqsa (approved):

ldmonster (approved):

deckhouse-BOaTswain:
Cherry pick PR [21856](https://github.com/deckhouse/deckhouse/pull/21856) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #21850, opened 2026-08-04
user-authz: report what a role grants, not only what a subject can do

Jabejixo:
The permission browser answers two questions about people -- who may do this, and what may this one do -- and none about the role model itself. The document a security officer keeps and a regulator reads is the third one: which resources are covered by which role. Today it is assembled by hand from ClusterRoles, which is why the tables in circulation still name the roles of the model before last.

RoleAccessReport is the catalogue side of SubjectAccessReport. Given a role, or a selection of them, it returns what they grant, in either model: the scope-based roles, or the access levels of ClusterAuthorizationRule.

The rows are produced by the same builder as the subject report. Sharing it is the point: a wildcard has to read the same way in the document about a role and in the document about a person, and two implementations of that expansion would drift into contradicting each other -- which is worse than having no document.

What the report does that reading a ClusterRole does not:

The rows come from the capabilities, not from the role's own rules. The aggregation controller fills a role by concatenating what its capabilities grant, without recording which one granted what, so expanding them one at a time is the only way to say which capability covers a resource -- the question a regulator actually asks. The aggregation is followed to the bottom: the levels include one another by aggregating the role below, and "the manager role" is not an answer either. Following it does not double count, because the rules of an included role are the copy the controller already made.

A namespace- or project-scoped role drops its cluster-scoped rows. The binding webhook refuses a ClusterRoleBinding carrying such a role, so those rules exist in RBAC and can never be exercised; listing them overstates access.

A legacy access level is resolved the way the fan-out hook resolves it: the level's own ClusterRole plus every ClusterRole annotated for that level or a lower one. That knowledge lives in a hook and moves with the platform, which is why it belongs here rather than in whatever assembles the document.

The snapshot is what makes two exports comparable. It carries the digest of the reported roles -- the timestamp deliberately left out of it, so an unchanged cluster hashes the same -- and the size of the discovery snapshot the wildcards were expanded against, without which a wildcard row cannot be read a year later.

Access: the report describes roles, never subjects, and every row of it is already readable in the ClusterRoles themselves, so it needs no self/non-self split. It is granted alongside the self-queries of the existing access levels, and `d8:user-authz:role-catalogue-reader` exists for identities that should produce the export without reading the whole RBAC tree.

## Description

A new resource, `RoleAccessReport` in `authorization.deckhouse.io/v1alpha1`, served by the permission browser API server. It is computed at request time and stored nowhere.

Around it:

- **Coverage.** `includeInventory` adds every resource of the discovery snapshot with the verbs the API server accepts for it, so a review can name what no role reaches. Roles alone cannot answer that: a resource nobody grants leaves no trace in them. Coverage is measured against the accepted verbs and not against a fixed list of eight -- `tokenreviews` only ever accept `create`, and "1 of 8" would read as a gap in the model where there is none.
- **Module attribution.** A resource is filed under the module that ships it: the CRD's own label, then the APIService of an aggregated API, then a group that exactly one module's capabilities claim. A resource with no definition of its own is never attributed -- reading a role as ownership of everything it names would file half of Kubernetes under `user-authz`. The inventory also carries whether a kind is a service one (`deckhouse.io/system-resource`) or holds cluster configuration, so a review can leave out the objects that exist for controllers.
- **Ordering.** Roles run from the narrowest access to the widest -- by scope, then by level inside it -- because that is how the model is read. Alphabetically the document opens with the administrator.
- **`excludeCustom`** reports the model the platform ships, with the count of what it skipped in `notes`.

The console side is [console!3449](https://github.com/deckhouse/deckhouse-external-modules/-/merge_requests/3449): the catalogue page, the coverage matrix, CSV and PDF.

One change outside the report: the multi-tenancy webhook answered "is this granted independently of a ClusterAuthorizationRule" by walking every ClusterRoleBinding on every request. It is an index now (99% off that check; the benchmark is in the tree), because the same question is now asked by more callers.

The compliance document exists today regardless -- it is assembled by hand, from `kubectl get clusterrole -o yaml` and institutional memory, and it goes stale the moment a module ships a capability. Nothing in the platform could produce it, so nothing could keep it current.

It also has to be produced for the model a cluster actually runs. Both are supported: the scope-based roles and the legacy access levels, resolved the way the fan-out hook resolves them, so a cluster that has not migrated is not excluded from its own audit.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: feature
summary: The permission browser reports what a role grants, not only what a subject can do, in either role model -- with the coverage of the cluster's resources and the module each of them belongs to.
```


modules/140-user-authz/templates/cluster-roles.yaml
```diff
@@ -400,6 +400,16 @@
...
   verbs:
   - create
+# The role catalogue: what the platform's own roles grant. Unlike the queries
+# above it says nothing about any subject -- it is the published role model,
+# assembled from ClusterRoles this access level can already read, so it is
+# granted here rather than through a checker role.
+- apiGroups:
+  - authorization.deckhouse.io
+  resources:
+  - roleaccessreports
```
AlwxSin, line 410: ### 2. 🟡 No rbacv2 grant — the API is unreachable in the model this branch is building

This is the legacy `user_authz_cluster_admin_rules`, and the grant is correct here: the same block reads `clusterroles` 27 lines above, so the report discloses nothing new. But the rbacv2 side is missing.

`modules/140-user-authz/templates/rbacv2/global/subsystem/capabilities/inspect_access.yaml:27-31` is the capability that grants the sibling report APIs:

```yaml
      - bulksubjectaccessreviews
      - bulksubjectaccessreviews/nonself
      - subjectaccessreports
      - subjectaccessreports/nonself
      - whocans
```

`roleaccessreports` is not added to it. So on a cluster running the model this PR targets, the new API is reachable only through the legacy ClusterAdmin access level or a hand-made binding of `d8:user-authz:role-catalogue-reader` — and the security officer this feature exists for is exactly the identity that would hold a `manage` / `inspect_access` role.

ee/be/modules/140-user-authz/images/webhook/src/internal/web/hook/rbac.go
```diff
@@ -76,19 +84,126 @@ func NewRBACEvaluator(logger *log.Logger, informerFactory informers.SharedInform
...
+		if err != nil {
+			logger.Printf("independent RBAC check: failed to index ClusterRoleBindings: %v", err)
+
+			return
+		}
+
+		evaluator.bindingsBySubject.rebuild(bindings)
+	}
+
+	if _, err := clusterRoleBindings.Informer().AddEventHandler(kcache.ResourceEventHandlerFuncs{
```
AlwxSin, line 123: ### 3. 🟡 The subject index is not part of `Synced()` — a startup window of wrong denials

`AllowsIndependently` gates on `cachesSynced()` (rbac.go:232), which is the informers' `HasSynced`. But the index is now filled from this event handler, and client-go's `processor.distribute` only *queues* notifications into a listener's `pendingNotifications` — it does not wait for the listener goroutine. So `HasSynced` can be true while `rebuild()` has not run yet.

In that window `bindingsFor` returns nothing → `AllowsIndependently` is `false` → the multi-tenancy deny stands → a cluster-scoped request that a non-CAR `ClusterRoleBinding` legitimately allows gets a 403.

Before this change the read came straight from the lister, which *is* in step with `HasSynced`, so this is new. Safe direction, and the window is short — but a wrong deny is user-visible on every webhook restart and leader change, which is not a great failure mode for an authorization webhook.

Cheapest fixes: append an index-ready `InformerSynced` to `Synced()`, or let `bindingsFor` fall through to the lister until the first rebuild lands.

ee/be/modules/140-user-authz/images/webhook/src/internal/web/hook/rbac.go
```diff
@@ -76,19 +84,126 @@ func NewRBACEvaluator(logger *log.Logger, informerFactory informers.SharedInform
...
 			clusterRoles.Informer().HasSynced,
 			clusterRoleBindings.Informer().HasSynced,
 		},
 	}
+
+	// Rebuilding the whole index on every event is O(bindings), and bindings
+	// change when a module is deployed, not when a request arrives. Tracking
+	// deltas would buy nothing here and would have to get subject removal on
+	// update exactly right.
+	rebuild := func() {
```
AlwxSin, line 115: ### 4. 🔵 `rebuild()` on every event makes the initial sync O(bindings²)

`AddFunc` / `UpdateFunc` / `DeleteFunc` all call this, and it does a full `List` plus a full rebuild.

The comment above argues that bindings change when a module is deployed, not when a request arrives — true in steady state, but the initial sync is precisely a burst of N `Add` events. A cluster with the "hundreds of ClusterRoleBindings" the comment cites therefore does hundreds of full passes at startup.

Still a large net win over the old per-request `List`, so this is only a nit. One `HasSynced` guard inside `rebuild` collapses the sync burst into a single pass.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/roleaccess.go
```diff
@@ -0,0 +1,821 @@
...
+}
+
+// digestOf hashes the canonical form of the reported roles.
+//
+// It is what makes two exports comparable: an unchanged cluster produces the
+// same digest, so "nothing changed since last quarter" is something a reader
+// can verify instead of having to diff two documents by eye. The snapshot
+// itself is not hashed -- the timestamp would make every digest unique, which
+// is the opposite of the point.
+// digestOf hashes the reported roles.
```
AlwxSin: ### 5. 🔵 Two stacked doc comments on `digestOf`

Both blocks open with `// digestOf hashes`, and godoc will render both. Leftover from an edit — the second one (the streaming rationale) reads as the keeper, the first one's "what makes two exports comparable" paragraph is worth folding into it.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/roleaccess.go
```diff
@@ -0,0 +1,821 @@
...
+//
+// The JSON is streamed into the hash rather than marshalled into memory first:
+// a report of a large cluster is tens of megabytes, and holding a second copy
+// of it only to throw it away is the most expensive thing this function could
+// do. The digest itself is opaque -- it is compared between two reports, never
+// to a constant -- so the framing of the stream does not matter.
+func digestOf(roles []v1alpha1.RoleAccess) string {
+	hasher := sha256.New()
+	if err := json.NewEncoder(hasher).Encode(roles); err != nil {
+		return ""
```
AlwxSin: ### 6. 🔵 `digestOf` swallows the encode error and returns `""`

Two reports that both failed to hash compare equal — which is exactly the conclusion ("nothing changed since last quarter") the digest exists to let a reader verify. The doc comment two lines up makes that the whole point of the field.

Return the error, or a value that cannot be mistaken for a digest.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/roleaccess.go
```diff
@@ -0,0 +1,821 @@
...
+		scopeCache:        scopeCache,
+		limits:            DefaultReportLimits,
+		now:               time.Now,
+	}
+}
+
+// WithModuleIndex attributes the inventory to the modules that ship it. Without
+// the index the inventory is still reported, just without module names: the CRDs
+// may be unreadable, and a coverage report without grouping beats none.
+func (r *RoleAccessResolver) WithModuleIndex(index *ModuleIndex) *RoleAccessResolver {
```
AlwxSin: ### 7. 🔵 `WithModuleIndex` mutates the receiver after construction

Builder shape, in-place write, no lock. Called once at wiring today (`apiserver.go:324-327`), before the server serves — but the signature invites calling it on a live resolver, and `Report` reads `r.moduleIndex` from request goroutines.

Taking the index in `NewRoleAccessResolver` removes the question entirely.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/roleaccess.go
```diff
@@ -0,0 +1,821 @@
...
+	// ExpandWildcards expands wildcard rules against discovery.
+	ExpandWildcards bool
+	// IncludeComposition reports what each role is assembled from.
+	IncludeComposition bool
+	// IncludeInventory adds every resource of the cluster to the report.
+	IncludeInventory bool
+}
+
+// Report builds the catalogue for the requested roles.
+func (r *RoleAccessResolver) Report(_ context.Context, req RoleAccessRequest) (v1alpha1.RoleAccessReportStatus, error) {
```
AlwxSin: ### 11. 🔵 `Report` drops the request context

Correct today — the lister, `scopeCache` and `moduleIndex` are all in-memory, so nothing here blocks. But a full-cluster report with `includeInventory` and wildcard expansion is bounded only by `MaxResourceRows`, and a client that hung up cannot stop it. Threading `ctx` into the per-role loop costs one check.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/roleaccess.go
```diff
@@ -0,0 +1,821 @@
...
+		LegacyNames: index.legacyNames[role.Name],
+		Namespaced:  namespaced,
+	}
+
+	components := r.componentsOf(index, role)
+
+	builder := newReportBuilder(r.limits, req.ExpandWildcards, r.scopeCache)
+	scope := newScopeAccumulator()
+
+	if len(components) == 0 {
```
AlwxSin, line 533: ### 12. 🔵 A role whose aggregation currently matches nothing reports zero rows, silently

A role that *has* an `aggregationRule` whose selector matches no capability right now — module not yet deployed, capability CRD not yet installed — lands in this branch and is reported from `role.Rules`, which the aggregation controller has emptied. Zero rows.

And the note that would explain it cannot fire: line 530 requires `len(components) > 0`.

So in the document a regulator reads, that role says "grants nothing" rather than "we could not resolve it". Distinguishing `role.AggregationRule != nil && len(components) == 0` from a genuinely self-carrying custom role would cover it.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/roleaccess.go
```diff
@@ -0,0 +1,821 @@
...
+
+// legacyRoles reports the access levels of ClusterAuthorizationRule.
+//
+// A level is not a role: the fan-out binds the level's own ClusterRole plus
+// every ClusterRole annotated for that level or a lower one. The report says
+// what the level grants, which is the union of those.
+func (r *RoleAccessResolver) legacyRoles(index *roleIndex, req RoleAccessRequest) []v1alpha1.RoleAccess {
+	wanted := setOf(req.AccessLevels)
+	for _, name := range req.Names {
+		wanted[name] = struct{}{}
```
AlwxSin, line 663: ### 10. 🔵 `Names` intersects in the primary model and unions in the legacy one

Here `Names` is merged into `AccessLevels`. In `primaryRoles` (line 424-432) `Names` and `Scopes` are ANDed instead.

`validateRoleAccessSpec` already rejects the cross-model combinations, so the only reachable case is `{names: [Admin], accessLevels: [User]}` — which returns both. Either intersect here too, or say in the `Names` godoc that in the legacy model it names access levels and merges with `accessLevels`.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/controller/namespace/controller.go
```diff
@@ -218,7 +218,7 @@ func (p customPredicate[T]) Delete(_ event.TypedDeleteEvent[T]) bool {
 }
 
 func isNil(arg any) bool {
-	if v := reflect.ValueOf(arg); !v.IsValid() || ((v.Kind() == reflect.Ptr ||
+	if v := reflect.ValueOf(arg); !v.IsValid() || ((v.Kind() == reflect.Pointer ||
```
AlwxSin: ### 8. 🔵 Scope creep: `reflect.Ptr` → `reflect.Pointer` in another module

Same change in `project/controller.go:193`. Both are unrelated to the role catalogue, in a PR of 33 files — and `reflect.Ptr` is a permanent alias for `reflect.Pointer`, so this changes nothing at all.

Worth dropping, especially since #21756 is concurrently touching this module.

AlwxSin (commented):
Three findings below have no line in the diff to anchor to, so they live here.

---

### 1. 🟡 The PR body is the unfilled template — the changelog block will land as placeholders

`## Description`, `## Why do we need it, and what problem does it solve?` and the whole checklist are empty, and the ```changes``` block still carries the literal template:

```
type: fix | feature | chore
```

`CHANGELOG/CHANGELOG-v<X.Y>.0.yml` is generated from that block, so this either fails the generator or writes the placeholders straight into the release changelog.

The prose above `## Description` is doing the description's job very well — it is just not in the fields the tooling reads. The checklist also says the code is not covered by unit tests, against ~1100 lines of them (`roleaccess_test.go`, `module_index_test.go`, `roleaccess_storage_test.go`, plus two bench files).

---

### 9. 🔵 Commit `f166f7f5` is titled `fix`

Every other commit in the branch reads like a sentence. Worth rewriting before merge.

---

### Cross-PR, because both bear on this branch

**🟡 C1 — rbacv2 hands back what #21672 withholds.** #21672's whole fix is that the cluster-wide `projects` read must not exist where the namespace ACL applies, or RBAC allows before the storage filter can narrow. `modules/160-multitenancy-manager/templates/rbacv2/manage/view.yaml:14-21` grants it unconditionally:

```yaml
- apiGroups: [deckhouse.io]
  resources: [projects]
  verbs: [get, list, …]
```

Defensible for a platform manager. But this PR sits on `integration/rbacv2`, and when that model becomes the default the filter goes inert for every `manage`-role holder — with nothing in the tree recording that the grant and the filter are coupled. #21672 put that reasoning in a comment on the user-authz side; the rbacv2 `manage/view.yaml` has no counterpart.

**🟡 C2 — textual conflict with #21672.** `git merge-tree` between the two heads:

```
CONFLICT (content): ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/apiserver/apiserver.go
CONFLICT (content): modules/140-user-authz/docs/README.md
CONFLICT (content): modules/140-user-authz/docs/README_RU.md
```

Both wire into the permission-browser's `apiserver.go` (#21672 in `initAuthorizers`, this one in `New`) and both regenerate the access-level doc tables — #21672 removes `deckhouse.io/projects` from the `read:` list, this one adds `roleaccessreports`. Mechanical, but whichever merges second should re-run the doc generation rather than resolve the tables by hand, or they end up describing neither state.

---

Evidence for the inline findings: `go test ./pkg/resolver/... ./pkg/registry/...` and `go vet ./...` in the permission-browser module, and `go test ./...` in the webhook module, all pass on `58895f8d78`. Nothing below is a test failure — the suite is green.

Jabejixo:
Thanks -- all of it was valid, and all of it is fixed. Force-pushed, so the inline threads are anchored to commits that no longer exist; the mapping is below.

**The report was unreachable in the primary model.** `roleaccessreports` is now in the `inspect_access` capability, next to the other reports of the permission browser. The capability is held by the identity that reviews access, and the catalogue of what each role grants is the other half of that review -- it discloses nothing elevated, every row of it is readable in the ClusterRoles themselves.

**The subject index answered before it was filled.** You were right that `Synced()` does not cover it, and I did not add it there: readiness of the webhook should not depend on an index whose failure to build is recoverable. Instead the events of the initial listing are ignored -- they were also what made the sync quadratic, one full rebuild per binding -- and the first read after the sync builds the index once. `TestSubjectIndexIsBuiltBeforeTheFirstAnswer` fails without it: a binding that predates the process was invisible, and on a settled cluster it stayed invisible until something happened to change a binding.

**`digestOf`** has one doc comment now, and returns an error instead of an empty string. An empty digest is worse than none: two failed reports compare equal, which is the one thing the digest exists to answer truthfully.

**An aggregation that matches nothing** now says so in `notes`. It still reports no rows, but "grants nothing" and "grants whatever gets labelled for it, and nothing is yet" are different findings, and the second one usually means a subsystem is not installed or a label was lost.

**`WithModuleIndex`** is gone; the index is a constructor argument. It was a write to a resolver that concurrent reports were reading.

**`Names` in the legacy model** stays a union with `accessLevels` -- the model has no role names to narrow by, so a name there can only be read as a level, and rejecting it would be the surprising behaviour. What was missing is that nothing said so: the field, the internal request type and the README now do.

**The context** is honoured in both role loops. Abandoned exports of a large cluster kept the API server working for a client that had gone.

**The `reflect.Ptr` change** is out of the branch entirely -- rebased out of the commit that carried it, not reverted on top, so the diff of `multitenancy-manager` is only the comment mentioned below.

**The commit named `fix`** is now "user-authz: leave the custom roles out of the catalogue on request", which is what it did.

**The PR body** is filled in, with a changelog entry.

On the cross-PR points: the `projects` grant in the manager capability is deliberate -- a platform operator manages the tenants and has to see all of them -- but you are right that nothing recorded the coupling with the visibility filter, so the template now says why it is cluster-wide and what removing it would mean. The textual conflict with #21672 is real and unavoidable; whichever merges second regenerates the docs.

PR #21855, opened 2026-08-04
[node-manager] Own bashible objects inside a virtual control plane tenant

AmazinMax:
## Description

In a virtual control plane (VCP) tenant cluster the `node-manager` module is now enabled and owns the node bootstrap objects that `control-plane-manager` used to write into the tenant on its behalf: the bashible CRDs, the bashible RBAC, the `bashible-bashbooster` Secret and the `bashible-apiserver-context` Secret. The corresponding reconcile steps and the embedded `manifests/crds.yaml` and `manifests/rbac.yaml` copies are deleted from `control-plane-manager`.

The tenant runs the module without any of its workloads. All templates that produce a Pod, a controller or an alert stay behind `global.deckhouseSelfHosted` — 55 of the module's 63 templates render nothing in a tenant. The eight that do are the bashible RBAC, the bashible-bashbooster Secret, the new context Secret, and the ClusterRoles describing access to the module's own resources (`user-authz-cluster-roles.yaml` and the two rbacv2/manage roles), which belong in a tenant because the CRDs they grant access to are installed there. Three hooks that write unconditionally (`create_master_node_group`, `order_bootstrap_token`, `gen_bashible_apiserver_certs`) return early in a nested cluster, where those objects belong to the VCP.

For the bashible context the host now publishes a contract instead of the finished document. `control-plane-manager` writes `d8-cloud-instance-manager/bashible-external-inputs` with the facts a tenant cannot compute — VCP PKI, signed `apiserverProxyCerts`, ALB addresses and RPP ports, the join token, the NodeGroup list. The new hook `bashible_context_vcp.go` assembles the context through `go_lib/bashiblecontext`, overlays the published inputs and hands the rendered `input.yaml` to `templates/bashible-apiserver/context-secret.yaml`. Missing inputs, an unknown contract `version` or absent credentials leave the value unset, so the Secret is not rendered at all rather than rendered incomplete.

`bashiblecontext` moved from `node-controller`'s `internal/` tree to `go_lib/bashiblecontext`, a separate Go module wired like `go_lib/dependency/k8s/drain`. Both writers of `input.yaml` — the `node-controller` Pod in a normal cluster and the hook in a tenant — now share one implementation instead of two that would drift.

**Impact on critical components.** A self-hosted cluster is unaffected: every change is behind `global.deckhouseSelfHosted` and the rendered chart is unchanged there. In a tenant, ownership of `bashible-apiserver-context` moves from `control-plane-manager` to the `node-manager` Helm release. The transitional hook `bashible_context_vcp_adopt.go` stamps Helm ownership metadata onto an already existing Secret before the chart renders, otherwise the release would fail with `invalid ownership metadata`. As a consequence, disabling `node-manager` in a tenant now deletes that Secret, which was harmless before.

Until now a tenant cluster had no `node-manager`, so `control-plane-manager` re-implemented the parts of it that nodes cannot bootstrap without, and kept them as copies inside its own image. Copies drift: the embedded bashible `manifests/crds.yaml` had already diverged by 393 lines from the module's own `crds/`, which means a tenant could be served CRDs the module no longer produces. The same risk applied to RBAC and to the hand-written `input.yaml` builder, whose format is a contract with `bashible-apiserver`.

Moving these objects to the module removes the second implementation in every case, and reduces the host's remaining responsibility to a small, versioned input contract. This also unblocks running the tenant's node lifecycle on the module's own code as more of it becomes usable there.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: feature
summary: The module now provides bashible CRDs, RBAC and the bashible context inside a virtual control plane tenant cluster, replacing the copies control-plane-manager used to write there.
impact_level: default
```


modules/040-node-manager/hooks/discover_kubernetes_ca.go
```diff
@@ -31,6 +31,13 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 const rootCAFile = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 
 func discoverKubernetesCAHandler(_ context.Context, input *go_hook.HookInput) error {
+	// TODO: delete after rebase https://github.com/deckhouse/deckhouse/pull/21769
+	// where serviceaccount is mounted to the pod
```
sprait: Is it time?

modules/040-node-manager/hooks/bashible_context_vcp.go
```diff
@@ -0,0 +1,257 @@
...
+// publishes it for templates/bashible-apiserver/context-secret.yaml to render.
+//
+// Every early return here leaves nodeManager.internal.bashibleContext unset, which leaves the
+// Secret unrendered. That is deliberate: a context missing the CA, the proxy certificates or the
+// bootstrap token passes schema validation and bootstraps nodes silently wrong, so publishing
+// nothing is the safe failure.
+func handleBashibleContextVCP(ctx context.Context, input *go_hook.HookInput) error {
+	if !nestedControlPlane(input) {
+		return nil
+	}
```
sprait, line 144: If get early return, nothing will be written to `values`, and the bashible-apiserver-context secret will be deleted. And bashibles will break?

ee/be/modules/040-node-manager/openapi/values.yaml
```diff
@@ -212,6 +212,14 @@ properties:
...
           Private key for API server used by bashible.
 
+      bashibleContext:
+        type: string
+        description: |
+          Rendered bashible input.yaml. Only set in a nested (virtual control plane)
+          cluster, where no node-controller Pod exists to write the Secret itself.
+        x-examples:
+          - "clusterDomain: cluster.virtual\n"
+
```
sprait, line 222: Previously we stored bashibleContext in a secret, now in memory (values). How much will dkp bloat with a large number of vcp/ng?

AmazinMax: Each VCP tenant runs its own Deckhouse process in the parent, so the number of tenants multiplies pods, not the footprint inside one of them — every process holds exactly one context, its own.

The context also ends up in the Helm release Secret and is visible in plaintext through `deckhouse-controller module values node-manager`, private key and tokens included. That is not new for this module — `bashibleApiServerKey` and `bootstrapTokens` have always been there — but it is the real cost of routing the document through values.

modules/040-node-manager/hooks/bashible_context_vcp.go
AmazinMax: true, will be fixed

modules/040-node-manager/hooks/bashible_context_vcp.go
```diff
@@ -0,0 +1,297 @@
...
+	if kubeconfigPath == "" {
+		return rest.InClusterConfig()
+	}
+
+	restCfg, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("load kubeconfig %q: %w", kubeconfigPath, err)
+	}
+	return restCfg, nil
+}
```
sprait: oudated?

ee/be/modules/040-node-manager/openapi/values.yaml
```diff
@@ -212,6 +212,14 @@ properties:
         description: |
           Private key for API server used by bashible.
 
+      bashibleContext:
```
sprait, line 215: also openapi in ee/cse

modules/040-node-manager/templates/bashible-apiserver/context-secret.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if and (not .Values.global.deckhouseSelfHosted) .Values.nodeManager.internal.bashibleContext }}
```
sprait, line 4: add comment that secret for vcp

z9r5 (approved):

sprait (approved):

PR #21896, opened 2026-08-06
[user-authn] reject reserved system: identities in basic-auth-proxy

Jabejixo:
## Description

`basic-auth-proxy` copied directory group names into `X-Remote-Group` verbatim. This change rejects identities in the reserved `system:` namespace at the point where the identity headers are written:

- Groups whose name begins with `system:` are dropped, and each drop is logged with the user and the group so an operator can see why a group vanished.
- A `system:`-prefixed username cannot be dropped, because it *is* the identity, so it is refused with the same `403 Forbidden` the proxy already returns for bad credentials. That keeps the response uniform and does not leak which usernames exist. It is logged distinctly.

The filtering lives in `modifyRequest` and nowhere else. That is the single point where `X-Remote-User` and `X-Remote-Group` are written and every provider path funnels through it, so no provider can bypass it. Filtering earlier, where provider groups enter the cache, would have to be repeated in each of the three provider implementations and a fourth provider added later could omit it. It would also turn the Crowd-specific "no allowed groups" check in `ServeHTTP` into a rejection for a user whose groups all happen to be reserved, whereas the intended behaviour is to authenticate that user with no groups at all.

No cluster component restarts beyond the `basic-auth-proxy` Deployment itself, which only exists when the feature is enabled.

The commit also realigns a struct in `proxy_test.go` that already failed `gofmt` on `main`, in a test this change does not otherwise touch.

### Scope notes

**`X-Remote-Extra-*` is already safe and is not changed.** The original report asked about it. `stripIdentityHeaders` deletes every header carrying the `X-Remote-Extra-` prefix on ingress, and the proxy never re-adds an extra of its own, so no directory-sourced value can reach `user.Info.Extra`. There is an existing regression test for this.

**A `system:`-prefixed username is not reachable over basic auth today.** RFC 7617 gives the first colon to the user-id/password separator, so `SetBasicAuth("system:masters", "pw")` is read back by `r.BasicAuth()` as user `system` with password `masters:pw`. The username guard is therefore defence in depth at the choke point for any future caller that does not source the login from `r.BasicAuth`, and it mirrors the rule the OIDC path already enforces. A test pins that parsing behaviour so the claim does not silently rot.

**Follow-up, not implemented here:** an `allowedGroups` allow-list exists for the Crowd provider only (`provider/crowd.go`), configured from `DexProvider.spec.crowd.groups`. `provider/oidc.go` and `provider/ldap.go` return `claims.Groups` with no equivalent. To be accurate about what it does: it is a general allow-list, not a `system:` filter, and it is inert when unconfigured — `GetGroups` only intersects when `len(allowedGroups) > 0` and otherwise appends every group. So it would **not** have prevented this defect on its own: an operator who never set `groups`, or who set it and also had a `system:masters` entry in it, was still exposed, and OIDC and LDAP users were exposed unconditionally. Extending an allow-list to the other two providers is reasonable work but is a different change with a different user-facing contract.

The proxy authenticates a user against an external directory and forwards their identity to kube-apiserver using request-header authentication. It presents a client certificate whose CN is `front-proxy-client` (`hooks/generate_basic_auth_proxy_cert.go:177`), which is exactly the identity in `--requestheader-allowed-names` (`candi/control-plane/kube-apiserver.yaml.tpl:137`). kube-apiserver therefore trusts `X-Remote-User` and `X-Remote-Group` from this proxy without further checks.

Group names came from the directory and were copied into `X-Remote-Group` with no filtering, so a user who belonged to a directory group literally named `system:masters` became a member of that Kubernetes group and picked up the `cluster-admin` binding that ships with it. The group name is chosen by the directory administrator, not by Kubernetes, and the directory may be shared with systems where `system:masters` is an unremarkable name.

Two things make this worth fixing rather than documenting:

The code twelve lines above the defect already named this exact escalation. The comment on `stripIdentityHeaders` explains that kube-apiserver trusts these headers and that leaving a client-supplied one through "would let any client become cluster-admin via `X-Remote-Group: system:masters`". But it describes a different attacker: the HTTP client spoofing the header on the way in, which is why the function deletes every client-supplied `X-Remote-*` value. The threat model did not cover the same value arriving from the trusted side, out of the directory, after authentication had already succeeded.

The project had already decided that `system:`-prefixed identities from an external IdP are unacceptable, and enforces it on the OIDC path: `modules/040-control-plane-manager/templates/_authentication_configuration.tpl:31-35` carries `userValidationRules` rejecting both a `system:` username and any `system:` group. Those rules sit inside the `jwt:` list item (and inside `{{- if .apiserver.oidcIssuerURL }}`), so they constrain JWT authenticators only and do not apply to the request-header path at all. This change makes the two paths agree.

### Severity, honestly bounded

This is not remotely reachable in a default installation. The code path only exists when `publishAPI` is enabled **and** a provider has `enableBasicAuth` set:

- `publishAPI.enabled` defaults to `false` in `openapi/config-values.yaml`.
- `enableBasicAuth` is an unset boolean with no default in the `DexProvider` CRD, so it is off unless an operator opts in.
- `templates/_helpers.tpl` (`is_basic_auth_enabled`) requires both. Note it accepts `Crowd`, `OIDC` **and** `LDAP`, so the exposure is slightly wider than a Crowd-only reading suggests, though `enableBasicAuth` requires exactly one such provider.

Exploiting it also requires the ability to create or join a group with a specific name in the external directory, which is normally an administrative action. The realistic scenario is not a determined attacker so much as a shared or pre-existing directory that already contains a group named `system:masters` for unrelated reasons, silently granting cluster-admin to everyone in it.

**Recommendation: yes, backport to `release-1.77`.**

The cherry-pick is clean. `proxy.go` and `proxy_test.go` are byte-identical on `main` and `release-1.77` (blob `c442acd3d5` and `6f3d7b7659` on both), so the commit applies without conflict.

Weighing it: the non-default configuration required to reach the path argues against urgency, and on its own would justify waiting for the next minor. What tips it the other way is that the failure is silent and the fix is small. An affected cluster shows no symptom — a directory group grants cluster-admin and nothing logs it — so operators cannot self-assess without being told to go and audit their directory. The change is a prefix check on a path that is off by default, so the regression risk to unaffected clusters is close to nil: a cluster not using basic auth cannot notice, and a cluster using it only notices if it was relying on the broken behaviour, which is the behaviour being removed.

The one genuine behaviour change to flag for a patch release is that a user who *is* currently getting privileges through a `system:`-prefixed group will lose them on update. That is the point of the fix, but it can look like a regression to an operator who does not know why, which is why the dropped group is logged and why the changelog impact is marked `high`.

## Checklist
- [ ] e2e tests passed.

Documentation is not affected: no user-facing parameter is added or changed, and the constraint enforced here is the one the OIDC path already documents.

## Changelog entries

```changes
section: user-authn
type: fix
summary: basic-auth-proxy no longer forwards reserved `system:` groups and usernames from the external directory to kube-apiserver.
impact: |
  Affects clusters where `publishAPI` is enabled together with `enableBasicAuth` on a
  Crowd, OIDC or LDAP `DexProvider`. Both are disabled by default, so a default
  installation is not affected.

  Previously the proxy copied directory group names into the `X-Remote-Group` header
  verbatim. Because kube-apiserver trusts that header from this proxy, a user in a
  directory group literally named `system:masters` was granted cluster-admin. Group
  names are chosen by the directory administrator, not by Kubernetes.

  After updating, any group whose name begins with `system:` is dropped before the
  request reaches kube-apiserver, and each drop is logged by the basic-auth-proxy pod
  with the user and the group name. A user authenticating with a `system:`-prefixed
  username is refused with `403 Forbidden`. This matches the `userValidationRules`
  already enforced on the OIDC/JWT path.

  Before updating, audit your directory for groups whose name begins with `system:`
  and check whether anyone currently relies on one to reach the Kubernetes API through
  basic auth. Such users will lose those privileges on update and must be granted
  access through a normal group bound by RBAC instead. Check the basic-auth-proxy logs
  after updating for `dropping group` messages.
impact_level: high
```

### DIFF
--- a/modules/150-user-authn/images/basic-auth-proxy/app/pkg/proxy/proxy.go
+++ b/modules/150-user-authn/images/basic-auth-proxy/app/pkg/proxy/proxy.go
@@ -415,6 +415,11 @@ func (h *Handler) validateCredentials(ctx context.Context, login, password strin
 
 	h.cache.SetWithTTL(key, cacheEntry{groups: groups}, h.GroupsCacheTTL)
 	h.logger.Printf("received groups for %s: %s", login, groups)
+	for _, group := range groups {
+		if isReservedIdentity(group) {
+			h.logger.Warningf("group %q of user %s will not be forwarded: the %q prefix is reserved by kube-apiserver", group, login, reservedIdentityPrefix)
+		}
+	}
 	return groups, nil
 }
 
@@ -441,12 +446,13 @@ func isReservedIdentity(name string) bool {
 	return strings.HasPrefix(name, reservedIdentityPrefix)
 }
 
-// filterReservedGroups drops directory groups that claim a reserved name.
-func (h *Handler) filterReservedGroups(login string, groups []string) []string {
+// filterReservedGroups drops directory groups that claim a reserved name. It is
+// silent by design: it runs on every proxied request, while validateCredentials
+// reports the same groups once per cache fill.
+func filterReservedGroups(groups []string) []string {
 	filtered := make([]string, 0, len(groups))
 	for _, group := range groups {
 		if isReservedIdentity(group) {
-			h.logger.Warningf("dropping group %q of user %s: the %q prefix is reserved by kube-apiserver", group, login, reservedIdentityPrefix)
 			continue
 		}
 		filtered = append(filtered, group)
@@ -463,7 +469,7 @@ func (h *Handler) modifyRequest(w http.ResponseWriter, r *http.Request, login st
 
 	stripIdentityHeaders(r.Header)
 	r.Header.Set("X-Remote-User", login)
-	for _, group := range h.filterReservedGroups(login, groups) {
+	for _, group := range filterReservedGroups(groups) {
 		r.Header.Add("X-Remote-Group", group)
 	}
 
--- a/modules/150-user-authn/images/basic-auth-proxy/app/pkg/proxy/proxy.go
+++ b/modules/150-user-authn/images/basic-auth-proxy/app/pkg/proxy/proxy.go
@@ -56,6 +56,11 @@ const (
 	// extraHeaderPrefix is the prefix kube-apiserver maps to user.Info.Extra
 	// (see --requestheader-extra-headers-prefix); must be stripped on ingress.
 	extraHeaderPrefix = "X-Remote-Extra-"
+
+	// reservedIdentityPrefix marks the identity namespace kube-apiserver keeps
+	// for itself, mirroring the userValidationRules the OIDC path enforces via
+	// AuthenticationConfiguration.
+	reservedIdentityPrefix = "system:"
 )
 
 var _ http.Handler = &Handler{}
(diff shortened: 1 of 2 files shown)


AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [21978](https://github.com/deckhouse/deckhouse/pull/21978) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #21925, opened 2026-08-06
[deckhouse] add module v2 validating webhook

Fkuloff:
## Description

Adds a validating webhook for `modules.deckhouse.io/v1alpha2` at `/validate/v1alpha2/modules`.

| Operation | ModuleConfig exists | No ModuleConfig |
|---|---|---|
| `UPDATE` | rejected | allowed |
| `DELETE` | rejected | allowed |

Creation is not restricted, so the rule does not list `CREATE`. The `deckhouse` service account is exempt — it
reconciles Modules and keeps their spec in sync with the ModuleConfig.

A failed ModuleConfig lookup is answered with an error rather than an allow, and the rule uses
`failurePolicy: Fail`, so a Module cannot be changed during a window in which the check could not run.

`matchPolicy` on the existing `modules` webhook changes from `Equivalent` to `Exact`. Both versions of the
resource now have their own webhook, and under `Equivalent` the API server also delivers a request for one
version to the other version's webhook, so both would fire on every write.

While a ModuleConfig exists it owns the module spec — its reconciler syncs it on every reconcile — so a manual
change is silently overwritten. Admission rejects the change and points at the ModuleConfig instead of leaving
the user to discover that their edit disappeared.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: add module v2 validating webhook
impact_level: low
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -672,9 +672,6 @@ manifests: controller-gen enrich-crds-local ## Generate WebhookConfiguration, Cl
 .PHONY: copy-crds
 copy-crds:
 	@echo "Copying CRDs to deckhouse-controller/crds..."
-	@cp bin/crd/bases/deckhouse.io_modules.yaml deckhouse-controller/crds/module.yaml
-	@cp bin/crd/bases/deckhouse.io_modulepackageversions.yaml deckhouse-controller/crds/modulepackageversion.yaml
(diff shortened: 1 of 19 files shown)


deckhouse-BOaTswain:
Aye, aye, @z9r5. I've started the workflow for label 'deploy/web/stage'...

PR #21948, opened 2026-08-07
[dhctl] Derive node group specs after CloudProviderVars is resolved

borg-z:
## Description

`MetaConfig.Prepare` derived `MasterNodeGroupSpec.Replicas` and `TerraNodeGroupSpecs` (`extractProviderClusterFields`, `meta.go:161`) *before* it populated `CloudProviderVars` from `ResourcesYAML` (`meta.go:165`). The derivation reads `CloudProviderVars`, so on the bootstrap-from-file path it always saw `nil` and did nothing.

The derivation is driven by `CloudProviderVars`, not by the provider cluster configuration, so this moves `applyNodeGroupReplicasFromCloudProviderVars` out of `extractProviderClusterFields` into `Prepare`, after every source of `CloudProviderVars` is resolved.

For providers whose node groups live in the resources documents rather than in a `<Provider>ClusterConfiguration` (mc-flow, e.g. DVP), both typed fields stayed zeroed on bootstrap:

* `TerraNodeGroupSpecs` empty → **no CloudPermanent node is created at all**. The bootstrap log shows `Waiting for NodeGroups [master] to become Ready` with the worker NodeGroup missing, even though the `NodeGroup` resource itself is applied to the cluster.
* `MasterNodeGroupSpec.Replicas` at 0 → the additional-master loop in `BootstrapAdditionalMasterNodes` (`steps_nodes.go:70`, `for i := 1; i < Replicas`) never iterates, so **a multi-master cluster comes up with a single master**, silently.

Converge and check were not affected: they parse from the cluster, and `base.go:303` preloads `CloudProviderVars` before `Prepare`. That is why the bug only showed on bootstrap.

It used to work by accident. `validateAndPrepareMetaConfig` re-ran `extractProviderClusterFields` at the very end of `Prepare` so that a provider *preparator* could mutate the provider cluster configuration, and that second pass happened to see a populated `CloudProviderVars`. Removing the preparator mechanism in 2647c34992 removed the second pass with it — nothing went red because no test called `Prepare()`.

Regression range: present in `main` and `release-1.77`, not in `release-1.76`.

Yes — 1.77 cannot bootstrap a DVP cluster with any CloudPermanent worker, and would bootstrap a 3-master cluster as a 1-master one.

## Checklist
- [x] e2e tests passed.

Tests are added at both levels: `Prepare` (derivation, provider-cluster-configuration precedence, idempotency on the `DeepCopy().Prepare()` re-runs done by check/converge) and `ParseConfigFromData`, which is where the `NodeGroup` documents actually reach `ResourcesYAML`. All of them fail on `main` and pass with this change.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Bootstrap creates the CloudPermanent nodes and the additional masters for providers configured through ModuleConfig.
impact_level: default
```

### DIFF
--- a/dhctl/pkg/config/base_test.go
+++ b/dhctl/pkg/config/base_test.go
@@ -1169,3 +1169,61 @@ func testCreateDeckhouseRegistrySecret(t *testing.T, kubeCl *client.KubernetesCl
 	_, err := kubeCl.CoreV1().Secrets("d8-system").Create(t.Context(), secret, metav1.CreateOptions{})
 	require.NoError(t, err)
 }
+
+// The mc-flow bootstrap path in one piece: NodeGroup documents have no schema,
+// so they land in ResourcesYAML, from where Prepare must derive the typed node
+// group fields. Without them dhctl bootstraps a single master and no workers.
+func TestParseConfigFromDataCloudDerivesNodeGroups(t *testing.T) {
+	config := `
+---
+apiVersion: deckhouse.io/v1
+kind: ClusterConfiguration
+clusterType: Cloud
+cloud:
+  provider: DVP
+  prefix: test
+kubernetesVersion: "Automatic"
+podSubnetCIDR: 10.222.0.0/16
+serviceSubnetCIDR: 10.111.0.0/16
+---
+apiVersion: deckhouse.io/v1
+kind: InitConfiguration
+deckhouse:
+   imagesRepo: test
+   devBranch: test
+   # {"auths": { "test": {}}}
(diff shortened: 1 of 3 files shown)


090809 (approved):

deckhouse-BOaTswain:
Cherry pick PR [21951](https://github.com/deckhouse/deckhouse/pull/21951) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #21976, opened 2026-08-09
candi: fix duplicated TOML table in rendered containerd config.toml

### LINKED ISSUE #13934: [deckhouse-controller] config.toml is incorrectly generated in containerd after multiple registry changes
### Preflight Checklist

### Version

1.69.11

### Expected Behavior

In a cluster installed in a closed contour that has the following InitConfiguration
```
apiVersion: deckhouse.io/v1
kind: InitConfiguration
deckhouse:
  imagesRepo: harbor.89.169.144.131.nip.io/d8s/deckhouse
  registryDockerCfg: ewoJImF1dGhzIjogewoJCSJoYXJib3IuODkuMTY5LjE0NC4xMzEubmlwLmlvIjogewoJCQkiYXV0aCI6ICJZV1J0YVc0NlNHRnlZbTl5TVRJek5EVT0iCgkJfQoJfQp9
  registryCA: |
     -----BEGIN CERTIFICATE-----
                <body if cert>
     -----END CERTIFICATE-----
```
when the registry is changed sequentially, at first without specifying the '--ca-file` option's.
```
kubectl -n d8-system exec -ti svc/deckhouse-leader -c deckhouse -- deckhouse-controller helper change-registry  harbor.89.169.144.131.nip.io/d8s/deckhouse
```
and after specifying the '--ca-file` flag,
```
CA_CONTENT=$(cat /home/skala/nexus.crt)
kubectl -n d8-system exec svc/deckhouse-leader -c deckhouse -- bash -c "echo '$CA_CONTENT' > /tmp/ca.crt && deckhouse-controller helper change-registry --ca-file /tmp/ca.crt harbor.89.169.144.131.nip.io/d8s/deckhouse"
```
 the registry for the cluster changes without errors.

### Actual Behavior

When changing the registry sequentially, first without specifying the '--ca-file` option, and then with the specifying '--ca-file`, an error occurs in the file  `/etc/containerd/config.toml`.
The following messages are observed in the bashible logs
```
Jun 10 13:10:36 galin-test-master-0 bashible.sh[303251]: bash [ERROR] 'containerd config dump' return error: containerd: failed to load TOML: /etc/containerd/config.toml: (108, 10): duplicated tables
```
Two parameters were found in the file `/etc/containerd/config.toml`
```
[plugins."io.containerd.grpc.v1.cri".registry.configs."harbor.89.169.144.131.nip.io".tls]
          ca_file = "/opt/deckhouse/share/ca-certificates/registry-ca.crt"
[plugins."io.containerd.grpc.v1.cri".registry.configs."harbor.89.169.144.131.nip.io".tls]
          insecure_skip_verify = true
```
Due to these duplicate parameters, containerd cannot restart when the registry is changed.

### Steps To Reproduce

1. Install deckhouse kubernetes cluster in close enviroment
2. Change registry without option`--ca-file`
3. Change registry with option `--ca-file`
4. Check containerd's work and bashible's logs.

### Additional Information

_No response_

### Logs

```shell

```

pujitha24:
## Description
This change fixes bashible's containerd config template (`032_configure_containerd.sh.tpl`, used for the "Containerd" v1 CRI) so it no longer emits the `[plugins."io.containerd.grpc.v1.cri".registry.configs."<host>".tls]` TOML table twice for the same mirror host. Previously, when a mirror had both a CA file and an `http` scheme configured at the same time, the template opened that table once to set `ca_file` and again to set `insecure_skip_verify`, producing invalid TOML (duplicate table definitions) that containerd refused to load. Both conditional blocks are now merged into a single `[...tls]` table, with `ca_file` and `insecure_skip_verify` each still independently conditional on the underlying mirror fields, so at most one table header is emitted per host. The `ContainerdV2` branch of the template is unaffected, since it always uses `config_path` and never generates per-mirror `[...configs."<host>".tls]` blocks. This does not affect control-plane, ingress-controller, or Prometheus restarts.

This is reachable via `deckhouse-controller helper change-registry`: running it first without `--ca-file` and then again with `--ca-file` for the same host can leave a mirror record with the `http` scheme from the first run and the CA from the second run, producing exactly this invalid config. As a result, bashible fails with the error `containerd: failed to load TOML: ... duplicated tables`, breaking containerd config regeneration on affected nodes.

Fixes https://github.com/deckhouse/deckhouse/issues/13934

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: Fix duplicated TOML table in rendered containerd config.toml when a registry mirror has both a CA file and an "http" scheme configured.
impact: Nodes whose registry mirror config had both a CA file and an "http" scheme set (e.g. after running `deckhouse-controller helper change-registry` twice with different flags for the same host) could fail to regenerate containerd's config.toml with a "duplicated tables" TOML error. After this fix, the generated config is valid and containerd loads it correctly.
impact_level: high
```

---
AI assistance: this change was drafted with Claude Code.

Fixes #13934

### DIFF
--- a/candi/bashible/common-steps/all/032_configure_containerd.sh.tpl
+++ b/candi/bashible/common-steps/all/032_configure_containerd.sh.tpl
@@ -390,13 +390,14 @@ oom_score = 0
           {{- else }}
           auth = {{ (($mirror).auth).auth | default "" | quote }}
           {{- end }}
-      {{- if $mirror.ca }}
+      {{- if or $mirror.ca (eq $mirror.scheme "http") }}
         [plugins."io.containerd.grpc.v1.cri".registry.configs."{{ $mirror.host }}".tls]
+      {{- if $mirror.ca }}
           ca_file = "/opt/deckhouse/share/ca-certificates/registry-{{ $mirror.host | lower }}-ca.crt"
       {{- end }}
       {{- if eq $mirror.scheme "http" }}
-        [plugins."io.containerd.grpc.v1.cri".registry.configs."{{ $mirror.host }}".tls]
           insecure_skip_verify = true
+      {{- end }}
       {{- end }}
     {{- end }}
   {{- end }}
--- a/dhctl/pkg/template/bashible_test.go
+++ b/dhctl/pkg/template/bashible_test.go
@@ -17,8 +17,10 @@ package template
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
+	"github.com/BurntSushi/toml"
 	"github.com/stretchr/testify/require"
 
 	"github.com/deckhouse/deckhouse/dhctl/pkg/app/options"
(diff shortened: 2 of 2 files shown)


pujitha24:
When someone has a moment, could a maintainer add `status/ok-to-test`? CI is currently skipping all jobs on this PR pending that label — happy to fix anything it turns up.

PR #21977, opened 2026-08-09
[cloud-provider-zvirt] Migrate legacy IsDefault key in discovery data before validation

### LINKED ISSUE #14919: [cloud-provider-zvirt] storageDomains. IsDefault is a forbidden property
### Preflight Checklist

### Version

v1.70.16

### Expected Behavior

Migration IsDefault-> isDefault
IsDefault doesn't break update.

### Actual Behavior

After update from 1.69 to 1.70.16 faced with issue, that queue stacks With error
```
1 error occurred:
storageDomains.IsDefault is a forbidden property
```

### Steps To Reproduce

_No response_

### Additional Information

[PR](https://github.com/deckhouse/deckhouse/pull/13386/files)

### Logs

```shell

```

pujitha24:
## Description
Discovery data secrets persisted before [#13386](https://github.com/deckhouse/deckhouse/pull/13386) used a mistagged `IsDefault` (capitalized) JSON key for storage domains, while the module's `cloud_discovery_data.yaml` schema has always required `isDefault` (lowercase). Since the schema's `storageDomains` item objects don't declare `additionalProperties` explicitly, lib-dhctl's validation loader defaults it to `false`, so the stale key is hard-rejected as an unknown property with `storageDomains.IsDefault is a forbidden property`.

This change normalizes the legacy `IsDefault` key to `isDefault` in the raw secret JSON before schema validation runs, so clusters carrying pre-existing discovery data can validate successfully again and the discoverer Pod gets a chance to self-heal the Secret going forward.

This does not affect critical cluster components such as ingress-controllers, control-plane, or Prometheus.

Because the discover hook runs `OnBeforeHelm`, the validation failure blocks the whole `cloud-provider-zvirt` module's Helm sync on every reconcile. This also prevents the `cloud-data-discoverer` Pod (the component that would otherwise rewrite the Secret with the corrected field) from being redeployed, leaving the `deckhouse-controller` queue permanently stuck after upgrading on any zvirt cluster whose discovery data predates the fix in #13386.

Fixes https://github.com/deckhouse/deckhouse/issues/14919

This fix unblocks the `deckhouse-controller` queue and restores Helm sync for the `cloud-provider-zvirt` module on any affected cluster, so it should be backported to the patch release to avoid clusters remaining stuck until the next minor release.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cloud-provider-zvirt
type: fix
summary: Fix a stuck deckhouse-controller queue caused by legacy `IsDefault` keys in zvirt discovery data failing schema validation.
impact: On zvirt clusters with discovery data persisted before the isDefault key fix in #13386, the cloud-provider-zvirt module's Helm sync was blocked on every reconcile due to a schema validation error, preventing the cloud-data-discoverer Pod from being redeployed. After this update, the legacy key is migrated automatically before validation, unblocking the queue and allowing the discoverer Pod to self-heal the Secret.
impact_level: high
```

---
AI assistance: this change was drafted with Claude Code.

Fixes #14919

### DIFF
--- a/ee/se-plus/modules/030-cloud-provider-zvirt/hooks/discover.go
+++ b/ee/se-plus/modules/030-cloud-provider-zvirt/hooks/discover.go
@@ -122,7 +122,7 @@ func handleCloudProviderDiscoveryDataSecret(_ context.Context, input *go_hook.Ho
 
 	secret := cloudSecrets[0]
 
-	discoveryDataJSON := secret.Data["discovery-data.json"]
+	discoveryDataJSON := migrateLegacyIsDefaultField(secret.Data["discovery-data.json"])
 
 	if err := validation.ValidateData([]string{"/deckhouse/ee/se-plus/modules/030-cloud-provider-zvirt/candi/openapi", "/deckhouse/candi/cloud-providers/zvirt/openapi"}, &discoveryDataJSON); err != nil {
 		return fmt.Errorf("failed to validate 'discovery-data.json' from 'd8-cloud-provider-discovery-data' secret: %v", err)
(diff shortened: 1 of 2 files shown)


pujitha24:
Could a maintainer add `status/ok-to-test` when convenient? All checks are waiting on that label — I'll address whatever CI finds.

PR #22037, opened 2026-08-11
[deckhouse] Serve validating webhooks on the controller-runtime webhook server

Fkuloff:
## Description

The nine validating admission handlers move from addon-operator's admission server to the webhook server built into the controller-runtime manager. Addon-operator's server is kept down so the port is bound once.

Unchanged: listen port, certificates directory, `tls.crt`/`tls.key` file names, all nine webhook paths. No chart or manifest changes.

The endpoint comes up earlier than before: the manager starts webhooks ahead of the caches, while the old server started only after the module loader had finished. With admission disabled (dhctl bootstrap, no certificates mounted) no webhook server is created.

The manager already runs in this binary and owns a webhook server, so a second HTTPS server for the same handlers is redundant. It also opens the controller-runtime admission machinery — typed handlers, decoders, per-kind builders — for the modulev2 Module and ModulePackage webhooks.

### Manual testing

Checked on a dev cluster.

```
# a single listener on the admission port: the manager's webhook server, no second one from addon-operator
ss -lntp | grep :4223
LISTEN 0      1000               *:4223             *:*    users:(("deckhouse-contr",pid=220569,fd=10))

# every route answers, 400 means the route matched and the empty body was rejected
for p in /validate/v1/deckhouse-registry-secret /validate/v1alpha1/module-configs /validate/v1alpha1/modules /validate/v1/configuration-secret /validate/v1/provider-configuration-secret /validate/v1/static-configuration-secret /validate/v1alpha1/update-policies /validate/v1alpha1/deckhouse-releases /validate/v1alpha1/applications; do printf '%s  %s\n' "$(curl -sk -o /dev/null -w '%{http_code}' -X POST -H 'Content-Type: application/json' -d '{}' https://127.0.0.1:4223$p)" "$p"; done
400  /validate/v1/deckhouse-registry-secret
400  /validate/v1alpha1/module-configs
400  /validate/v1alpha1/modules
400  /validate/v1/configuration-secret
400  /validate/v1/provider-configuration-secret
400  /validate/v1/static-configuration-secret
400  /validate/v1alpha1/update-policies
400  /validate/v1alpha1/deckhouse-releases
400  /validate/v1alpha1/applications

# control: a path that is not registered
curl -sk -o /dev/null -w '%{http_code}\n' -X POST -d '{}' https://127.0.0.1:4223/validate/nope
404

# the webhook validates end to end
k apply -f - <<'YAML'
kind: ModuleConfig
metadata:
  name: no-such-module
spec:
  enabled: true
YAML
The request is invalid: : the 'no-such-module' module not found
```

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: chore
summary: Serve validating webhooks on the controller-runtime webhook server
impact_level: low
```

### DIFF
--- a/deckhouse-controller/pkg/controller/controller.go
+++ b/deckhouse-controller/pkg/controller/controller.go
@@ -219,18 +219,16 @@ func NewDeckhouseController(
 		opts.Cache.ByObject[&v1alpha2.Module{}] = cache.ByObject{}
 	}
 
-	// The manager's webhook server takes over addon-operator's admission server and
-	// reuses its settings. Skipped where no certificates are mounted (dhctl bootstrap).
-	admissionEnabled := app.AdmissionServerEnabled()
-	if admissionEnabled {
-		listenPort, err := strconv.Atoi(app.AdmissionServerListenPort())
+	admission, serveWebhooks := app.TakeOverAdmissionServer()
+	if serveWebhooks {
+		listenPort, err := strconv.Atoi(admission.ListenPort)
 		if err != nil {
 			return nil, fmt.Errorf("parse admission server listen port: %w", err)
 		}
 
 		opts.WebhookServer = webhook.NewServer(webhook.Options{
 			Port:    listenPort,
-			CertDir: app.AdmissionServerCertsDir(),
+			CertDir: admission.CertsDir,
 		})
 	}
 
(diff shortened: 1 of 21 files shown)


deckhouse-controller/pkg/controller/controller.go
```diff
@@ -217,6 +219,21 @@ func NewDeckhouseController(
 		opts.Cache.ByObject[&v1alpha2.Module{}] = cache.ByObject{}
 	}
 
+	// The manager's webhook server takes over addon-operator's admission server and
+	// reuses its settings. Skipped where no certificates are mounted (dhctl bootstrap).
+	admission := operator.Config().Admission
```
ipaqsa: Just use adapp consts

ipaqsa (approved):

ipaqsa (approved):

ldmonster (approved):

PR #22043, opened 2026-08-11
[cni-cilium] Document AWS internal load balancer being unreachable from inside the cluster

tarabrind:
## Description

Add a troubleshooting section to the `cni-cilium` module documentation about an operational specific on AWS: traffic from inside the cluster may fail to reach an internal load balancer, while the same load balancer stays reachable from outside.

The new section explains the exact conditions that trigger the behavior, describes the cause (the `preserve_client_ip` option combined with return traffic routed through the VXLAN tunnel), and gives two ways to fix it: through the AWS web console and through the AWS CLI. It also warns that disabling `preserve_client_ip` makes the load balancer perform SNAT, so the application no longer sees the real client IP at the network level, though for HTTP/HTTPS the address is still passed in the `X-Forwarded-For` header.

Changes cover both the English and Russian versions and add a screenshot of the AWS console setting. Documentation only, no code changes and no component restarts.

Users who run Cilium on AWS in VXLAN mode with an internal load balancer hit a confusing situation: pods and nodes cannot reach the load balancer IP, but external clients can. Without guidance, this looks like a Cilium or AWS bug and is hard to diagnose.

The section documents the root cause and the fix, so users can resolve the issue themselves instead of spending time on debugging or opening a support request.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: cni-cilium
type: chore
summary: Documented the AWS internal load balancer being unreachable from inside the cluster and how to fix it.
impact_level: low
```


AndreyPavlovFlant (approved):

valeria-brazhnikova (approved):

z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [22265](https://github.com/deckhouse/deckhouse/pull/22265) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #21990, opened 2026-08-10
[registrypackages] Build olcedar sysext images only for non-CSE editions

borg-z:
## Description

Wrap the olcedar sysext image definitions (`containerd`, `kubelet`, `kubernetes-cni`) in `{{- if ne $.Env "CSE" }}`, so they are only built for non-CSE editions. Drops the now-unreachable CSE-only `pidumount` import from the containerd sysext.

`WERF_ENV=CSE werf config list` renders no `*-sysext-*` images; EE still renders all of them.

## Why do we need it, and what problem does it solve?

The immutable (olcedar) OS is not shipped in CSE, so the sysext images are dead weight there: extra build time and an extra attack surface in an edition that has to be certified.

## Why do we need it in the patch release (if we do)?

Not necessarily.

## Checklist
- [ ] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [ ] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: registrypackages
type: chore
summary: Build olcedar sysext images only for non-CSE editions.
impact_level: low
```

PR #21557, opened 2026-07-24
[node-manager] Immutable OS (olcedar) node support

borg-z:
## Description

Immutable (olcedar) nodes are reconciled by an on-node agent (nodelet) from a `NodeConfig`
object instead of by bashible, and the first master of such a cluster cannot be provisioned
over SSH at all. This PR makes such nodes creatable, bootstrappable and operable day-2.

**node-manager**

- `systemType: Mutable | Immutable` on NodeGroup, immutable after create (CEL + webhook).
- New APIs and controllers: `NodeConfig`, `NodeBootstrapConfig`, `NodeOperation`,
  `NodeExtensionRequest`. `NodeConfig.status` reports per-extension and per-unit outcomes
  instead of aggregate conditions.
- Day-2 reconciliation: the nodeconfig controller renders per-node desired state (system
  extensions pinned by digest, kubelet settings), rolls it out group by group, and requests
  disruption approvals as `NodeOperation`; draining goes through the existing controller.
- CAPI bootstrap provider: for an immutable NodeGroup the controller renders a
  `NodeBootstrapConfigTemplate` per group and points the MachineDeployment at it. No
  group-wide bootstrap Secret, and node-controller names no infrastructure kind.
- The agent itself ships as a signed sysext delivered by digest, so it updates without a
  reboot; the OS image carries a seed copy under `/usr/share/nodelet`.

**dhctl**

- The new `pkg/immutable` package builds the cloud-init payload that boots the first
  control-plane node: a `NodeConfig` plus control-plane manifests. The node runs no sshd,
  no bash and no bashible — it generates the cluster PKI itself. No cluster key travels in
  the payload.
- dhctl then collects the admin kubeconfig over a one-shot TLS handoff, retargets it at a
  reachable address and finishes the install over the Kubernetes API. An SSH bastion is
  supported through a local tunnel; `--kubeconfig-out` hands the kubeconfig to automation.
- Additional masters join through a per-node payload instead of the group-wide Secret.
- `dhctl destroy` works from a kubeconfig alone, with no SSH.

## Why do we need it, and what problem does it solve?

Without it an immutable cluster cannot be installed at all: there is no bashible on the
node and no sshd for the installer to reach.

## Why do we need it in the patch release (if we do)?

Not necessarily.

## ⚠️ Before merging

Two temporary pins must be replaced, **agent first, image second** — the image merges
sysexts before the agent starts, so an image arriving first would launch a build the agent
would have refused:

[...]

PR #22101, opened 2026-08-12
[node-manager] Build nodelet as a signed system extension

borg-z:
## Description

Builds the Olcedar node agent (`nodelet`) as a signed system extension inside Deckhouse and delivers its digest through `NodeConfig`, alongside `containerd`, `kubelet` and `kubernetes-cni`.

- new `registrypackages/nodelet-sysext` image: the agent source is cloned from the private repository over SSH with the credentials the build job already carries (`DECKHOUSE_PRIVATE_REPO` + `SOURCE_REPO_SSH_KEY`), built, wrapped into erofs + dm-verity and signed with the existing Olcedar signer. No CI change needed. The agent has no releases yet, so it is pinned by commit;
- the image ships the agent binary **and** its unit: the agent restarts only the units it discovers inside an extension image (`systemd-dissect --list`), so an image without one would be merged and never take effect;
- `node-controller` reads `nodeletSysext` from `images_digests.json` and renders it into `spec.extensions`; the name is reserved against `NodeExtensionRequest`;
- **dhctl renders the same four extensions.** A node prunes every installed extension its document does not list, and pruning `nodelet` runs `systemctl stop nodelet.service` — the agent stops itself, and an explicitly stopped unit is not restarted, so the node loses the API proxy kubelet talks through and never gets it back. Before the API is reachable the node reads dhctl's document, so a three-extension document from the installer and a four-extension one from the cluster is enough to kill every node on its next boot;
- it also watches that ConfigMap. It was read on every pass but nothing enqueued one, and no resync period is set, so a release that changed only a digest reached the fleet whenever some unrelated object happened to move;
- **`status.osImage` is declared too.** The agent publishes how far a rootfs update has got; a status field the CRD does not declare is not merely pruned — the whole status apply is refused with `field not declared in schema`, so the node then reports no conditions, no extensions and no units at all. `spec.osImage` also joins the spec's `required` list, which the Go type already demanded;
- **`spec.osImage` becomes an object pinned by digest** instead of a string naming a tag. The merged agent and initramfs both require that shape, and a scalar fails their parse and drops the node into an emergency shell. The digest comes from `nodeManager.olcedar`, which #22059 (cherry-picked here) puts into the release;
- also cherry-picks the CSE gate for the three existing sysext images (#21990), which this base branch predates — without it a CSE render would see three digests out of four.

Affects Olcedar nodes only. No restarts of ingress, control-plane or Prometheus.

**Rollout gate:** every Olcedar node must already run an agent that handles its own restart. A node running an older agent restarts, and on prune stops, its own unit.

The agent changes more often than the OS it runs on. Baked into the read-only root, it could only be updated by rebuilding the image, writing it into the second boot slot and rebooting the node. As an extension it updates the same way every other node component does — by digest — and the node keeps running.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

Verified on a live cluster (1 master + 2 workers): all four sidecars including `nodelet.roothash.p7s` land in `/var/lib/extensions`, the `/usr` overlay puts the extension's copy above the image's, and the agent process starts 294 s after boot on a node that never rebooted — that is the clean-exit self-restart.

The extension-set mismatch above was found the same way, on `zykov-ab-19be2a59`: the node went NotReady with `dial tcp 127.0.0.1:6445: connect: connection refused` while `nodelet.service` was still installed, and the serial console named the cause — `stop nodelet.service before pruning extension nodelet: signal: killed`.

Re-verified after the fix on `zykov-ab-57a656c4` (1 master + 2 workers, all Ready): no prune, the agent survives its own extension being installed, and the node reports all four extensions `Ready` including `nodelet`. Changing the digest in `images_digests.json` reaches the rendered `NodeConfig` in about ten seconds with nothing else touched.

Two defects outside this diff surfaced on that cluster and are **not** fixed here:

- the rootfs update cannot stage. `extractTar` (nodelet `internal/registry/client.go`) treats the largest tar member as the image, and the `nodeManager.olcedar` artifact ships `efi.img` (64 MiB) beside `rootfs.erofs` (28 MiB), so the sidecars are looked for under the `efi` stem and the slot ends up empty (`rootfs-b.erofs is empty`). Fixable from either side: select the member by name, or package only the four rootfs files;
- nodelet's `pruneExtension` stops an extension's units without checking whether one of them is the agent's own. The matching digest sets above make that unreachable in practice, but a node that falls back to a stale on-disk config still self-stops; the guard exists on the `nodelet-self-update` branch (`297a454`) and is not in `main`.

## Changelog entries

```changes
section: node-manager
type: feature
summary: The Olcedar node agent is delivered as a signed system extension and updates without rebuilding the OS image or rebooting the node.
impact_level: low
```


borg-z:
Сведено в #21557: доставка агента подписанным расширением перенесена туда целиком (14 коммитов), плюс гейт выката. Отдельный PR больше не нужен.

PR #22103, opened 2026-08-13
[node-manager] Rework derived_status: single I/O boundary, typed sources, no swallowed read failures

borg-z:
## Description

Rewrites node-controller's `derived_status` — the code that resolves a NodeGroup into the element published to the bashible context and to the provider machine templates — around one I/O boundary and a typed spec, and fixes three bugs that made an unreadable source look like an absent one.

### What was done

- `snapshot.go` is now the only file that talks to the API, and it reads every source exactly once.
- `derive.go` and `validate.go` became pure functions over that snapshot, guarded by `purity_test.go`.
- The cloud-provider registration Secret decodes into a typed `CloudProviderRegistration` instead of a `map[string]interface{}`.
- `InstanceClass` and `NodeCapacity` are typed, which removes nine blind type assertions and a `RawExtension` round-trip from the callers.
- `Service.Reader`, the last uncached live GET, is gone.
- `ResolveNodeGroup` no longer takes the NodeGroup twice: the published passthrough is built from the typed spec's own JSON form, so `rawSpec` disappears.
- Integers in that passthrough are decoded with `UseNumber` and narrowed back to `int64`, which is what the API server's unstructured decode produces.
- The resolved InstanceClass is round-tripped back to `float64` on purpose, because that rendering feeds the checksum naming an immutable machine template.
- `unstructured` leaves the NodeGroup path entirely, so both controllers now do a plain typed `Get`/`List`.
- `v1.NodeGroupSpec` gained `fencing`, `mig.customConfigs` and `topologyManager.enabled`, and four fields were renamed to the spelling the CRD actually uses.
- A new drift guard (`internal/schema`) walks the shipped CRD and fails when a spec path has no home in the Go type, and it also runs against a live cluster's CRD via `LIVE_CRD=<dump>`.

### Bugs fixed

- An unreadable source was encoded as a valid-looking empty value, so the checks returned `Processed=false` with an empty `Error` and the bashible context published the NodeGroup without its `instanceClass` — a configuration-checksum shift on every node.
- The cgroup-v2 guard for `LimitedSwap` never fired, because the CRD calls the field `swapBehavior` while the Go type called it `behavior`.
- An InstanceClass deleted between the List and the Get of one pass was published as `null`, because both checks saw the stale name and passed.

The package was a verbatim port of the `get_crds` hook: readers interleaved with logic, every source read twice, and errors encoded as valid-looking values. Each swallowed error publishes a smaller element than the truth, and that element is hashed into every node's `configuration-checksum`, so "the API blipped" became "re-run bashible on every node".

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Verification

The published element must not move: it is hashed into every node's configuration checksum and rendered into the name of an immutable machine template.

- All unit tests and every envtest suite pass.
- The golden corpus, a shape-invariant property test and the provider checksum goldens pass.
- A running cluster's `checksum/instance-class` was reproduced byte for byte, and that value names an immutable template.
- Every NodeGroup of that cluster was verified against the published bashible context: key set, passthrough subtrees, serialized labels and taints.
- The drift guard passes against that cluster's live CRD.
- A cluster was bootstrapped from this branch and exercised end to end: NodeGroups resolve, the cloud overlay is published for CloudEphemeral and withheld elsewhere, a node came up through MachineDeployment → Machine → Node, a live `swapBehavior` edit reached the published element, every node's `configuration-checksum` annotation equals the published value, the context Secret is not rewritten between passes, and node-controller logged no errors with zero restarts.

Golden fixtures changed in two ways, both fixture artifacts rather than output changes:

- `nodeCapacity` gained `name` and `rootDisk`, because `capacity.InstanceType` has no `omitempty` tags and the previous code marshalled the whole struct, so a real cluster always saw all four keys.
- Four fixtures described NodeGroups the API server rejects: `cloudInstances` without its required `minPerZone`/`maxPerZone`/`classReference`, plus a fractional `nodeDrainTimeoutSecond` and `overprovisioningRate` where the CRD declares `type: integer`.

## Changelog entries

```changes
section: node-manager
type: fix
summary: node-controller no longer treats an unreadable source as an absent one, which could publish a NodeGroup without its instanceClass and re-run bashible on every node
---
section: node-manager
type: fix
summary: the validating webhook now enforces that kubelet.memorySwap.swapBehavior LimitedSwap requires cgroup v2, a check that never fired because the field was misspelled in the Go type
impact_level: high
impact: |
  A NodeGroup created while the check was dead — `swapBehavior: LimitedSwap` with nodes that do not
  support cgroup v2 — is now rejected on every update, including the finalizer removal that a
  delete needs, so such a NodeGroup can hang in Terminating. Removing `swapBehavior: LimitedSwap`
  from its spec passes and unblocks it.
```


deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:nm-derived-status-rewrite` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/31677053815) in 34m46s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.2.200`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 33m26s.

</details>

deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:nm-derived-status-rewrite` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/31677060313) in 25m5s.



GCP-WithoutNAT-Containerd-1.34 - Connection string: `ssh  user@136.92.14.186`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.34` succeeded in 23m59s.

</details>

deckhouse-BOaTswain:
:red_circle: `e2e: Yandex.Cloud` for `deckhouse:nm-derived-status-rewrite` [failed](https://github.com/deckhouse/deckhouse/actions/runs/31677061911) in 42m6s.




  `ssh -i ~/.ssh/e2e-id-rsa  cloud-user@111.88.250.214
` - connect for debugging;

  `/e2e/abort yandex-cloud;WithoutNAT;containerd;1.34 31677061911 31677061911-1-con-1-34 /sys/deckhouse-oss/install:pr22103 cloud-user@111.88.250.214
` - for abort failed cluster




Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@111.88.250.214`


:red_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` failed in 40m43s.

</details>

090809 (approved):

deckhouse-BOaTswain:
:green_circle: `e2e: Static` for `deckhouse:nm-derived-status-rewrite` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/31691506154) in 39m39s.



Static-Static-Containerd-1.34 - Connection string: `ssh -J astra@139.100.218.12 astra@192.168.199.108`


:green_circle: `e2e: Static, Containerd, Kubernetes 1.34` succeeded in 38m15s.

</details>

PR #22106, opened 2026-08-13
[deckhouse-controller] add x-deckhouse-ui-resource-name type to package settings schemas

rilonx:
## Description

Type the `x-deckhouse-ui-resource-name` OpenAPI extension in the Application package settings schema so the controller preserves it in `ApplicationPackageVersion.status.packageSchemas` instead of dropping it as an unknown `x-*` key on unmarshal.

The extension binds a `type: string` settings field to a live selection of cluster resources of a given kind (`apiVersion` + `kind`), optionally narrowed by a standard Kubernetes `labelSelector`. The web console renders a dropdown of matching resource names from the application's namespace instead of a free-text input.

New type: `UIResourceNameSelector` (`apiVersion`, `kind`, optional `labelSelector *metav1.LabelSelector`); deepcopy regenerated with `make controller-gen-generate`. No effect on critical cluster components. The CRD is unchanged — `settingsSchema` stays `Schemaless` + `x-kubernetes-preserve-unknown-fields`, so the new type does not appear in any CRD (verified by generation). The field is `omitempty` and present only when a package uses the extension, so existing clusters are unaffected.

Untyped `x-*` keys are stripped when the controller unmarshals `settings.yaml` into the typed `OpenAPIV3Schema`, so `x-deckhouse-ui-resource-name` never reaches the schema stored in the package version status. Typing it makes the extension available to Application package developers: a settings field can offer a selection of existing cluster resources (for example, pick a `Secret` for a TLS certificate) instead of a free-text input, optionally filtered by labels. Same mechanism as `x-deckhouse-grantable-resource` and the other typed `x-deckhouse-ui-*` extensions.

Not necessarily.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```yaml
section: deckhouse-controller
type: feature
summary: Application packages can bind a settings field to a live selection of cluster resources with the x-deckhouse-ui-resource-name schema extension.
impact_level: low
```


z9r5 (approved):

ldmonster (approved):

deckhouse-BOaTswain:
Cherry pick PR [22268](https://github.com/deckhouse/deckhouse/pull/22268) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22117, opened 2026-08-13
[test] Broken nodelet extension for unhappy-path testing

borg-z:
Throwaway branch: builds the agent extension from a deliberately broken agent (nodelet `bbbf3053`, tag `test-broken-1`) so the refusal paths can be exercised on a live cluster. The build is the only deliverable — the digest, not the code.

Do not merge. Close once the digest is recorded.


borg-z:
Тестовый артефакт для испытаний immutable-узлов. Прогоны завершены, итоги в #21557 — PR больше не нужен.

PR #22126, opened 2026-08-13
[docs] Add docs for integrity control in deckhouse

Nikita-Voskresenskiy:
## Description

Rewrites the "Integrity control" architecture page (RU + EN) so that it describes what DKP actually does, split by mechanism and edition, and fixes the same incorrect claims duplicated in the installation guide.

**`architecture/security/integrity-control.html`** is restructured into three mechanisms, as agreed in the thread:

1. **Integrity control of images at the CRI level** — new section stating explicitly that signature verification and DM-Verity protection of unpacked layers exist **only in CSE Lite / CSE Pro**, and that the signing key belongs to the vendor so a custom key cannot be added. For other editions it explains what containerd v2 + the EROFS snapshotter do and do not give.
2. **Integrity control of user workloads at startup** — previously **not documented at all**. Describes the Cosign + SecurityPolicy `verifyImageSignatures` + Gatekeeper/Ratify flow with the SE+/EE/CSE edition marker, and links to the existing configuration guide.
3. **Integrity control of running containers** — unchanged (runtime audit, available everywhere).

The block about **module integrity** is moved out of the intro alert into its own section, renamed to "Integrity **protection** of DKP modules", and corrected against the implementation (details below).

Removed from the page: the claim that DKP verifies application container integrity at the CRI level by comparing SHA-256 hashes and that "a container can only be started if the checksum verification succeeds". This described default containerd behaviour, not a DKP integrity mechanism.

Docs only, no code changes.

Four separate problems were confirmed:

1. **CRI-level integrity control was described as a general DKP feature.** It only exists in CSE. In CE/BE/SE/SE+/EE there is no image signature verification at the CRI level at all.
2. **The mechanism that does exist outside CSE was missing from the docs.** Ratify + `admission-policy-engine` provides signature verification for user workloads at the admission stage, and the page never mentioned it — which is what left the customer thinking nothing was configurable.
3. **The module integrity section promised an alert that does not exist.** Verified: no Prometheus rule and no metric anywhere in `modules/` or `ee/` covers this; the fallback to the symlink installer at [`installer.go:65`](deckhouse-controller/internal/module/installer/installer.go#L65) is completely silent. The only `erofs`-related alert is `D8NodeContainerdV2NotSupported`, which is about the containerd v2 snapshotter — a different feature.
4. **The `erofs` prerequisite was described inaccurately.** `verity.IsSupported()` reads `/proc/filesystems`, so the module must be *loaded or built in*, not merely available — and DKP runs `modprobe erofs` and persists it **only** under `{{- if eq .cri "ContainerdV2" }}` ([`005_load_mandatory_modules.sh.tpl:23-29`](candi/bashible/common-steps/all/005_load_mandatory_modules.sh.tpl#L23), the only such place in the repo). On clusters with containerd v1 on master nodes, module integrity protection therefore silently does not engage. The docs also said "on the nodes where DKP is installed", while the check happens in the Deckhouse controller pod, i.e. on master nodes by default.

The module section additionally now states that there is **no cryptographic signature verification for modules** — the only trust anchor is an unsigned manifest annotation, so the mechanism protects against post-installation modification, not against a compromised registry — and that it works in all editions.

Not necessarily, though the page is currently misleading for every non-CSE edition, so backporting to supported patch releases would be reasonable.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: docs
type: fix
summary: Clarify which integrity control mechanisms are available in which DKP editions and correct the DKP module integrity description.
impact_level: low
```


sprait (approved):

ipaqsa (approved):

Suselz (approved):

voitenkov (approved):

denmaloyreb (approved):

Zhbert (approved):

z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [22179](https://github.com/deckhouse/deckhouse/pull/22179) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Backport failed ⚠️.

z9r5:
Cherry-picked into release-1.76, release-1.75, release-1.74

PR #22130, opened 2026-08-13
[test] Broken OS image for unhappy-path testing

borg-z:
Throwaway branch: republishes the deliberately broken Olcedar image (`v0.1.3-broken`, an agent that never converges) into the release so a node can pull it by digest. The digest is the only deliverable.

Do not merge.


borg-z:
Тестовый артефакт для испытаний immutable-узлов. Прогоны завершены, итоги в #21557 — PR больше не нужен.

PR #22142, opened 2026-08-13
Reduce the in-core GPU node preparation to a non-blocking minimum

serg-su:
## Description

Reduces the in-tree GPU node preparation (`NodeGroup.spec.gpu` without the `gpu` module) to
a non-blocking minimum, and adds a deprecation alert for it. Six files, +593/−54.

`candi/bashible/common-steps/all/006_check_gpu.sh.tpl`:

- every failure path now exits 0 and only reports through the log. The step used to `exit 1`
  on five branches whose prerequisites are external to Deckhouse — the NVIDIA driver, the
  container toolkit, device readiness, DKMS after a kernel change;
- every `nvidia-smi` call is bounded with `timeout -k`. Plain `timeout` only sends SIGTERM
  and then keeps waiting, which a process wedged in a driver ioctl ignores, so the call was
  not actually bounded;
- the compute capability comparison no longer uses `bc` (pure-shell integer comparison), the
  validation pattern accepts both two- and three-component driver versions, and
  "cannot parse the value" and "value is below the minimum" are now distinct messages;
- a PCI gate keeps the step silent on a node without an NVIDIA display controller.

`candi/bashible/common-steps/all/032_configure_containerd.sh.tpl`:

- the containerd default runtime is resolved at run time and falls back to `runc` while
  `/usr/bin/nvidia-container-runtime` is absent;
- `net.core.bpf_jit_harden=1` is persisted in an own `/etc/sysctl.d` drop-in instead of
  editing distribution-owned files that may not exist; `/etc/sysctl.conf` is still fixed up
  conditionally, because procps applies it after all drop-in directories and RED OS ships the
  hardened value there.

`modules/040-node-manager`:

- new hook exporting `d8_node_group_gpu_in_core_deprecated` and the
  `NodeGroupUsesDeprecatedInCoreGPU` alert for NodeGroups that still use `spec.gpu` while the
  `gpu` module is disabled. The hook is bound to `OnAfterHelm` in addition to NodeGroup
  events, because enabling the module only changes `global.enabledModules` and would
  otherwise never expire the metric.

### Influence on critical cluster components

- **containerd on GPU nodes may be reconfigured and restarted.** If
  `/usr/bin/nvidia-container-runtime` is absent, the rendered config now selects `runc`
  instead of `nvidia` as the default runtime. On a node whose config previously selected
  `nvidia`, this is a config change, so the existing `containerd-config-file-changed` event
  fires and step 033 restarts containerd. This is the intended effect: with a missing binary
  as the default runtime, containerd cannot start **any** container on the node.
- **Non-GPU and CSE nodes are not affected.** The rendered `/etc/containerd/config.toml` is
  byte-identical to the previous version for ContainerdV1/V2 on non-GPU nodes and on CSE,
  with and without the toolkit present.
- No changes to ingress controllers, control-plane or Prometheus.
- Node convergence behaviour changes for GPU nodes: a GPU problem no longer stops a node
  from converging.

The in-tree GPU check ran as bashible step 006, long before kubelet, and failed the whole
run on any problem. bashible aborts a run on a failing step and retries it forever, so a GPU
problem blocked CRI configuration, kubelet and every other module's
NodeGroupConfiguration on that node.

The consequences are not limited to GPU functionality. Since every later step is skipped, a
node that trips the check never configures kubelet, never finishes bootstrap or an update,
and never applies the node configuration of unrelated modules. A GPU driver broken by a
kernel upgrade is therefore enough to leave a storage module's kernel module uninstalled on
that node, after which replicated volumes stop being attached and the workloads using them
lose their data volumes.

The visible symptom is a node stuck in `UPDATING` and an unrelated subsystem failing on it,
which points an investigation away from the GPU check.

Two more failures of the same class are fixed here:

- the compute capability check compared values with `bc -l`. `bc` is not among the mandatory
  packages and is missing on several supported distributions, so the substitution was empty,
  `(( ))` received an empty string and every GPU was declared unsupported. A healthy H100
  failed with `bc: command not found` followed by a misleading `compute_cap 9.0 is invalid`,
  and the node stopped converging;
- the GPU branch of step 032 edited `/etc/sysctl.d/99-sysctl.conf` and `/etc/sysctl.conf`
  in place with `sed -i`. Neither file exists on many distributions, so the step failed with
  `sed: can't read ...: No such file or directory` and was retried forever. The edit was also
  unreliable when the file did exist: without a literal `net.core.bpf_jit_harden = 2` line
  `sed` succeeded and the value was never persisted.

The in-tree GPU path is deprecated and is replaced by the `gpu` module, but until it is
removed it must not be able to take a node or a cluster down. The new alert makes clusters
still using it visible, since that is the path where these defects live.

Yes, the three convergence-blocking fixes are worth backporting: each of them can stop a node
from converging indefinitely, which also blocks the node configuration of every other module
on that node. Observed on 1.76.x.

The deprecation alert is safe to backport together with them: it only adds a metric and an
alerting rule and changes no behaviour, and the sooner clusters still on the deprecated path
become visible, the better.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: candi
type: fix
summary: A GPU problem no longer blocks node convergence.
impact: >
  On a NodeGroup with `spec.gpu`, a failing GPU check used to abort the bashible run and be
  retried forever, so the node never configured kubelet and never received updates from any
  other module. Such a node now converges and only its GPU stays unusable.
  If `/usr/bin/nvidia-container-runtime` is absent, containerd is now configured with `runc`
  as the default runtime instead of `nvidia`; on a node whose configuration previously
  selected `nvidia` this changes the containerd config and therefore restarts containerd.
  Previously such a node could not start any container at all.
impact_level: high
---
section: candi
type: fix
summary: GPU compute capability check no longer requires `bc`.
---
section: candi
type: fix
summary: Persist `net.core.bpf_jit_harden` for GPU nodes in an own sysctl drop-in instead of editing distribution files that may be absent.
---
section: node-manager
type: feature
summary: Alert when a NodeGroup uses the deprecated built-in GPU support while the `gpu` module is disabled.
```

### DIFF
--- a/docs/documentation/_data/deckhouse-alerts.yml
+++ b/docs/documentation/_data/deckhouse-alerts.yml
@@ -4297,6 +4297,21 @@ alerts:
         NodeGroup {{ $labels.node_group }} has unavailable instances.
       severity: "8"
       markupFormat: markdown
+    - name: NodeGroupUsesDeprecatedInCoreGPU
+      sourceFile: modules/040-node-manager/monitoring/prometheus-rules/node-group-deprecate.yaml
+      moduleUrl: 040-node-manager
+      module: node-manager
+      edition: ce
+      description: |
+        NodeGroup {{ $labels.name }} has `spec.gpu` configured while the `gpu` module is disabled,
+        so its GPU nodes are served by the built-in GPU support.
+        Built-in GPU support is deprecated and will be removed.
+
+        The `gpu` module must be enabled.
+      summary: |
+        NodeGroup {{ $labels.name }} uses the deprecated built-in GPU support.
+      severity: "9"
+      markupFormat: markdown
     - name: NodeRequiresDisruptionApprovalForUpdate
       sourceFile: modules/040-node-manager/monitoring/prometheus-rules/node-update.yaml
       moduleUrl: 040-node-manager
--- a/docs/documentation/_data/deckhouse-alerts.yml
+++ b/docs/documentation/_data/deckhouse-alerts.yml
@@ -4303,11 +4303,11 @@ alerts:
       module: node-manager
       edition: ce
       description: |
-        NodeGroup {{ $labels.name }} has `spec.gpu` configured while the `gpu` module is disabled,
-        so its GPU nodes are served by the built-in GPU support.
-        Built-in GPU support is deprecated and will be removed.
+        NodeGroup `{{ $labels.name }}` has [`spec.gpu`](/modules/node-manager/cr.html#nodegroup-v1-spec-gpu) configured while the `gpu` module is disabled.
+        As a result, GPU nodes in the NodeGroup use the built-in GPU support,
+        which is deprecated and will be removed in a future DKP release.
 
-        The `gpu` module must be enabled.
+        To use the supported GPU implementation, enable the [`gpu`](/modules/gpu/) module.
       summary: |
         NodeGroup {{ $labels.name }} uses the deprecated built-in GPU support.
       severity: "9"
--- a/modules/040-node-manager/monitoring/prometheus-rules/node-group-deprecate.yaml
+++ b/modules/040-node-manager/monitoring/prometheus-rules/node-group-deprecate.yaml
@@ -18,3 +18,24 @@
 
         To resolve this issue, delete the field from NodeGroup {{ $labels.name }}.
         This is safe, as the setting has already been migrated automatically.
+- name: d8.node-group-gpu-in-core
+  rules:
+  - alert: NodeGroupUsesDeprecatedInCoreGPU
+    expr: |
+      max by (name) (d8_node_group_gpu_in_core_deprecated) > 0
+    for: 15m
+    labels:
+      tier: cluster
+      severity_level: "9"
+    annotations:
+      plk_markup_format: markdown
+      plk_protocol_version: "1"
+      plk_create_group_if_not_exists__cluster_has_node_groups_deprecation_alerts: "NodeGroupsDeprecationAlerts,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
+      plk_grouped_by__cluster_has_node_groups_deprecation_alerts: "NodeGroupsDeprecationAlerts,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
+      summary: NodeGroup {{ $labels.name }} uses the deprecated built-in GPU support.
+      description: |
+        NodeGroup {{ $labels.name }} has `spec.gpu` configured while the `gpu` module is disabled,
+        so its GPU nodes are served by the built-in GPU support.
+        Built-in GPU support is deprecated and will be removed.
+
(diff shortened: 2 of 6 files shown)


z9r5 (approved):

modules/040-node-manager/monitoring/prometheus-rules/node-group-deprecate.yaml
```diff
@@ -18,3 +18,24 @@
...
+      plk_protocol_version: "1"
+      plk_create_group_if_not_exists__cluster_has_node_groups_deprecation_alerts: "NodeGroupsDeprecationAlerts,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
+      plk_grouped_by__cluster_has_node_groups_deprecation_alerts: "NodeGroupsDeprecationAlerts,tier=cluster,prometheus=deckhouse,kubernetes=~kubernetes"
+      summary: NodeGroup {{ $labels.name }} uses the deprecated built-in GPU support.
+      description: |
+        NodeGroup {{ $labels.name }} has `spec.gpu` configured while the `gpu` module is disabled,
+        so its GPU nodes are served by the built-in GPU support.
+        Built-in GPU support is deprecated and will be removed.
+
+        The `gpu` module must be enabled.
```
maxinsky:
```suggestion
        NodeGroup `{{ $labels.name }}` has [`spec.gpu`](/modules/node-manager/cr.html#nodegroup-v1-spec-gpu) configured while the `gpu` module is disabled.
        As a result, GPU nodes in the NodeGroup use the built-in GPU support,
        which is deprecated and will be removed in a future DKP release.

        To use the supported GPU implementation, enable the [`gpu`](/modules/gpu/) module.
```

docs/documentation/_data/deckhouse-alerts.yml
```diff
@@ -4377,6 +4297,21 @@ alerts:
...
+      sourceFile: modules/040-node-manager/monitoring/prometheus-rules/node-group-deprecate.yaml
+      moduleUrl: 040-node-manager
+      module: node-manager
+      edition: ce
+      description: |
+        NodeGroup {{ $labels.name }} has `spec.gpu` configured while the `gpu` module is disabled,
+        so its GPU nodes are served by the built-in GPU support.
+        Built-in GPU support is deprecated and will be removed.
+
+        The `gpu` module must be enabled.
```
maxinsky:
```suggestion
        NodeGroup `{{ $labels.name }}` has [`spec.gpu`](/modules/node-manager/cr.html#nodegroup-v1-spec-gpu) configured while the `gpu` module is disabled.
        As a result, GPU nodes in the NodeGroup use the built-in GPU support,
        which is deprecated and will be removed in a future DKP release.

        To use the supported GPU implementation, enable the [`gpu`](/modules/gpu/) module.
```

maxinsky (approved):

z9r5 (approved):

090809 (approved):

PR #22151, opened 2026-08-14
[user-authn] New crd UserAccount,  Move user, password, and provider-check logic from hooks to a controller

Jabejixo:
## Description

Moves Dex user and provider-check work out of addon-operator hooks into a `user-authn-controller` Deployment in `d8-user-authn`.

The controller now owns:

- `User` → Dex `Password` create/update/rename/orphan cleanup and `User.status` (groups, lock, `expireAt`)
- `UserOperation` (lock / unlock / reset password / reset 2FA) and 24h retention
- expiry: strip the user from `Group` members, then delete the `User`
- `DexProviderCheck` connectivity probes
- `d8_dex_provider_ldap_tls_conflict` metric
- `UserAccount` projection of Dex `Password` / `OfflineSessions` (no secrets)

`get_dex_user_crds` stays, names-only, so Helm can still toggle `enablePasswordDB` and deploy `user-api`.

Does not restart ingress, control-plane, or Prometheus. The user-authn Helm release adds the controller (HA, 2 replicas when HA mode is on).

User, password, lock, expiry, and provider-check logic shared the `/modules/user-authn` hook queue and a `*/5` cron. A new `User` waited on that queue; `User.status.lock` after a Dex lockout could lag up to 5 minutes (`Password` was snapshotted with events off). Provider checks blocked the same queue.

Per-object reconcile gives Password and status updates on the watch, not on the next cron tick, and keeps probes off the module hook queue.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: feature
summary: Reconcile users, passwords, operations, and Dex provider checks in a dedicated controller instead of module hooks.
impact_level: default
```


deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:user-authn/new-controller` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/31787009884) in 33m34s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.4.128`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 32m3s.

</details>

modules/150-user-authn/images/user-authn-controller/src/internal/controller/user/controller.go
```diff
@@ -0,0 +1,275 @@
...
+		Lock:     lock,
+	}); err != nil {
+		return 0, err
+	}
+	return controller.RequeueAfterTime(live.LockedUntil, now), nil
+}
+
+func (r *Reconciler) namespaceExists(ctx context.Context) (bool, error) {
+	ns := controller.Object(controller.NamespaceGVK)
+	err := r.client.Get(ctx, types.NamespacedName{Name: naming.DexNamespace}, ns)
```
AlwxSin: 🔴 **1. Namespace and Secret are read through the informer cache, but RBAC grants only `get`**

`cmd/main.go:133` sets `Client.Cache.Unstructured: true`, so every read of an `*unstructured.Unstructured` is served by the informer cache (`client.shouldBypassCache` returns false). Two call sites read objects that have no `ByObject` entry and no list/watch permission:

- this line — `Get` on `v1/Namespace`, while `templates/controller/rbac-for-us.yaml:164` grants `namespaces: [get]`;
- `internal/controller/providercheck/ldap.go:113` — `Get` on `v1/Secret`, while `templates/controller/rbac-for-us.yaml:192` grants `secrets: [get]` in `d8-user-authn` only (and that informer would be cluster-wide anyway).

The first read lazily starts an informer whose reflector `LIST`/`WATCH` gets 403, and `Informers.Get` then blocks in `WaitForCacheSync(ctx.Done(), ...)`. A reconcile context has no deadline, so all 8 workers of the `user` controller block permanently on the first `User` event: static users never get a `Password`, `User.status` is never synced. `/readyz` is `cache.WaitForCacheSync` over *all* informers with a 2s budget, so the pod also flips NotReady for good → 0 available replicas → `D8UserAuthnControllerUnavailable` + `D8UserAuthnControllerTargetDown` — plus a reflector retrying forbidden `LIST`/`WATCH` forever (~2 requests per 1.6s). The other five controllers keep working, because the informer map lock is released before the wait; that is why e2e stayed green. The Kerberos step is bounded by the 1-minute probe context, so it stalls 60s and then reports `ldapKerberosKeytab: Failed` for every LDAP+Kerberos provider.

Reproduced against a fake apiserver that 403s `LIST namespaces` and serves `GET namespaces/<name>`, with the same client/cache wiring as `newManagerOptions`:

```
apiserver:    GET /api/v1/namespaces?...&watch=true        -> 403
apiserver:    GET /api/v1/namespaces?limit=500&resourceVersion=0 -> 403
cached Get:   err=Timeout: failed waiting for *unstructured.Unstructured Informer to sync after 5.001s
uncached Get: ok, name=d8-user-authn
```

In-cluster check: create one static `User`, then `kubectl -n d8-user-authn get pod -l app=user-authn-controller` — the pod stays `0/1 Running`.

Suggested fix: drop `namespaceExists` altogether — it guarded a hook that ran *before* Helm created the namespace, and is meaningless for a Deployment that lives in that namespace — and read the keytab Secret through `mgr.GetAPIReader()`, the pattern already used for `passwordFromAPI`. `client.CacheOptions.DisableFor` on both GVKs would work too.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/user/password.go
```diff
@@ -0,0 +1,346 @@
...
+		expectedName = passwordName(user.Email)
+	}
+	for _, pw := range passwords {
+		if !isModuleManagedPassword(pw) {
+			continue
+		}
+		if expectedName != "" && pw.Name == expectedName {
+			continue
+		}
+		if pw.Username != "" && pw.Username != reqName {
```
AlwxSin: 🟡 **2. A module-managed Password with an empty `username` is deleted by any user's reconcile, reverting that account's password**

`username == ""` falls through this guard, so the object is treated as belonging to whichever user is currently being reconciled. Sweeping nameless leftovers is clearly intended (`controller_test.go:196`), but the check ignores the object *name*: the old hook computed `expectedPasswordNames` across **all** users, so a nameless `Password` sitting at user B's canonical name (`ToFnvLikeDex(lower(email_B))`) was kept, and `reconcileExistingPassword` then adopted it and filled in `username`.

Now user A's reconcile deletes it. B's next reconcile misses by name, misses the rename fallback (which requires `pw.Username == user.Name`), falls into `createPassword` and re-seeds `hash` from the immutable `User.spec.password` — B's current password, `previousHashes`, `hashUpdatedAt` and lock state are gone, silently.

Narrow trigger (a `heritage: deckhouse` Password without `username`), but the outcome is credential loss. Suggested fix: sweep a nameless Password only when its name is not the canonical name of any existing `User`, or match positively on `pw.Username == reqName` and let the canonical-name object be adopted as before.

modules/150-user-authn/templates/rbacv2/use/view.yaml
```diff
@@ -14,6 +14,7 @@ rules:
   resources:
   - dexauthenticators
   - dexclients
+  - useraccounts
```
AlwxSin: 🟡 **3. `useraccounts` in the `use` capability roles cannot apply to a cluster-scoped resource**

`UserAccount` is `scope: Cluster` (`crds/user-account.yaml:10`), while the sibling resources in this very rule (`dexauthenticators`, `dexclients`) are namespaced. A `use` capability is namespaced by construction: it aggregates into `d8:use:role:*`, which `user-authz` binds with a namespaced `RoleBinding` (`modules/140-user-authz/hooks/handle_manage_bindings.go:234-257`), and a RoleBinding to a ClusterRole never grants cluster-scoped resources. So this block — and the identical one in `rbacv2/use/edit.yaml:21-29` — is dead. The grant that does work was already added in `rbacv2/manage/{view,edit}.yaml`.

Should a use-role ever be bound cluster-wide, the rule silently becomes identity disclosure: every project-level viewer could enumerate every cluster identity, emails and groups included, external accounts too. Suggested fix: delete both blocks.

modules/150-user-authn/templates/rbacv2/manage/view.yaml
```diff
@@ -17,6 +17,7 @@ rules:
   - dexproviderchecks
   - dexproviders
   - groups
+  - useraccounts
```
AlwxSin, line 20: 🟡 **4. Legacy (RBAC v1) `user-authz` ClusterAdmin role was not updated with `useraccounts`**

*(the real place is `modules/150-user-authn/templates/user-authz-cluster-roles.yaml:45`, which this PR does not touch — commenting on the rbacv2 counterpart instead)*

`useraccounts` was added to `rbacv2/manage/{view,edit}`, `rbacv2/use/{view,edit}` and `d8:user-authn:admin-kubeconfig`, but not to `d8:user-authz:user-authn:cluster-admin`, which enumerates `users`, `groups`, `dexproviders`, `dexproviderchecks` and `useroperations`. Clusters still on the old role model cannot read the new resource, and the console is its intended consumer. Either add it there as well, or state explicitly that the legacy role model is out of scope.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/secrets.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	toolscache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"user-authn-controller/internal/naming"
+)
+
+var secretFields = []string{"hash", "previousHashes", "connectorData", "totp", "token"}
```
AlwxSin: 🟡 **5. `obsoleteToken` is not stripped from the informer cache**

Dex's `RefreshToken` stores both `token` and `obsoleteToken` (`storage/kubernetes/types.go` upstream); `obsoleteToken` is the previous refresh token and is still honoured during the rotation window. RefreshToken objects are fully cached via `InformerCacheByObject`, so live credential material stays in the controller's memory — which is exactly what this strip list exists to prevent. Suggested fix: add `obsoleteToken`, and re-check the fork's `RefreshToken` field set while there.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/useraccount/controller.go
```diff
@@ -0,0 +1,549 @@
...
+		if naming.ExternalName(sess.ConnID, sess.UserID) != name {
+			continue
+		}
+		desired := projectExternal(sess, provider.Type, now)
+		return &desired, nil
+	}
+	return nil, nil
+}
+
+func (r *Reconciler) passwordForLocalAccount(ctx context.Context, name string) (passwordView, bool, error) {
```
AlwxSin, line 252: 🟡 **6. UserAccount reconcile does full cache lists per object — O(N²) JSON round-trips on restart or on any DexProvider change**

For any non-local account name the named `Get` here misses, so the `listPasswords()` fallback always runs, followed by `listSessions()` + `listProviders()` in `desiredExternal` (line 233). Every listed item goes through `DecodeInto` (`internal/controller/unstructured.go:55`), i.e. `MarshalJSON` + `json.Unmarshal` per object — not a cheap deep copy.

The fan-out is per object: on startup every `Password` and every `OfflineSessions` add event enqueues one account, and `mapDexProvider` enqueues every account of a connector. So the work is N reconciles × O(N) JSON round-trips. With a few thousand external users that is minutes of CPU and a lot of garbage against the VPA ceiling of `memory: 150Mi` (`templates/controller/deployment.yaml:28`).

Suggested fix: skip the `listPasswords` fallback when the name cannot be a local one, and look sessions up by `naming.ExternalName(connID, userID)` — a field index, or a map built once — instead of scanning the full list per account.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/useroperation/execute.go
```diff
@@ -0,0 +1,397 @@
...
+		"user", userPassword.Username,
+		"for", op.Spec.Lock.For,
+		"lockedUntil", lockedUntil.UTC().Format(time.RFC3339),
+	)...)
+
+	if err := r.patchMerge(ctx, controller.PasswordGVK, userPassword.Namespace, userPassword.Name, map[string]any{
+		"lockedUntil": lockedUntil.UTC().Format(time.RFC3339),
+		"metadata": map[string]any{
+			"annotations": map[string]any{
+				lockedByAdministratorAnnot: "",
```
AlwxSin: 🔵 **7. Lock marker annotation is written with two different values**

The local lock path writes `deckhouse.io/locked-by-administrator: ""`, the external one writes `"true"` (`execute.go:257`). Every reader is presence-only (`hasAdminLock`, `lockFromPassword`), so nothing breaks today, but any future value-based check will. Suggested fix: pick one — `"true"` reads better in `kubectl get -o yaml`.

modules/150-user-authn/monitoring/prometheus-rules/controller.yaml
```diff
@@ -0,0 +1,96 @@
...
+        Check deployment and pod states:
+
+        ```bash
+        d8 k -n d8-user-authn get deploy user-authn-controller
+        d8 k -n d8-user-authn get pods -l app=user-authn-controller
+        d8 k -n d8-user-authn logs deploy/user-authn-controller --all-pods=true
+        ```
+
+  - alert: D8UserAuthnControllerTargetDown
+    expr: sum(up{job="user-authn-controller", namespace="d8-user-authn"}) == 0
```
AlwxSin: 🔵 **8. This alert cannot detect the missing PodMonitor its own description promises**

`sum(up{...}) == 0` returns an empty vector when no series match at all, so the alert never fires in exactly the case the description lists ("The controller is down, not Ready ..., or the PodMonitor is missing"). Suggested fix: add a second rule on `absent(up{job="user-authn-controller", namespace="d8-user-authn"})`, or drop that clause from the description.

modules/150-user-authn/templates/controller/deployment.yaml
```diff
@@ -0,0 +1,103 @@
...
+{{- if (include "helm_lib_ha_enabled" .) }}
+          value: "true"
+{{- else }}
+          value: "false"
+{{- end }}
+        ports:
+        - containerPort: 9090
+          name: healthz
+          protocol: TCP
+        - containerPort: 9091
```
AlwxSin, line 74: 🔵 **9. Controller metrics are exposed unauthenticated, unlike the module's other metrics endpoints**

Port 9091 serves `/metrics` over plain HTTP with no kube-rbac-proxy, and `templates/controller/podmonitor.yaml` scrapes it directly. The repo convention is mixed (`templates/dex/podmonitor.yaml:18` in this very module goes through `https-metrics` + `prometheus-token`, while `002-deckhouse` webhook-handler uses a plain port), but inside this module it is now inconsistent. Exposure is low — controller-runtime metrics plus DexProvider names — so this is a convention call, not a leak.

modules/150-user-authn/images/user-authn-controller/mount-points.yaml
```diff
@@ -0,0 +1,2 @@
+dirs:
+- /certs
```
AlwxSin: 🔵 **10. Dead `/certs` mount point**

The Deployment mounts no volumes and every TLS config is built from in-memory PEM (`providercheck/tls.go:145`), so this directory is never used. Suggested fix: drop the file.

modules/150-user-authn/hooks/get_dex_user_crds.go
```diff
@@ -18,699 +18,45 @@ package hooks
...
-			}
-			input.Logger.Info("Deleting orphaned Password",
-				slog.String("name", password.Name), slog.String("username", password.Username))
-			input.PatchCollector.Delete("dex.coreos.com/v1", "Password", password.Namespace, password.Name)
+			return fmt.Errorf("cannot iterate over 'users' snapshot: %w", err)
 		}
+		users = append(users, dexUserName{Name: name})
 	}
-
 	input.Values.Set("userAuthn.internal.dexUsersCRDs", users)
```
AlwxSin, line 56: 🔵 **11. Dead schema left in values for `dexUsersCRDs`**

*(the real place is `modules/150-user-authn/openapi/values.yaml:217-251`, which this PR does not touch)*

The hook now emits `[{name: ...}]` only, but the values schema still describes `encodedName`, `spec` and `status` (including the lock reason enum). Harmless — nothing is `required` — but misleading now. Removing it also makes a nice side effect of this PR explicit: no bcrypt hash reaches addon-operator values any more.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/user/password.go
```diff
@@ -0,0 +1,346 @@
...
+func annotationStripped(existing passwordView, now time.Time) bool {
+	lockActive := existing.LockedUntil != nil && existing.LockedUntil.After(now)
+	if lockActive {
+		return false
+	}
+	_, ok := existing.Annotations[lockedByAdministratorAnnot]
+	return ok
+}
+
+func (r *Reconciler) stripExpiredAdminLock(ctx context.Context, existing passwordView, now time.Time, isRename bool) error {
```
AlwxSin: 🔵 **12. Dead parameter and a misleading predicate name**

`stripExpiredAdminLock` is only reachable from the `passwordExists` branch of `reconcilePassword`, where `isRename` is always false — the parameter never does anything. And `annotationStripped` (line 151) reads as "was stripped" while it actually answers "should be stripped"; `shouldStripAdminLock` would say it.

modules/150-user-authn/images/user-authn-controller/src/go.mod
```diff
@@ -0,0 +1,71 @@
...
+require (
+	github.com/go-ldap/ldap/v3 v3.4.13
+	github.com/go-logr/logr v1.4.3
+	github.com/prometheus/client_golang v1.22.0
+	github.com/prometheus/client_model v0.6.1
+	go.uber.org/zap v1.27.0
+	golang.org/x/crypto v0.55.0
+	k8s.io/apimachinery v0.35.1
+	k8s.io/client-go v0.35.1
+	sigs.k8s.io/controller-runtime v0.21.0
```
AlwxSin: 🔵 **13. go.mod mixes controller-runtime v0.21 with k8s.io v0.35**

controller-runtime v0.21.0 targets k8s 0.33, while this module pins `k8s.io/apimachinery` and `k8s.io/client-go` at v0.35.1 and `k8s.io/apiextensions-apiserver` at v0.33.0. It builds and the tests pass, but the combination is unsupported upstream and drifts on list/watch semantics (client-go 1.32+ enables WatchList by default). Suggested fix: move to controller-runtime v0.22.x.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/user/controller.go
Jabejixo: Dropped `namespaceExists` — the Deployment already lives in that namespace. Kerberos keytab Secret is read via `GetAPIReader()`, same pattern as Password hashes.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/user/password.go
Jabejixo: Nameless module-managed Password is kept when its name is the canonical name of any User. That User's next reconcile adopts it and fills `username` instead of reseeding the hash.

modules/150-user-authn/templates/rbacv2/use/view.yaml
Jabejixo: Removed `useraccounts` from both `rbacv2/use` roles. Cluster-scoped grant stays on `manage` only.

modules/150-user-authn/templates/rbacv2/manage/view.yaml
Jabejixo: Added `useraccounts` to `d8:user-authz:user-authn:cluster-admin` so the legacy role model can list the CRD for Console.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/secrets.go
Jabejixo: `obsoleteToken` is now in the informer strip list with `token` and the other Dex credential fields.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/useraccount/controller.go
Jabejixo: External account names skip the Password list fallback. OfflineSessions is a named Get by Dex `OfflineTokenName`, DexProvider is a named Get by `connID`.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/useroperation/execute.go
Jabejixo: Local and external lock paths both write `deckhouse.io/locked-by-administrator: "true"`.

modules/150-user-authn/monitoring/prometheus-rules/controller.yaml
Jabejixo: The rule now also fires on `absent(up{job="user-authn-controller", namespace="d8-user-authn"})`, so a missing PodMonitor is visible.

modules/150-user-authn/templates/controller/deployment.yaml
Jabejixo: Left `:9091` plain HTTP, same as the 002-deckhouse webhook-handler. Exposure is controller-runtime metrics plus DexProvider names — not wrapping it in kube-rbac-proxy in this PR.

modules/150-user-authn/images/user-authn-controller/mount-points.yaml
Jabejixo: `/certs` is gone. The file stays with empty `dirs` because the werf `image mount points` include requires it.

modules/150-user-authn/hooks/get_dex_user_crds.go
Jabejixo: `dexUsersCRDs` schema is names-only now, matching what the hook actually writes.

modules/150-user-authn/images/user-authn-controller/src/internal/controller/user/password.go
Jabejixo: Dropped the unused `isRename` parameter. The predicate is `shouldStripAdminLock`.

modules/150-user-authn/images/user-authn-controller/src/go.mod
Jabejixo: Bumped to controller-runtime v0.23.3, which matches client-go/apimachinery 0.35. v0.22 targets 0.34.

AlwxSin (approved):

AlwxSin (approved):

Nikolay1224 (approved):

z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [22361](https://github.com/deckhouse/deckhouse/pull/22361) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #21936, opened 2026-08-06
[user-authz] Fix validating webhooks

Suselz:
## Description
Refactored the multitenancy validating webhook to read `enableMultiTenancy` from the Module `v1alpha2` CR (`status.lastAppliedConfiguration`) instead of `ModuleConfig.spec.settings`.
Also removed the incorrect "(EE Only)" suffix from error messages.

## Why do we need it, and what problem does it solve?

In some editions (e.g. CSE), `enableMultiTenancy` defaults to `true` in the OpenAPI schema. When a user has not explicitly set this parameter in the ModuleConfig, the field is absent from `spec.settings` — but the effective value is still `true`. The webhook was reading from `ModuleConfig.spec.settings`, causing it to treat multitenancy as disabled when it was actually enabled by default.

The fix uses the `Module` v1alpha2 CR's `status.lastAppliedConfiguration`, which contains the user config merged with schema defaults — the effective runtime configuration.

Additionally, the "(EE Only)" label was inaccurate since multitenancy is available in multiple editions (BE, SE, SE+, EE, CSE-lite, CSE-pro), not just EE.

## Checklist
- [ ] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [ ] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: user-authz
type: fix
summary: Fix multitenancy validatiing webhook to read effective enableMultiTenancy from the Module `v1alpha2` CR 
impact_level: default
```

PR #22158, opened 2026-08-14
[user-authz] Fix multitenancy check hooks

Suselz:
## Description

The mechanism for determining multi-tenancy involvement has been fixed.

Previously, the `multitenancy.py` validating webhook read the effective `enableMultiTenancy` value from a `Module` CR (`deckhouse.io/v1alpha2`) field that doesn't actually exist for built-in modules, so the webhook always treated MultiTenancy as disabled regardless of the real setting. The fix adds a small Go hook (`discoverMultitenancyState`) that mirrors the already defaults-merged `userAuthz.enableMultiTenancy` value into a ConfigMap in the module's own `d8-user-authz` namespace (with a Kubernetes binding on that namespace so the ConfigMap is republished immediately once it's created, without waiting for the next unrelated reconcile), and the webhook now reads that ConfigMap instead.

Fixed bug added in #21936

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: fix
summary:  Fix multitenancy check hooks
impact_level: default
```


Jabejixo (commented):
Reviewed the diff plus the surrounding module (chart templates, webhook-handler RBAC, existing tests). The fix is right in substance — the effective value genuinely cannot be read from `ModuleConfig` when the edition's config-schema supplies the default, and `Module` `deckhouse.io/v1alpha2` is not served for built-in modules. The notes are about how the value is published and about one behaviour change; they are left inline on the relevant lines:

- 🔴 removing `enableMultiTenancy` from `spec.settings` now skips CAR validation, which it did not on `main`;
- 🟡 the hook could be a chart template — same value, fewer moving parts;
- 🟡 the bootstrap window right after MultiTenancy is enabled;
- 🟢 unexported constants, mutable default argument in the test factory.

Checked and fine: the webhook-handler has cluster-wide `get/list/watch` on ConfigMaps, so the snapshot works; the `heritage: deckhouse` label puts the object under the existing ValidatingAdmissionPolicy that keeps non-system identities from patching it; the jq filter is strict (`.data.enableMultiTenancy == "true"`), so any other value reads as disabled; an empty snapshot fails closed. The Python tests cover the new paths well, and the CSE case dropped from the CAR tests is genuinely a duplicate now that the state has a single source.

modules/140-user-authz/webhooks/validating/multitenancy.py
```diff
@@ -104,20 +129,32 @@ def main(ctx: hook.Context):
...
+        field_present = "enableMultiTenancy" in settings
+
+        # don't check existing CARs if the request explicitly enables MultiTenancy...
+        if field_present and settings.enableMultiTenancy is True:
+            return [], []
+        # ...or if the request doesn't touch enableMultiTenancy at all (it's absent from
+        # spec.settings, e.g. an unrelated ModuleConfig edit) and it's still effectively
+        # enabled — explicitly, or via the edition's config-schema default.
+        if not field_present and is_multitenancy_enabled(ctx):
             return [], []
```
Jabejixo, line 168: 🔴 **Skipping CAR validation when the field is removed is a regression.**

On `main` this branch read `req.object.spec.settings.enableMultiTenancy is True`, so an absent field fell through to CAR validation — removing the field was still treated as a possible disable. Now an absent field defers to the ConfigMap, which at admission time still reflects the **previous** state.

On an edition where the schema default is `false`, dropping `enableMultiTenancy` from `spec.settings` *is* the disable, and this path skips the check: MultiTenancy goes off while CARs keep `limitNamespaces` / `namespaceSelector` / `allowAccessToSystemNamespaces`, whose restrictions then stop being enforced — exactly what this webhook exists to prevent.

The webhook cannot compute the schema default, but it does not have to: `oldObject` is in the request, and the test factories already build it.

```python
old_settings = req.oldObject.spec.settings
was_present = "enableMultiTenancy" in old_settings

# an unrelated edit: the field was absent before and stays absent,
# so the effective value does not change
if not field_present and not was_present and is_multitenancy_enabled(ctx):
    return [], []
```

Worth a test for "the field was explicitly `true`, the request removes it" — the current suite does not cover it.

modules/140-user-authz/hooks/discover_multitenancy_state.go
```diff
@@ -0,0 +1,114 @@
...
+			Kind:       "Namespace",
+			NameSelector: &types.NameSelector{
+				MatchNames: []string{internal.Namespace},
+			},
+			FilterFunc: func(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
+				return obj.GetName(), nil
+			},
+		},
+	},
+}, discoverMultitenancyState)
```
Jabejixo: 🟡 **This could be a chart template, and that removes two problems at once.**

`templates/namespace.yaml` is already gated on the same merged value this hook reads, and the chart sees it too:

```yaml
{{- if .Values.userAuthz.enableMultiTenancy }}
```

Putting the ConfigMap inside that block is ~10 lines and replaces 114 lines of Go plus 134 lines of test:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: d8-user-authz-multitenancy-state
  namespace: d8-{{ .Chart.Name }}
  {{- include "helm_lib_module_labels" (list .) | nindent 2 }}
data:
  enableMultiTenancy: "true"
```

It also drops the "namespace may not exist yet" special case and the Namespace watch that exists only to re-run the hook, the object is deleted with the release instead of being left behind, and it closes the bootstrap window (see the comment below). If a template does not work here for some reason, it is worth saying so in the header comment — it does not follow from the code.

modules/140-user-authz/hooks/discover_multitenancy_state.go
```diff
@@ -0,0 +1,114 @@
...
+
+func discoverMultitenancyState(_ context.Context, input *go_hook.HookInput) error {
+	if len(input.Snapshots.Get(multitenancyNamespaceSnapshot)) == 0 {
+		// The module's own namespace doesn't exist yet — nothing to publish. The
+		// webhook treats a missing ConfigMap the same as enableMultiTenancy=false,
+		// which is correct: this namespace only exists when it's true.
+		return nil
+	}
```
Jabejixo: 🟡 **Bootstrap window: a legitimate CAR is rejected right after MultiTenancy is enabled.**

`OnBeforeHelm` runs before the chart creates the namespace, so the first pass publishes nothing; the ConfigMap appears only once the Namespace event re-runs the hook. In between, the webhook reads "disabled" and denies CARs that carry multitenancy fields.

It fails closed and heals itself, so this is not a security problem — but from the outside it looks like "I enabled MultiTenancy and it still refuses". Rendering the ConfigMap from the chart puts it in the same apply as the namespace and the window disappears.

modules/140-user-authz/hooks/discover_multitenancy_state.go
```diff
@@ -0,0 +1,114 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/deckhouse/deckhouse/modules/140-user-authz/hooks/internal"
+)
+
+const (
+	MultitenancyStateConfigMapName = "d8-user-authz-multitenancy-state"
+	MultitenancyStateDataKey       = "enableMultiTenancy"
```
Jabejixo: 🟢 Neither constant is used outside this file — the Go test hardcodes the same strings, so renaming one would not even fail the test. Unexport them and use them from the test instead.

modules/140-user-authz/webhooks/validating/multitenancy_test_factories.py
```diff
@@ -258,9 +258,32 @@ def build_three_not_multitenancy_related_cars() -> list[CAR]:
     ]
 
 
-def prepare_module_config_binding_context(module_enable_multitenancy_field: Optional[bool], cars: list[CAR] = []) -> str:
+def prepare_module_config_binding_context(
+        module_enable_multitenancy_field: Optional[bool],
+        cars: list[CAR] = [],
```
Jabejixo: 🟢 `cars: list[CAR] = []` is a mutable default argument — inherited from `main`, but the signature is being touched anyway, so `None` plus a local default is the safer form.

modules/140-user-authz/webhooks/validating/multitenancy.py
Suselz: Fixed in 8645cff — added the `oldObject` comparison you suggested: the ConfigMap-fallback shortcut now only applies when `enableMultiTenancy` was absent from `spec.settings` both before and after the request. If it was explicitly set and is being removed, it falls through to CAR validation regardless of what the ConfigMap currently says. Added a test for exactly the "was true, request removes it" case, plus a sanity test for the genuinely-unrelated-edit path. Verified live: removing an explicit `enableMultiTenancy: true` while a CAR still has `limitNamespaces` is now denied.

modules/140-user-authz/hooks/discover_multitenancy_state.go
Suselz: Agreed — done in 8645cff. Removed the Go hook and its test entirely; the ConfigMap is now rendered directly in `templates/namespace.yaml`, in the same `if .Values.userAuthz.enableMultiTenancy` block as the namespace. Verified live: the ConfigMap now carries `app.kubernetes.io/managed-by: Helm`, and appears/disappears atomically with the namespace.

modules/140-user-authz/hooks/discover_multitenancy_state.go
Suselz: Closed by the chart-template move in 8645cff (see the sibling comment) — the ConfigMap is applied in the same Helm release as the namespace now, so there's no separate hook pass to wait for. Verified live: namespace + ConfigMap both showed up within ~6s of enabling MultiTenancy, and a CAR with `limitNamespaces` was accepted right after.

modules/140-user-authz/hooks/discover_multitenancy_state.go
Suselz: Moot now — the whole file is gone as part of the chart-template move in 8645cff.

modules/140-user-authz/webhooks/validating/multitenancy_test_factories.py
Suselz: Fixed in 8645cff: `cars: Optional[list[CAR]] = None` with `cars = cars or []` inside.

nabokihms (approved):

deckhouse-BOaTswain:
Cherry pick PR [22178](https://github.com/deckhouse/deckhouse/pull/22178) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22165, opened 2026-08-14
[deckhouse] fix values validation failing on the injected registry

Fkuloff:
## Description

`InjectRegistryValue` wrote a `*structpb.Struct` into the package values tree; it now writes a plain `map[string]any` with the same keys. The `flant/addon-operator` bump carries the same fix for the module path, where the injected value was a `*modules.Registry` struct.

CEL validation of `x-deckhouse-validations` converts values with `structpb`, which accepts only maps, slices and scalars, so a Go struct anywhere in the values tree fails the whole validation with `proto: invalid type`. On the module path this broke every hook values patch of any module that has a root-level validation rule and a registry injected from a `ModuleSource`, and the module never converged; on the package path the same failure was waiting for the first package with a root-level rule.

### Notes

Pinned to `addon-operator v1.24.10`, the release carrying flant/addon-operator#820.

### Manual testing

Converged the same external module twice in one cluster — with `addon-operator v1.24.9` and with the pin from this PR. The module carries a root-level `x-deckhouse-validations` rule and an onStartup hook that patches module values, which is the combination that triggers the defect. Module name sanitized to `<module>`, stacktrace elided.

Before, every retry of the hook's values patch failed:

```
k -n d8-system logs deployment/deckhouse | grep "invalid type"
{"level":"error","logger":"addon-operator.module-run","msg":"ModuleRun failed. Requeue task to retry after delay.","binding":"ConvergeModules","count":8,"error":"2 errors occurred:\n\t* cannot apply values patch for module values\n\t* convert values to struct: proto: invalid type: *modules.Registry\n\n","event.type":"OperatorStartup","module":"<module>","phase":"Startup","queue":"main","task.id":"<task-id>","stacktrace":[...]}

k get module <module> -o jsonpath='{.status.phase}'
Error
```

After, same module and same rule:

```
k -n d8-system logs deployment/deckhouse | grep -c "invalid type"
0

k get module <module> -o jsonpath='{.status.phase}'
Ready
```

The rule is evaluated rather than silently skipped — admission rejects a value it forbids and accepts one it allows:

```
k patch mc <module> --type=merge -p '{"spec":{"version":1,"settings":{"replicas":5}}}' --dry-run=server
The request is invalid: : spec.settings are not valid (version 1):  validate config values: replicas must not exceed 3

k patch mc <module> --type=merge -p '{"spec":{"version":1,"settings":{"replicas":2}}}' --dry-run=server
moduleconfig.deckhouse.io/<module> patched
```

Not covered in the cluster: the packages path in `internal/packages/values`, which had the same defect with a `*structpb.Struct` and is fixed by the same change.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: fix values validation failing on the injected registry
impact_level: low
```

### DIFF
--- a/deckhouse-controller/internal/packages/values/storage.go
+++ b/deckhouse-controller/internal/packages/values/storage.go
@@ -24,7 +24,6 @@ import (
 	addonutils "github.com/flant/addon-operator/pkg/utils"
 	"github.com/go-openapi/spec"
 	"github.com/go-openapi/swag/conv"
-	"google.golang.org/protobuf/types/known/structpb"
 
 	sdkutils "github.com/deckhouse/module-sdk/pkg/utils"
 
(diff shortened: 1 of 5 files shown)


ipaqsa (approved):

ldmonster (approved):

deckhouse-BOaTswain:
Cherry pick PR [22194](https://github.com/deckhouse/deckhouse/pull/22194) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22166, opened 2026-08-14
test: rollout gate and self-updating agent trial stand

borg-z:
Test stand only — never merge.

Builds two images for the live trials on the immutable stand:

- `node-controller` with the rollout-gate membership fix (a slot is held only by a node carrying the current, unproven spec; unhealthy nodes are served first).
- `nodelet` sysext pinned to the tip of `selfupdate-on-main` (trial-and-demote self-update).

Base is `nodelet-sysext` (#22101), so the diff shows only the gate commits plus the pin.

The real gate change will come as its own PR from `rollout-gate-membership`.

https://claude.ai/code/session_0146tQC21Qc8YED924aALToL


borg-z:
Тестовый стенд, своё дело сделал: на его образах прошли живые испытания гейта выката и самообновления агента. Итог сведён в #21557.

PR #22182, opened 2026-08-15
[dhctl] Phases rewrite

090809:
## Description

`dhctl bootstrap` no longer executes a hand-written slice of nine phase methods while announcing a second, separately hand-maintained phase list. Both now come from one declarative phase tree in `pkg/operations/phases`, gated on cluster type plus a new "does this config carry a ClusterConfiguration" axis: the walker resolves the tree, announces each node itself, and looks the body up by name — a declared node with no implementation is a hard error instead of a phase Commander is told about that nothing runs. Riding on top of that: five new bootstrap phases that were already running but never declared, `--skip-phase` on the CLI and gRPC, per-cluster state-cache identities for static clusters, an abort that never destroys a running cluster, and support for bootstrapping a cluster whose control plane dhctl did not create.

**Phase state machine (`pkg/operations/phases`)**

- Bootstrap declares 14 top-level phases for a cloud cluster (was 9), 11 for static, 7 for a config with no ClusterConfiguration. New: `Preparation` (with `ImagesDownload`/`ConfigValidation`/`CachePreparation`/`StatePreparation`), `FirstMaster`, `ParseResources`, `WaitForSSHOnMaster`, `WaitForControlPlaneManagerReadiness`; `InstallKubernetes` gained the `ModulesPreparation` sub-phase. All of these were already announced at runtime — undeclared, so `calculatePhaseProgress` resolved index `-1` and dropped the event silently.
- `FirstMaster` and `WaitForControlPlaneManagerReadiness` moved from sub-phases to top-level phases with their strings unchanged, so they need gating independent of their former parents. A consumer keying on `completedSubPhase` stops seeing those two strings; one keying on `completedPhase` starts seeing them.
- Destroy declares 6 phases for static (was 2, since `BaseInfra` was gated out) and 4 for cloud (was 3): `CreateStaticDestroyerNodeUser`, `WaitStaticDestroyerNodeUser`, `SetDeckhouseResourcesDelete` and `UpdateStaticDestroyerIPs` were also announced-but-undeclared. Converge drops `ScaleToMultiMaster`, which was only ever a `ConvergeState` value and could never complete, and moves `CheckInfra` to a Cloud-gated sub-phase of `Check`.
- The single negative gate `ifNotStatic` was replaced by positive `ifCloud`/`ifStatic`/`ifHasClusterConfiguration`, so an unresolved cluster type no longer reads as "cloud". `SetClusterConfig` no longer short-circuits on the zero value, because "no ClusterConfiguration, type unknown" is now a legitimate answer rather than an absence.
- Three progress bugs fixed: a phase run nested inside another phase's `Run` scope no longer overwrites the enclosing phase (a successful static destroy used to finish reporting progress 0.0 and an empty `completedPhase`); `Complete()` now names the last phase that actually ran instead of the last declared one; an undeclared phase now logs a WARN instead of moving nothing in silence.
- Terminal progress bar: reaches 100% on runs whose declared tail was skipped, and renders an unknown phase name raw instead of a blank 60-char pad that also froze the bar.
- Refactor: six flat `[]PhaseWithSubPhases` literals became six private `node` trees pruned by `gate()` and flattened by `project()` into the unchanged wire struct; `PhaseWithSubPhases` is now plain copyable data and `OperationSubPhase` is a type alias of `OperationPhase`. New exported `SkippablePhases`/`ResolveSkipPhases`/`PhasesFor` back `--skip-phase`.

**Bootstrap orchestration**

- Every phase past `Preparation` gained a `requireStateCache` guard plus an optional `validate` hook that refuses to run on missing inputs and names the phase that produces them. Several of these previously produced silent wrong outcomes rather than failures — a missing `installDeckhouseResult` made `CreateResources` and `Finalization` no-op while the bootstrap still reported success.
- `bashibleNodeInterface` now requires `GetSSHProvider` to succeed whenever the cluster is Cloud or `CheckHosts` is set, instead of falling through to the local-node interface — that fallback installed Kubernetes inside the installer container and reported success.
- Static `PostInfraPreflights` records every connection host to the cache (`first-master`, `master-1`, …) instead of only `Hosts[0]`, and a failed cache write now fails the phase instead of logging a warning. Hosts left out of that record keep a half-installed control plane running that abort never cleans.
- `KubeadmBootstrap` and `MasterNodeSelector` follow `HasClusterConfiguration()` instead of being hardcoded `true`; the deferred `Finalize`/`finishProgress`/cleanup block is now registered before the walk, so a config-load failure no longer leaks the progress goroutine.
- For a config with no ClusterConfiguration, `InstallDeckhouse` skips `WaitForFirstMasterNodeBecomeReady` outright and returns as soon as the manifests are applied and Deckhouse is Ready; it also persists the cluster UUID under the `uuid` state-cache key, which `commander.ParseMetaConfig` then reads for every later operation on that cluster.
- `bootstrap-phase base-infra` is now the same walk restricted to one node, so it picks up x-extension validation, `ApplyCNIBootstrap` and the pipeline it had drifted away from. It no longer opens a terminal progress box, and `runProgress` is left with no callers.
- `RunBashiblePipeline` takes a `CompleteSubPhase` func instead of a whole `PhasedExecutionContext`, with a nil check in `Validate()` (the panic from #20102).

**Cache identity and bootstrap state**

- The state-cache directory name is now resolved by `cacheIdentity`: cloud clusters keep the byte-identical `<prefix>-<provider>-terraform-state-cache`, while static and managed clusters get `ssh-<hash of sorted host names>`, `kubeconfig-<hash>` or `in-cluster`. Previously every static and managed cluster on a machine collapsed into one directory — sharing one cluster UUID, one master-host list and one tombstone, so a second static bootstrap met the first's tombstone.
- A one-time adoption path lets the first post-upgrade cluster keep using the old collapsed directory, claimed via an `O_EXCL` `<hash>.owner` marker beside it. Nothing is moved or merged.
- `cache.initCache` installs the global cache only on success. It used to store a failed `choiceCache` result — a nil `*StateCache` inside a non-nil interface — so every later `cache.Global()` call, including the caller's own deferred `Finalize`, dereferenced nil. A cache that cannot be opened now reports its reason instead of SIGSEGV.
- `InstallDeckhouse` resolves the cluster UUID against `kube-system/d8-cluster-uuid` before checking it: dhctl's own value wins only when it holds one *and* the config carries a ClusterConfiguration, otherwise the cluster is the authority. Standalone `bootstrap-phase install-deckhouse` used to fail with `Cluster UUIDs are not equal in the cluster (X) and in the cache ()`.
- `ClusterUUIDConfigMap` now carries `heritage: deckhouse`, matching what the Deckhouse global hook creates — dhctl's full `Update` used to strip it.

**Destroy / converge**

- Static abort cleans every control-plane host recorded in the cache, reconnecting per host with the user-passed credentials. On main the abort path left the target list empty, so a multi-master static bootstrap ran the cleanup script N times against one host and never touched the others. New hard failures: a cache naming hosts the session is not connected to is refused ("belongs to another cluster"), as is an empty cache with no connected hosts.
- `dhctl destroy` sets the tracker's cluster config before `CheckCommanderUUID`, the first call that can announce a phase, so every progress frame carries the real phase list from the first frame. Side effect: a commander-UUID mismatch is now detected after the cluster config has been fetched and written to the cache.
- `ConvergeMigration` now calls `SetClusterConfig`; both converge entry points extract the dhctl state after `cache.InitWithOptions` instead of before, where `cache.Global()` was still a DummyCache and always yielded nil.

**CLI**

- `dhctl bootstrap` gains `--skip-phase` (repeatable, `DHCTL_CLI_SKIP_PHASES`) and `--kubeconfig` / `--kubeconfig-context` / `--kube-client-from-cluster`. Skip names are validated in a kingpin PreAction against the ungated tree (unknown name, `Preparation`/`PreInfraPreflights`, `Phase/SubPhase` paths are refused) and again inside the walk for phases the cluster's gates exclude — before any infrastructure is created.
- Because `dhctl bootstrap` now always passes `WithKubeConfig`, `resolveKubeConfig` returns those values unconditionally and lib-connection's own flag parser is no longer reached; it falls back to the bare `KUBECONFIG` environment variable. On a static bootstrap `KubeProvider` is never replaced later (only the cloud-gated `bootstrapFirstMaster` replaces it), so a static bootstrap run with `KUBECONFIG` exported installs into the SSH-tunnelled cluster.
- The state-cache flag group is now defined on `converge`, `converge-periodical`, `converge-migration`, `terraform check` and `bootstrap-phase create-resources`, where the operations layer already read those values but nothing could set them.
- `dhctl config render {bashible-bundle,master-bootstrap-scripts,control-plane-manifests}` keep refusing a config with no ClusterConfiguration, now with a message naming the command and the reason.
- **Removed:** `dhctl bootstrap-phase execute-bashible-bundle` together with its required `--internal-node-ip` and `--device-path` flags — it was reachable but broken on main (it built `BashiblePipelineParams` with no `PhasedExecutionContext` and the pipeline dereferenced that nil interface immediately, regression #20102); and `dhctl bootstrap-phase exec-post-bootstrap`, which worked but was help-texted as a test aid ("Test scp upload and ssh execution of the uploaded script").
- **Changed contract:** `bootstrap-phase abort` no longer destroys a running cluster. It used to read the `tf-state-and-manifests-in-cluster` cache key and, when Deckhouse manifests already existed, build a full `ClusterDestroyer` that connected over the Kube API and deleted in-cluster resources before tearing down infrastructure. It now always aborts from cache with a kube provider that refuses — for gRPC clients too, since the Abort RPC has no force field and always took that branch. Use `dhctl destroy` to delete a running cluster. `--force-abort-from-cache` still parses (dhctl-e2e passes it on every abort and an unknown flag is a parse error) but is now a no-op, and the cache key `tf-state-and-manifests-in-cluster` plus `State.SaveManifestsCreated`/`IsManifestsCreated`/`PostBootstrapScriptResult` are gone. Existing caches may still hold that file; nothing reads it.

**gRPC / config / preflight**

- `BootstrapStartOptions` gains `repeated string skip_phases = 6` — additive and wire-compatible; old clients omit it, old servers ignore it. Names are validated at operation start, not at request time, because the gate inputs are produced by `Preparation` itself. No gRPC field was removed or renumbered.
- `config.LoadConfigFromFile` no longer requires a ClusterConfiguration. `MetaConfig.HasClusterConfiguration()` is the new predicate for "dhctl did not create this control plane"; the registry CRI check, `CidrIntersectionCheck` and `PublicDomainTemplateCheck` skip work that has no input, and the refusal is re-added where it genuinely applies (render commands, abort).
- `ValidateClusterSettingsChanges` accepts `FirstMaster` alongside `BaseInfra`, since master-0 creation is now its own top-level phase — the permitted window in the bootstrap timeline is unchanged.

**Tests**

- ~4,240 lines across 20 new and 10 modified `_test.go` files, plus the `eks_without_cluster_configuration.yaml` fixture. They drive the real walk, the real gates, `cacheIdentity`, `resolveClusterUUID`, `initCache`, `Destroyer.destroyCluster` and `consumeProgress`, and assert ordering, error text, wire JSON and filesystem side effects. The projected JSON for 12 operation/cluster-type combinations is frozen, so moving the wire has to be a deliberate step.
- Not covered: the `converge/converger.go` reordering, and the gRPC `skip_phases` plumbing.

**Impact on critical components**

No impact on running cluster components (no restarts of ingress-controllers, control-plane, or Prometheus) — dhctl is an installer CLI and gRPC server and does not run in the cluster. The affected surfaces are:

- **Persisted state-cache directories.** Static and managed clusters move to `ssh-<hash>` / `kubeconfig-<hash>` directories. An in-flight bootstrap started by an older dhctl and resumed or aborted by this one depends on the `.owner` adoption path landing on the legacy directory; if it does not, a resume finds an empty cache and an abort refuses with "No UUID found in the cache" (the message now explains that a static abort must repeat the bootstrap's `--ssh-host` list).
- **gRPC / progress-log consumers (Commander, CI parsing `--progress-log-file-path`).** The `phases` array content changed for every operation and the progress denominators moved with it; the struct shape (field names, types, order, `omitempty`) and the `Action` string set are unchanged, so parsing does not break. A consumer with a hardcoded phase list, a hardcoded count, or an expectation about which phase is last needs updating. Whether Commander tolerates the new names and the two relocated sub-phase strings cannot be determined from this repo.
- **A completed static bootstrap re-run after upgrade.** The old directory is left holding only `.tombstone`, which is not adoptable, so a re-run opens a fresh directory and the "cache marked as exhausted" refusal no longer fires; the run is instead stopped by the cluster-UUID check inside `InstallDeckhouse`. This gap exists only for the first re-run after upgrading.

Bootstrap kept two hand-maintained lists — one of what runs, one of what is announced — and they had drifted. The concrete cost, all visible in this diff:

- Phases that ran but were never declared had their progress events silently dropped, because an undeclared name resolves to index `-1` and the tracker returns the record unchanged. That covered `Preparation` (a commented-out block on main, so a bootstrap that died during config load or cache init reported no phase at all), `ModulesPreparation`, and four destroy phases.
- Sub-phases whose gating differed from their parent left `CurrentPhase` pointing at the wrong node.
- A successful static destroy reported progress 0.0 and an empty completed phase, because a nested `Run` leaked the inner phase out and that name was undeclared.
- A standalone `bootstrap-phase` command that was a near-copy of a phase body drifted from it (no extension validation, no CNI bootstrap, no pipeline), and one of them panicked on every invocation.

Making the tree the single source for both execution and the wire removes the drift by construction: a declared node with no implementation fails at run start, and the projected JSON is frozen in tests.

The other changes were reachable only once that was in place. Positive gates mean an unresolved cluster type stops reading as "cloud". A second gating axis plus a loader that no longer demands a ClusterConfiguration makes a bootstrap into a cluster dhctl did not create expressible at all. Per-cluster cache identities fix a real collision — every static cluster on a machine shared one directory, one UUID and one tombstone, and meeting that tombstone crashed with SIGSEGV rather than explaining itself. And abort stopped straddling two jobs: rolling back what dhctl provisioned, from cache and without the Kube API, is now its whole contract; deleting a running cluster is `dhctl destroy`.

Not necessarily. The branch does fix several real defects (the failed-cache-init nil dereference, the static abort cleaning only one of N masters, the dropped progress events, `Complete()` naming a phase that never ran), but they arrive on top of a restructuring that changes the phase list every gRPC consumer sees and moves the state-cache directory for static clusters. Backporting the fixes in isolation would mean a different, much smaller change.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: feature
summary: Bootstrap phases are driven by one declarative tree, with `--skip-phase`, per-cluster state-cache identities, and support for clusters without ClusterConfiguration; `bootstrap-phase abort` no longer destroys a running cluster.
impact: |
  `dhctl bootstrap-phase abort` now always rolls back from the state cache and never deletes in-cluster resources, for CLI and gRPC clients alike. Use `dhctl destroy` to delete a running cluster. `--force-abort-from-cache` still parses but has no effect.

  The `bootstrap-phase` subcommands `execute-bashible-bundle` and `exec-post-bootstrap` are removed, along with `--internal-node-ip` and `--device-path`.

  The announced phase list changed for every operation (bootstrap gains 5 top-level phases, destroy gains 4, converge loses `ScaleToMultiMaster`), and `FirstMaster` / `WaitForControlPlaneManagerReadiness` moved from sub-phases to top-level phases. The gRPC message shape is unchanged; consumers with a hardcoded phase list or count need updating.

  Static and managed clusters now name their state-cache directory after the SSH host list instead of sharing one directory per machine. A static `bootstrap-phase abort` must be given the same `--ssh-host` set as its bootstrap.
impact_level: high
```


deckhouse-BOaTswain:
:red_circle: `e2e: DVP` for `deckhouse:dhctl/phase-state-clump` [failed](https://github.com/deckhouse/deckhouse/actions/runs/31942901241) in 2m47s.




  `/e2e/abort dvp;WithoutNAT;containerd;1.34 31942901241 31942901241-1-con-1-34 /sys/deckhouse-oss/install:pr22182 ` - for abort failed cluster




:red_circle: `e2e: DVP, Containerd, Kubernetes 1.34` failed in 1m26s.

</details>

deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:dhctl/phase-state-clump` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/32021930183) in 28m43s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.2.44`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 27m26s.

</details>

deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:dhctl/phase-state-clump` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/32356031335) in 27m10s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.5.0`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 25m55s.

</details>

borg-z (approved):

PR #22189, opened 2026-08-17
[deckhouse] Make Module v1alpha2 primary and fill it from its sources

ipaqsa:
## Description

First step of the Module v2 migration. `v1alpha2` becomes the main version of the Module resource, and Deckhouse now keeps it filled with real data: package version, origin and config.

### How it works

**CRD**

- `v1alpha2` is now served and the storage version; `v1alpha1` stays served.
- No conversion webhook: both views show the same object, and the v2 spec passes through the v1alpha1 view untouched.

**At startup - new package `internal/modulesync`**

- `Sync()` runs first in `Start`, before any controller. It fails fast: no sync - no start.
- It resolves one origin per module. Priority: Deckhouse image (embedded) > pull override > newest deployed release.
- It demotes duplicate Deployed releases of one module to Superseded.
- It mirrors the ModuleConfig fields in the same pass.
- Writes are merge patches; a module that did not change gets no write.

**At runtime - the three old-stack controllers mirror their events**

- config controller: mirrors the ModuleConfig on every reconcile. It mirrors the spec as stored, before schema conversion.
- override controller: mirrors the MPO on reconcile and again every scan interval, so drift heals itself.
- release controller: mirrors a deployed release on reconcile and right after a deploy. A pull override always wins over a release.

### Field map

| Source object | Source field | Module v1alpha2 field | Written at |
|---|---|---|---|
| Deckhouse image | `app.Version` (ldflags) | `spec.packageVersion` | startup |
| Deckhouse image | constant `embedded` | `spec.packageRepositoryName` | startup |
| Deckhouse image | module ships in the image | `metadata.annotations["modules.deckhouse.io/embedded"]` | startup |
| ModulePullOverride | `spec.imageTag` | `spec.packageVersion` | startup, override controller |
| ModulePullOverride | `Module(v1alpha1).properties.source` | `spec.packageRepositoryName` | startup, override controller |
| ModulePullOverride | the override is Ready | `metadata.annotations["modules.deckhouse.io/dev"]` | startup, override controller |
| ModuleRelease (Deployed) | `spec.version` | `spec.packageVersion`, as `"v"` + semver | startup, release controller |
| ModuleRelease (Deployed) | `metadata.ownerReferences[ModuleSource].name`, fallback `metadata.labels["source"]` | `spec.packageRepositoryName` | startup, release controller |
| ModuleConfig | `spec.source` | `spec.packageRepositoryName` | startup, config controller |
| ModuleConfig | `spec.enabled` | `spec.enabled` | startup, config controller |
| ModuleConfig | `spec.settings` | `spec.settings` | startup, config controller |
| ModuleConfig | `spec.version` | `spec.settingsVersion` | startup, config controller |
| ModuleConfig | `spec.updatePolicy` | `spec.updatePolicy` | startup, config controller |
| ModuleConfig | `spec.maintenance` | `spec.maintenance` | startup, config controller |

**Who wins when two sources describe the same module**

- Version and origin: the image beats a pull override, and a pull override beats a deployed release. One origin per module.
- Repository: a ModuleConfig naming a source overrides that origin, because the name is the user's choice while the origin only reports where the package came from last. An embedded module keeps `embedded` - it ships in the image and no repository serves it.

**Other rules**

- Config fields are written only when the module already has `spec.packageVersion`: a config alone says nothing about where the package comes from, so it must not materialize the spec.
- When the version arrives on an empty module, the config fields are seeded in the same write.
- Every origin also sets `metadata.name`: the definition name, the MPO name, or `ModuleRelease.spec.moduleName`. A config alone never creates a module.
- `spec.releaseChannel` is in the schema but nothing writes it yet.

### Left out on purpose

- `ModuleRelease.metadata.labels["modules.deckhouse.io/update-policy"]` is not mirrored. `spec.updatePolicy` comes from the ModuleConfig only.
- A module with a config but no origin (never downloaded, not embedded) gets no mirror at all.
- Deletes do not clean up: removing a ModuleConfig keeps the mirrored fields, removing an MPO keeps the `dev` annotation. The old flow behaved the same way.
- Everything else on the source objects stays where it is. `properties` and `status` are written by the old stack, and the mirror never touches them.

### Before / After

**Before:** the Module CRD serves only `v1alpha1`, and it is the storage version; the `v1alpha2` view does not exist on a regular cluster.

**After:** every module with a known origin carries its version, origin annotations and config in the v2 spec; `v1alpha2` is the storage version.

### Tests

- `internal/modulesync`: unit tests for the sync and the mirror entry points (origin priority, config seeding, the `dev` annotation survives later writes).
- `TestFillModuleV2`: the config source wins over the origin, an empty source keeps it, and an embedded module stays on `embedded`.
- config controller: golden tests for enable/disable mirroring, plus `TestMirrorSeesStoredConfig` - the mirror writes the config as stored, not the converted one.
- release controller: golden files now include the Module v2 object for every release scenario.
- override controller: no unit tests, its mirror is verified live only.
- Checked live on a dev cluster: the filled Module v2 resources appear at startup; ModuleConfig edits, MPO create/delete and deployed releases all mirror correctly.

- Module v1 only shows state; the config lives in a separate ModuleConfig.
- In the target model everything about a module lives in the Module resource itself.
- Later stages switch the module engine to the package runtime and deprecate ModuleConfig and ModulePullOverride. They need v2 objects that already carry real data.
- The engine switch can only happen after v2 objects exist and are filled. This PR is that groundwork.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Introduce modulev2.
impact_level: low
```


deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:feat/modulev2-crd-controllers` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/32475882037) in 25m3s.



GCP-WithoutNAT-Containerd-1.34 - Connection string: `ssh  user@34.141.2.135`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.34` succeeded in 23m48s.

</details>

ipaqsa:
We can get rid of sync.go if its not needed and move load function to controller.go

deckhouse-controller/pkg/apis/deckhouse.io/v1alpha2/module.go
```diff
@@ -76,8 +81,19 @@ type Module struct {
...
+	// +optional
+	Spec ModuleSpec `json:"spec,omitempty"`
+
+	// Properties carries the v1alpha1 `properties` of the same stored object
+	// verbatim. The versions share storage without conversion, so this
+	// passthrough is required: a v1alpha2 read or full update would otherwise
+	// drop the v1alpha1 fields.
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +optional
+	Properties *runtime.RawExtension `json:"properties,omitempty"`
```
ipaqsa, line 99: Why did you add it ?

PR #22202, opened 2026-08-17
[deckhouse] fix package yaml parsing

ipaqsa:
## Description

Switches the three package-manifest parsers from `sigs.k8s.io/yaml` to `gopkg.in/yaml.v3`:

| File | Parses |
| --- | --- |
| `deckhouse-controller/internal/packages/loader/loader.go` | on-disk `package.yaml` / legacy `module.yaml` |
| `deckhouse-controller/pkg/controller/packages/module-package-version/parser.go` | `package.yaml` / `module.yaml` from the registry image |
| `deckhouse-controller/pkg/controller/packages/application-package-version/parser.go` | `package.yaml` from the registry image |

Also refreshes one golden file whose expected condition message contained the
`error converting YAML to JSON:` prefix that `sigs.k8s.io/yaml` used to add.

No API type, CRD or schema changes — `Module.requirements.bootstrapped` stays `type: string`.

A module whose `module.yaml` writes a YAML boolean for a string-typed requirement — the common
case being `requirements.bootstrapped: true` — fails metadata parsing. The
`ModulePackageVersion` never becomes ready, and its `MetadataLoaded` condition carries:

```
fetch package metadata: decode module.yaml: error unmarshaling JSON: while decoding JSON:
json: cannot unmarshal bool into Go struct field
ModuleRequirements.requirements.ModulePlatformRequirements.bootstrapped of type string
```

The module is simply not installable via the package runtime.

### Root cause

`sigs.k8s.io/yaml` is not a YAML decoder — it converts YAML to JSON and delegates to
`encoding/json`, which is strictly typed and will not put a JSON `true` into a `string` field.

It does ship a compensating coercion (`yaml.go:329-355`): when the JSON target field is a
`reflect.String`, an int / int64 / float64 / uint64 / **bool** scalar is stringified before the
JSON decode. **That coercion only fires for fields declared directly on the struct being
decoded — it does not follow anonymous-embedded promotion.** Reproduced against
`sigs.k8s.io/yaml v1.6.0`:

| target shape | `bootstrapped: true` |
| --- | --- |
| `struct{ Bootstrapped string }` | coerced to `"true"` |
| `struct{ Requirements Flat }` | coerced to `"true"` |
| `struct{ Plat }` (embedded) | `cannot unmarshal bool` |
| `struct{ Requirements InlineTagged }` | `cannot unmarshal bool` |

`ModuleRequirements` embeds `ModulePlatformRequirements`, and `Bootstrapped` lives on the
embedded struct (`pkg/apis/deckhouse.io/v1alpha1/module.go:160-172`) — so we land in the
broken column. `deckhouse:` and `kubernetes:` are string fields on the same embedded struct
and never tripped this, only because version constraints are always written quoted and are
therefore already YAML strings.

`gopkg.in/yaml.v3` decodes YAML nodes straight into Go values and resolves a `!!bool` scalar
to `"true"` whenever the destination is a `string`, at any nesting or embedding depth.

### Why yaml.v3 rather than a schema change or a local workaround

- **It removes a divergence instead of adding one.** Everything else in the repo that parses
  these same files already uses `gopkg.in/yaml.v3` — `pkg/controller/moduleloader/loader.go`
  (the addon-operator path) and `pkg/controller/module-controllers/downloader/downloader.go`
  (the `ModuleRelease` path). The three package-runtime parsers were the outliers, which is
  exactly why a `module.yaml` that loads fine on the legacy path fails on the package path.
- **No field changes binding.** `sigs.k8s.io/yaml` reads `json:` tags, yaml.v3 reads `yaml:`
  tags. Every field on `dto.*`, `moduleloader/types.Definition` and the nested `v1alpha1`
  requirement types (`ModuleRequirements`, `ModulePlatformRequirements`, `ModuleDisableOptions`,
  `ModuleAccessibility`, `ModuleDisableMessages`) already carries both tags — verified, zero
  `json:`-only fields in the reachable graph.
- **Behaviour-preserving for the value itself.** `Requirements.Bootstrapped` is stored but never
  read as a value anywhere in the tree; the bootstrapped extender keys off the `critical` flag
  (`go_lib/dependency/extenders/extenders.go:110-113` → `bootstrapped.AddFunctionalModule`).
  Coercing `true` to `"true"` changes no enablement decision.

## Behaviour changes worth knowing

Verified by direct comparison of the two libraries; all are consequences of decoding YAML
natively rather than through JSON:

1. **Error messages lose the `error converting YAML to JSON: ` prefix** and gain yaml.v3's
   `line N:` positions. This is what the golden-file update in this PR covers, and it affects
   the `MetadataLoaded` condition message users see on a malformed manifest.
2. **Duplicate mapping keys are now an error.** `sigs.k8s.io/yaml` silently took last-wins;
   yaml.v3 reports `mapping key "x" already defined at line N`. Stricter, and aligned with the
   legacy loader.
3. **Key matching is now case-sensitive.** `encoding/json` matched keys case-insensitively, so a
   manifest with `Name:` or `exclusivegroup:` used to bind; under yaml.v3 it does not. A
   manifest relying on that would go from silently-working to silently-empty. Correctly-cased
   manifests — every one in-tree — are unaffected.
4. **YAML 1.1 vs 1.2 scalars.** `sigs.k8s.io/yaml` (via goyaml.v2) resolves `yes`/`no`/`on`/`off`
   as booleans; yaml.v3 keeps them as the literal strings. So `name: yes` used to decode as
   `"true"` and now decodes as `"yes"` — closer to author intent.

Yes — affected modules cannot be installed through the package runtime at all, and the failure
is in manifest parsing, so no cluster-side change works around it.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Accept YAML booleans in string-typed package manifest requirements.
impact_level: low
```


ldmonster (approved):

deckhouse-BOaTswain:
Cherry pick PR [22282](https://github.com/deckhouse/deckhouse/pull/22282) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22214, opened 2026-08-17
[user-authn] Quote DexProvider-derived values in rendered manifests

Jabejixo:
## Description

Two templates of the `user-authn` module interpolated string fields of the cluster-scoped `DexProvider` resource into rendered YAML without quoting. Every such string-typed interpolation is now rendered as a quoted scalar, reusing the idiom that both files already used for their neighbouring fields.

### 1. `templates/basic-auth-proxy/deployment.yaml`

Values were interpolated into the `args:` list of the `proxy` container. The existing idiom for `--ldap-client-secret` is reused:

```
- {{ printf "--oidc-scope=%s" $scope | quote }}
```

| Argument | Source field |
| --- | --- |
| `--crowd-application-login` | `spec.crowd.clientID` |
| `--crowd-application-password` | `spec.crowd.clientSecret` |
| `--crowd-base-url` | `spec.crowd.baseURL` |
| `--crowd-allowed-group` | `spec.crowd.groups[]` |
| `--oidc-base-url` | `spec.oidc.issuer` |
| `--oidc-client-id` | `spec.oidc.clientID` |
| `--oidc-client-secret` | `spec.oidc.clientSecret` |
| `--oidc-scope` | `spec.oidc.scopes[]` |
| `--ldap-scope` | `spec.ldap.scopes[]` |

### 2. `templates/dex/config.yaml`

Values were interpolated into the connector blocks of the Dex configuration file. The existing idiom for `bindDN`, one line above `bindPW`, is reused:

```
bindDN: {{ $provider.ldap.bindDN | quote }}
```

| Key | Source field | Previous form |
| --- | --- | --- |
| `bindPW` | `spec.ldap.bindPW` | single-quoted |
| `claimMapping.email` | `spec.oidc.claimMapping.email` | unquoted |
| `claimMapping.groups` | `spec.oidc.claimMapping.groups` | unquoted |
| `claimMapping.preferred_username` | `spec.oidc.claimMapping.preferred_username` | unquoted |

A single-quoted YAML scalar is terminated by a `'` in the value, so `bindPW` was affected by a quote as well as by a line break.

### Fields deliberately left alone

A full pass over both templates found no other unsafe interpolation. The following were checked and excluded:

* `--oidc-get-user-info`, `--oidc-basic-auth-unsupported`, `--ldap-basic-auth-unsupported`, `saml.filterGroups`, `saml.insecureSkipSignatureValidation` and the remaining `default false` flags — the corresponding CRD fields are `type: boolean`, so they cannot carry a line break, and quoting them would turn the rendered value into a string that Dex would reject.
* `github.teamNameField` — `type: string`, but constrained by `enum` in the CRD (`['name', 'slug', 'both']` in `v1alpha1`, `['Name', 'Slug', 'Both']` in `v1`), so no other value can be admitted. Left unquoted to keep the existing `| lower` rendering intact.
* `ldap.rootCAData`, `gitlab.rootCAData`, `saml.rootCAData` — piped through `b64enc`, whose output cannot leave the scalar.
* `ldap.userSearch`, `ldap.groupSearch` — rendered with `toYaml`, which quotes as needed.

No behaviour changes for existing configurations: single-line values render exactly as before.

### 3. One schema constraint, and why only one

The templates above close the injection. A third commit additionally constrains `spec.oidc.scopes` items with `pattern: '^\S+$'`, because that constraint follows from the protocol rather than from this defect: OAuth2 passes scopes as a space-delimited list, so a scope containing whitespace cannot work at all, and rejecting it at admission is better than rendering a value that silently breaks authentication. The `oidc` schema is shared between `v1alpha1` and `v1` through the `&oidc` / `*oidc` YAML anchor, so the constraint applies to both served versions and cannot drift apart; that was verified by parsing the CRD.

No other `DexProvider` field gets a schema constraint, and that is a deliberate decision rather than an omission:

* Adding `pattern: '^[^\r\n]*$'` across the roughly sixty unconstrained string fields would **break** `ldap.rootCAData`, `gitlab.rootCAData`, `oidc.rootCAData` and `saml.rootCAData`, whose values are PEM blocks and therefore multi-line by nature.
* The group-shaped lists — `oidc.allowedGroups`, `saml.allowedGroups`, `crowd.groups`, `gitlab.groups`, `bitbucketCloud.teams` — hold group names coming from an external identity provider, which legitimately contain spaces, slashes and non-ASCII characters. A whitespace ban there would reject configurations that work today.
* For everything else the marginal value is low: `DexProvider` is cluster-scoped, so writing it already requires authority over the module's identity configuration, and the rendering path is now quoted regardless of the value.

The same reasoning is why no unit test accompanies the schema change: a test asserting that a declarative `pattern` equals the string it was just set to proves nothing, and the cross-version invariant is guaranteed structurally by the anchor rather than by a hand-maintained duplicate.

Both sinks let a `DexProvider` author inject arbitrary YAML into a rendered manifest.

In the Deployment, a field containing a line break terminated the `args:` block sequence and the remaining lines were parsed as sibling keys of the `proxy` container, allowing `image`, `command`, `args` and `workingDir` of a system container to be replaced. The pod mounts the `basic-auth-cert` secret, so controlling that container matters.

In the Dex configuration the injected YAML lands in Dex's own configuration file, so it can add keys to a connector or append an entire connector — for example enabling `insecureNoSSL` on an LDAP connector, or changing how identities are mapped. Dex mints the identities the kube-apiserver trusts, so silently altering its configuration from a custom resource is worth closing even though it grants nothing beyond what write access to `DexProvider` already implies.

Both paths are covered by template tests that fail against the previous templates and pass with this change:

* `template_tests/basic_auth_proxy_test.go` renders the Deployment with a multi-line value in each of the nine affected fields and asserts the container keeps its own image, gains no `command` or `workingDir`, and that the value stays a single argv element. It also pins the boolean flags to their unquoted form.
* `template_tests/connector_quoting_test.go` decodes the rendered Dex configuration out of its Secret, parses it back as YAML, and asserts that values containing both a quote and a line break round-trip as single scalars, that no extra connector keys appear, and that the connector list stays the expected length.

This is a hardening fix for a manifest-rendering issue that is reachable from a custom resource, so it should be backported to 1.77 and 1.76.

All three commits cherry-pick onto `release-1.77` and `release-1.76` with no conflicts. `basic-auth-proxy/deployment.yaml` is byte-identical on all three branches; `dex/config.yaml` differs only by lines added elsewhere in the file on `main`, so the change applies by context. For the schema commit, `crds/dex-provider.yaml` and `crds/doc-ru-dex-provider.yaml` diverge from `main` on `release-1.76`, but the `scopes` block and its surrounding context are byte-identical on all three branches and the `&oidc` anchor is present on each, so that hunk applies by context too.

On `release-1.77` the whole `template_tests` suite passes as-is. On `release-1.76` the templates and the new Dex configuration test apply and pass unchanged; only `basic_auth_proxy_test.go` needs `userAuthn.publishAPI.*` instead of `userAuthn.internal.publishAPI.*`, because that value moved between the two releases, and with those two paths adjusted the suite passes there as well. That two-line adaptation is the only reason `release-1.76` needs a dedicated backport pull request rather than a straight pick.

## Checklist
- [ ] e2e tests passed.

The `scopes` constraint is described in the CRD `description` in both languages, which is where the reference documentation for this field is generated from. Any further client-facing documentation for this series of changes is handled in a separate documentation pull request, so that the documentation team reviews it in one place.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Quote DexProvider-derived values rendered into the basic-auth-proxy Deployment and the Dex configuration file to prevent YAML injection.
impact_level: default
---
section: user-authn
type: fix
summary: Reject whitespace in DexProvider OIDC scopes, which the OAuth2 protocol cannot represent.
impact: |
  `spec.oidc.scopes` items are now validated against `^\S+$` in both served versions of DexProvider. A scope containing whitespace could never work, because OAuth2 passes scopes as a space-delimited list, but it was previously admitted and silently broke authentication. Stored objects are not invalidated, however a DexProvider holding such a value is rejected on its next write until the value is corrected, which also affects a GitOps flow that reapplies the object unchanged.
impact_level: high
```

### DIFF
--- a/modules/150-user-authn/crds/dex-provider.yaml
+++ b/modules/150-user-authn/crds/dex-provider.yaml
@@ -303,8 +303,11 @@ spec:
                       default: ["openid", "profile", "email", "groups", "offline_access"]
                       description: |
                         List of [additional scopes](https://github.com/dexidp/website/blob/main/content/docs/configuration/custom-scopes-claims-clients.md) to request in token response.
+
+                        A scope is a single token and must not contain whitespace: the OAuth2 protocol passes scopes as a space-delimited list.
                       items:
                         type: string
+                        pattern: '^\S+$'
                     allowedGroups:
                       type: array
                       description: |
--- a/modules/150-user-authn/crds/doc-ru-dex-provider.yaml
+++ b/modules/150-user-authn/crds/doc-ru-dex-provider.yaml
@@ -209,6 +209,8 @@ spec:
                     scopes:
                       description: |
                         Список [полей](https://github.com/dexidp/website/blob/main/content/docs/configuration/custom-scopes-claims-clients.md) для включения в ответ при запросе токена.
+
+                        Scope — это одиночный токен, он не может содержать пробельные символы: протокол OAuth2 передаёт scope'ы списком, разделённым пробелами.
                     promptType:
                       description: |
                         Определяет — должен ли Issuer запрашивать подтверждение и давать подсказки при аутентификации.
--- a/modules/150-user-authn/template_tests/basic_auth_proxy_test.go
+++ b/modules/150-user-authn/template_tests/basic_auth_proxy_test.go
@@ -1,0 +1,167 @@
+/*
+Copyright 2026 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package template_tests
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	. "github.com/deckhouse/deckhouse/testing/helm"
+)
+
+// multilineValue is a DexProvider field value spanning several lines, indented to match the
+// `args:` block of the proxy container. Unless it is rendered as a quoted scalar, the newline
+// ends the args sequence and the remaining lines become sibling keys of that container.
+const multilineValue = "value\n" +
+	"        image: registry.example.invalid/injected\n" +
+	"        command: [\"/bin/sh\", \"-c\"]\n" +
+	"        args: [\"injected\"]\n" +
+	"        workingDir: /injected"
+
+const basicAuthProxyImage = "registry.example.com@imageHash-userAuthn-basicAuthProxy"
+
+const crowdProvider = `
+- id: crowdID
+  displayName: crowdName
+  type: Crowd
+  crowd:
+    enableBasicAuth: true
+    clientID: clientID
+    clientSecret: secret
+    baseURL: https://example.com
+`
+
+const oidcProvider = `
+- id: oidcID
+  displayName: oidcName
+  type: OIDC
+  oidc:
+    enableBasicAuth: true
+    issuer: https://example.com
+    clientID: clientID
+    clientSecret: secret
+`
+
+const ldapProvider = `
+- id: ldapID
+  displayName: ldapName
+  type: LDAP
+  ldap:
+    enableBasicAuth: true
+    host: ldap.example.com:636
+    userSearch:
+      baseDN: cn=users,dc=example,dc=com
+      username: uid
+      idAttr: uid
+      emailAttr: mail
+`
+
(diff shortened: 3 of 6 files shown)


AlwxSin (approved):

z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [22354](https://github.com/deckhouse/deckhouse/pull/22354) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [22355](https://github.com/deckhouse/deckhouse/pull/22355) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #22216, opened 2026-08-17
[user-authn] Issue a dedicated OAuth2 client secret per DexAuthenticator

Jabejixo:
## Description

Two internal-only changes in `modules/150-user-authn`. No CRD/API change.

**1. Stop copying the shared `kubernetes` client secret into application namespaces.**

`templates/dex-authenticator/secret.yaml` rendered `client-secret` from
`userAuthn.internal.kubernetesDexClientAppSecret` whenever the authenticator carried the
`dexauthenticator.deckhouse.io/allow-access-to-kubernetes` annotation. That value is the single secret of the
privileged `kubernetes` OAuth2 client, also used by `templates/dex/secret.yaml`,
`templates/kubeconfig-generator/oauth2client.yaml`, `templates/dex/kubeconfig-oauth2clients.yaml`,
`templates/dex/kubernetes-dex-client-configuration.yaml` and `templates/basic-auth-proxy/deployment.yaml`.
The Secret is created in `{{ $crd.namespace }}`, so anyone able to read Secrets in an application namespace
obtained a cluster-level credential. The template now always renders the per-authenticator
`credentials.appDexSecret`.

Because `hooks/get_dex_authenticator_crds.go` reads `client-secret` back out of that same Secret into
`credentials.appDexSecret` (`applyDexAuthenticatorSecretFilter`), on clusters where the annotation was already
in use `appDexSecret` *is* the shared secret, and dropping the template branch alone would not rotate anything.
The hook therefore regenerates `appDexSecret` once when it equals the current secret of the `kubernetes`
client, reusing the existing `go_lib/pwgen` generator (`pwgen.AlphaNum(20)`) already used for `appDexSecret`
and `cookieSecret`.

The comparison value is read from the hook's own `kubernetesClientSecret` snapshot of
`d8-user-authn/kubernetes-dex-client-app-secret`, reusing `applyKubernetesSecretFilter` from
`hooks/generate_kubernetes_dex_client_app_secret.go`. It is deliberately not read from
`userAuthn.internal.kubernetesDexClientAppSecret`: that values path is written by
`kubernetesDexClientAppSecret`, which runs in the `main` queue, while this hook declares
`Queue: "/modules/user-authn"`, and `ModuleRun` queues the Synchronization tasks of the two queues
independently, so nothing orders them on a cold start. The cluster Secret is also the authoritative source,
being the object whose content was copied out. An absent Secret now means the shared secret does not exist, so
nothing can be carrying it. Every rotation is logged with `input.Logger.Warn`, since it rolls the
authenticator's pods.

**2. Remove dead tenant-controlled input from the privileged client.**

`templates/dex/oauth2client.yaml` appended `https://{{ $crd.spec.applicationDomain }}/dex-authenticator/callback`
to the `redirectURIs` of the `kubernetes` client itself. The hook watches `deckhouse.io/v2alpha1`, whose schema
has `spec.applications[].domain` and no `spec.applicationDomain` (`crds/dex-authenticator.yaml`), so the entry
always rendered as the broken literal `https:///dex-authenticator/callback`. The block is removed, and the
`redirectURIs` key now renders as `[]` instead of a bare key when the client exists only to hold the
`trustedPeers` of annotated authenticators.

### Verification that matching client secrets are not required

Cross-client audience issuance in Dex depends only on `trustedPeers` membership, never on a matching client
secret. In dex v2.44.0 (the version pinned in `modules/150-user-authn/oss.yaml`),
`server/oauth2.go:634-652` `validateCrossClientTrust` loads the peer client and returns true purely on
`peer.TrustedPeers` containing the requesting `clientID`; `server/oauth2.go:321-342` `getAudience` then adds the
peer to the audience. The three call sites (`server/oauth2.go:426`, `server/oauth2.go:541`,
`server/handlers.go:1150`) pass only client IDs. `templates/dex/oauth2client.yaml` already lists
`<name>-<namespace>-dex-authenticator` in `trustedPeers`, and `templates/dex-authenticator/oauth2client.yaml:12`
already registers the authenticator's own client under `secret: {{ $crd.credentials.appDexSecret }}` — which is
exactly what `OAUTH2_PROXY_CLIENT_SECRET` in `templates/dex-authenticator/deployment.yaml:240-244` consumes,
paired with `--client-id={{ $crd.name }}-{{ $crd.namespace }}-dex-authenticator`. So the authenticator
authenticates as itself and the shared secret was never needed.

### Verification that the removed redirect URI breaks nothing

oauth2-proxy authenticates under its own `client_id` (see above), and Dex validates `redirect_uri` against the
client resolved from the request's `client_id` (`server/oauth2.go:481-492`: `GetClient(ctx, clientID)` followed
by `validateRedirectURI(client, redirectURI)`), i.e. against
`templates/dex-authenticator/oauth2client.yaml:25-28`, where the genuine callback lives. The removed entry was
never a valid URI anyway.

### Operational consequence

Regenerating `appDexSecret` changes the `checksum/config` annotation on the authenticator Deployment
(`templates/dex-authenticator/deployment.yaml:150`), so authenticator pods roll once and users of those
applications may have to authenticate again. The `cookieSecret` is not touched.

### Follow-up work

Admission control over who may set the `dexclient.deckhouse.io/allow-access-to-kubernetes` /
`dexauthenticator.deckhouse.io/allow-access-to-kubernetes` annotations, and a namespace allowlist for them, are
deliberately out of scope here — they are behaviour-breaking and handled separately.

It removes an over-shared credential: a cluster-level OAuth2 client secret was being distributed into
tenant-readable namespaces with no user interaction required to obtain it. It also removes the single place
where tenant-controlled input reached the privileged client's `redirectURIs` — harmless today only because the
field name is wrong, and a real defect the moment someone "repairs" it.

The over-sharing exists on any cluster that uses the annotation outside system namespaces, so it is worth
backporting to 1.77 and 1.76.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Issue a dedicated OAuth2 client secret per DexAuthenticator instead of reusing the shared cluster client secret.
impact: |
  On upgrade, every DexAuthenticator that carries the
  `dexauthenticator.deckhouse.io/allow-access-to-kubernetes` annotation gets a freshly generated client
  secret, because it previously received the client secret of the privileged `kubernetes` OAuth2 client
  instead. Its pods roll once and users of those applications may have to sign in again. Each rotation is
  written to the Deckhouse log as `Rotating DexAuthenticator client secret that reused the shared kubernetes
  client secret`, with the authenticator name and namespace.

  If that annotation was ever used outside the `d8-*` and `kube-*` namespaces, the shared secret must be
  considered disclosed to anyone who could read Secrets in those namespaces, and has to be rotated by hand.
  Deleting the Secret alone does not rotate it: while the value is non-empty in memory the owning hook
  re-renders the same one. Rotate it as follows.

  1. If a GitOps tool syncs the `d8-user-authn` namespace, exclude
     `Secret/kubernetes-dex-client-app-secret` from that sync first, otherwise it is restored to the old
     value.
  2. Blank the stored value:
     `kubectl -n d8-user-authn patch secret kubernetes-dex-client-app-secret --type merge -p '{"data":{"secret":""}}'`
  3. Drop the in-memory copy: `kubectl -n d8-system rollout restart deployment/deckhouse`
  4. Check that `kubectl -n d8-user-authn get secret kubernetes-dex-client-app-secret -o jsonpath='{.data.secret}'`
     now differs from the previous value. If it does not, the module re-rendered the old value before the
     restart took effect; repeat steps 2 and 3.

  The blast radius of that rotation is wider than the `kubernetes` client alone. The same value is the client
  secret of the `kubeconfig-generator`, `kubeconfig-publish-api` and every `kubeconfig-<slug>` OAuth2 client,
  and it is passed to basic-auth-proxy as `--ldap-client-secret`. In-cluster consumers are re-rendered and
  their pods roll automatically, but kubeconfig files already downloaded from the kubeconfig generator carry
  the old client secret and can no longer refresh their tokens, so their users must download a new
  kubeconfig. Already issued ID tokens keep working until they expire, at most `settings.idTokenTTL`.

  Deleting and recreating a DexAuthenticator is always a safe way to force a fresh per-authenticator secret:
  the credentials Secret is regenerated with the resource. Its content is owned by Deckhouse, so no GitOps
  change is needed for the per-authenticator rotation itself.
impact_level: high
---
section: user-authn
type: fix
summary: Drop the dead per-DexAuthenticator redirect URI from the privileged kubernetes OAuth2 client.
```

### DIFF
--- a/modules/150-user-authn/hooks/generate_kubernetes_dex_client_app_secret.go
+++ b/modules/150-user-authn/hooks/generate_kubernetes_dex_client_app_secret.go
@@ -29,6 +29,11 @@ import (
 	"github.com/deckhouse/deckhouse/go_lib/pwgen"
 )
 
+// kubernetesDexClientAppSecretPath holds the secret of the privileged kubernetes OAuth2Client.
+// It is shared by every consumer of the kubernetes client and must never be handed out to
+// per-application clients.
+const kubernetesDexClientAppSecretPath = "userAuthn.internal.kubernetesDexClientAppSecret"
+
 type KubernetesSecret []byte
 
 func applyKubernetesSecretFilter(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
@@ -61,7 +66,7 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 }, kubernetesDexClientAppSecret)
 
 func kubernetesDexClientAppSecret(_ context.Context, input *go_hook.HookInput) error {
-	secretPath := "userAuthn.internal.kubernetesDexClientAppSecret"
+	secretPath := kubernetesDexClientAppSecretPath
 	if input.Values.Exists(secretPath) && input.Values.Get(secretPath).String() != "" {
 		return nil
 	}
--- a/modules/150-user-authn/hooks/generate_kubernetes_dex_client_app_secret.go
+++ b/modules/150-user-authn/hooks/generate_kubernetes_dex_client_app_secret.go
@@ -29,10 +29,15 @@ import (
 	"github.com/deckhouse/deckhouse/go_lib/pwgen"
 )
 
-// kubernetesDexClientAppSecretPath holds the secret of the privileged kubernetes OAuth2Client.
-// It is shared by every consumer of the kubernetes client and must never be handed out to
-// per-application clients.
-const kubernetesDexClientAppSecretPath = "userAuthn.internal.kubernetesDexClientAppSecret"
+const (
+	// kubernetesDexClientAppSecretPath holds the secret of the privileged kubernetes OAuth2Client.
+	// It is shared by every consumer of the kubernetes client and must never be handed out to
+	// per-application clients.
+	kubernetesDexClientAppSecretPath = "userAuthn.internal.kubernetesDexClientAppSecret"
+
+	kubernetesDexClientAppSecretNamespace = "d8-user-authn"
+	kubernetesDexClientAppSecretName      = "kubernetes-dex-client-app-secret"
+)
 
 type KubernetesSecret []byte
 
@@ -54,11 +59,11 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 			Kind:       "Secret",
 			NamespaceSelector: &types.NamespaceSelector{
 				NameSelector: &types.NameSelector{
-					MatchNames: []string{"d8-user-authn"},
+					MatchNames: []string{kubernetesDexClientAppSecretNamespace},
 				},
 			},
 			NameSelector: &types.NameSelector{
-				MatchNames: []string{"kubernetes-dex-client-app-secret"},
+				MatchNames: []string{kubernetesDexClientAppSecretName},
 			},
 			FilterFunc: applyKubernetesSecretFilter,
 		},
@@ -66,8 +71,7 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 }, kubernetesDexClientAppSecret)
 
 func kubernetesDexClientAppSecret(_ context.Context, input *go_hook.HookInput) error {
-	secretPath := kubernetesDexClientAppSecretPath
-	if input.Values.Exists(secretPath) && input.Values.Get(secretPath).String() != "" {
+	if input.Values.Exists(kubernetesDexClientAppSecretPath) && input.Values.Get(kubernetesDexClientAppSecretPath).String() != "" {
 		return nil
 	}
 
@@ -81,14 +85,14 @@ func kubernetesDexClientAppSecret(_ context.Context, input *go_hook.HookInput) e
 
 		// if secret field was removed, generate a new one
 		if len(secretContent) == 0 {
-			input.Values.Set(secretPath, pwgen.AlphaNum(20))
+			input.Values.Set(kubernetesDexClientAppSecretPath, pwgen.AlphaNum(20))
 			return nil
 		}
 
-		input.Values.Set(secretPath, string(secretContent))
+		input.Values.Set(kubernetesDexClientAppSecretPath, string(secretContent))
 		return nil
 	}
 
-	input.Values.Set(secretPath, pwgen.AlphaNum(20))
+	input.Values.Set(kubernetesDexClientAppSecretPath, pwgen.AlphaNum(20))
 	return nil
 }
--- a/modules/150-user-authn/hooks/get_dex_authenticator_crds.go
+++ b/modules/150-user-authn/hooks/get_dex_authenticator_crds.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/flant/addon-operator/pkg/module_manager/go_hook"
 	"github.com/flant/addon-operator/sdk"
+	"github.com/flant/shell-operator/pkg/kube_events_manager/types"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -144,8 +145,47 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 			},
 			FilterFunc: applyDexAuthenticatorSecretFilter,
 		},
+		{
+			Name:       "kubernetesClientSecret",
+			ApiVersion: "v1",
+			Kind:       "Secret",
+			NamespaceSelector: &types.NamespaceSelector{
+				NameSelector: &types.NameSelector{
+					MatchNames: []string{kubernetesDexClientAppSecretNamespace},
+				},
+			},
+			NameSelector: &types.NameSelector{
+				MatchNames: []string{kubernetesDexClientAppSecretName},
+			},
+			FilterFunc: applyKubernetesSecretFilter,
+		},
 	},
 }, getDexAuthenticator)
+
+// sharedKubernetesClientSecret returns the secret of the privileged kubernetes OAuth2Client as it
+// exists in the cluster.
+//
+// It deliberately reads this hook's own snapshot instead of kubernetesDexClientAppSecretPath: that
+// values path is written by kubernetesDexClientAppSecret, which runs in the "main" queue while this
+// hook runs in "/modules/user-authn", so nothing orders the two and the value may still be absent
+// here on a cold start. The cluster Secret is also the authoritative source, since it is the object
+// whose content was copied into the application namespaces.
+//
+// An empty result means the shared secret does not exist in the cluster, so no authenticator can be
+// carrying it and there is nothing to migrate.
+func sharedKubernetesClientSecret(input *go_hook.HookInput) (string, error) {
+	snapshots := input.Snapshots.Get("kubernetesClientSecret")
+	if len(snapshots) == 0 {
+		return "", nil
+	}
+
+	var secretContent []byte
+	if err := snapshots[0].UnmarshalTo(&secretContent); err != nil {
+		return "", fmt.Errorf("cannot convert kubernetes client secret to bytes: failed to unmarshal 'kubernetesClientSecret' snapshot: %w", err)
+	}
+
+	return string(secretContent), nil
+}
 
 func getDexAuthenticator(_ context.Context, input *go_hook.HookInput) error {
 	authenticators := input.Snapshots.Get("authenticators")
@@ -166,7 +206,10 @@ func getDexAuthenticator(_ context.Context, input *go_hook.HookInput) error {
 	// Secret, and the "credentials" snapshot above reads it straight back into appDexSecret.
 	// Keep it to detect and rotate those, otherwise the shared secret would never leave the
 	// application namespace.
-	sharedClientSecret := input.Values.Get(kubernetesDexClientAppSecretPath).String()
+	sharedClientSecret, err := sharedKubernetesClientSecret(input)
+	if err != nil {
+		return err
+	}
 
 	dexAuthenticators := make([]DexAuthenticator, 0, len(authenticators))
 	// Build computed names map: key "<name>@<namespace>" => {name, truncated, hash}
@@ -190,9 +233,16 @@ func getDexAuthenticator(_ context.Context, input *go_hook.HookInput) error {
 			existedCredentials.CookieSecret = pwgen.AlphaNum(24)
 		}
 
-		// Migrate authenticators that reuse the shared kubernetes client secret to their own one
+		// Migrate authenticators that reuse the shared kubernetes client secret to their own one.
+		// The non-empty guard keeps an authenticator whose credentials Secret carries no
+		// client-secret from being rotated against an absent shared secret.
 		if sharedClientSecret != "" && existedCredentials.AppDexSecret == sharedClientSecret {
 			existedCredentials.AppDexSecret = pwgen.AlphaNum(20)
+			// Rotating logs every user of this authenticator out and rolls its pods, so make the
+			// upgrade traceable in the Deckhouse log.
+			input.Logger.Warn("Rotating DexAuthenticator client secret that reused the shared kubernetes client secret",
+				slog.String("dexauthenticator", dexAuthenticator.Name),
+				slog.String("namespace", dexAuthenticator.Namespace))
 		}
 
 		if raw := dexAuthenticator.AllowAccessToKubernetesAnnotation; raw != nil {
--- a/modules/150-user-authn/hooks/get_dex_authenticator_crds.go
+++ b/modules/150-user-authn/hooks/get_dex_authenticator_crds.go
@@ -161,6 +161,13 @@ func getDexAuthenticator(_ context.Context, input *go_hook.HookInput) error {
 		credentialsByID[dexSecret.ID] = dexSecret.Credentials
 	}
 
+	// Authenticators created before the templates stopped copying the kubernetes OAuth2Client
+	// secret into the application namespace carry that shared secret in their credentials
+	// Secret, and the "credentials" snapshot above reads it straight back into appDexSecret.
+	// Keep it to detect and rotate those, otherwise the shared secret would never leave the
+	// application namespace.
+	sharedClientSecret := input.Values.Get(kubernetesDexClientAppSecretPath).String()
+
 	dexAuthenticators := make([]DexAuthenticator, 0, len(authenticators))
 	// Build computed names map: key "<name>@<namespace>" => {name, truncated, hash}
 	namesMap := make(map[string]interface{})
@@ -181,6 +188,11 @@ func getDexAuthenticator(_ context.Context, input *go_hook.HookInput) error {
 		// Migrate all cookie secret from 20 bytes length to 24 bytes
 		if len(existedCredentials.CookieSecret) < 24 {
 			existedCredentials.CookieSecret = pwgen.AlphaNum(24)
+		}
+
+		// Migrate authenticators that reuse the shared kubernetes client secret to their own one
+		if sharedClientSecret != "" && existedCredentials.AppDexSecret == sharedClientSecret {
+			existedCredentials.AppDexSecret = pwgen.AlphaNum(20)
 		}
 
 		if raw := dexAuthenticator.AllowAccessToKubernetesAnnotation; raw != nil {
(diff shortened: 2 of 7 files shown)


modules/150-user-authn/hooks/get_dex_authenticator_crds.go
```diff
@@ -161,6 +161,13 @@ func getDexAuthenticator(_ context.Context, input *go_hook.HookInput) error {
 		credentialsByID[dexSecret.ID] = dexSecret.Credentials
 	}
 
+	// Authenticators created before the templates stopped copying the kubernetes OAuth2Client
+	// secret into the application namespace carry that shared secret in their credentials
+	// Secret, and the "credentials" snapshot above reads it straight back into appDexSecret.
+	// Keep it to detect and rotate those, otherwise the shared secret would never leave the
+	// application namespace.
+	sharedClientSecret := input.Values.Get(kubernetesDexClientAppSecretPath).String()
```
AlwxSin: 🟡 **The migration depends on a value another hook populates, with nothing declaring or enforcing that order**

`sharedClientSecret` is read from `input.Values`, but that path is written by a different hook (`kubernetesDexClientAppSecret`). Nothing in this hook's config expresses the dependency, and the two hooks sit in different queues — `generate_kubernetes_dex_client_app_secret.go` declares no `Queue` (so: `main`), while this hook declares `Queue: "/modules/user-authn"` — so their synchronization order on a cold start is not guaranteed.

Lose that race and `sharedClientSecret` is `""`, the `!= ""` guard skips every authenticator, and helm then renders the un-rotated `appDexSecret` for that whole module run. The leak survives the upgrade, silently, on exactly the reconcile where the migration was supposed to happen; it only heals on some later unrelated DexAuthenticator/Secret event, which on a quiet cluster may be a long time away.

The sibling hook already shows the self-contained pattern: bind `d8-user-authn/kubernetes-dex-client-app-secret` into this hook's `Kubernetes` config and read it from the snapshot. Two lines, no cross-hook ordering — and it is the more authoritative source anyway, since the cluster Secret is what actually got copied out.

modules/150-user-authn/hooks/get_dex_authenticator_crds.go
```diff
@@ -183,6 +190,11 @@ func getDexAuthenticator(_ context.Context, input *go_hook.HookInput) error {
 			existedCredentials.CookieSecret = pwgen.AlphaNum(24)
 		}
 
+		// Migrate authenticators that reuse the shared kubernetes client secret to their own one
+		if sharedClientSecret != "" && existedCredentials.AppDexSecret == sharedClientSecret {
+			existedCredentials.AppDexSecret = pwgen.AlphaNum(20)
```
AlwxSin, line 240: 🟡 **A credential rotation that restarts pods and logs users out happens with zero log output**

The rotation is invisible: no `input.Logger` call, no metric, no event. An operator upgrading a cluster sees kiali, hubble-ui and every annotated application authenticator roll and their users bounced, with nothing in the Deckhouse log tying it to this change.

The pattern and the logger are already right there — a few lines below, the same function logs a `Warn` for a far less consequential thing (an unparseable annotation):

```go
input.Logger.Warn("Rotating DexAuthenticator client secret that reused the shared kubernetes client secret",
    slog.String("dexauthenticator", dexAuthenticator.Name),
    slog.String("namespace", dexAuthenticator.Namespace))
```

modules/150-user-authn/templates/dex-authenticator/secret.yaml
```diff
@@ -19,11 +19,7 @@ metadata:
   {{- end }}
   {{- include "helm_lib_module_labels" (list $context $labels) | nindent 2 }}
 data:
-  {{- if $crd.allowAccessToKubernetes }}
-  client-secret: {{ $context.Values.userAuthn.internal.kubernetesDexClientAppSecret | b64enc }}
```
AlwxSin, line 23: 🟡 **The changelog instructs operators to rotate the shared secret, but gives no procedure and understates the blast radius**

(The real target of this comment is the `changes` block in the PR description; anchoring here because that text is not part of the diff. This removed line is what handed the shared secret out in the first place.)

`impact_level: high` with *"Rotate the shared `kubernetes` client secret if the annotation was previously used outside system namespaces"* — and no steps. It is not a `delete secret` away: `kubernetesDexClientAppSecret` early-returns while the in-memory value is non-empty, so a running Deckhouse just re-renders the same value. The rotation needs the Secret `d8-user-authn/kubernetes-dex-client-app-secret` gone (or its `data.secret` blanked) **and** a Deckhouse restart to clear the value.

The blast radius is also bigger than the entry implies: the same value is the client secret of `kubeconfig-generator`, `kubeconfig-publish-api` and every `kubeconfig-<slug>` client (`templates/dex/kubernetes-dex-client-configuration.yaml:27,37`, `templates/kubeconfig-generator/oauth2client.yaml:12,31`), plus `--ldap-client-secret` in basic-auth-proxy. Rotating it invalidates those flows too. For a high-impact security entry that is the difference between an operator doing it and skipping it.

modules/150-user-authn/hooks/get_dex_authenticator_crds_test.go
```diff
@@ -379,6 +387,57 @@ spec:
...
+				Expect(appDexSecret).NotTo(Equal(sharedKubernetesClientSecret))
+				Expect(appDexSecret).To(HaveLen(20))
+			} else {
+				Expect(appDexSecret).To(Equal(in.existing))
+			}
+		},
+		Entry("is regenerated when it equals the shared kubernetes client secret",
+			inputDexAuthenticatorClientSecret{existing: sharedKubernetesClientSecret, rotated: true},
+		),
+		Entry("is preserved when it is unrelated to the shared kubernetes client secret",
```
AlwxSin, line 477: 🔵 **No test for the case where the shared secret value is absent**

The new table covers "equals shared → rotated" and "unrelated → preserved", but not the `sharedClientSecret == ""` guard — the exact branch that makes the ordering hazard above observable. One `Entry` that leaves `userAuthn.internal.kubernetesDexClientAppSecret` unset and asserts the existing secret is kept would pin that guard down.

modules/150-user-authn/template_tests/kubernetes_oauth2client_test.go
```diff
@@ -80,9 +80,12 @@ var _ = Describe("Module :: user-authn :: helm template :: kubernetes oauth2clie
...
 			hec.HelmRender()
 		})
-		It("Should deploy kubernetes OAuth2Client", func() {
+		It("Should deploy kubernetes OAuth2Client without redirect URIs", func() {
 			Expect(hec.RenderError).ShouldNot(HaveOccurred())
-			Expect(hec.KubernetesResource("OAuth2Client", "d8-user-authn", "nn2wezlsnzsxizltzpzjzzeeeirsk").Exists()).To(BeTrue())
+
+			oauth2Client := hec.KubernetesResource("OAuth2Client", "d8-user-authn", "nn2wezlsnzsxizltzpzjzzeeeirsk")
+			Expect(oauth2Client.Exists()).To(BeTrue())
+			Expect(oauth2Client.Field("redirectURIs").Array()).To(BeEmpty())
```
AlwxSin: 🔵 **Fixtures in this file still set `spec.applicationDomain` — the field this PR proves cannot exist**

Lines 68, 102, 134 and 143 hand-feed a v1alpha1-shaped `spec` that the hook can no longer produce: the conversion webhook's `v1_to_v2alpha1` does `del(.spec.applicationDomain, ...)`, which is exactly why the removed template line rendered as the literal `https:///dex-authenticator/callback` in production. Those fixtures are the only reason the old `"https://with-access.example.com/dex-authenticator/callback"` assertion ever passed.

Now that nothing reads the field it is dead fixture data that reads as if the field were real — the next person touching this template will trust it. Worth switching to `applications: [{domain: ...}]`, matching what the hook actually emits.

modules/150-user-authn/hooks/generate_kubernetes_dex_client_app_secret.go
```diff
@@ -61,7 +66,7 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
 }, kubernetesDexClientAppSecret)
 
 func kubernetesDexClientAppSecret(_ context.Context, input *go_hook.HookInput) error {
-	secretPath := "userAuthn.internal.kubernetesDexClientAppSecret"
+	secretPath := kubernetesDexClientAppSecretPath
```
AlwxSin: 🔵 **`secretPath` is now a pointless alias**

`secretPath := kubernetesDexClientAppSecretPath` — use the constant directly and drop the local.

modules/150-user-authn/templates/dex/oauth2client.yaml
```diff
@@ -55,11 +55,4 @@ redirectURIs:
   {{- if $context.Values.userAuthn.internal.publishAPI.enabled }}
 - https://{{ include "helm_lib_module_public_domain" (list $context "kubeconfig") }}/callback/
   {{- end }}
-  {{- if $context.Values.userAuthn.internal.dexAuthenticatorCRDs }}
-    {{- range $crd := $context.Values.userAuthn.internal.dexAuthenticatorCRDs }}
-      {{- if $crd.allowAccessToKubernetes }}
-- https://{{ $crd.spec.applicationDomain }}/dex-authenticator/callback
```
AlwxSin, line 61: 🔵 **`redirectURIs:` now renders as a bare key when the client exists only for authenticators**

With `publishAPI` disabled and no `kubeconfigGenerator`, the `kubernetes` OAuth2Client renders solely because some authenticator carries the annotation — and after this removal the `redirectURIs:` key (line 49) emits with nothing under it, i.e. `redirectURIs: null`.

Harmless in practice: `crds/external/oauth2clients.yaml` is `x-kubernetes-preserve-unknown-fields: true` with no validation, and dex unmarshals it to a nil slice, so the new `.Array()).To(BeEmpty())` assertion passes either way. Still, guarding the key or emitting `[]` is cleaner than shipping a null.

modules/150-user-authn/hooks/get_dex_authenticator_crds.go
Jabejixo: Done, exactly as suggested. The hook now has its own `kubernetesClientSecret` binding on `d8-user-authn/kubernetes-dex-client-app-secret`, reusing `applyKubernetesSecretFilter` from the sibling hook (`hooks/get_dex_authenticator_crds.go:148-161`), and the comparison value comes from that snapshot (`hooks/get_dex_authenticator_crds.go:176-188`). No cross-hook ordering left.

Confirmed the hazard you described: `ModuleRun` puts `main`-queue Synchronization tasks at the head of the ModuleRun queue while queueing the parallel-queue ones independently (`addon-operator@v1.24.11/pkg/task/tasks/module-run/task.go:304-386`), so the `main` hook and the `/modules/user-authn` hook are not ordered on a cold start.

On a run where the shared secret is absent, the guard now means something different: an empty snapshot means the Secret does not exist in the cluster, so no authenticator can be carrying that value and there is nothing to migrate. The `!= ""` guard is kept for the other side of the comparison — an authenticator whose credentials Secret has no `client-secret` key reads back as `""` and must not be rotated against an absent shared secret. That is pinned by a new test (see the reply on the test thread).

modules/150-user-authn/hooks/get_dex_authenticator_crds.go
Jabejixo: Added, with your wording and fields: `hooks/get_dex_authenticator_crds.go:243-245`. Asserted in both directions in the client-secret table (`hooks/get_dex_authenticator_crds_test.go:462,466`) — present when a rotation happens, absent when it does not.

modules/150-user-authn/hooks/generate_kubernetes_dex_client_app_secret.go
Jabejixo: Dropped. `kubernetesDexClientAppSecretPath` is now used directly at `hooks/generate_kubernetes_dex_client_app_secret.go:74,88,92,96`. The namespace and name of the Secret are constants too now (`:38-39`), shared with the new binding in `get_dex_authenticator_crds.go`.

modules/150-user-authn/hooks/get_dex_authenticator_crds_test.go
Jabejixo: Added, plus one more. The table is now parameterised on the credentials `client-secret`, the cluster Secret and the values path separately (`hooks/get_dex_authenticator_crds_test.go:40-53`):

- `is preserved when the shared kubernetes client secret does not exist in the cluster` (`:494`) — your entry, nothing set, existing secret kept.
- `is left empty when neither the credentials Secret nor the shared secret carries one` (`:499`) — credentials Secret without a `client-secret` key. This is the one that actually pins the guard: dropping `sharedClientSecret != ""` makes it fail, because `"" == ""` rotates. Verified by removing the guard locally.
- `is regenerated from the cluster Secret when the values path is not populated yet` (`:487`) — the ordering case from the other thread; fails before the snapshot change, passes after.

modules/150-user-authn/template_tests/kubernetes_oauth2client_test.go
Jabejixo: Switched all four to `applications: [{domain, ingressClassName, ingressSecretName}]` (`template_tests/kubernetes_oauth2client_test.go:68,119,157,168`), matching what the hook emits and what `template_tests/authenticator_test.go:54` already used. `dexAuthenticatorNames` gained the matching `ingressNames."0"` entries, since the Ingress template keys off the application index.

This also un-weakens the negative assertion. With the old `applicationDomain` fixtures, a template "repaired" to read `spec.applications[].domain` would have rendered nothing and `NotTo(ContainElement(ContainSubstring("/dex-authenticator/callback")))` would have passed vacuously. I checked it now catches that: re-adding a `range $crd.spec.applications` redirect URI block to the template fails `Should render OAuth2Client with all expected trustedPeers, redirectURIs and secret`.

modules/150-user-authn/templates/dex/oauth2client.yaml
Jabejixo: Changed to emit `[]`. The `redirectURIs` block is now guarded on `kubeconfigGenerator` or `publishAPI.enabled`, with an explicit `redirectURIs: []` otherwise (`templates/dex/oauth2client.yaml:49-64`).

Also tightened the assertion, since as you noted `.Array()).To(BeEmpty())` passed either way: the test now asserts `redirectURIs.IsArray()` as well (`template_tests/kubernetes_oauth2client_test.go:96`), which fails on the null, and additionally asserts the authenticator is still in `trustedPeers` so the empty list is not just an empty render.

modules/150-user-authn/templates/dex-authenticator/secret.yaml
Jabejixo: Rewrote the `changes` block in the PR description.

It now spells out the procedure you described — exclude the Secret from GitOps sync first, blank `data.secret`, `rollout restart deployment/deckhouse` to drop the in-memory copy, then verify the value changed and repeat if the module re-rendered the old one before the restart landed. Your point about the early return is stated explicitly (`hooks/generate_kubernetes_dex_client_app_secret.go:74`), so nobody tries `delete secret` alone.

The blast radius is listed: the same value is the client secret of `kubeconfig-generator`, `kubeconfig-publish-api` and every `kubeconfig-<slug>` client, and `--ldap-client-secret` for basic-auth-proxy (`templates/dex/kubernetes-dex-client-configuration.yaml:27,37`, `templates/kubeconfig-generator/oauth2client.yaml:12,31`, `templates/dex/kubeconfig-oauth2clients.yaml:12,30`, `templates/kubeconfig-generator/config.yaml:61`, `templates/basic-auth-proxy/deployment.yaml:148`). In-cluster consumers re-render and roll on their own; already downloaded kubeconfig files carry the old client secret and have to be downloaded again. Already issued ID tokens keep working until `settings.idTokenTTL` expiry.

Also covered: the per-authenticator rotation only touches a Deckhouse-owned Secret, so no GitOps change is needed for it, and delete-and-recreate of a DexAuthenticator is always a way to force a fresh secret. Validated the block against `.github/scripts/js/validators/pr-changes-validator.js` before pushing — one block, two documents, `impact_level: high` with `impact`.

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [22360](https://github.com/deckhouse/deckhouse/pull/22360) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22218, opened 2026-08-17
[user-authz] Reject identities that collide with authorization rule subjects

Jabejixo:
## Description

A local identity name reaches the token Dex issues, and kube-apiserver maps it to a Kubernetes identity with an **empty prefix** (`modules/040-control-plane-manager/templates/_authentication_configuration.tpl:17-23` — `claimMappings.username.claim: "email"`, `claimMappings.groups.claim: "groups"`, both with `prefix: ""`). There is therefore no namespacing between locally managed identities and identities asserted by an external identity provider, in either direction:

- `Group.spec.name` (`deckhouse.io/v1alpha1`) becomes the group name in the `groups` claim.
- `User.spec.email` (`deckhouse.io/v1`) becomes the username. The `AuthorizationRule` and `ClusterAuthorizationRule` CRDs say so themselves: "Use the user's `email` as the username to grant privileges to the specific user".

Creating such an object with a name that an existing rule already lists as a subject grants that rule's privileges to it, and nothing in either object records that it happened. Nothing validated that.

This PR adds a validating webhook, `modules/140-user-authz/webhooks/validating/identity_collision.py`, that refuses to introduce such a name and reports the rule responsible.

### Why the check lives in `user-authz` and not in `user-authn`

The obvious placement is the existing `user-authn` hooks that already validate these objects, and the first version of this PR did that. It was wrong, and the reason is worth stating because it is a trap anyone extending an admission hook can fall into.

Those hooks would need `kubernetes` bindings on `AuthorizationRule` and `ClusterAuthorizationRule`, which `user-authz` owns. Module CRDs are installed per module (`deckhouse-controller/internal/packages/runtime/tasks/ensurecrd/task.go`), so in a cluster where `user-authz` is disabled those CRDs do not exist. A binding on an absent CRD is not a no-op: `createSharedInformer` fails with "Cannot get GroupVersionResource info … Possibly CRD is not created before informers are started" (`shell-operator/pkg/kube_events_manager/resource_informer.go:120`) and the error propagates out of `CreateInformers`, taking down the whole hook — including the checks it performed before this change. `allowFailure` does not help; it governs hook execution, not informer creation.

Putting the check in `user-authz` inverts the dependency into a harmless one. The hook subscribes only to the two CRDs its own module ships, so they are present whenever the hook is loaded at all. What it needs from `user-authn` is only the incoming object, which arrives in the admission request. Its `kubernetesValidating` rules name `groups` and `users`, and if `user-authn` is disabled those CRDs are absent and the rules simply never match — the API server does not require a rule's target resource to exist.

This is also the established shape in this module rather than a new invention: `webhooks/validating/system_roles` validates `clusterroles` and `webhooks/validating/role_binding` validates `clusterrolebindings`, both resources this module does not own. `webhooks/validating/multitenancy.py` is the closest structural precedent — one Python hook validating two kinds, dispatching on `request.kind.kind`, with `includeSnapshotsFrom` attaching snapshots to individual rules — and this hook follows it.

The symmetry is deliberate and complete in the direction that matters. The other direction is not covered: adding a subject to a rule when a matching local `Group` or `User` already exists produces no warning. Covering it would require informers on the `user-authn` CRDs, reintroducing exactly the fragility described above, and it is the less important direction — whoever edits an authorization rule is the privileged party making a deliberate grant.

### Behaviour

**Snapshots.** Two `kubernetes` bindings with `executeHookOnEvent: []`, `executeHookOnSynchronization: false`, `keepFullObjectsInMemory: false`, and a `jqFilter` projecting only the rule identity and its `Group` and `User` subjects, so the snapshots stay small. They carry no `group` and no `queue` — nothing has to be coordinated with another binding, and `includeSnapshotsFrom` names both snapshots on each validating rule explicitly. `ServiceAccount` subjects are filtered out; they are not affected by this class of collision. No RBAC change is needed: the webhook handler already holds `get`/`list`/`watch` on `deckhouse.io` `*` (`modules/002-deckhouse/templates/webhook-handler/rbac-for-us.yaml`).

**Deny vs warn.** The collision is denied, with two deliberate scope limits and one escape hatch:

1. **Denied only when the colliding identity is being introduced** — on `CREATE`, or on `UPDATE` when the identity actually changes into a colliding value. An object that already carries a colliding name keeps working and only produces a warning. Denying unconditionally would lock existing legitimate setups out of any further modification, such as adding a group member, and the privileges are already effective for such an object anyway, so a hard error would not take them away. An `UPDATE` whose `oldObject` cannot be read counts as new, so an unreadable previous value fails closed.
2. **Escape hatch** — the `user-authz.deckhouse.io/allow-authorization-rule-collision: "true"` annotation on the `Group` or `User` acknowledges the collision and downgrades the denial to a warning. This keeps the legitimate workflow of writing a rule in advance and then creating the local object that populates it. The prefix is `user-authz` because the constraint and its enforcement belong to this module: with `user-authz` disabled there are no rules and the annotation means nothing.
3. **No silent allow** — every path either denies or emits a warning naming the rule that grants the privileges.

The alternative of denying only when the requester is not permitted to manage the colliding rule was rejected: without `SubjectAccessReview` it can only be approximated from `request.userInfo`, which is guessing rather than authorization, and adding an API call on the write path trades a correctness problem for an availability one. The chosen design is honest about what it protects. It deterministically stops the *accidental* silent privilege grant, and it turns the *deliberate* one into an explicit, greppable, audit-visible annotation instead of an invisible name match. It does not by itself stop someone who can already create these objects, since they can also set the annotation; closing that needs `SubjectAccessReview` and is worth a major release.

**Deletion warns that the rule survives it.** Deleting a `Group` or a `User` does not touch the rule, so the name stays granted: recreating an object with the same name restores the privileges, and until then the name is free for anyone allowed to create one. Both kinds warn and neither blocks the deletion.

### The email comparison is case-insensitive on one side only

Rule subjects are matched against the identity that ends up in the token, not against the spec field verbatim, and for `User.spec.email` the two differ. The two directions are not symmetric and both matter:

- **A mixed-case incoming email is a real collision and is denied.** Deckhouse lowercases the email before it reaches the `Password` object — `email := strings.ToLower(dexUser.Spec.Email)` at `modules/150-user-authn/hooks/get_dex_user_crds.go:276`, and that variable is what populates `Password.email` (`:496`, `:540`, `:593`). The username claim the API server consumes is that email, so the username is unconditionally `spec.email.lower()`. Comparing verbatim made the check bypassable by case alone: a `User` with `Privileged@Example.com` matched no subject, and the issued token still carried `privileged@example.com`, which the rule grants. The hook now compares `spec.email.lower()`.
- **A mixed-case rule subject is not a collision and is not reported.** Subjects reach the RoleBinding verbatim (`modules/140-user-authz/templates/cluster-role-bindings.yaml:38`, `modules/140-user-authz/hooks/handle_manage_bindings.go:256`) and RBAC matches subjects by exact string, so a subject that is not already lowercase can never match an issued token. Normalising that side too would report a collision that does not exist, so subjects stay indexed verbatim.

Consistently with the same rule, the "is the identity being introduced" comparison for `UPDATE` is on the token identity as well: rewriting an already colliding email in a different case does not count as introducing it and only warns, so a legacy object stays modifiable.

**`Group.spec.name` is compared verbatim in both directions**, and deliberately not the same way. Nothing normalises a group name on its way to the `groups` claim: `makeUserGroupsMap` accumulates `group.Spec.Name` byte for byte (`modules/150-user-authn/hooks/get_dex_user_crds.go:697,706`), the result is written to `Password.groups` unchanged (`:503`, `:563`, `:596`), and there is no `strings.ToLower` on any group path in either module — the only one in `user-authz` is on an annotation value used for naming, not on a subject (`modules/140-user-authz/hooks/handle_dict_bindings.go:167`). A case-only difference in a group name is therefore a genuinely different group, and treating it as a collision would be a false positive. Two tests pin this down so that the email fix cannot later be copied onto groups by reflex.

This hook does not rely on `modules/150-user-authn/webhooks/validating/user.py` rejecting uppercase emails on `CREATE`: that check has a legacy path that allows an `UPDATE` between two uppercase values (`user.py:117` only fires when the old email was lowercase, `user.py:122-125` then allows and returns), and it belongs to a different module that may be disabled.

`object` is null on `DELETE` and `oldObject` is null on `CREATE`, so both are treated as absent rather than dereferenced.

### Tests

`identity_collision_test.py` with an `identity_collision_test_factories.py` context factory, following the structure of `multitenancy_test.py` and `multitenancy_test_factories.py` in the same directory. 28 tests cover, for both `Group` and `User`: denial against a `ClusterAuthorizationRule` subject and against a namespaced `AuthorizationRule` subject; a non-colliding name; no rules at all; the acknowledging annotation; renaming into a collision; a pre-existing collision producing a warning instead of a denial; an `UPDATE` whose `oldObject` is unreadable failing closed; both deletion outcomes; and both directions of the email case question plus both directions of the group case question.

The `jqFilter` programs are no longer assumed. `TestSnapshotJQFilters` parses `CONFIG` as YAML and runs each filter through `jq` against real rule objects, checking that `ServiceAccount` subjects are dropped, that the `[]?` guard survives a rule with no `spec.subjects`, and that the filter output feeds `granting_rule_for` unchanged. `jq` was added to `testing/webhooks/test.sh` so the CI container has it; the class is skipped rather than failing if it is absent.

Every guard was checked by mutation — each of the following makes the suite fail, and reverting it makes it pass again:

| mutation | failing tests |
| --- | --- |
| compare `spec.email` verbatim | `test_user_denied_when_email_reaches_the_subject_only_after_lowercasing`, `test_user_deletion_warns_for_a_mixed_case_email`, `test_filter_output_feeds_the_lookup` |
| index rule subjects lowercased | `test_user_allowed_when_the_rule_subject_itself_is_mixed_case` |
| lowercase group names too | `test_group_allowed_when_name_differs_from_the_subject_only_by_case` |
| handle `DELETE` for `User` only | `test_group_deletion_warns_that_the_rule_survives_it` |
| drop the fail-closed `oldObject` guard | `test_user_denied_when_the_old_object_cannot_be_read` |
| compare old and new identity verbatim | `test_user_allowed_with_warning_when_a_colliding_email_only_changes_case` |
| drop `select(.kind == "Group")` from the filters | both `*_filter_keeps_only_group_and_user_subjects` tests |
| drop the `?` from `.spec.subjects[]?` | `test_filters_tolerate_a_rule_without_subjects` (both subtests) |

Picking a name that is already a subject in an authorization rule silently grants that rule's privileges, with no indication anywhere that it happened. It is reachable by accident by an administrator who simply chooses a name already present in a rule. It is also the mechanism by which someone who may create `Group` or `User` objects but has no authority over `user-authz` can inherit privileges intended for an external identity — for a `Group` by adding themselves as a member, for a `User` by taking over an email a rule already trusts — and then re-authenticating.

Yes. It closes an unintended privilege escalation path that needs no special conditions to hit, and the change is one new validating hook that adds no new failure mode to any existing one. The check is deliberately scoped so that it cannot break already existing objects: they stay updatable and only produce a warning.

Backporting is nearly trivial. Three of the four touched files are new and their paths do not exist on either release branch; the fourth is a one-line `apk add` in `testing/webhooks/test.sh`. Both commits were cherry-picked with `-x` into throwaway detached worktrees and the whole webhook test directory was run there:

- **`release-1.77`** (at `89786b5386`): both picks clean, no conflict. `identity_collision_test.py` 28 tests OK, `multitenancy_test.py` 12 tests OK.
- **`release-1.76`** (at `d9659de1fc`): the first pick is clean; the second conflicts on `testing/webhooks/test.sh`, because that branch does not yet carry the `DISTRO_PACKAGES_PROXY` block. The resolution is to keep the branch's own version of the file and append `jq` to its `apk add` line — one line, no logic. After it: `identity_collision_test.py` 28 tests OK, `multitenancy_test.py` 8 tests OK.

The runtime prerequisites were verified on both branches: both `AuthorizationRule` and `ClusterAuthorizationRule` CRDs with their `subjects` schema, `Group.spec.name`, the webhook handler's per-enabled-module hook copying (`images/webhook-handler/src/entrypoint.sh`), the same `deckhouse==0.4.11`, `dotmap==1.3.30` and `PyYAML==6.0.2` webhook dependencies, and the same `deckhouse.io` RBAC. The premise of the email fix holds on both: `release-1.77` lowercases the email in the hook exactly as `main` does, and `release-1.76` does it in the Helm template that renders the `Password` object instead (`modules/150-user-authn/templates/dex/passwords.yaml:14` — `email: {{ $crd.spec.email | lower | quote }}`), while the groups on the same object are rendered verbatim (`:21`).

## Checklist
- [ ] e2e tests passed.

Client-facing documentation for this change is handled in a separate documentation PR, together with the docs for the other changes in this series, so that the documentation team reviews them in one place.

## Changelog entries

```changes
section: user-authz
type: fix
summary: Reject local Group names and User emails that collide with a subject already granted privileges by an authorization rule.
impact: |
  Creating a `Group` or a `User`, or renaming an existing one, so that `Group.spec.name` matches a group subject or `User.spec.email` matches a user subject of an existing AuthorizationRule or ClusterAuthorizationRule is now rejected, preventing an unintended privilege grant. The email is compared after lowercasing, since that is the form that reaches the token. Set the `user-authz.deckhouse.io/allow-authorization-rule-collision` annotation to `"true"` on the object if the collision is intentional. Already existing objects keep working and only produce a warning. Any declarative flow that manages both the rule and the identity is affected, not only one that deletes and recreates the identity: a first-time apply denies the `Group` or `User` whenever the rule happens to be applied first, and nothing guarantees the order. Add the annotation to such objects before upgrading. Deleting a `Group` or a `User` whose name is still a subject of a rule now produces a warning, since the rule keeps granting that name.
impact_level: high
---
section: testing
type: chore
summary: Install jq in the python webhook test container so hooks can test their own jqFilter programs.
impact_level: low
```

### DIFF
--- a/modules/140-user-authz/webhooks/validating/identity_collision.py
+++ b/modules/140-user-authz/webhooks/validating/identity_collision.py
@@ -1,0 +1,227 @@
+#!/usr/bin/python3
+
+# Copyright 2026 Flant JSC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# This hook guards the authorization rules owned by this module against a local identity silently
+# landing in one of them.
+#
+# A local identity name reaches the token Dex issues verbatim, and kube-apiserver maps it to a
+# Kubernetes identity with an empty prefix, so there is no namespacing between locally managed
+# identities and identities asserted by an external identity provider:
+#
+# - Group.spec.name becomes the group name in the "groups" claim.
+# - User.spec.email becomes the username. The AuthorizationRule and ClusterAuthorizationRule CRDs
+#   say so themselves: "Use the user's `email` as the username to grant privileges to the specific
+#   user".
+#
+# Creating such an object with a name that an existing rule already lists as a subject therefore
+# grants that rule's privileges to it, with nothing in either object recording that it happened.
+#
+# The check lives in this module rather than in user-authn on purpose. It needs the authorization
+# rules, which this module owns, and it only needs the incoming object from the admission request.
+# A hook in user-authn would have to open informers on the CRDs of this module, and a missing CRD
+# makes informer creation fail (shell-operator, pkg/kube_events_manager/resource_informer.go), which
+# would take the whole hook down in a cluster where user-authz is disabled. Here the dependency is
+# the other way round and is harmless: the validating rules below name resources owned by
+# user-authn, and if that module is disabled its CRDs are absent, the rules simply never match.
+
+from typing import Optional
+
+from deckhouse import hook
+from dotmap import DotMap
+
+CLUSTER_RULES_SNAPSHOT_NAME = "d8-user-authz-collision-cluster-authorization-rules"
+NAMESPACED_RULES_SNAPSHOT_NAME = "d8-user-authz-collision-authorization-rules"
+
+CONFIG = f"""
+configVersion: v1
+kubernetesValidating:
+- name: d8-user-authz-group-authorization-rule-collision.deckhouse.io
+  includeSnapshotsFrom: ["{CLUSTER_RULES_SNAPSHOT_NAME}", "{NAMESPACED_RULES_SNAPSHOT_NAME}"]
+  rules:
+  - apiGroups:   ["deckhouse.io"]
+    apiVersions: ["*"]
+    operations:  ["CREATE", "UPDATE"]
+    resources:   ["groups"]
+    scope:       "Cluster"
+- name: d8-user-authz-user-authorization-rule-collision.deckhouse.io
+  includeSnapshotsFrom: ["{CLUSTER_RULES_SNAPSHOT_NAME}", "{NAMESPACED_RULES_SNAPSHOT_NAME}"]
+  rules:
+  - apiGroups:   ["deckhouse.io"]
+    apiVersions: ["*"]
+    operations:  ["CREATE", "UPDATE", "DELETE"]
+    resources:   ["users"]
+    scope:       "Cluster"
+kubernetes:
+- name: {CLUSTER_RULES_SNAPSHOT_NAME}
+  apiVersion: deckhouse.io/v1alpha1
+  kind: ClusterAuthorizationRule
+  executeHookOnEvent: []
+  executeHookOnSynchronization: false
+  keepFullObjectsInMemory: false
+  jqFilter: |
+    {{
+      "name": .metadata.name,
+      "groupSubjects": [.spec.subjects[]? | select(.kind == "Group") | .name],
+      "userSubjects": [.spec.subjects[]? | select(.kind == "User") | .name]
+    }}
+- name: {NAMESPACED_RULES_SNAPSHOT_NAME}
+  apiVersion: deckhouse.io/v1alpha1
+  kind: AuthorizationRule
+  executeHookOnEvent: []
+  executeHookOnSynchronization: false
+  keepFullObjectsInMemory: false
+  jqFilter: |
+    {{
+      "name": .metadata.name,
+      "namespace": .metadata.namespace,
+      "groupSubjects": [.spec.subjects[]? | select(.kind == "Group") | .name],
+      "userSubjects": [.spec.subjects[]? | select(.kind == "User") | .name]
+    }}
+"""
+
+# Acknowledges that a collision with an authorization rule is intentional. The prefix is
+# user-authz because the constraint and its enforcement belong to this module: with user-authz
+# disabled there are no rules and the annotation means nothing.
+COLLISION_ANNOTATION = "user-authz.deckhouse.io/allow-authorization-rule-collision"
+
+
+def main(ctx: hook.Context):
+    try:
+        # DotMap is a dict with dot notation
+        binding_context = DotMap(ctx.binding_context)
+        errmsg, warnings = validate(binding_context)
+        if errmsg is None:
+            ctx.output.validations.allow(*warnings)
+        else:
+            ctx.output.validations.deny(errmsg)
+    except Exception as e:
+        ctx.output.validations.error(str(e))
+
+
+def validate(ctx: DotMap) -> tuple[Optional[str], list[str]]:
+    req = ctx.review.request
+    kind = req.kind.kind.lower()
+
+    if kind == "group":
+        return validate_identity(ctx, resource="groups.deckhouse.io", field=".spec.name",
+                                 subjects_key="groupSubjects",
+                                 name=spec_value(req.object, "name"),
+                                 old_name=spec_value(req.oldObject, "name"))
+    if kind == "user":
+        if req.operation == "DELETE":
+            return warn_rule_outlives_user(ctx, spec_value(req.oldObject, "email"))
+        return validate_identity(ctx, resource="users.deckhouse.io", field=".spec.email",
+                                 subjects_key="userSubjects",
+                                 name=spec_value(req.object, "email"),
+                                 old_name=spec_value(req.oldObject, "email"))
+
+    return None, []
+
+
+def spec_value(obj: Optional[DotMap], field: str):
+    """
+    Read a spec field of an admission review object.
+
+    `object` is null on DELETE and `oldObject` is null on CREATE, so both have to be treated as
+    absent rather than dereferenced.
+    """
+    if not obj:
+        return None
+
+    return obj.spec[field]
+
+
+def validate_identity(ctx: DotMap, resource: str, field: str, subjects_key: str,
+                      name, old_name) -> tuple[Optional[str], list[str]]:
+    warnings = []
+
+    if not isinstance(name, str) or not name:
+        # The field is required by both CRDs, so an absent value is not something this hook has an
+        # opinion about — the owning module's own schema and webhooks reject it.
+        return None, warnings
+
+    granting_rule = granting_rule_for(ctx, subjects_key, name)
+    if granting_rule is None:
+        return None, warnings
+
+    collision = f"{resource} \"{field}\" \"{name}\" is already granted privileges by {granting_rule}"
+
+    # An already existing collision is only reported: the privileges are effective anyway, and
+    # denying updates would lock such an object out of any further modification. An UPDATE whose
+    # oldObject is missing counts as new, so an unreadable previous value fails closed.
+    name_is_new = ctx.review.request.operation == "CREATE" or old_name != name
+    if name_is_new and not is_collision_acknowledged(ctx.review.request.object):
+        return (
+            f"{collision}; use a different value or set the "
+            f"\"{COLLISION_ANNOTATION}: true\" annotation to confirm the collision"
+        ), warnings
+
+    warnings.append(f"{collision}; it inherits the privileges granted by that rule")
+    return None, warnings
+
+
+def warn_rule_outlives_user(ctx: DotMap, email) -> tuple[Optional[str], list[str]]:
+    """
+    Report that deleting the User does not revoke the rule.
+
+    The email stays granted, so recreating a User with the same email restores the privileges, and
+    until then the email is free for anyone allowed to create a User.
+    """
+    if not isinstance(email, str) or not email:
+        return None, []
+
+    granting_rule = granting_rule_for(ctx, "userSubjects", email)
+    if granting_rule is None:
+        return None, []
+
+    return None, [f"{granting_rule} still grants privileges to \".spec.email\" \"{email}\""]
+
+
+def granting_rule_for(ctx: DotMap, subjects_key: str, name: str) -> Optional[str]:
+    """
+    Find an authorization rule that already grants privileges to the given identity name.
+
+    The comparison is exact. Dex lowercases the email before putting it into the token, so a rule
+    subject written in another case never matches an issued token and reporting it would be a false
+    positive.
+
+    Returns:
+        Optional[str]: description of the rule granting the privileges, or None when there is none.
+                       The first rule found for a name wins, so the reported rule is stable.
+    """
+    privileged = {}
+
+    for rule in ctx.snapshots.get(CLUSTER_RULES_SNAPSHOT_NAME, []):
+        rule_name = f"clusterauthorizationrules.deckhouse.io \"{rule.filterResult.name}\""
+        for subject in rule.filterResult[subjects_key]:
+            privileged.setdefault(subject, rule_name)
+
+    for rule in ctx.snapshots.get(NAMESPACED_RULES_SNAPSHOT_NAME, []):
+        rule_name = (f"authorizationrules.deckhouse.io "
+                     f"\"{rule.filterResult.namespace}/{rule.filterResult.name}\"")
+        for subject in rule.filterResult[subjects_key]:
+            privileged.setdefault(subject, rule_name)
+
+    return privileged.get(name)
+
+
+def is_collision_acknowledged(obj: DotMap) -> bool:
+    return obj.metadata.annotations.get(COLLISION_ANNOTATION, "").lower() == "true"
+
+
+if __name__ == "__main__":
+    hook.run(main, config=CONFIG)
--- a/modules/140-user-authz/webhooks/validating/identity_collision.py
+++ b/modules/140-user-authz/webhooks/validating/identity_collision.py
@@ -18,14 +18,14 @@
 # This hook guards the authorization rules owned by this module against a local identity silently
 # landing in one of them.
 #
-# A local identity name reaches the token Dex issues verbatim, and kube-apiserver maps it to a
-# Kubernetes identity with an empty prefix, so there is no namespacing between locally managed
-# identities and identities asserted by an external identity provider:
+# A local identity name reaches the token Dex issues, and kube-apiserver maps it to a Kubernetes
+# identity with an empty prefix, so there is no namespacing between locally managed identities and
+# identities asserted by an external identity provider:
 #
-# - Group.spec.name becomes the group name in the "groups" claim.
-# - User.spec.email becomes the username. The AuthorizationRule and ClusterAuthorizationRule CRDs
-#   say so themselves: "Use the user's `email` as the username to grant privileges to the specific
-#   user".
+# - Group.spec.name becomes the group name in the "groups" claim, byte for byte.
+# - User.spec.email becomes the username, lowercased on the way. The AuthorizationRule and
+#   ClusterAuthorizationRule CRDs say so themselves: "Use the user's `email` as the username to
+#   grant privileges to the specific user".
 #
 # Creating such an object with a name that an existing rule already lists as a subject therefore
 # grants that rule's privileges to it, with nothing in either object recording that it happened.
@@ -38,7 +38,7 @@
 # the other way round and is harmless: the validating rules below name resources owned by
 # user-authn, and if that module is disabled its CRDs are absent, the rules simply never match.
 
-from typing import Optional
+from typing import Callable, NamedTuple, Optional
 
 from deckhouse import hook
 from dotmap import DotMap
@@ -54,7 +54,7 @@ kubernetesValidating:
   rules:
   - apiGroups:   ["deckhouse.io"]
     apiVersions: ["*"]
-    operations:  ["CREATE", "UPDATE"]
+    operations:  ["CREATE", "UPDATE", "DELETE"]
     resources:   ["groups"]
     scope:       "Cluster"
 - name: d8-user-authz-user-authorization-rule-collision.deckhouse.io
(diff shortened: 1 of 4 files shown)


modules/140-user-authz/webhooks/validating/identity_collision.py
```diff
@@ -0,0 +1,227 @@
...
+        for subject in rule.filterResult[subjects_key]:
+            privileged.setdefault(subject, rule_name)
+
+    for rule in ctx.snapshots.get(NAMESPACED_RULES_SNAPSHOT_NAME, []):
+        rule_name = (f"authorizationrules.deckhouse.io "
+                     f"\"{rule.filterResult.namespace}/{rule.filterResult.name}\"")
+        for subject in rule.filterResult[subjects_key]:
+            privileged.setdefault(subject, rule_name)
+
+    return privileged.get(name)
```
AlwxSin: 🔴 **The email comparison is case-sensitive, but the identity that actually reaches the token is lowercased — the new denial is bypassable**

`granting_rule_for` compares `User.spec.email` verbatim against rule subjects. Deckhouse itself lowercases the email before it ever reaches Dex: `modules/150-user-authn/hooks/get_dex_user_crds.go:276` — `email := strings.ToLower(dexUser.Spec.Email)` — and that lowercased variable is what populates the Password object's `email` field (`:496`, `:540`, `:593`). So the username the API server sees is unconditionally `spec.email.lower()`.

The docstring above (lines 198-200) only covers one direction. A mixed-case *rule subject* never matching an issued token is correct — reporting it would be a false positive. A mixed-case *incoming email* is the opposite case, and there the exact comparison misses a real collision.

Concrete bypass, given a pre-existing `User` whose `spec.email` contains uppercase (explicitly supported legacy state):

1. `modules/150-user-authn/webhooks/validating/user.py:117` denies an UPDATE that introduces uppercase only when `not old_has_upper` — for a legacy object that guard does not fire.
2. `user.py:122-124` then allows the update with a warning and returns, so the email can be changed to *another* uppercase value, e.g. `ADMIN@COMPANY.COM`.
3. This hook compares `ADMIN@COMPANY.COM` against the rule subject `admin@company.com`, finds nothing, and allows silently — no denial, no warning.
4. Deckhouse writes the Password with `admin@company.com`; the token username matches the rule subject and the privileges are granted.

The same gap swallows the DELETE warning and the pre-existing-collision warning for any legacy uppercase-email `User`.

Fix: compare `name.lower()` for the `userSubjects` path while keeping subjects verbatim. That catches a lowercase subject against an email in any case, and still ignores a mixed-case subject that can never match a token. Note that `identity_collision_test.py:124` currently asserts the wrong behaviour as correct, with a comment justifying it — that test needs inverting, not extending, and the docstring should name both directions so this cannot be reintroduced.

modules/140-user-authz/webhooks/validating/identity_collision.py
```diff
@@ -0,0 +1,227 @@
...
+    if granting_rule is None:
+        return None, warnings
+
+    collision = f"{resource} \"{field}\" \"{name}\" is already granted privileges by {granting_rule}"
+
+    # An already existing collision is only reported: the privileges are effective anyway, and
+    # denying updates would lock such an object out of any further modification. An UPDATE whose
+    # oldObject is missing counts as new, so an unreadable previous value fails closed.
+    name_is_new = ctx.review.request.operation == "CREATE" or old_name != name
+    if name_is_new and not is_collision_acknowledged(ctx.review.request.object):
```
AlwxSin, line 203: 🟡 **The changelog's GitOps caveat covers only delete/recreate, but a first-time apply hits the same denial**

(The real target of this comment is the `changes` block in the PR description; anchoring here because that text is not part of the diff.)

The impact text says a GitOps flow that deletes and recreates such an object will hit the denial. A first-time apply has the same problem: a repository that manages the rule and the identity together has no guarantee about which one lands first, and any order that applies the rule before the `Group`/`User` denies the identity. Nothing about that requires a prior delete.

That changes the guidance from "add the annotation when recreating" to "add the annotation on any declarative flow that manages both objects" — a materially different instruction for anyone reading the changelog before upgrading.

modules/140-user-authz/webhooks/validating/identity_collision.py
```diff
@@ -0,0 +1,227 @@
...
+
+CONFIG = f"""
+configVersion: v1
+kubernetesValidating:
+- name: d8-user-authz-group-authorization-rule-collision.deckhouse.io
+  includeSnapshotsFrom: ["{CLUSTER_RULES_SNAPSHOT_NAME}", "{NAMESPACED_RULES_SNAPSHOT_NAME}"]
+  rules:
+  - apiGroups:   ["deckhouse.io"]
+    apiVersions: ["*"]
+    operations:  ["CREATE", "UPDATE"]
```
AlwxSin: 🔵 **`Group` deletion has no counterpart to the `User` deletion warning**

The `users` rule below includes `DELETE` and produces the "the rule still grants this email" warning; this `groups` rule covers only `CREATE`/`UPDATE`. Deleting a `Group` whose name is a rule subject leaves the rule granting that name just the same, and nothing says so. The asymmetry is not explained in the description.

modules/140-user-authz/webhooks/validating/identity_collision_test_factories.py
```diff
@@ -0,0 +1,137 @@
...
+CLUSTER_RULE_GROUP_SUBJECT = "privileged-group"
+CLUSTER_RULE_USER_SUBJECT = "privileged@example.com"
+NAMESPACED_RULE_GROUP_SUBJECT = "team-devs"
+NAMESPACED_RULE_USER_SUBJECT = "team-lead@example.com"
+
+CLUSTER_RULE_DESCRIPTION = 'clusterauthorizationrules.deckhouse.io "admin-rule"'
+NAMESPACED_RULE_DESCRIPTION = 'authorizationrules.deckhouse.io "team-a/team-rule"'
+
+
+def _snapshots(with_rules: bool = True) -> dict:
```
AlwxSin: 🔵 **Two paths the new tests never reach**

- The factories hand-build `filterResult`, so neither `jqFilter` is ever executed. The `ServiceAccount` filtering and the `[]?` guard against a rule with no `subjects` are verified only by the manual `jq` run described in the PR body — a change to either filter would not be caught. (Same shape as `multitenancy_test.py`, so this is the module's existing convention rather than a regression.)
- The `name_is_new` fail-closed path for an UPDATE whose `oldObject` cannot be read is called out in a comment (`identity_collision.py:164-166`) but has no test: `prepare_group_binding_context` and `prepare_user_binding_context` always supply an `oldObject` for UPDATE.

modules/140-user-authz/webhooks/validating/identity_collision.py
```diff
@@ -0,0 +1,227 @@
...
+  - apiGroups:   ["deckhouse.io"]
+    apiVersions: ["*"]
+    operations:  ["CREATE", "UPDATE", "DELETE"]
+    resources:   ["users"]
+    scope:       "Cluster"
+kubernetes:
+- name: {CLUSTER_RULES_SNAPSHOT_NAME}
+  apiVersion: deckhouse.io/v1alpha1
+  kind: ClusterAuthorizationRule
+  executeHookOnEvent: []
```
AlwxSin, line 72: 🔵 **The description claims `group: main` on the kubernetes bindings; the code has no such field**

*"Two `kubernetes` bindings with `group: main`, `executeHookOnEvent: []`, …"* — the `CONFIG` has no `group` key. It is not needed, since `includeSnapshotsFrom` names both snapshots explicitly, but the description is the review artifact for a security change and should not assert something the code does not do.

modules/140-user-authz/webhooks/validating/identity_collision.py
Jabejixo: Fixed in bb7edadd92. `granting_rule_for` now compares the identity that reaches the token, not the spec field: for `User` that is `spec.email.lower()`, while rule subjects stay indexed verbatim, so a lowercase subject is caught against any case of the incoming email and a mixed-case subject is still not reported.

The docstring on `granting_rule_for` now names both directions explicitly, with the citations for each: `get_dex_user_crds.go:276` for the lowercasing, and `cluster-role-bindings.yaml:38` / `handle_manage_bindings.go:256` for subjects reaching the RoleBinding verbatim.

`identity_collision_test.py:124` is gone. It is replaced by two tests: `test_user_denied_when_email_reaches_the_subject_only_after_lowercasing` (mixed-case incoming email, lowercase subject, now denied) and `test_user_allowed_when_the_rule_subject_itself_is_mixed_case` (mixed-case subject, lowercase email, still allowed). `test_user_deletion_warns_for_a_mixed_case_email` covers the DELETE warning you mention.

Two follow-ups from this that were not in the comment:

- The `name_is_new` check for UPDATE now compares token identities too, so rewriting an already colliding email from one case to another warns instead of denying. Otherwise the fix would have started denying updates to exactly the legacy objects the design promises to keep modifiable.
- `Group.spec.name` deliberately keeps the verbatim comparison in both directions. Nothing normalises a group name on its way to the `groups` claim: `makeUserGroupsMap` accumulates `group.Spec.Name` byte for byte (`get_dex_user_crds.go:697,706`) and it is written to `Password.groups` unchanged (`:503`, `:563`, `:596`). `test_group_allowed_when_name_differs_from_the_subject_only_by_case` and `test_group_allowed_when_the_rule_subject_differs_only_by_case` pin that down so the email fix is not copied onto groups later.

modules/140-user-authz/webhooks/validating/identity_collision.py
Jabejixo: Correct, the caveat was too narrow. The `impact:` text now says that any declarative flow managing both the rule and the identity is affected, that a first-time apply denies the `Group`/`User` whenever the rule happens to be applied first, that nothing guarantees the order, and that the annotation should be added to such objects before upgrading. The delete-and-recreate wording is dropped as a special case of that.

The changelog block was validated with the repo's own tooling before pushing: `md-parser.js` extracts one fenced `changes` block under the heading and `pr-changes-validator.js` reports "Changes is valid" for it.

modules/140-user-authz/webhooks/validating/identity_collision.py
Jabejixo: Fixed. `group: main` is removed from the description; the sentence now lists only the fields `CONFIG` actually sets and states why no `group` and no `queue` are needed.

The rest of the body was re-read against the code for the same class of drift. Two other things were corrected: the `_authentication_configuration.tpl` reference now cites the concrete lines and claim mappings instead of just the file, and the "email comparison is exact" paragraph is replaced, since it no longer describes the code and its reasoning was wrong (see the thread on `granting_rule_for`). The backport paragraph is also corrected: it claimed the pick could not conflict by construction, which stopped being true once `testing/webhooks/test.sh` was touched, and it now reports the actual per-branch result.

modules/140-user-authz/webhooks/validating/identity_collision.py
Jabejixo: Fixed rather than explained: the asymmetry was an oversight, not a decision. The `groups` rule now includes `DELETE`, and `warn_rule_outlives_user` became `warn_rule_outlives_identity`, driven by the same `IDENTITY_KINDS` table as the rest of the dispatch, so both kinds take the same path and the two cannot drift apart again.

Deleting a `Group` whose name is a rule subject now warns with the rule name, exactly like the `User` case, and does not block. Tests: `test_group_deletion_warns_that_the_rule_survives_it` and `test_group_deletion_is_silent_when_no_rule_grants_the_name`. Restricting the `DELETE` branch back to `User` makes the first of them fail.

The description no longer describes deletion as a `User`-only behaviour.

modules/140-user-authz/webhooks/validating/identity_collision_test_factories.py
Jabejixo: Both gaps are closed.

**jqFilter.** A new `TestSnapshotJQFilters` class parses `CONFIG` as YAML, pulls both `jqFilter` programs out of it and runs them through `jq` against real rule objects, so the filters are executed rather than assumed. It checks that `ServiceAccount` subjects are dropped from a rule carrying all three subject kinds, that the `[]?` guard survives a rule with no `spec.subjects` (both filters), and that the filter output feeds `granting_rule_for` unchanged, which also ties the projection to the consumer. `jq` was added to the `apk add` line in `testing/webhooks/test.sh` so the CI container has it; the class is skipped, not failed, if it is missing. Dropping `select(.kind == "Group")` fails the two subject tests and dropping the `?` fails both subtests of the no-subjects test.

**`name_is_new` fail-closed.** `prepare_group_binding_context` and `prepare_user_binding_context` take `with_old_object=False`, which builds an UPDATE with a null `oldObject`. Covered by `test_group_denied_when_the_old_object_cannot_be_read` and `test_user_denied_when_the_old_object_cannot_be_read`; removing the `isinstance(old_name, str)` guard breaks the user one.

Every new guard was checked by mutation, and the failing test for each is listed in a table in the PR description. Suite is 28 tests, and `multitenancy_test.py` still passes unchanged.

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [22352](https://github.com/deckhouse/deckhouse/pull/22352) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit de2abe5b4ea07e6f6536d358772267b21d3e4da3 to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/32460011853) for details.

PR #22219, opened 2026-08-17
[user-authn] Gate the allow-access-to-kubernetes annotation on DexClient and DexAuthenticator

Jabejixo:
> Based on #22222 (`chore/helm-test-harness-api-versions`), which carries the Helm test harness change this PR's tests need. Merge #22222 first; this PR targets that branch and should be retargeted to `main` once it lands.

## Description

`DexClient` and `DexAuthenticator` are namespaced resources. Two annotations on them act as cluster-level privilege grants:

- `dexclient.deckhouse.io/allow-access-to-kubernetes`
- `dexauthenticator.deckhouse.io/allow-access-to-kubernetes`

The annotation adds the tenant's OAuth2 client to `trustedPeers` of the privileged `kubernetes` OAuth2Client (`modules/150-user-authn/templates/dex/oauth2client.yaml`). Being a trusted peer lets the client ask Dex for cross-client ID tokens with `aud: kubernetes` — tokens the kube-apiserver accepts, mapping `username` from the `email` claim and `groups` from the `groups` claim with an empty prefix (`modules/040-control-plane-manager/templates/_authentication_configuration.tpl`).

Until now nothing constrained who set those annotations. The hooks (`hooks/get_dex_client_crds.go`, `hooks/get_dex_authenticator_crds.go`) only read the annotation off the object; they check neither the subject nor the namespace. The module's only ValidatingAdmissionPolicies target `users`.

This PR adds a `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` named `dex-allow-access-to-kubernetes.deckhouse.io` to `modules/150-user-authn/templates/validation.yaml`, matching `deckhouse.io` `dexclients` and `dexauthenticators` on `CREATE` and `UPDATE`, with `failurePolicy: Fail` and `validationActions: [Deny]`, inside the existing `helm_lib_kind_exists` guard.

### What is gated: presence, not truthiness

The policy gates the **presence** of the annotation key, not the value written into it.

Two hook generations read this annotation differently. Since #21897 (`41ef802ec8` on `main`, `0488dada7b` on `release-1.77`) the hooks parse the value with `strconv.ParseBool` and fail closed on anything else. Before that — and still on `release-1.76`, `hooks/get_dex_client_crds.go:98` and `hooks/get_dex_authenticator_crds.go:78` — the hooks grant on the mere presence of the key, whatever the value is.

A policy that only looked at truthiness would therefore be sound on one branch and bypassable on the other: on the presence-honouring hook a value the policy reads as harmless still switches the grant on. Gating presence removes the dependency on the hook's parsing altogether, so the same policy is correct wherever it is deployed and stays correct if the parsing is ever changed again.

Truthiness is still used, but only in one direction: on `UPDATE` it decides whether the change makes an already annotated object *more* permissive. That keeps objects that already carry the annotation — including ones carrying a non-truthy value — editable by their owners, while still denying a non-truthy value being flipped to a truthy one.

### Authorisation predicate

```
authorizer.group("deckhouse.io").resource("moduleconfigs").name("user-authn").check("update").allowed()
```

The right to update the `user-authn` ModuleConfig is the cluster-wide authority over this module, and it is exactly what the shipped `d8:manage:permission:module:user-authn:edit` role grants (`resourceNames: [user-authn]` on `moduleconfigs`). It is *not* granted by the roles that let a tenant create these objects in the first place: legacy access level `Editor` (`d8:user-authz:user-authn:editor`) and rbacv2 `d8:use:capability:module:user-authn:edit`, which aggregates into `d8:namespace:manager`. A cluster administrator passes; a namespace manager does not. Using the CEL `authorizer` is what makes a policy sufficient here — a webhook would need its own RBAC informer to answer the same question.

The Deckhouse service account is exempted explicitly, the same way `system-users.deckhouse.io` does it in the same file. That covers the DexAuthenticators shipped by modules with the annotation already set (Kiali, Cilium Hubble UI, console), because the Deckhouse controller runs as `system:serviceaccount:d8-system:deckhouse` and applies them itself.

### Backward compatibility

The policy gates the *transition*, not the presence of the annotation on an existing object: `oldObject` is consulted so that an object which already carries the annotation stays fully editable by its owner. Case by case:

| Case | Outcome |
| --- | --- |
| CREATE, no annotation | Allow |
| CREATE, annotation present with any value, unauthorised subject | Deny |
| CREATE, annotation present, subject may update the `user-authn` ModuleConfig | Allow |
| CREATE, annotation present, `system:serviceaccount:d8-system:deckhouse` | Allow |
| UPDATE, annotation present before and after, value unchanged | Allow |
| UPDATE, `"true"` rewritten as `"1"` | Allow |
| UPDATE, `"false"` rewritten as `"false"` | Allow |
| UPDATE, annotation newly added | Deny |
| UPDATE, a value no hook reads as true flipped to `"true"` | Deny |
| UPDATE, `"true"` flipped to `"false"` | Allow |
| UPDATE, annotation removed | Allow |
| `DexClient` carrying the `dexauthenticator.*` key, and the mirror case | Allow (neither hook reads it) |

The one behaviour that changes relative to a truthiness-based gate is that an unauthorised subject can no longer *introduce* the annotation with a non-truthy value such as `"false"`. That value is a grant on the presence-honouring hook, and omitting the annotation expresses the same intent everywhere.

### Operational hazard: delete-and-recreate, and GitOps

Existing objects keep working **only for as long as they are updated in place**. The annotation is evaluated on `CREATE` as a new grant regardless of what existed before, because a deleted object leaves no `oldObject` to compare against. Anything that replaces an object rather than patching it will therefore lose Kubernetes API access:

- ArgoCD or Flux prune-and-recreate, a `replace`-style sync, or recreation triggered by an immutable-field change.
- A tenant re-applying a manifest from scratch after deleting the object.

In each case the `CREATE` carrying the annotation is denied, and if the annotation is stripped to get the object accepted, it comes back **without** access to the Kubernetes API.

Operators managing such objects through GitOps should do one of the following:

- Grant the GitOps controller's service account permission to manage the `user-authn` ModuleConfig (`update` on `moduleconfigs` named `user-authn`), so its `CREATE` requests satisfy the policy — noting this is a genuine cluster-level privilege and should be a deliberate decision.
- Or ensure these objects are updated in place rather than replaced, and have a cluster administrator perform any re-creation.

This tightening is intentional: allowing recreation to restore the grant would leave the problem open to any tenant who can delete and recreate their own object.

### Not included

No hook-side namespace allowlist. It would change behaviour for existing legitimate objects in arbitrary namespaces and therefore need an opt-out ModuleConfig setting, while the admission policy already closes the gap without breaking anything that exists. Worth a follow-up as defence in depth.

No gating of `spec` changes on an object that already carries the grant (`redirectURIs`, `allowedGroups`, `allowedEmails`, `applicationDomain`). Doing that would freeze those fields on every already-granted object, retroactively and with no self-service remedy, which is a wider break than the one this PR makes — the gate here only affects grants being introduced. It is a separate decision and belongs in its own change, most likely behind a ModuleConfig setting.

### Tests

`modules/150-user-authn/template_tests/validation_test.go` asserts that the policy and the binding render when the VAP kinds are available and are skipped when they are not, and evaluates the rendered CEL through the same compiler and evaluator kube-apiserver uses (`k8s.io/apiserver/pkg/admission/plugin/cel`), with a stub authorizer, over all the cases in the table above. The policy is decoded strictly into `admissionregistration.k8s.io/v1`, so a misspelled field fails the test too. Making the VAP kinds visible to the renderer relies on `WithAPIVersions` from #22222.

Two properties are pinned beyond the case table:

- The policy must deny every annotation value that *any* hook generation reads as a grant. Both hook models are expressed as code in the test and the value corpus is run through them, so widening a hook without widening the policy turns into a test failure rather than a silently reopened gap. The same case also asserts that an authorised subject is still allowed to set each of those values, so the assertion cannot be satisfied by a policy that denies everything.
- The CEL environment is pinned to the lowest Kubernetes version listed in `candi/version_map.yml` for the branch under test, and expressions are compiled in `environment.NewExpressions` mode. `environment.StoredExpressions` deliberately exposes every CEL library so persisted policies keep evaluating, so it would accept an expression the oldest supported cluster cannot compile. With `failurePolicy: Fail` that would deny every `DexClient` and `DexAuthenticator` write on such a cluster.

A namespace-scoped subject allowed to create `DexClient`/`DexAuthenticator` could grant itself cluster-level access to the Kubernetes API by adding an annotation to its own namespaced object, with no cluster-level authority involved. Granting Kubernetes API access is a cluster-administration decision, and it now requires cluster-level authority.

It closes a privilege boundary that a namespace-level role could cross, and it is non-breaking for objects that already carry the annotation and are updated in place.

Because the gate is on the presence of the annotation rather than on its value, it does not depend on the hook parsing introduced by #21897, and therefore needs no prerequisite backport. Verified by cherry-picking #22222's commit, this PR's original commit, and the presence-gating commit into detached worktrees off both release branches and running `go test ./modules/150-user-authn/... -count=1`:

- `release-1.77` — three cherry-picks clean, module suite green; the CEL environment resolves to 1.32 from that branch's `candi/version_map.yml`.
- `release-1.76` — three cherry-picks clean, module suite green; the CEL environment resolves to 1.31, and the policy's expressions compile there.

`ValidatingAdmissionPolicy` is GA since Kubernetes 1.30 and both release branches support 1.31 and newer, so the `helm_lib_kind_exists` guard only matters for clusters where the API is genuinely absent.

## Checklist
- [ ] e2e tests passed.
- [x] Documentation updated according to the changes. Handled in #22250, which covers the client documentation for this series; the CI/CD admin guide needs the new privilege requirement and the new troubleshooting case.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Adding the allow-access-to-kubernetes annotation to a DexClient or a DexAuthenticator now requires cluster-level authority over the user-authn module.
impact: Only a subject allowed to update the user-authn ModuleConfig, or the Deckhouse service account, may add the `dexclient.deckhouse.io/allow-access-to-kubernetes` or `dexauthenticator.deckhouse.io/allow-access-to-kubernetes` annotation. Objects that already carry it keep working and stay editable by their owners as long as they are updated in place. A GitOps controller that deletes and recreates such an object instead of patching it will have the recreation denied, and the object will come back without access to the Kubernetes API unless the controller's service account is allowed to update the user-authn ModuleConfig.
impact_level: high
```

### DIFF
--- a/modules/150-user-authn/template_tests/validation_test.go
+++ b/modules/150-user-authn/template_tests/validation_test.go
@@ -1,0 +1,457 @@
+/*
+Copyright 2026 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package template_tests
+
+import (
+	"context"
+
+	celgo "github.com/google/cel-go/cel"
+	celtypes "github.com/google/cel-go/common/types"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/admission"
+	admissioncel "k8s.io/apiserver/pkg/admission/plugin/cel"
+	celconfig "k8s.io/apiserver/pkg/apis/cel"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	"k8s.io/apiserver/pkg/cel/environment"
+	"sigs.k8s.io/yaml"
+
+	. "github.com/deckhouse/deckhouse/testing/helm"
+)
+
+const (
+	validatingAdmissionPolicyAPIVersion        = "admissionregistration.k8s.io/v1/ValidatingAdmissionPolicy"
+	validatingAdmissionPolicyBindingAPIVersion = "admissionregistration.k8s.io/v1/ValidatingAdmissionPolicyBinding"
+
+	allowAccessPolicyName = "dex-allow-access-to-kubernetes.deckhouse.io"
+
+	dexClientAnnotation        = "dexclient.deckhouse.io/allow-access-to-kubernetes"
+	dexAuthenticatorAnnotation = "dexauthenticator.deckhouse.io/allow-access-to-kubernetes"
+
+	deckhouseServiceAccount = "system:serviceaccount:d8-system:deckhouse"
+	namespaceEditor         = "editor@example.com"
+)
+
+var _ = Describe("Module :: user-authn :: helm template :: validation", func() {
+	hec := SetupHelmConfig("")
+
+	BeforeEach(func() {
+		hec.ValuesSet("global.discovery.kubernetesVersion", "1.32.0")
+		hec.ValuesSet("global.modules.publicDomainTemplate", "%s.example.com")
+		hec.ValuesSet("global.modules.https.mode", "CertManager")
+		hec.ValuesSet("global.modules.https.certManager.clusterIssuerName", "letsencrypt")
+		hec.ValuesSet("global.modulesImages.registry.base", "registry.example.com")
+		hec.ValuesSet("global.enabledModules", []string{"cert-manager", "vertical-pod-autoscaler"})
+		hec.ValuesSet("global.discovery.d8SpecificNodeCountByRole.system", 2)
+		hec.ValuesSet("global.discovery.kubernetesCA", "plainstring")
+
+		hec.ValuesSet("userAuthn.internal.kubernetesDexClientAppSecret", "plainstring")
+		hec.ValuesSet("userAuthn.internal.dexTLS.crt", "plainstring")
+		hec.ValuesSet("userAuthn.internal.dexTLS.key", "plainstring")
+		hec.ValuesSet("userAuthn.internal.dexTLS.ca", "plainstring")
+	})
+
+	Context("Without ValidatingAdmissionPolicy support in the cluster", func() {
+		BeforeEach(func() {
+			hec.HelmRender()
+		})
+
+		It("Should not render the policies", func() {
+			Expect(hec.RenderError).ShouldNot(HaveOccurred())
+			Expect(hec.KubernetesGlobalResource("ValidatingAdmissionPolicy", allowAccessPolicyName).Exists()).To(BeFalse())
+			Expect(hec.KubernetesGlobalResource("ValidatingAdmissionPolicyBinding", allowAccessPolicyName).Exists()).To(BeFalse())
+			Expect(hec.KubernetesGlobalResource("ValidatingAdmissionPolicy", "system-users.deckhouse.io").Exists()).To(BeFalse())
+		})
+	})
+
+	Context("With ValidatingAdmissionPolicy support in the cluster", func() {
+		BeforeEach(func() {
+			hec.HelmRender(WithAPIVersions(
+				validatingAdmissionPolicyAPIVersion,
+				validatingAdmissionPolicyBindingAPIVersion,
+			))
+		})
+
+		It("Should render the policy gating the allow-access-to-kubernetes annotation", func() {
+			Expect(hec.RenderError).ShouldNot(HaveOccurred())
+
+			policy := hec.KubernetesGlobalResource("ValidatingAdmissionPolicy", allowAccessPolicyName)
+			Expect(policy.Exists()).To(BeTrue())
+			Expect(policy.Field("apiVersion").String()).To(Equal("admissionregistration.k8s.io/v1"))
+			Expect(policy.Field("metadata.labels.heritage").String()).To(Equal("deckhouse"))
+			Expect(policy.Field("metadata.labels.module").String()).To(Equal("user-authn"))
+			Expect(policy.Field("spec.failurePolicy").String()).To(Equal("Fail"))
+
+			rule := policy.Field("spec.matchConstraints.resourceRules.0")
+			Expect(rule.Get("apiGroups").String()).To(MatchJSON(`["deckhouse.io"]`))
+			Expect(rule.Get("apiVersions").String()).To(MatchJSON(`["*"]`))
+			Expect(rule.Get("operations").String()).To(MatchJSON(`["CREATE","UPDATE"]`))
+			Expect(rule.Get("resources").String()).To(MatchJSON(`["dexclients","dexauthenticators"]`))
+
+			Expect(policy.Field("spec.validations.0.reason").String()).To(Equal("Forbidden"))
+			Expect(policy.Field("spec.validations.0.messageExpression").String()).To(
+				ContainSubstring("grants access to the Kubernetes API"),
+			)
+			Expect(policy.Field("spec.validations.0.messageExpression").String()).To(
+				ContainSubstring("must be applied by a cluster administrator"),
+			)
+
+			binding := hec.KubernetesGlobalResource("ValidatingAdmissionPolicyBinding", allowAccessPolicyName)
+			Expect(binding.Exists()).To(BeTrue())
+			Expect(binding.Field("apiVersion").String()).To(Equal("admissionregistration.k8s.io/v1"))
+			Expect(binding.Field("spec.policyName").String()).To(Equal(allowAccessPolicyName))
+			Expect(binding.Field("spec.validationActions").String()).To(MatchJSON(`["Deny"]`))
+		})
+
+		It("Should keep rendering the pre-existing user validation policy", func() {
+			Expect(hec.KubernetesGlobalResource("ValidatingAdmissionPolicy", "system-users.deckhouse.io").Exists()).To(BeTrue())
+			Expect(hec.KubernetesGlobalResource("ValidatingAdmissionPolicyBinding", "system-users.deckhouse.io").Exists()).To(BeTrue())
+		})
+
+		It("Should evaluate the rendered CEL expressions as intended", func() {
+			policy := compileRenderedPolicy(
+				hec.KubernetesGlobalResource("ValidatingAdmissionPolicy", allowAccessPolicyName).ToYaml(),
+			)
+
+			cases := []struct {
+				name           string
+				resource       string
+				kind           string
+				operation      admission.Operation
+				username       string
+				canManage      bool
+				annotations    map[string]string
+				oldAnnotations map[string]string
+				allowed        bool
+			}{
+				{
+					name:        "unauthorised subject grants access on create",
+					annotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:     false,
+				},
+				{
+					name:        "unauthorised subject creates a client without the annotation",
+					annotations: nil,
+					allowed:     true,
+				},
+				{
+					name:        "annotation explicitly disabled",
+					annotations: map[string]string{dexClientAnnotation: "false"},
+					allowed:     true,
+				},
+				{
+					name:        "annotation value is not a boolean the hooks would honour",
+					annotations: map[string]string{dexClientAnnotation: "yes"},
+					allowed:     true,
+				},
+				{
+					name:        "truthy value in another casing",
+					annotations: map[string]string{dexClientAnnotation: "True"},
+					allowed:     false,
+				},
+				{
+					name:        "truthy value as a single character",
+					annotations: map[string]string{dexClientAnnotation: "t"},
+					allowed:     false,
+				},
+				{
+					name:        "truthy value as a digit",
+					annotations: map[string]string{dexClientAnnotation: "1"},
+					allowed:     false,
+				},
+				{
+					name:        "deckhouse service account grants access",
+					username:    deckhouseServiceAccount,
+					annotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:     true,
+				},
+				{
+					name:        "subject allowed to update the user-authn ModuleConfig grants access",
+					canManage:   true,
+					annotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:     true,
+				},
+				{
+					name:           "existing grant is carried over unchanged",
+					operation:      admission.Update,
+					annotations:    map[string]string{dexClientAnnotation: "true"},
+					oldAnnotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:        true,
+				},
+				{
+					name:           "existing grant is rewritten with an equivalent value",
+					operation:      admission.Update,
+					annotations:    map[string]string{dexClientAnnotation: "1"},
+					oldAnnotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:        true,
+				},
+				{
+					name:           "annotation is added by an update",
+					operation:      admission.Update,
+					annotations:    map[string]string{dexClientAnnotation: "true"},
+					oldAnnotations: nil,
+					allowed:        false,
+				},
+				{
+					name:           "annotation is flipped from false to true",
+					operation:      admission.Update,
+					annotations:    map[string]string{dexClientAnnotation: "true"},
+					oldAnnotations: map[string]string{dexClientAnnotation: "false"},
+					allowed:        false,
+				},
+				{
+					name:           "annotation is flipped from true to false",
+					operation:      admission.Update,
+					annotations:    map[string]string{dexClientAnnotation: "false"},
+					oldAnnotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:        true,
+				},
+				{
+					name:           "annotation is dropped by an update",
+					operation:      admission.Update,
+					annotations:    nil,
+					oldAnnotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:        true,
+				},
+				{
+					name:        "dex authenticator grants access on create",
+					resource:    "dexauthenticators",
+					kind:        "DexAuthenticator",
+					annotations: map[string]string{dexAuthenticatorAnnotation: "true"},
+					allowed:     false,
+				},
+				{
+					name:           "dex authenticator keeps an existing grant",
+					resource:       "dexauthenticators",
+					kind:           "DexAuthenticator",
+					operation:      admission.Update,
+					annotations:    map[string]string{dexAuthenticatorAnnotation: "true"},
+					oldAnnotations: map[string]string{dexAuthenticatorAnnotation: "true"},
+					allowed:        true,
+				},
+				{
+					name:        "annotation of the other kind is not gated",
+					annotations: map[string]string{dexAuthenticatorAnnotation: "true"},
+					allowed:     true,
+				},
+			}
+
+			for _, tc := range cases {
+				By(tc.name)
+
+				resource, kind := tc.resource, tc.kind
+				if resource == "" {
+					resource, kind = "dexclients", "DexClient"
+				}
+				operation := tc.operation
+				if operation == "" {
+					operation = admission.Create
+				}
+				username := tc.username
+				if username == "" {
+					username = namespaceEditor
+				}
+
+				allowed, err := policy.validate(dexAdmissionInput{
+					resource:       resource,
+					kind:           kind,
+					operation:      operation,
+					username:       username,
+					canManage:      tc.canManage,
+					annotations:    tc.annotations,
+					oldAnnotations: tc.oldAnnotations,
+				})
+
+				Expect(err).ToNot(HaveOccurred(), tc.name)
+				Expect(allowed).To(Equal(tc.allowed), tc.name)
+			}
+		})
+	})
+})
+
+// compiledPolicy evaluates the validation expressions of a ValidatingAdmissionPolicy through the
+// same CEL machinery kube-apiserver uses, so the expressions the module ships are under test.
+type compiledPolicy struct {
+	validations admissioncel.ConditionEvaluator
+}
+
+func compileRenderedPolicy(renderedPolicy string) compiledPolicy {
+	// Strict decoding also asserts that the rendered manifest carries no field the API does not know.
+	var policy admissionregistrationv1.ValidatingAdmissionPolicy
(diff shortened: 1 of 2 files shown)


modules/150-user-authn/templates/validation.yaml
```diff
@@ -27,6 +27,59 @@ metadata:
...
+    # The annotation is named after the kind it belongs to, so the key depends on the matched resource.
+    - name: annotation
+      expression: 'request.resource.resource == "dexclients"
+        ? "dexclient.deckhouse.io/allow-access-to-kubernetes"
+        : "dexauthenticator.deckhouse.io/allow-access-to-kubernetes"'
+    - name: annotations
+      expression: 'object.metadata.?annotations.orValue({})'
+    - name: oldAnnotations
+      expression: 'oldObject == null ? {} : oldObject.metadata.?annotations.orValue({})'
+    # The truthy set is a superset of what strconv.ParseBool accepts in the hooks, so no value that
```
AlwxSin: 🔴 **The proposed `release-1.76` backport leaves the gate fully bypassable**

This "superset of `strconv.ParseBool`" invariant is a property of the *hook*, and it only exists on `main` / `release-1.77` — it arrived with #21897 (`41ef802ec8`). On `release-1.76` the hooks still grant on mere key presence:

- `modules/150-user-authn/hooks/get_dex_client_crds.go:98` — `_, allowAccessToKubernetes := obj.GetAnnotations()["dexclient.deckhouse.io/allow-access-to-kubernetes"]`
- `modules/150-user-authn/hooks/get_dex_authenticator_crds.go:78` — same shape

`git grep allow-access-to-kubernetes origin/release-1.76 -- modules/150-user-authn/` returns only those two sites plus hook tests, so there is no other gate on that branch. A namespace-level subject on 1.76 can therefore create:

```yaml
metadata:
  annotations:
    dexclient.deckhouse.io/allow-access-to-kubernetes: "false"
```

`grantsAccess` evaluates false, `!variables.grantsAccess` short-circuits the validation, the object is admitted — and the 1.76 hook then sees the key exists, sets `AllowAccessToKubernetes: true`, so `templates/dex/oauth2client.yaml:31` puts the client into `trustedPeers` of the privileged `kubernetes` OAuth2Client. The exact escalation this PR closes stays open behind one character.

The diff is correct on its own base; the defect is in the backport plan stated in the description ("cherry-picks cleanly onto both release branches"). Either backport #21897 to `release-1.76` first, or ship a 1.76-specific variant that gates on annotation *presence* rather than truthiness.

modules/150-user-authn/template_tests/validation_test.go
```diff
@@ -0,0 +1,457 @@
...
+					annotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:     false,
+				},
+				{
+					name:        "unauthorised subject creates a client without the annotation",
+					annotations: nil,
+					allowed:     true,
+				},
+				{
+					name:        "annotation explicitly disabled",
```
AlwxSin: 🟡 **Nothing enforces the policy ↔ hook truthy-set invariant**

The coupling the policy's soundness rests on — policy truthy set ⊇ hook truthy set — lives only in a YAML comment (`templates/validation.yaml:55`) and in the PR description. Hook tests sit in `hooks/`, policy tests here in `template_tests/`, and neither references the other. If the hook is later widened (accepts `"yes"` / `"on"`, or reverts to a presence check), the policy silently reopens and every existing test stays green.

Suggestion: a case that enumerates the values `strconv.ParseBool` reports as true — `1 t T TRUE true True` — and asserts the rendered CEL denies each one, so the two sides cannot drift apart unnoticed.

modules/150-user-authn/templates/validation.yaml
```diff
@@ -27,6 +27,59 @@ metadata:
...
+  validations:
+    # Only a newly granted access is gated, so objects that already carry the annotation stay editable
+    # by their owners. Authority is proven by the right to update the user-authn ModuleConfig, which
+    # only cluster-wide administrators of this module hold.
+    - expression: 'request.userInfo.username == "system:serviceaccount:d8-system:deckhouse"
+        || !variables.grantsAccess
+        || variables.grantedBefore
+        || authorizer.group("deckhouse.io").resource("moduleconfigs").name("user-authn").check("update").allowed()'
+      reason: Forbidden
+      messageExpression: '"the " + variables.annotation + " annotation grants access to the Kubernetes API
```
AlwxSin: 🟡 **Documentation not updated for a breaking behavioural change**

(Real location: `docs/documentation/pages/admin/configuration/access/authorization/CI_CD.md:304` and `:479`, plus `CI_CD_RU.md:305` / `:480`. Those files are not in this diff, so the comment is anchored on the message the change produces.)

The CI/CD admin guide hands the user a ready-to-run `d8 k apply` manifest carrying `dexclient.deckhouse.io/allow-access-to-kubernetes: "true"`, with no mention of the privilege now required — after this PR that command fails for anyone without `update` on the `user-authn` ModuleConfig. The troubleshooting list at `:479` also gains a new failure mode: the annotation silently absent because the `CREATE` was denied, or because GitOps recreated the object without it. The RU copies need the same treatment, and the Documentation checkbox in the PR is still unchecked.

modules/150-user-authn/templates/validation.yaml
```diff
@@ -27,6 +27,59 @@ metadata:
...
+    - name: grantedBefore
+      expression: 'variables.annotation in variables.oldAnnotations
+        && variables.oldAnnotations[variables.annotation].lowerAscii() in ["1", "t", "true"]'
+  validations:
+    # Only a newly granted access is gated, so objects that already carry the annotation stay editable
+    # by their owners. Authority is proven by the right to update the user-authn ModuleConfig, which
+    # only cluster-wide administrators of this module hold.
+    - expression: 'request.userInfo.username == "system:serviceaccount:d8-system:deckhouse"
+        || !variables.grantsAccess
+        || variables.grantedBefore
```
AlwxSin: 🟡 **The grant binds to the object, not to its spec**

`|| variables.grantedBefore` keeps a granted object fully editable by its namespace owner — deliberately, per the description. But the grant yields `aud: kubernetes` for whatever the spec says *now*, and `crds/dex-client.yaml` leaves `redirectURIs` (:54), `allowedGroups` (:47) and `trustedPeers` (:59) mutable; the same holds for DexAuthenticator's `applicationDomain` / `allowedEmails`.

So once an administrator grants a tenant's DexClient, the tenant can repoint `redirectURIs` at an attacker-controlled callback and widen `allowedGroups` to allow-all, while retaining tokens the kube-apiserver accepts. Narrower than the escalation being closed — it needs one admin grant first — but the same primitive, and this policy is the natural place to gate it: e.g. also require authority when `redirectURIs` / `allowedGroups` change on an object that already carries the grant.

modules/150-user-authn/template_tests/validation_test.go
```diff
@@ -0,0 +1,457 @@
...
+	// Strict decoding also asserts that the rendered manifest carries no field the API does not know.
+	var policy admissionregistrationv1.ValidatingAdmissionPolicy
+	Expect(yaml.UnmarshalStrict([]byte(renderedPolicy), &policy)).To(Succeed())
+
+	// kube-apiserver declares the authorizer for validations but not for message expressions.
+	validationVars := admissioncel.OptionalVariableDeclarations{HasAuthorizer: true, StrictCost: true}
+	messageVars := admissioncel.OptionalVariableDeclarations{StrictCost: true}
+
+	compiler, err := admissioncel.NewCompositedCompiler(
+		environment.MustBaseEnvSet(environment.DefaultCompatibilityVersion(), true),
```
AlwxSin: 🔵 **Test pins the vendored compat version, not the branch minimum**

`environment.DefaultCompatibilityVersion()` on `k8s.io/apiserver v0.34.8` resolves to 1.33 (`MinCompatibilityVersion()` of the build version), while `candi/version_map.yml` supports k8s 1.32+ on `main` / `release-1.77` and 1.31+ on `release-1.76`. Today's expressions are safe — everything used is gated at ≤1.29, `cel.OptionalTypes()` at 1.28 — but with `failurePolicy: Fail`, a future expression relying on a 1.32+ CEL feature would compile green here and then deny *every* DexClient/DexAuthenticator write on a 1.31 cluster. Pinning the env set to the branch's minimum supported version turns that into a test failure instead of a cluster-wide outage.

modules/150-user-authn/template_tests/validation_test.go
```diff
@@ -0,0 +1,457 @@
...
+					annotations: nil,
+					allowed:     true,
+				},
+				{
+					name:        "annotation explicitly disabled",
+					annotations: map[string]string{dexClientAnnotation: "false"},
+					allowed:     true,
+				},
+				{
+					name:        "annotation value is not a boolean the hooks would honour",
```
AlwxSin: 🔵 **Small gaps in the case table**

Two cheap additions: the empty-string value (`annotation: ""` — allowed by the policy, fails closed in the hook, currently untested), and the mirror of "annotation of the other kind is not gated" (line 253) — a `DexAuthenticator` carrying the `dexclient.*` key. One struct literal each, and together they pin the `variables.annotation` ternary from both sides.

modules/150-user-authn/templates/validation.yaml
Jabejixo: Confirmed and fixed by making the policy self-sufficient rather than by ordering the backports.

`git show origin/release-1.76:modules/150-user-authn/hooks/get_dex_client_crds.go` line 98 and `get_dex_authenticator_crds.go` line 78 do grant on key presence, while `origin/main` (41ef802ec8) and `origin/release-1.77` (0488dada7b) parse the value with `strconv.ParseBool`. 41ef802ec8 is an ancestor of `origin/main` only, so the two release branches genuinely differ.

The policy now gates the presence of the annotation key, not its value (`modules/150-user-authn/templates/validation.yaml:57-70`). Truthiness is kept for one purpose only: on UPDATE it decides whether the change makes an already annotated object more permissive, so an object that already carries the key — including with a non-truthy value — stays editable, while a non-truthy value flipped to a truthy one is still denied. The behavioural cost is that an unauthorised subject can no longer introduce the annotation with `"false"`; that value is a grant on the presence-honouring hook, and omitting the annotation expresses the same intent on every branch.

This removes the dependency on the hook entirely, so no prerequisite backport is needed and the gate cannot reopen if the parsing changes again. Cherry-picked #22222, the original commit and 91f9ea416d into detached worktrees off both release branches: clean on 1.77 and 1.76, `go test ./modules/150-user-authn/... -count=1` green on both. The backport section of the description has been rewritten accordingly.

modules/150-user-authn/template_tests/validation_test.go
Jabejixo: Turned into a test, and the coupling is largely gone anyway.

Both hook generations are now expressed as code in `modules/150-user-authn/template_tests/validation_test.go` (`hookGenerations`): the value-honouring one calls `strconv.ParseBool` — the same function the hook calls — and the presence-honouring one grants unconditionally. A value corpus covering everything `strconv.ParseBool` accepts in both polarities, plus `""`, `" "`, `"yes"`, `"on"`, `"2"`, `"TrUe"`, `"true "` and others, is run through both models, and every value either model reads as a grant must be denied for a subject without authority. Widening a hook means adding to that list, and the assertions start failing until the policy is widened too.

The same case also asserts that a subject holding `update` on the `user-authn` ModuleConfig is still allowed to set each of those values, so it cannot be satisfied by a policy that denies everything.

Evidence: run against the previous truthiness-only policy the case fails at `value "0" is honoured as a grant by the [presence honouring] hook`; against the current policy the suite is green (47 specs).

A test that imports the hook package and calls its parsing directly would be stronger still, but the logic is inline and unexported in `filterDexClient`/`filterDexAuthenticator`, and the hook package differs between the branches this change is backported to, so the model above is as close as it gets without a refactor of the hooks.

Since the policy now gates presence rather than truthiness, its soundness no longer depends on the hook at all; the test guards the weaker remaining property that the truthiness relaxation on UPDATE cannot drift.

modules/150-user-authn/templates/validation.yaml
Jabejixo: Agreed on the analysis, declining to implement it here — recorded in the description under "Not included".

The concern is real: `crds/dex-client.yaml` leaves `redirectURIs` (:54), `allowedGroups` (:47) and `trustedPeers` (:59) mutable, and DexAuthenticator likewise, so a granted object keeps yielding `aud: kubernetes` for whatever the spec says now.

What makes it a separate change is the blast radius. This PR only affects grants being *introduced*: everything that exists today keeps working and stays editable. Gating spec fields would apply to every object that already carries the grant, retroactively, with no self-service remedy — a tenant rotating a callback URL on an already-granted DexClient would need a cluster administrator, and there is no `oldObject` comparison that can grandfather it, because the fields being gated are exactly the ones that legitimately change. With `failurePolicy: Fail` and `validationActions: [Deny]` that is a wider break than the one being made here, and it needs a ModuleConfig opt-out to be shippable in a patch release.

It is also not a regression introduced by this PR: those spec fields are mutable today, with or without the policy.

Happy to do it as a follow-up — the natural shape is a second validation on the same policy, gated by a ModuleConfig setting, requiring the same authorizer check when `object.spec.redirectURIs != oldObject.spec.redirectURIs || object.spec.allowedGroups != oldObject.spec.allowedGroups` on an object where `variables.grantedBefore` holds.

modules/150-user-authn/templates/validation.yaml
Jabejixo: Agreed. Not editing those files here — the client documentation for this series is being collected in #22250, so the text goes there to avoid two PRs touching the same paragraphs.

The exact wording, anchored to the current lines, is below; it has been handed over to #22250.

**`CI_CD.md`, after line 309** (after "The annotation `dexclient.deckhouse.io/allow-access-to-kubernetes` allows the client to request tokens with `aud=kubernetes`."):

> {% alert level="warning" %}
> Adding the `dexclient.deckhouse.io/allow-access-to-kubernetes` annotation grants cluster-level access to the Kubernetes API, so it may only be set by a subject allowed to update the `user-authn` ModuleConfig — for example a holder of the `d8:manage:permission:module:user-authn:edit` role. A namespace-level role that permits creating a DexClient is not sufficient, and the request will be rejected by admission with a message naming the annotation.
>
> The check applies to the annotation being added or changed to a value that enables access, not to its presence on an existing object: a DexClient that already carries it stays editable by its owner. Deleting and recreating the object counts as adding the annotation again, so a GitOps controller that replaces objects instead of patching them needs `update` on the `moduleconfigs` resource named `user-authn` as well.
> {% endalert %}

**`CI_CD.md`, line 479**, replace the bullet:

> - Annotation `dexclient.deckhouse.io/allow-access-to-kubernetes` on DexClient. If it is missing, check whether the `d8 k apply` that created the object was rejected by admission, or whether a GitOps controller recreated the object without it: the annotation may only be added by a subject allowed to update the `user-authn` ModuleConfig.

**`CI_CD_RU.md`, after line 310** (after "Аннотация `dexclient.deckhouse.io/allow-access-to-kubernetes` позволяет клиенту запрашивать токены с `aud=kubernetes`."):

> {% alert level="warning" %}
> Добавление аннотации `dexclient.deckhouse.io/allow-access-to-kubernetes` выдаёт доступ к Kubernetes API на уровне кластера, поэтому её может установить только субъект, которому разрешено обновлять ModuleConfig `user-authn`, — например обладатель роли `d8:manage:permission:module:user-authn:edit`. Роли уровня пространства имён, позволяющей создать DexClient, недостаточно: запрос будет отклонён admission-политикой с сообщением, содержащим имя аннотации.
>
> Проверка выполняется при добавлении аннотации или изменении её значения на включающее доступ, а не при её наличии на существующем объекте: DexClient, который её уже содержит, остаётся доступным для редактирования владельцу. Удаление и повторное создание объекта считается повторным добавлением аннотации, поэтому GitOps-контроллеру, который пересоздаёт объекты вместо их обновления, также требуется право `update` на ресурс `moduleconfigs` с именем `user-authn`.
> {% endalert %}

**`CI_CD_RU.md`, line 480**, replace the bullet:

> - Аннотация `dexclient.deckhouse.io/allow-access-to-kubernetes` на DexClient. Если её нет, проверьте, не был ли отклонён admission-политикой `d8 k apply`, создавший объект, и не пересоздал ли GitOps-контроллер объект без неё: аннотацию может добавить только субъект, которому разрешено обновлять ModuleConfig `user-authn`.

The manifest at `CI_CD.md:297-307` / `CI_CD_RU.md:298-308` itself needs no change — it is addressed to a cluster administrator, who passes the check.

modules/150-user-authn/template_tests/validation_test.go
Jabejixo: Fixed, and the investigation turned up a second half to the problem.

The environment is now built from the lowest Kubernetes version listed in `candi/version_map.yml` for the branch under test (`minimalSupportedKubernetesVersion` in `validation_test.go`), so it follows the branch instead of the vendored apiserver: 1.32 on `main` and `release-1.77`, 1.31 on `release-1.76`.

Pinning alone was not enough. The test compiled with `environment.StoredExpressions`, which deliberately exposes every CEL library so that policies already persisted in etcd keep evaluating; it accepts a too-new expression regardless of the compatibility version. `environment.NewExpressions` is what kube-apiserver validates a freshly submitted policy against, and the test now uses it.

Measured, with a 1.33-gated construct (`semver()`) temporarily inserted into a policy variable:

- `DefaultCompatibilityVersion()` (resolves to 1.33) + `StoredExpressions` — suite green, the regression is invisible.
- `DefaultCompatibilityVersion()` + `NewExpressions` — suite green, still invisible.
- branch minimum (1.32) + `NewExpressions` — fails with `undeclared reference to 'semver'`.

And with a 1.32-gated construct (`transformList`), showing the pin follows the branch: green in a `release-1.77` worktree (minimum 1.32), failing with `undeclared reference to 'transformList'` in a `release-1.76` worktree (minimum 1.31). Both probes were reverted; the committed policy uses nothing above 1.29.

modules/150-user-authn/template_tests/validation_test.go
Jabejixo: Both added, and the empty-string case now denies rather than allows.

- `annotation value is empty` (`dexClientAnnotation: ""`, CREATE) — expected `false`. With the policy gating presence, an empty value is a grant on the presence-honouring hook, so it is denied instead of merely failing closed downstream.
- `dex authenticator carrying the dex client annotation is not gated` — a `DexAuthenticator` with the `dexclient.*` key, expected `true`, mirroring the existing DexClient case, which is now named `dex client carrying the dex authenticator annotation is not gated`.

Three more went in while the transition logic was reworked, to pin the update path from both sides: `object that already carries a disabled annotation stays editable` (UPDATE `"false"` → `"false"`, allowed), `disabled annotation is added to an object that had none` (denied), and `annotation is flipped from a value no hook reads as true to true` (UPDATE `"yes"` → `"true"`, denied).

Checked for vacuity by flipping each new expectation one at a time: `annotation value is empty`, the mirror case and `object that already carries a disabled annotation stays editable` each produce exactly one failure when inverted, and the suite is green otherwise (47 specs).

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [22358](https://github.com/deckhouse/deckhouse/pull/22358) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [22371](https://github.com/deckhouse/deckhouse/pull/22371) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #22221, opened 2026-08-17
[cni-cilium] Fix agent-wide deadlock and crash in ipcache/policy under identity churn.

### LINKED ISSUE #21647: cni-cilium: cilium-agent deadlock on cilium 1.17.x (ipcache/endpoint lock inversion), fixed only in cilium 1.18
**Deckhouse:** 1.75.14, `cni-cilium` (cilium **1.17.4**). Also applies to the 1.17.17 bump in #20720.

### Problem

`cilium-agent` deadlocks and never recovers on its own. Symptoms on the affected node:

* every endpoint stuck in `regenerating` / `waiting-to-regenerate`, none reach `ready`
* CNI returns `[PUT /endpoint/{id}][429] putEndpointIdTooManyRequests`, pod sandbox creation hangs 30–90 min
* agent RSS grows ~250–300 MiB/h (0.7 → 5.1 GiB observed), no OOM
* **liveness probe stays green, 0 restarts** — it does not touch the deadlocked locks
* the only recovery is deleting the agent pod

Seen twice in 10 days on a 60-node production cluster (3 nodes total), 15–19 h per occurrence, uncorrelated with any config change.

### Root cause (upstream, cilium 1.17.x)

Lock inversion between the IPCache and endpoint locks, line numbers for `v1.17.4`:

* `pkg/ipcache/metadata.go:541` — `doInjectLabels` takes `ipc.mutex.Lock()` (`defer Unlock()`)
* `:610` — still holding it, calls `UpdatePolicyMaps(ctx, nil, idsToDelete)`
* `:646` — `policyImplementedWG.Wait()`, no ctx/timeout: waits for **all** endpoints to apply their policy maps
* the other direction: endpoint regeneration holds the endpoint lock and needs the IPCache lock — upstream documents exactly this hazard in `pkg/endpoint/bpf.go:634` (*"Endpoint lock must not be held during 'GetDNSRules' as it locks IPCache, which leads to a deadlock"*)

The add path was deliberately moved out of the lock (`metadata.go:531`, comment mentions the deadlock); the delete path was not.

Related upstream report: cilium/cilium#40219 (same version, same 429, closed as stale).

### Fixed upstream in 1.18.0

Commit [`47ace2de6`](https://github.com/cilium/cilium/commit/47ace2de6) (PR cilium/cilium#39970, merged 2025-06-13) removes `UpdatePolicyMaps` from `doInjectLabels` and awaits `IdentityUpdater.UpdateIdentities()` **before** taking `ipc.mutex`. First tag: `v1.18.0-rc.0`, first stable: **`v1.18.0`**.

Not backported — `v1.17.17` and `v1.17.18` still carry the pattern (`:548` lock / `:617` `UpdatePolicyMaps` / `:653` `Wait`), so the 1.17.4 → 1.17.17 bump in #20720 does not address this.

### Ask

1. Ship cilium 1.18.x in `cni-cilium`, or backport `47ace2de6` to the vendored 1.17.
2. Until then: a health check that detects endpoints stuck in `regenerating` (e.g. `cilium_endpoint_state{endpoint_state=~"regenerating|waiting-to-regenerate"}` non-zero for >10 min) — the current liveness probe cannot see this state.

AndreyPavlovFlant:
## Description

Adds two backport patches for the Cilium 1.17.x agent used in the cni-cilium module:

- `019-ipcache-no-deadlock-on-label-injection.patch` — moves the delete-path UpdatePolicyMaps() call in doInjectLabels() out of the ipc.mutex critical section (matching the add path above it) and bounds the previously unbounded wait with ctx and a 3-minute timeout.
- `020-policy-nil-safe-selector-policy-detach.patch` — adds a nil check in policyCache.delete() before calling selectorPolicy.detach(), so a cache entry whose policy is still being resolved by updateSelectorPolicy() (without holding the cache lock) is not dereferenced.
Both patches document the exact upstream fix and are marked for removal on upgrade to Cilium 1.18+.

Fixes https://github.com/deckhouse/deckhouse/issues/21647

Under identity churn the Cilium agent could deadlock: doInjectLabels() held ipc.mutex while UpdatePolicyMaps() waited for all endpoints to regenerate, but endpoint regeneration acquires the same mutex (GetDNSRules → LookupByIdentity). As a result endpoint regeneration stopped node-wide, CNI requests piled up, the endpoint-create limiter started returning 429 putEndpointIdTooManyRequests, and pods failed to get networking — while agent liveness stayed green, so the agent was not restarted automatically.

Once regeneration ran at full speed, a second latent bug surfaced: policyCache.delete() dereferenced a nil SelectorPolicy for an endpoint whose identity changed mid-regeneration, crashing the agent with a SIGSEGV in selectorPolicy.detach.

Together these caused node-wide loss of pod networking and agent crashes on clusters with high identity churn.

Both are production-impacting stability bugs (node-wide connectivity outage and agent crash) that affect existing clusters running the current Cilium 1.17.x. The fixes are self-contained backports and should be delivered to affected clusters without waiting for the Cilium upgrade, where both are fixed upstream.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: cni-cilium
type: fix
summary: Fix agent-wide deadlock in ipcache label injection and a nil-pointer crash in policy cache under identity churn.
impact: Cilium agent pods are restarted to apply the fixes.
impact_level: high
```

### DIFF
--- a/modules/021-cni-cilium/images/bin-cilium/patches/019-ipcache-no-deadlock-on-label-injection.patch
+++ b/modules/021-cni-cilium/images/bin-cilium/patches/019-ipcache-no-deadlock-on-label-injection.patch
@@ -1,0 +1,64 @@
+diff --git a/pkg/ipcache/metadata.go b/pkg/ipcache/metadata.go
+index 1f41ae64..d8be860f 100644
+--- a/pkg/ipcache/metadata.go
++++ b/pkg/ipcache/metadata.go
+@@ -34,6 +34,9 @@ var (
+
+ 	LabelInjectorName = "ipcache-inject-labels"
+
++	// policyMapUpdateTimeout bounds the wait in UpdatePolicyMaps().
++	policyMapUpdateTimeout = 3 * time.Minute
++
+ 	injectLabelsControllerGroup = controller.NewGroup("ipcache-inject-labels")
+ )
+
+@@ -546,7 +549,6 @@ func (ipc *IPCache) doInjectLabels(ctx context.Context, modifiedPrefixes []netip
+ 	}
+
+ 	ipc.mutex.Lock()
+-	defer ipc.mutex.Unlock()
+ 	for p, entry := range entriesToReplace {
+ 		prefix := p.String()
+ 		meta := ipc.getK8sMetadata(prefix)
+@@ -613,14 +615,17 @@ func (ipc *IPCache) doInjectLabels(ctx context.Context, modifiedPrefixes []netip
+ 			}
+ 		}
+ 	}
++	for prefix, id := range entriesToDelete {
++		ipc.deleteLocked(prefix.String(), id.Source)
++	}
++	// Don't hold lock while calling UpdatePolicyMaps, as it will otherwise run into a deadlock
++	ipc.mutex.Unlock()
++
+ 	if len(idsToDelete) > 0 {
+ 		if ipc.UpdatePolicyMaps(ctx, nil, idsToDelete) {
+ 			mutated = true
+ 		}
+ 	}
+-	for prefix, id := range entriesToDelete {
+-		ipc.deleteLocked(prefix.String(), id.Source)
+-	}
+
+ 	return remainingPrefixes, mutated, err
+ }
+@@ -650,7 +655,19 @@ func (ipc *IPCache) UpdatePolicyMaps(ctx context.Context, addedIdentities, delet
+ 	}
+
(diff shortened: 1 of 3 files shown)


apolovov (approved):

deckhouse-BOaTswain:
Cherry pick PR [22586](https://github.com/deckhouse/deckhouse/pull/22586) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22222, opened 2026-08-17
[testing] Stop sharing Helm render capabilities between tests

Jabejixo:
## Description

`testing/library/helm.Renderer` built a capability set for each render but never used it: `chartutil.ToRenderValues` was called with `nil` capabilities. It only appeared to work because `caps := chartutil.DefaultCapabilities` copies a package-level **pointer**, so `caps.APIVersions = vers` mutated the process-wide default that Helm then fell back to. Every render extended the same list for the rest of the test binary.

This copies the capabilities per render and passes them explicitly, and adds an opt-in `WithAPIVersions` render option to `testing/helm`.

The option makes a whole class of templates testable for the first time. Helm's default capabilities carry bare `group/version` strings, so any template guarded by `helm_lib_kind_exists` rendered nothing in template tests regardless of the values supplied.

Does not restart any cluster component; test-only change.

Two defects in shared test infrastructure:

- **Cross-test leakage.** Mutating `chartutil.DefaultCapabilities` means one package's render can change what a later, unrelated render sees. Copying per render removes the shared mutable state.
- **Dead code path.** The assembled capabilities were discarded at the call site, so the intent of the existing `autoscaling.k8s.io/v1/VerticalPodAutoscaler` handling was only satisfied as a side effect of the leak. The pre-existing loop guarding that append was also ineffective — it reassigned its result on every iteration instead of accumulating, so only the last comparison survived.

`WithAPIVersions` is additive and opt-in: renders that do not pass it keep Helm's defaults, so existing tests are unaffected.

Not necessarily. It is test-only, but it cherry-picks cleanly and is a prerequisite for backporting admission-policy tests to 1.77 and 1.76.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: fix
summary: Copy Helm render capabilities per render instead of mutating the shared defaults, and allow tests to declare available API kinds.
impact_level: low
```


AlwxSin (approved):

testing/library/helm/render.go
```diff
@@ -35,6 +35,10 @@ type Renderer struct {
 	Name      string
 	Namespace string
 	LintMode  bool
+	// APIVersions extends the capabilities the chart is rendered with. Helm's defaults carry
+	// group/version strings only, so a template guarded by helm_lib_kind_exists renders nothing
+	// unless the group/version/Kind it looks for is named here.
```
ldmonster: **Doc nit: the `helm_lib_kind_exists` description here isn't accurate** (same wording on the `WithAPIVersions` doc in `testing/helm/init.go` L333-335).

This comment says a template guarded by `helm_lib_kind_exists` "renders nothing unless the group/version/Kind it looks for is named here". But `helm_lib_kind_exists` neither looks up a full `group/version/Kind` nor uses `.Capabilities.APIVersions.Has`:

https://github.com/deckhouse/deckhouse/blob/930bdf4cd60c6720a22e339fe093f48ffd0cd94e/helm_lib/charts/deckhouse_lib_helm/templates/_api_version_and_kind.tpl#L3-L15

It takes a **bare Kind** (`$kind_name`, "Kind name portion") and does a case-insensitive `hasSuffix "/<kind>"` scan over `.Capabilities.APIVersions`. The full-G/V/K-via-`.Has` contract described here actually belongs to the sibling helper `helm_lib_api_version_exists` (and to inline `.Capabilities.APIVersions.Has "..."` deployment guards — which is exactly what the always-injected `autoscaling.k8s.io/v1/VerticalPodAutoscaler` string serves):

https://github.com/deckhouse/deckhouse/blob/930bdf4cd60c6720a22e339fe093f48ffd0cd94e/helm_lib/charts/deckhouse_lib_helm/templates/_api_version_and_kind.tpl#L19-L25

Not a functional problem — passing a full `group/version/Kind` to `WithAPIVersions` works for both cases: it satisfies the case-insensitive suffix match in `helm_lib_kind_exists` and the exact `.Has` check in `helm_lib_api_version_exists`/inline guards. It just names `helm_lib_kind_exists` as the helper that needs a G/V/K string via `.Has`, which misrepresents its contract (and its case-insensitivity). Worth rewording to point at `helm_lib_api_version_exists` / the inline `.Has` guards, or to describe the bare-Kind suffix match accurately.

Jabejixo: You are right, and it is my wording that is wrong rather than a shortcut: `helm_lib_kind_exists` takes a bare kind and matches it as a case-insensitive `/<Kind>` suffix over the list, while it is `helm_lib_api_version_exists` and the inline `.Capabilities.APIVersions.Has` guards that compare a whole `group/version/Kind`. Naming the first helper as the one with the `.Has` contract misdescribes it.

Reworded both doc comments in d655ca8199 to say which guard reads the list which way, and to keep the practical advice that an entry spelled `group/version/Kind` satisfies both.

The premise the comment rests on is unchanged and worth keeping: Helm's default capabilities come from `allKnownVersions()`, which appends `gv.String()` for every known group version (`chartutil/capabilities.go:115-128`), so the defaults hold no kinds at all and any kind-seeking guard answers no until a kind is named.

ldmonster (approved):

yalosev (approved):

z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [22357](https://github.com/deckhouse/deckhouse/pull/22357) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 2025c964ab8954663e25c5404cd5fb9d60f94540 to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/32464422241) for details.

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 2025c964ab8954663e25c5404cd5fb9d60f94540 to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/32464444380) for details.

PR #22237, opened 2026-08-18
[docs-builder] add hugo react to context and timeouts

ldmonster:
## Description

`ModuleDocumentation` resources intermittently stay in `RenderResult: Error`, with a status condition like:

```
build documentation: POST "http://<pod>.d8-system.pod.cluster.local:8081/api/v1/build":
do request: ... net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

even though the individual module content renders fine in the docs-builder logs — the aggregate build step is what times out. On a cluster with many modules this typically shows up as a handful of modules `Rendered` and most stuck in `Error`.

### Root cause

Every `POST /api/v1/build` triggers a **full-site Hugo rebuild** of all modules (content + the aggregated search index), serialized on the builder. The `module-documentation` controller reconciles modules one at a time (`MaxConcurrentReconciles: 1`) and, for each not-yet-rendered module, uploads its docs and triggers one such full rebuild. Three properties turn a slow build into a self-sustaining congestion collapse:

- The controller's HTTP client used a fixed **10s** timeout (the `go_lib/dependency/http` default). A full-site build that legitimately takes longer is canceled client-side; the module is marked `Error` and retried, which re-uploads and re-triggers another full rebuild.
- `docs-builder`'s build handler **ignored the request context** and `Service.Build` held a plain `sync.Mutex`, so a build whose client had already given up kept running to completion, holding the build lock, while every queued reconcile stacked up behind it and timed out in turn.
- The self-heal loop in `buildHugo` reruns the whole Hugo build once per broken module it strips, multiplying the work inside a single request.

Modules that happened to finish a build under 10s were cached by `version`+`checksum` and stayed `Rendered`; the rest churned in `Error`, keeping the builder saturated — which matches the observed split.

### Fix

**docs-builder — make a build cancelable, so abandoned work stops instead of piling up:**

- Thread the request context from the `/api/v1/build` handler through `Service.Build` → `buildHugo` → the Hugo wrapper (`Build` → `Run` → `build` → `fullBuild` → `buildSites`).
- Replace the build mutex with a context-aware semaphore: a caller whose request was already canceled drops out of the queue instead of running a now-redundant full rebuild.
- Check the context between self-heal passes and immediately before each Hugo pass.
- Treat a canceled build as *not a failure*: readiness and the last good render are preserved, and it is recorded as `status="canceled"` rather than `fail`.

**deckhouse-controller — give the build a longer, configurable timeout (now safe because builds are cancelable):**

- The build request timeout is configurable via the `DOCUMENTATION_BUILD_TIMEOUT` environment variable (default `120s`, wired in the `deckhouse` Deployment), scoped to the build call so that upload and health checks keep their tight failure detection against an unreachable builder.

### Known limitation

A single in-flight Hugo render pass cannot be aborted mid-pass: Hugo `v0.163.3`'s `HugoSites.Build` takes no context and builds under an internal `context.Background()`. Cancellation therefore happens at boundaries (slot wait, self-heal loop, before each pass) — enough to stop the pile-up and free the build slot promptly, but one already-started pass runs to completion.

On clusters with many modules, module documentation could get permanently stuck in `Error` and the documentation site would stop updating, because the controller and the builder drove each other into a full-rebuild congestion collapse. This change restores reliable documentation rendering and makes the build timeout tunable for large docs sites without rebuilding the image.

Operational note: before raising the timeout, check the `docs_builder_build_duration_seconds` metric. If a clean single build already fits well under the previous 10s, the cancellation fix alone removes the pile-up; the larger timeout is the safety margin for genuinely long builds.

Yes. This is a production issue — documentation rendering wedges under load with no real workaround other than reducing the module count or restarting the controller — and the fix is self-contained, so it should be backported.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: docs
type: fix
summary: Module documentation no longer gets stuck in Error under load; docs-builder builds are cancelable and no longer pile up redundant full-site rebuilds behind requests whose clients already timed out.
impact_level: low
```
```changes
section: deckhouse-controller
type: fix
summary: The module-documentation build request timeout to docs-builder is configurable via DOCUMENTATION_BUILD_TIMEOUT (default 120s) instead of a fixed 10s that canceled healthy full-site builds.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/internal/app/env.go
+++ b/deckhouse-controller/internal/app/env.go
@@ -17,6 +17,7 @@ package app
 import (
 	"os"
 	"strconv"
+	"time"
 )
 
 // Environment variable names the deckhouse controller reads at runtime. They are
@@ -66,6 +67,13 @@ const (
 	EnvEditor              = "DECKHOUSE_EDITOR"
 	EnvKubeConfigInCluster = "DECKHOUSE_KUBE_CONFIG_IN_CLUSTER"
 	EnvTmpDir              = "DECKHOUSE_TMP_DIR"
+
+	// EnvDocumentationBuildTimeout caps a single upload+build round-trip from the
+	// module-documentation controller to a docs-builder. Every build triggers a
+	// full-site Hugo rebuild serialized across all modules, so this must exceed the
+	// worst-case build time; too low a value cancels healthy builds and churns the
+	// builder. A Go duration string (e.g. "120s", "2m").
+	EnvDocumentationBuildTimeout = "DOCUMENTATION_BUILD_TIMEOUT"
 )
 
 // EnvOr returns the value of the env var name, or defaultValue when it is unset or empty.
@@ -84,6 +92,20 @@ func EnvBoolOr(name string, defaultValue bool) bool {
 		return defaultValue
 	}
 	parsed, err := strconv.ParseBool(v)
+	if err != nil {
+		return defaultValue
+	}
+	return parsed
+}
+
+// EnvDurationOr parses the env var as a Go duration (per time.ParseDuration), or
+// returns defaultValue when unset, empty, or unparseable.
+func EnvDurationOr(name string, defaultValue time.Duration) time.Duration {
+	v, ok := os.LookupEnv(name)
+	if !ok || v == "" {
+		return defaultValue
+	}
+	parsed, err := time.ParseDuration(v)
 	if err != nil {
 		return defaultValue
 	}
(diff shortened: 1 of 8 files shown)


z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [22258](https://github.com/deckhouse/deckhouse/pull/22258) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit c44ddf816f57d7ff99ab8f8d805dbd1e5fb67608 to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/32221532382) for details.

PR #22246, opened 2026-08-18
[deckhouse] Read a legacy module's versions from its release history

Fkuloff:
## Description

**How a module is scanned.** A package that publishes `<package>/version` is read from there, as
before. A package with no `/version` path — or, on a full scan, one whose `/version` carries no
semver tags — is a legacy module, and its versions are read from `<package>/release` instead.

```
package scan
   |
   +- /version has semver tags -----> type from the newest tag
   |                                    +- Application -> APV per tag
   |                                    +- Module      -> MPV per tag
   |
   +- /version missing --------------> walk /release
   +- /version has no semver tags ---> walk /release (full scan only)
```

The walk goes from the newest release down and creates a `ModulePackageVersion` for every release
image that carries a module definition. It stops at the first image without one:

```
v0.4.15 ... v0.4.0 ... v0.3.0 | v0.2.4 ... v0.1.0
\---- module.yaml is there ---/ \-- no module.yaml -/
        MPV created            |    nothing created
                              ^
                       the walk ends here
```

That version and everything older are not offered — the metadata to describe them is not there.

A version the cluster already holds keeps its release image unread and the walk carries on past it.
Skipping rather than stopping is what makes the walk repeatable: a version whose creation failed is
simply missing next time, and the walk reaches it again instead of ending at the newer ones above it.

The release service now embeds the version service — a release image carries the same metadata files,
so it answers the same reads, `HasModuleDefinition` among them.

**A version whose image comes back.** The version controllers mark a version whose image is gone with
`packages.deckhouse.io/exist-in-registry: "false"`. Until now only an `ApplicationPackageVersion` was
ever unmarked again, and only if a scan happened to list its tag. Now a `ModulePackageVersion` gets
the same treatment, and an incremental scan re-checks the marked versions explicitly instead of
leaving them below its watermark.

The watermark itself is unchanged: still the newest processed version, so no scan lists more tags
than before. The marked versions come from the cluster's own CRs, so the extra work is proportional
to the number of versions actually missing rather than to the length of the tag list.

The re-check is a pass of its own — `rediscoverMissingVersions`, before the ordered walk over the
listed tags — and it is keyed on the kind of CR carrying the mark, not on a type read from the
registry. Both halves matter. A marked version sits at or below the watermark, so merging it into
the walk would make it the tag the package type is detected from, even once that tag has been pulled
from `/version`; and because the walk stops at the first version it fails to create, one unreadable
image would withhold every newer version instead of costing a single re-check.
`ensureApplicationPackageVersion` / `ensureModulePackageVersion` therefore leave an existing version
alone and never re-check it.

**Every patch is offered.** An incremental scan no longer thins a minor line down to its newest
patch: `filterLatestTags` is gone, so every semver tag a repository offers becomes a package
version. A repository with a long patch history therefore produces more version objects than before.

**The partial-listing warning is gone.** `WarnUnavailablePartialScan` is removed. Tags are still
filtered on the controller when a registry ignores the `last=` parameter, so scan results are
unchanged — only the warning is. `PackageRepositoryStatus.PartialScanAvailable` is still written by
the `PackageRepository` controller and currently has no reader.

**Refactor.** `EnsureApplicationPackage` and `EnsureModulePackage` moved to the operation reconciler
as `ensureApplicationPackage` / `ensureModulePackage`: creating the package CR and listing the
repository under it is queue-level work, and the operations service is left with the registry reads
and the version CRs. `DiscoverPackage` became `Discover` and returns package names;
`PackageProcessResult` became `Result`; `foundTagsToProcess` and `performIncrementalScan` collapsed
into `tagsToProcess`, which returns the listed tags and the marked versions as separate fields.

Reaching `/release` only as a fallback meant a module without `/version` had every release tag turned
into a package version — including images that carry nothing but a version file, which cannot be
described, let alone installed. The walk offers only the versions whose image actually carries a
module definition, and says where that history stops.

Separately, a version whose image disappeared from the registry stayed marked as missing until
someone ran a full scan, because an incremental scan resumes above its watermark and the marked
version sits below it. Modules had no unmarking path at all.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: feature
summary: Read a legacy module's versions from its release history and stop at images without a module definition
impact_level: low
```


deckhouse-controller/internal/registry/service/package_service.go
```diff
@@ -294,17 +294,18 @@ func (s *PackageVersionService) ReadPackageDefinition(ctx context.Context, tag s
...
 
-// HasModuleDefinition checks whether the version image contains a module.yaml (or module.yml) file.
-// This is used as a fallback to identify legacy modules when neither type labels nor package.yaml are present.
+// HasModuleDefinition checks whether the image contains a module.yaml (or module.yml) file.
+// It tells a module image apart from one that carries the version alone, on both the version
+// and the release path.
 //
 // Returns (false, nil) if the image does not exist.
-func (s *PackageVersionService) HasModuleDefinition(ctx context.Context, tag string) (bool, error) {
+func (s *BasicService) HasModuleDefinition(ctx context.Context, tag string) (bool, error) {
```
ipaqsa: better to do this instead:
```
// PackageReleaseService provides access to the <package>/release path for legacy v1alpha1 modules.
type PackageReleaseService struct {
	*PackageVersionService
}

func NewPackageReleaseService(basicService *BasicService) *PackageReleaseService {
	return &PackageReleaseService{
		PackageVersionService: NewPackageVersionService(basicService),
	}
}
```

ipaqsa (approved):

ldmonster (approved):

PR #22217, opened 2026-08-17
[deckhouse] Reserve the platform's own public hostnames from tenant namespaces

Jabejixo:
> Based on #22222, which carries the test-harness change this PR's tests need. The base branch is `chore/helm-test-harness-api-versions`, so the diff here is only the module files; merge #22222 first.

## Description

Deckhouse publishes its own web interfaces under hostnames rendered from `global.modules.publicDomainTemplate` (`api`, `console`, `dex`, `kubeconfig`, `grafana`, …). Those hostnames belong to the platform, but nothing has ever said so to the API server: an `Ingress` is a namespaced object with no cluster-wide uniqueness check on `spec.rules[].host`, so anybody who may write an Ingress in their own namespace can name one of them and have the shared ingress controller merge that rule into the routing of a hostname the platform owns. The Gateway API primitives carry hostnames too, on `spec.hostnames` for `HTTPRoute`, `GRPCRoute` and `TLSRoute`, and on `spec.listeners[].hostname` for `Gateway` and `ListenerSet`.

This adds six `ValidatingAdmissionPolicy` objects to the `deckhouse` module, next to the guardrails already in `modules/002-deckhouse/templates/validation.yaml`, plus the ConfigMap they read their parameters from:

- `d8-reserved-public-hosts` (ConfigMap in `d8-system`) — the parameters: a regex for the namespace `publicDomainTemplate` carves out, the hostnames matched exactly on top of it, the hostnames allowed back out of it, and the one object the reservation does not cover.
- `reserved-public-hosts-ingress.deckhouse.io` — denies `networking.k8s.io/Ingress` CREATE/UPDATE claiming a reserved hostname.
- `reserved-public-hosts-httproute.deckhouse.io`, `-grpcroute`, `-tlsroute` — the same on `gateway.networking.k8s.io/HTTPRoute`, `GRPCRoute` and `TLSRoute`, which all carry `spec.hostnames`.
- `reserved-public-hosts-listenerset.deckhouse.io`, `-gateway` — the same on `gateway.networking.k8s.io/ListenerSet` and `Gateway`, which both carry `spec.listeners[].hostname`.

[...]

PR #22250, opened 2026-08-18
[docs] Document identity collisions, Kubernetes API access, client secret rotation and public hostname reservation

Jabejixo:
## Description

Client-facing documentation for a series of `user-authn` / `user-authz` changes, collected into one pull request so that the documentation team reviews the prose once instead of once per change. The changes themselves are in #22214, #22215, #22216, #22218, #22219, #22222 and #22234; this pull request contains no code.

Reference documentation that lives inside CRD `description` fields and `openapi/` schemas is **not** here — it stays in the pull request that changes the schema, because there it is part of the API contract and is reviewed together with the field it describes.

### `user-authn` usage

**A new section documenting the Kubernetes API access annotation.** `dexclient.deckhouse.io/allow-access-to-kubernetes` and `dexauthenticator.deckhouse.io/allow-access-to-kubernetes` were not documented anywhere: they existed only in hook code and in the manifests of the modules that use them. They are a cluster-level privilege grant, so the section states what the annotation actually does — it makes the client a trusted peer of the privileged `kubernetes` OAuth2 client, and the resulting token maps the username from the `email` claim and the groups from the `groups` claim — and it states who may set it after #22219: only a subject allowed to update the `user-authn` module configuration. It also records that resources which already carry the annotation stay editable by their owner, so an operator is not left guessing whether an existing setup breaks.

**Identity collisions with authorization rules**, for #22218. Documented next to the `User` and the `Group` examples, where someone about to create one of those objects will actually read it: why the rejection happens, the `user-authz.deckhouse.io/allow-authorization-rule-collision` annotation that acknowledges an intentional match, and the fact that existing objects only produce a warning. Both notes call out the delete-and-recreate case explicitly, because an object that is currently only warned about will be rejected when a GitOps flow recreates it, and that is the failure an operator is most likely to hit without being told.

The `User` note additionally records that deleting a user does not revoke the rule: the email stays granted, so recreating a user with the same email restores the privileges, and until then the email is available to anyone allowed to create users.

### `user-authz`

**A correction, not an addition.** The overview claimed that a manage role holder "will not be able to manage other cluster functions … or change settings in the namespaces of users applications". For the `security` subsystem the first half of that promise does not hold, and the documentation should not imply a boundary the roles do not enforce. The sentence is trimmed to what is true, and a warning explains why: managing `user-authn` allows registering an identity provider or resetting any local user's credentials, and managing `user-authz` allows writing authorization rules, so either one yields an identity with arbitrary privileges. The guidance is to plan access treating a `security` subsystem manage role as a cluster administrator role.

This is worth stating plainly because the promise, not any single defect, is what made several reported findings look like escalations: the roles behave as designed, and it was the documentation that described a boundary that was never there.

**A cross-reference** next to the `ClusterAuthorizationRule` example, explaining that `User` and `Group` subjects are matched by name against the identity in the token and that a local identity is indistinguishable from one asserted by an external provider. It also states the asymmetry honestly: creating a colliding `User` or `Group` is rejected, but adding a subject to a rule is not restricted the same way, and if a matching object already exists it receives the privileges immediately.

### Deliberately not documented

* #22217 was closed and handed over to the team that owns ingress and network, so the documentation of the hostname reservation goes with it rather than shipping here.
* #22222 changes only the Helm render test harness. Nothing user-visible.
* #22214 and #22234 are template-quoting fixes. Single-line values render exactly as before, so there is no behaviour to describe; the field-level constraints they add are documented in the CRD `description` fields in those pull requests.
* #22216 is internal: the per-`DexAuthenticator` client secret and the removal of a dead redirect URI. Operators take no action, so it belongs in the changelog rather than in the documentation.
* #22215 removes rules from two ClusterRoles. The published role listing in `modules/140-user-authz/docs/README.md` is generated from the roles themselves and is regenerated inside that pull request, which is the right place for generated output.

Two of the behaviours described here are new restrictions that can reject a write, so an operator who meets them without documentation has no way to tell an intended guard from a bug. One of them — the `allow-access-to-kubernetes` annotation — has been a cluster-level privilege grant with no documentation at all. And one is a correction of a statement that is currently wrong in a way that affects how administrators plan access.

It should follow whichever of the referenced changes are backported, and land in the same releases as they do. The documentation describes behaviour that only exists once those changes are merged, so this pull request should not merge before them.

## Checklist
- [ ] e2e tests passed.

`markdownlint` was run over the six changed files with the repository's own `testing/markdownlint.yaml`. The only findings are two pre-existing `MD051` warnings about a `#get_rules` link fragment in the `user-authz` overview, present unchanged on `main`; this pull request adds none.

## Changelog entries

```changes
section: docs
type: fix
summary: Documented the identity collision rules for local Group and User objects and the allow-access-to-kubernetes annotation, and corrected the described scope of a security subsystem manage role.
impact_level: low
```


AlwxSin (approved):

maxinsky (approved):

z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [22635](https://github.com/deckhouse/deckhouse/pull/22635) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [22636](https://github.com/deckhouse/deckhouse/pull/22636) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #22299, opened 2026-08-19
[node-manager] keep MachineDeployment replicas owned by deckhouse-hook

ldmonster:
## Description

Adds a hook: `force_machine_deployment_replicas_ownership`. It runs `OnBeforeHelm`, right before the node-manager release is applied.

What it does:

- looks at MCM and CAPI MachineDeployments in `d8-cloud-instance-manager`;
- picks only those where `spec.replicas` is still owned by the old `deckhouse-controller` field manager;
- moves ownership of that field to `deckhouse-hook` by rewriting `metadata.managedFields` (JSON patch, `resourceVersion` pinned).

The replica count is not changed. Other fields of the old manager are kept. Errors are only logged and never block the release.

Server-side apply does not work here. SSA reports a conflict only for values an apply changes, so applying the same replica count adds a second owner but keeps the old one.

Nothing restarts, no nodes roll.

1.76.9 turns on `NELM_FEAT_ADOPT_DECKHOUSE_CONTROLLER_FIELDS` (#21831). nelm takes over fields owned by `deckhouse-controller` and drops the ones that are not in the manifest.

`MachineDeployment.spec.replicas` is not in the template. It is set by `set_replicas_on_machine_deployment.go` and by cluster-autoscaler. That hook patches only when the value changes, so for NodeGroups with a stable replica count the field is still owned by `deckhouse-controller` (the hook manager was renamed to `deckhouse-hook` in Dec 2025).

So on upgrade nelm takes the field and deletes it. MCM reads a missing `spec.replicas` as 0, scales the MachineSet to zero, and every CloudEphemeral node is deleted and created again.

This happened in production on a 1.76.8 -> 1.76.9 upgrade: all cloud nodes were recreated at once.

The hook runs before the apply, so it prevents the scale to zero instead of fixing it afterwards.

Any cluster with MCM or CAPI nodes that upgrades to 1.76.9 or later can hit this, and it recreates all cloud nodes at once.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: MachineDeployment `spec.replicas` is no longer dropped on upgrade, so cloud nodes are not recreated.
impact_level: high
impact: |
  Fixes recreation of all CloudEphemeral nodes on upgrade to 1.76.9. The MachineDeployment
  replica count was dropped and the MachineDeployment was scaled to zero.
```

### DIFF
--- a/modules/040-node-manager/hooks/force_machine_deployment_replicas_ownership.go
+++ b/modules/040-node-manager/hooks/force_machine_deployment_replicas_ownership.go
@@ -31,15 +31,24 @@ limitations under the License.
 // The FilterFunc keeps only MachineDeployments whose spec.replicas is still owned
 // by the legacy field manager (everything else is dropped from the snapshot by
 // returning a nil result), so the handler's snapshots contain exactly the objects
-// that must be fixed. For each of them it force-claims the field for
-// "deckhouse-hook" via server-side apply, re-applying the current value so the
-// count is unchanged. It runs OnBeforeHelm, right before nelm applies the
-// node-manager release, so ownership is migrated BEFORE the apply that would
-// otherwise prune it — that is what prevents the scale-to-zero instead of only
-// healing it afterwards. WaitForSynchronization is left at its default (true) so
-// the snapshots are populated before beforeHelm runs; both CRDs ship in
-// node-manager/crds and node-manager is critical, so they are always present and
-// the synchronization cannot stall.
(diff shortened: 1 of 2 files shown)


borg-z (approved):

deckhouse-BOaTswain:
Cherry pick PR [22338](https://github.com/deckhouse/deckhouse/pull/22338) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22316, opened 2026-08-20
[DO NOT MERGE] Combined build of the bug bounty fixes for cluster testing

Jabejixo:
## DO NOT MERGE

This pull request exists only to produce a build that carries every bug bounty fix at once, so the whole set can be exercised in one cluster instead of eight. It is closed once the testing is done, and nothing is merged from here: every change belongs to the pull request it came from, and any fix found during testing is made there first and merged back into this branch.

The branch is assembled with merges rather than cherry-picks, so an updated source branch is picked up by merging it again and the diff against each source stays empty.

## What it carries

| Source | What it changes | What only a cluster can answer |
| --- | --- | --- |
| #22222 | Copies Helm render capabilities per render instead of mutating the shared defaults | Nothing, test-only. Included because #22217 and #22219 branch from it |
| #22217 | Reserves the platform's public hostnames from tenant namespaces: three ValidatingAdmissionPolicies plus a one-shot hook that records what tenants already serve | Real admission verdicts against a real apiserver, the cost accounting for `matches()` with a pattern that comes from a ConfigMap, and the hook's cluster-wide list with pagination and discovery |
| #22219 | Gates the `allow-access-to-kubernetes` annotation behind the right to update the `user-authn` ModuleConfig | The `authorizer` call inside CEL. Nothing outside an apiserver evaluates it |
| #22218 | Rejects a `Group` name or a `User` email that collides with a subject of an authorization rule | That the webhook is wired and reached from `user-authz`, and that its failure path does not block object creation |
| #22216 | Issues a dedicated OAuth2 client secret per DexAuthenticator and migrates the ones that carry the shared one | The migration on live objects: whether pods roll, whether sign-in still works, and whether the privileged `kubernetes` client is untouched |
| #22214 | Quotes DexProvider-derived values in the rendered manifests, constrains `oidc.scopes` | Whether Dex still starts on the quoted config, and whether the new schema constraint rejects a DexProvider that already exists |
| #22234 | Quotes custom resource values in rendered manifests, constrains DexClient and DexAuthenticator fields | The same upgrade question for objects that already exist |
| #22215 | Removes list access to Dex credential storage from read-only roles | That the console and the user operations page keep working without the grant |

Not included: #22250, which is documentation and changes no behaviour.

## State

Every file in this branch is byte-identical to the source branch it came from, checked per file across all eight. `go test ./modules/002-deckhouse/... ./modules/150-user-authn/... ./modules/140-user-authz/... -count=1` passes, the controller builds, and `make generate` leaves no diff.

One conflict the merge surfaced is already fixed at the source: `main` gained a policy harness in `modules/002-deckhouse/template_tests/validation_policy_test.go` (#22257) that declares `allowed`, `denied` and `skipped` in the same package as the verdict values #22217 added, so that package stopped compiling. #22217 now names them `verdictAllowed`, `verdictDenied` and `verdictSkipped`. Nothing else collided.

`modules/140-user-authz/template_tests` fails locally on both this branch and a clean `main`, because it symlinks paths under `/deckhouse` and only runs inside the build container.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: chore
summary: Temporary branch that merges the bug bounty fixes into one build for cluster testing. Never merged.
impact_level: low
```


Jabejixo:
Testing is done, closing as planned — nothing here was ever meant to merge.

What this build was used for, on the `integration/rbacv2` cluster:

- #22217: the reservation was exercised on real admission across all six kinds. `console.<domain>`, `shop.<domain>`, `my-console.<domain>` and the wildcard were denied on Ingress, HTTPRoute, GRPCRoute, TLSRoute, ListenerSet and Gateway; a two-label name, a deeper name, the bare domain tail and another domain were allowed. The namespace bypass label worked and was reversible. The grandfathering snapshot recorded nothing and was correct to: all 41 covered Ingresses live in namespaces labelled `heritage: deckhouse`, including `kube-system`. After a full Deckhouse restart with the Gateway API kinds present the record stayed put, which is the anti-self-widening property. The shared Gateway exemption was verified by identity: with the key set, that exact Gateway passed with a wildcard listener while another name and the same name on an HTTPRoute were denied.
- #22219: the `authorizer` branch of the CEL, which nothing outside an apiserver evaluates. A subject with namespace DexClient rights and no ModuleConfig rights was denied at any annotation value including `"false"`, allowed without the annotation; a subject that may update the `user-authn` ModuleConfig was allowed.
- #22218: a new Group named after a real rule subject was denied, a name differing only in case was allowed, and the pre-existing colliding Group stayed editable with the intended warning.
- #22216: six authenticators, six distinct client secrets, none equal to the shared one, and every one of them present in Dex's OAuth2Client storage — so both sides of the rotation agree.
- #22214 and #22234: every existing DexProvider and DexAuthenticator re-applied cleanly against the new CRD constraints, so the schema tightening is not an upgrade hazard.
- #22215: the read-only roles no longer grant `passwords` or `offlinesessionses`, and the console stayed healthy.

Gateway API CRDs were installed on that cluster to exercise the five Gateway API policies. They are additive and no module renders Gateway API objects without a configured gateway, which was verified: zero objects appeared.

PR #22333, opened 2026-08-20
[user-authn] Stop copying the platform registry credential into DexAuthenticator namespaces

Jabejixo:
## Description

Stacked on #22216 because that PR already rewrites `templates/dex-authenticator/secret.yaml`. This change is only the extra `Secret/registry-dex-authenticator` document and the matching `imagePullSecrets` on the authenticator Deployment. Retarget to `main` after #22216 merges.

`templates/dex-authenticator/secret.yaml` copied `global.modulesImages.registry.dockercfg` into a `kubernetes.io/dockerconfigjson` Secret named `registry-dex-authenticator` in every namespace that hosts a `DexAuthenticator`. The authenticator Deployment then referenced that Secret as `imagePullSecrets`. The Secret lives in `{{ $crd.namespace }}`, so anyone who can `get`/`list` Secrets there obtained the platform registry credential.

The copy is removed. The cookie/client-secret Secret is unchanged.

Pull path after this change:

- A DexAuthenticator in a `d8-*` / `kube-*` namespace gets `imagePullSecrets: [{name: deckhouse-registry}]` instead of `registry-dex-authenticator`. That Secret already exists in those namespaces — the same credential path as other module workloads. The name change is a pod-template change, so those Deployments roll once.
- A DexAuthenticator in a tenant namespace has no `imagePullSecrets`. On `defaultCRI: Containerd` / `ContainerdV2` the node CRI already has registry auth (`hosts.toml`). On `defaultCRI: NotManaged` Deckhouse does not write node registry auth: CE anonymous pull still works; EE/FE/SE and private registries will fail a cold pull of a tenant authenticator. Dropping `imagePullSecrets` is also a pod-template change, so those Deployments roll once as well.

Every DexAuthenticator in the cluster therefore rolls once: platform authenticators on the ingress auth path (kiali, hubble-ui, openvpn, deckhouse-tools, documentation) and every tenant authenticator.

This PR only stops the **user-authn** copy. Istio still renders `Secret/d8-istio-sidecar-registry` with the same dockercfg into every mesh-enrolled namespace; that is out of scope here.

Do not mix this with #22219: that PR is about who may set `allow-access-to-kubernetes`, not about registry credentials.

It stops user-authn from distributing a cluster-level registry pull credential into tenant-readable namespaces. Creating any `DexAuthenticator` was enough; no extra annotation or cluster-admin right was required.

Leaving the Secret in place whenever `spec.nodeSelector` is not the system selector does not close the leak: that field is tenant-controlled. Moving the pods into `d8-user-authn` would break the documented in-cluster `auth-url` of the form `https://<svc>.<tenant-ns>.svc`.

The copy is created on any cluster that has a `DexAuthenticator` outside `d8-*` / `kube-*`, so it is worth backporting to 1.77 and 1.76 together with #22216.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Stop copying the platform registry credential into DexAuthenticator namespaces.
impact: |
  On upgrade Helm deletes Secret/registry-dex-authenticator from every namespace that hosts a DexAuthenticator. Every DexAuthenticator Deployment changes imagePullSecrets — d8-* / kube-* switch the name from registry-dex-authenticator to deckhouse-registry (that Secret already exists there); tenant Deployments drop imagePullSecrets — so every authenticator rolls once. That includes platform authenticators on the ingress auth path (kiali, hubble-ui, openvpn, deckhouse-tools, documentation) and every tenant authenticator.

  Tenant pods then pull with the node's CRI registry credentials when defaultCRI is Containerd or ContainerdV2. That node path is not configured when defaultCRI is NotManaged; CE anonymous pull is unaffected, EE/FE/SE and private registries can fail a cold pull of a tenant DexAuthenticator. The Secret that holds cookie-secret and client-secret is unchanged.

  If a DexAuthenticator already existed in a namespace where non-platform identities can read Secrets, the platform registry credential that was copied there must be treated as disclosed. Rotating that credential (the in-cluster deckhouse-registry Secret and the matching CRI config on nodes) is a separate operator action; this change only stops further copies from user-authn and removes the per-namespace ones.
impact_level: high
```

### DIFF
--- a/modules/150-user-authn/template_tests/authenticator_test.go
+++ b/modules/150-user-authn/template_tests/authenticator_test.go
@@ -115,6 +115,17 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
   allowAccessToKubernetes: true
   spec:
     keepUsersLoggedInFor: "2h20m4s"
+- name: test-tenant
+  encodedName: justForTestTenant
+  namespace: my-app
+  credentials:
+    appDexSecret: dexSecret
+    cookieSecret: cookieSecret
+  spec:
+    applications:
+    - domain: tenant.example.com
+      ingressClassName: test
+      ingressSecretName: test
 `)
 			hec.ValuesSet("userAuthn.idTokenTTL", "2h20m4s")
 
@@ -166,6 +177,18 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
       name: "test-05f0e90e-dex-authenticator"
       truncated: false
       hash: ""
+"test-tenant@my-app":
+  name: "test-tenant-dex-authenticator"
+  truncated: false
+  hash: ""
+  secretName: "dex-authenticator-test-tenant"
+  secretTruncated: false
+  secretHash: ""
+  ingressNames:
+    "0":
+      name: "test-tenant-dex-authenticator"
+      truncated: false
+      hash: ""
 `)
 			hec.HelmRender()
 		})
@@ -178,6 +201,7 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
 			Expect(hec.KubernetesResource("PodDisruptionBudget", "d8-test", "test-dex-authenticator").Exists()).To(BeTrue())
 			Expect(hec.KubernetesResource("VerticalPodAutoscaler", "d8-test", "test-dex-authenticator").Exists()).To(BeTrue())
 			Expect(hec.KubernetesResource("Secret", "d8-test", "registry-dex-authenticator").Exists()).To(BeFalse())
+			Expect(hec.KubernetesResource("Secret", "my-app", "registry-dex-authenticator").Exists()).To(BeFalse())
 
 			secret := hec.KubernetesResource("Secret", "d8-test", "dex-authenticator-test")
 			Expect(secret.Exists()).To(BeTrue())
@@ -218,7 +242,7 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
 
 			deploymentTest := hec.KubernetesResource("Deployment", "d8-test", "test-dex-authenticator")
 			Expect(deploymentTest.Exists()).To(BeTrue())
-			Expect(deploymentTest.Field("spec.template.spec.imagePullSecrets").Exists()).To(BeFalse())
+			Expect(deploymentTest.Field("spec.template.spec.imagePullSecrets").String()).To(MatchJSON(`[{"name":"deckhouse-registry"}]`))
 			Expect(deploymentTest.Field("spec.template.spec.nodeSelector").String()).To(MatchJSON(`{"testnode": ""}`))
 			Expect(deploymentTest.Field("spec.template.spec.tolerations").String()).To(MatchYAML(`
 - key: foo
@@ -275,7 +299,7 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
 
 			deploymentTest2 := hec.KubernetesResource("Deployment", "d8-test", "test-2-dex-authenticator")
 			Expect(deploymentTest2.Exists()).To(BeTrue())
-			Expect(deploymentTest2.Field("spec.template.spec.imagePullSecrets").Exists()).To(BeFalse())
+			Expect(deploymentTest2.Field("spec.template.spec.imagePullSecrets").String()).To(MatchJSON(`[{"name":"deckhouse-registry"}]`))
 			Expect(deploymentTest2.Field("spec.template.spec.nodeSelector").String()).To(MatchJSON(`{"node-role.deckhouse.io/system": ""}`))
 			Expect(deploymentTest2.Field("spec.template.spec.tolerations").Exists()).To(BeTrue()) // default taints
 
@@ -311,6 +335,11 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
 			}
 			Expect(oauth2proxyArgTest4).Should(ContainElement("--cookie-expire=2h20m5s"))
 			Expect(oauth2proxyArgTest4).Should(ContainElement("--cookie-refresh=2h20m4s"))
+
+			Expect(hec.KubernetesResource("Secret", "my-app", "dex-authenticator-test-tenant").Exists()).To(BeTrue())
+			deploymentTenant := hec.KubernetesResource("Deployment", "my-app", "test-tenant-dex-authenticator")
+			Expect(deploymentTenant.Exists()).To(BeTrue())
+			Expect(deploymentTenant.Field("spec.template.spec.imagePullSecrets").Exists()).To(BeFalse())
 		})
 	})
 
--- a/modules/150-user-authn/template_tests/authenticator_test.go
+++ b/modules/150-user-authn/template_tests/authenticator_test.go
@@ -177,7 +177,7 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
 
 			Expect(hec.KubernetesResource("PodDisruptionBudget", "d8-test", "test-dex-authenticator").Exists()).To(BeTrue())
 			Expect(hec.KubernetesResource("VerticalPodAutoscaler", "d8-test", "test-dex-authenticator").Exists()).To(BeTrue())
-			Expect(hec.KubernetesResource("Secret", "d8-test", "registry-dex-authenticator").Exists()).To(BeTrue())
+			Expect(hec.KubernetesResource("Secret", "d8-test", "registry-dex-authenticator").Exists()).To(BeFalse())
 
 			secret := hec.KubernetesResource("Secret", "d8-test", "dex-authenticator-test")
 			Expect(secret.Exists()).To(BeTrue())
@@ -218,6 +218,7 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
 
 			deploymentTest := hec.KubernetesResource("Deployment", "d8-test", "test-dex-authenticator")
 			Expect(deploymentTest.Exists()).To(BeTrue())
+			Expect(deploymentTest.Field("spec.template.spec.imagePullSecrets").Exists()).To(BeFalse())
 			Expect(deploymentTest.Field("spec.template.spec.nodeSelector").String()).To(MatchJSON(`{"testnode": ""}`))
 			Expect(deploymentTest.Field("spec.template.spec.tolerations").String()).To(MatchYAML(`
 - key: foo
@@ -274,6 +275,7 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
 
 			deploymentTest2 := hec.KubernetesResource("Deployment", "d8-test", "test-2-dex-authenticator")
 			Expect(deploymentTest2.Exists()).To(BeTrue())
+			Expect(deploymentTest2.Field("spec.template.spec.imagePullSecrets").Exists()).To(BeFalse())
 			Expect(deploymentTest2.Field("spec.template.spec.nodeSelector").String()).To(MatchJSON(`{"node-role.deckhouse.io/system": ""}`))
 			Expect(deploymentTest2.Field("spec.template.spec.tolerations").Exists()).To(BeTrue()) // default taints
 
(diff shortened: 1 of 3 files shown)


modules/150-user-authn/templates/dex-authenticator/deployment.yaml
```diff
@@ -165,8 +165,6 @@ spec:
       {{- include "helm_lib_priority_class" (tuple $context "cluster-low") | nindent 6 }}
       {{- include "helm_lib_pod_anti_affinity_for_ha" (list $context (dict "app" "dex-authenticator" "name" $crd.name)) | nindent 6 }}
       {{- include "helm_lib_module_pod_security_context_run_as_user_deckhouse" . | nindent 6 }}
-      imagePullSecrets:
-      - name: registry-dex-authenticator
```
AlwxSin, line 169: ### 1. 🟡 On `cri: NotManaged` the pods lose their only registry credential — and it fails late, not on upgrade

Both bashible steps that write registry auth to the node are gated on the same condition:

- `candi/bashible/common-steps/all/030_configure_containerd_registry.sh.tpl:15`
- `candi/bashible/common-steps/all/032_configure_containerd.sh.tpl:15`

```
{{- if or ( eq .cri "Containerd") ( eq .cri "ContainerdV2") }}
```

`NotManaged` is a documented `defaultCRI` value (`candi/openapi/cluster_configuration.yaml:148`) where Deckhouse configures nothing on the node. After this change a DexAuthenticator pod in *any* namespace has no credential at all on such nodes. CE is unaffected (anonymous `registry.deckhouse.io` pull); EE/FE/SE and private or airgapped registries are.

It will not show up during the upgrade: `helm_lib_module_image` emits digest refs and no `imagePullPolicy` is set, so the default is `IfNotPresent` and the rolled pod reuses the cached image. It surfaces on the first cold pull — new node, image GC, re-provision. That is also why the green `Tests` job says nothing about it; e2e runs on Containerd.

Suggested remedy, which also removes the collateral damage: keep `imagePullSecrets: [{name: deckhouse-registry}]` when `$crd.namespace` has a `d8-` / `kube-` prefix. Every platform DexAuthenticator sits in such a namespace — `d8-system` (deckhouse-tools, documentation), `d8-istio`, `d8-cni-cilium`, `d8-openvpn` — and `deckhouse-registry` already exists in all of them, so a three-line conditional closes the tenant leak while leaving the platform authenticators on the same credential path as every other module pod. Tenant namespaces then get no `imagePullSecrets` reference at all, which also sidesteps any question about kubelet behaviour on a missing pull secret.

modules/150-user-authn/templates/dex-authenticator/secret.yaml
```diff
@@ -21,18 +21,4 @@ metadata:
...
----
-apiVersion: v1
-kind: Secret
-metadata:
-  name: registry-dex-authenticator
-  namespace: {{ $crd.namespace }}
-  {{- include "helm_lib_module_labels" (list $context (dict "app" "dex-authenticator")) | nindent 2 }}
-type: kubernetes.io/dockerconfigjson
-data:
-  .dockerconfigjson: {{ $context.Values.global.modulesImages.registry.dockercfg }}
```
AlwxSin, line 35: ### 2. 🟡 The pattern is not fixed — istio still copies the same credential into every mesh namespace

The description says this change "stops distributing a cluster-level registry pull credential into tenant-readable namespaces". It stops one of at least two distributors.

`modules/110-istio/templates/registry-secret.yaml:13-32` renders `Secret/d8-istio-sidecar-registry` with

```
.dockerconfigjson: {{ $.Values.istio.registry.dockercfg | default $.Values.global.modulesImages.registry.dockercfg }}
```

into every entry of `istio.internal.applicationNamespaces`, and the injected sidecars reference it (`modules/110-istio/files/v1x27/templates/sidecar-injection-values.yaml:70`, `modules/110-istio/templates/control-plane/iop/iop.yaml:184`). `istio.registry.dockercfg` is internal-only — declared at `modules/110-istio/openapi/values.yaml:5` with no `config-values.yaml` counterpart — so there is no operator-facing way to point it at a scoped credential; it always falls back to the platform dockercfg.

Blast radius is larger than the one fixed here: every mesh-enrolled tenant namespace, not just the ones that happen to host a DexAuthenticator.

Not a change request on this branch — out of its scope. But either the description's claim narrows to user-authn, or a companion issue lands for istio; otherwise the fix reads as complete when it is not.

modules/150-user-authn/template_tests/authenticator_test.go
```diff
@@ -177,7 +177,7 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
 
 			Expect(hec.KubernetesResource("PodDisruptionBudget", "d8-test", "test-dex-authenticator").Exists()).To(BeTrue())
 			Expect(hec.KubernetesResource("VerticalPodAutoscaler", "d8-test", "test-dex-authenticator").Exists()).To(BeTrue())
-			Expect(hec.KubernetesResource("Secret", "d8-test", "registry-dex-authenticator").Exists()).To(BeTrue())
+			Expect(hec.KubernetesResource("Secret", "d8-test", "registry-dex-authenticator").Exists()).To(BeFalse())
```
AlwxSin, line 226: ### 4. 🟡 The new assertions never exercise a tenant namespace

Every DexAuthenticator in the fixture lives in `d8-test` (`namespace: d8-test` at lines 49, 86, 102, 111, 322, 340, 365). So all three flipped assertions — this one plus `imagePullSecrets` absent on both Deployments at lines 221 and 278 — are verified only for a system-prefixed namespace. The scenario the PR exists for, a DexAuthenticator in an operator-created tenant namespace, has no template test at all. Adding one CRD in something like `my-app` and asserting there would cover it.

This matters concretely if finding 1's conditional is adopted: the correct expectations then differ by namespace prefix, and the current fixture can only express one of the two branches — it would keep passing with the tenant branch entirely unimplemented.

AlwxSin (commented):
### 3. 🟡 The changelog and the description assert the wrong thing about how other module workloads pull

This one has no line to anchor to — it is about the PR body and the `changes` block, since `CHANGELOG/CHANGELOG-v*.yml` is generated from them.

Both say module images "continue to be pulled with the node's CRI registry credentials (`hosts.toml` / equivalent), the same path used by other module workloads that do not place a dockercfg in the tenant namespace."

`grep -rn imagePullSecrets --include='*.yaml' modules/ ee/` returns ~50 platform workloads, and every one of them references `imagePullSecrets: deckhouse-registry` in its own `d8-*` namespace, backed by a per-module `templates/registry-secret.yaml`. Nothing in the platform relies on node CRI auth as its pull path today — after this PR, DexAuthenticator becomes the first workload that does.

On an `impact_level: high` entry that goes into the release notes and is slated for backport to 1.77/1.76, that unconditional sentence needs the CRI qualifier from finding 1: it holds for `Containerd`/`ContainerdV2`, not for `NotManaged`.

The rest of the `impact` text — Secret deletion, the single pod roll, the copied credential having to be treated as disclosed, rotation being a separate operator action — is accurate and unusually thorough.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
Jabejixo: Adopted. `d8-*` / `kube-*` keep `imagePullSecrets: [{name: deckhouse-registry}]`; tenant namespaces still get none. Pushed in 80633f96fd.

modules/150-user-authn/templates/dex-authenticator/secret.yaml
Jabejixo: Agreed, out of scope for this branch. The description now says this PR only stops the user-authn copy and names the remaining istio `d8-istio-sidecar-registry` path. Not opening a companion issue from here unless you want one.

modules/150-user-authn/template_tests/authenticator_test.go
Jabejixo: Added a `DexAuthenticator` in `my-app`: no `registry-dex-authenticator`, no `imagePullSecrets`. The `d8-test` fixtures now expect `deckhouse-registry`.

Jabejixo (commented):
Changelog and description updated: platform authenticators stay on `deckhouse-registry`; tenant pods rely on node CRI only for Containerd/ContainerdV2, not NotManaged. This is not the pull path used by other module workloads.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
```diff
@@ -165,8 +165,10 @@ spec:
       {{- include "helm_lib_priority_class" (tuple $context "cluster-low") | nindent 6 }}
       {{- include "helm_lib_pod_anti_affinity_for_ha" (list $context (dict "app" "dex-authenticator" "name" $crd.name)) | nindent 6 }}
       {{- include "helm_lib_module_pod_security_context_run_as_user_deckhouse" . | nindent 6 }}
+{{- if or (hasPrefix "d8-" $crd.namespace) (hasPrefix "kube-" $crd.namespace) }}
```
AlwxSin, line 168: ### 6. 🔵 The `kube-` half of the conditional is untested and currently dead

No fixture in `modules/150-user-authn/template_tests/authenticator_test.go` uses a `kube-`-prefixed namespace — the new coverage is `d8-test` and `my-app` — and no in-tree DexAuthenticator lives in one either: the five platform authenticators sit in `d8-istio`, `d8-cni-cilium`, `d8-openvpn` and `d8-system`. So a typo in that half of the `or` would pass CI unnoticed.

Either add a third CRD in something like `kube-something` and assert `imagePullSecrets` there, or drop the branch. Keeping it for symmetry with how the platform describes system namespaces elsewhere is defensible — `modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/namespaces/is_system.go:22` and `modules/160-multitenancy-manager/hooks/alert_on_grant_forbidden_resource_use.go:48` both treat `kube-` as system — but then it is worth one line of coverage.

The `d8-` half is properly guarded in both directions by the new fixtures, so this is minor.

AlwxSin (commented):
### 5. 🟡 The changelog rewrite lost the rollout line, and it is now broader than it was

No line to anchor to — this is about the PR body and the `changes` block, since `CHANGELOG/CHANGELOG-v*.yml` is generated from them.

The previous `impact` ended with: "the authenticator Deployments lose imagePullSecrets, **so those pods roll once**." The rewritten text has no rollout sentence at all, and for system namespaces it says Deployments "keep imagePullSecrets: deckhouse-registry" — which reads as *unchanged* for those namespaces.

They do not keep it. The name in the pod spec goes from `registry-dex-authenticator` (base, `deployment.yaml:169`) to `deckhouse-registry`. That is a `spec.template` change, so a new ReplicaSet and a rollout — and now for **every** DexAuthenticator in the cluster, not just the tenant ones: kiali (`d8-istio`), hubble-ui (`d8-cni-cilium`), openvpn (`d8-openvpn`), deckhouse-tools and documentation (`d8-system`), plus every tenant authenticator.

DexAuthenticator sits on the ingress auth path, and this is an `impact_level: high` entry slated for backport to 1.77/1.76, so the rollout line should come back — worded broader than the original, since it now covers all of them rather than only the tenant ones.

The rest of the rewrite is a clear improvement: splitting `d8-*` / `kube-*` from tenant namespaces, naming `Containerd` / `ContainerdV2` as where the node pull path exists, and spelling out the `NotManaged` consequence per edition all read accurately now.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
Jabejixo: Kept the `kube-` branch (same system-namespace cut as multitenancy-manager). Added a `kube-test` DexAuthenticator that expects `imagePullSecrets: deckhouse-registry` and no `registry-dex-authenticator` Secret.

Jabejixo (commented):
Changelog and description now say every DexAuthenticator rolls once: d8-*/kube-* switch imagePullSecrets from registry-dex-authenticator to deckhouse-registry; tenant Deployments drop it. That covers the platform authenticators on the ingress auth path as well as tenant ones.

PR #22212, opened 2026-08-17
Bootstrap a static cluster of immutable nodes, and immutable converge

borg-z:
Bootstraps a Deckhouse control plane on bare metal or on static VMs that run Deckhouse Engine and have no cloud-init. Such a machine boots, finds no configuration and waits on port 50000. dhctl pushes the configuration there. Everything after that (the handoff channel, the kubeconfig, installing Deckhouse) works the way it already does in a cloud.

## What is in the PR

- `dhctl bootstrap` for static immutable clusters. `--master-host <node-name>=<address>` names the machines, dhctl pushes each one its `NodeConfig` document and records the push in the state cache, so a rerun never sends a second document to a machine that already has one.
- The operator describes a machine with a `NodeConfig` document next to `ClusterConfiguration`: network, storage and the node IP, the fields only the machine knows. The machine serves its own inventory on `/inventory`, so those fields can be copied from it.
- `NodeConfigTemplate` is served by an aggregated API in node-controller. `d8 k get nodeconfigtemplate <ng>` renders the cluster's half of a node's document on demand, so no bootstrap token is stored in etcd. Reading it needs the same rights as adding a node, in rbacv2 and in legacy user-authz.
- A NodeGroup with `systemType: Immutable` gets its document from dhctl whatever kind of master the cluster has.
- `dhctl converge` learns about immutable nodes. It converges the base infrastructure, refuses the node phases for immutable NodeGroups, and runs over the Kubernetes API through the bastion when no SSH host is given. Older converge defects found on the way are fixed: a pure VM delete was not seen as destruction, an etcd learner counted as a member, and a failed control-plane hook was silently replaced by a no-op hook.
- The node moves to Deckhouse Engine v0.1.9 (sysext ID `deckhouse-engine`).

## Verified live

On DVP, with the CI install image of this branch: a static cluster of three immutable masters, a cloud cluster with three immutable masters plus immutable CloudPermanent and CloudEphemeral groups, and a cluster with a classic master and immutable groups. Converge over the API through the bastion and converge over SSH both exit 0 without touching a node.

One thing learned on the way: the installer image (ISO or qcow2) must be the same engine version the cluster installs. An older ISO copies its own rootfs to the disk and the node never gets a kubelet. That is an installer defect in the engine repository, not in dhctl.

## Node address as a property of an interface

[...]

PR #22348, opened 2026-08-20
[dhctl] Converge: refuse immutable node groups, restore the master guards

borg-z:
## Description

Stacked on #22212 (base branch `immutable-static-bootstrap`).

`dhctl converge` was never taught about immutable (olcedar) nodes. Walking that path today
ends in one of two ways, both silent: the master hook is never built, so a master VM is
recreated while it still holds its etcd membership and its Node object; or a new VM is
handed the group-wide bashible cloud-init and never joins the cluster. This PR closes the
node phases for such groups and fixes the classic defects that made the failure silent.

**Classic defects (reachable without any immutable node, cherry-pickable to `main`)**

- `Match` ignored `Change.Before`, so a pure delete — including deleting the master
  `VirtualMachine` itself — did not raise `HasVMDestruction()`, and the control plane hook
  did not run at all.
- `waitEtcdHasMember` accepted a member that is still a learner. A learner does not vote,
  so a rolling replace of three masters could walk the quorum down to zero while every step
  looked successful.
- `newHookForUpdatePipeline` returned a nil hook on error, and the runner substitutes
  `DummyHook` for a nil one. Any SSH provider failure silently turned a guarded master
  update into an unguarded one: no control plane role removed, no Node object deleted, no
  wait for the member to come back, and no `IsAllNodesReady` gate before destroying the
  next master. It now returns an error. `getHook()` keeps its `DummyHook` — it is
  legitimate for pipelines that carry no hook by construction.
- `AfterAction` panicked (`panic("Node interface is not ssh")`) where an error belongs.
- `CheckSSHHosts` asks to confirm the node-to-host mapping on **every** run, and with no
  terminal `Confirmation.Ask()` answers "no" by default. That silent "no" was the reason the
  hook was never built in automation: a converge with no TTY recreated masters with zero
  guards. The mapping is now accepted without a terminal and written to the log, with the
  full node → host list, so an incident has something to read. A destructive plan still has
  its own approval.
- The node index was parsed twice: once in `sortNodeNames`, then again per node, where an
  unparsable name returned nil — reporting success for a node that was skipped. The second
  parse is gone; the index is carried from the single place that computes it. Those
  branches were unreachable (`sortNodeNames` fails the whole pass first), so this is dead
  code removal, not a behaviour fix.

**Immutable**

- The node phases refuse to run while a group they walk is immutable, in both entry points
  (`converge` and `converge-migration`). The set is exactly what converge walks: the master
  group, the CloudPermanent groups of the config, and groups that still hold state after
  being dropped from it. An immutable CloudEphemeral group does not trip the gate — the
  bootstrap provider owns those nodes and converge never touches them. Fail-closed: an
  unreadable NodeGroup list refuses the phase too. `systemType` is read from live NodeGroup
  objects, never guessed: the bashible bootstrap Secret exists for an immutable group as
  well, so its presence proves nothing.
- The base infrastructure phase is untouched and keeps converging, which is what the
  in-cluster auto-converger already does on such clusters.

**SSH-less transport**

`SSHless` is a property of how dhctl was started — Kubernetes credentials of its own and no
SSH host known — not a property of the cluster, and a classic cluster converged with
`--kubeconfig` alone gets the same behaviour. "Credentials of its own" is
`providerinitializer.KubeAuthMode`, not "any kube flag": `--kube-config-context` alone still
tunnels over SSH. Hosts are looked up when the answer is used rather than when the context
is built — the state cache they come from is initialized later, and a converge with hosts in
the cache would otherwise be mistaken for an SSH-less one. With it the client switcher skips
creating the `NodeUser`: that user is delivered by bashible, so on a node without it the
wait never ends. This is
deliberately *not* built on `CommanderMode`, which carries about fifteen other behaviours,
four of them harmful here (the converge lock is not taken, MetaConfig and states come from
the local cache, and the Deckhouse configuration phase overwrites cluster configuration).

Running `dhctl converge` against an immutable cluster today can destroy a master while it is
still an etcd member. The refusal makes that impossible, and the classic fixes restore the
guards that protect a rolling master replace on any cluster.

## What is intentionally not here

- Joining a master to an immutable cluster (rendering the join payload instead of the
  bashible Secret) and removing or replacing an immutable master. Both need a live run that
  has never been done, and no test that fails on the old code can prove them. The refusal
  is a scaffold with an expiry date and is removed by that work.
- The per-node liveness probe that replaces the SSH check (a direct TLS call to
  `https://<masterIP>:6443` verifying the cluster UUID) and reading
  `NodeConfig.status.appliedGeneration` as proof-of-applied. Both only serve node
  operations, which this PR refuses, so they would be code with no reachable call site.

## Verified

Every fix ships with a check that fails on the old code (shown failing, then passing):

| Fix | Check |
|---|---|
| `Match` reads `Before` | `TestMatch/pure_delete_carries_the_kind_in_Before_only` |
| etcd learner | `TestHasVotingMember` on real `member list -w json` output, plus `TestHasVotingMemberWithStaleDuplicate` for a name listed twice while the stale member is removed |
| hook error instead of nil | `TestNewHookForUpdatePipelineFailsWithoutSSHHosts` |
| mapping confirmed with no TTY | `TestHostsMappingConfirmedWithoutTerminal` |
| panic → error | `TestAfterActionReportsUnavailableSSHClient` |
| immutable gate | `TestRefuseImmutableNodeGroups` (4 cases incl. fail-closed and a group converge never walks), `TestConvergedNodeGroupNames` |
| sshless switcher | `TestSwitcherSkipsNodeUserWhenSSHless`, `TestSSHlessIsFalseWhileHostsAreKnown` |

`golangci-lint` reports no issues for the changed packages. `go test ./...` for dhctl is
green except three tests that fail identically on the base branch and are untouched here:
`TestRootWritesFileAndTTY` (`cmd/dhctl`), `TestGetCloudApiURLFromMetaConfig` and
`TestChecker_CheckDeckhouseUser` (`pkg/preflight/checks`).

## Verified on a live cluster

Three immutable masters on DVP (Deckhouse Olcedar, containerd 2.2.7), bootstrapped by an
installer built from this branch. Converge was run over the Kubernetes API alone — those nodes
answer no sshd — through a bastion tunnel and a retargeted admin kubeconfig.

| | this branch | the installer without these commits |
|---|---|---|
| base infrastructure phase | converged, 3.7s | converged |
| node phases | refused: `converge of nodes is not supported yet for immutable NodeGroups: master …` | proceeded |
| converge lock | taken and released | taken |
| `NodeUser` in the cluster | never created | **created** `d8-dhctl-converger`, which no immutable node will ever apply |
| outcome | clean exit in seconds | **panic**: `nil pointer dereference` in `KubeClientSwitcher.createNodeUser` |

The panic is the same class as the one fixed in `AfterAction`, on a second code path: with no
SSH provider the switcher dereferences it. This PR does not patch that line — it never reaches
it — so the panic still guards nothing on a path that no longer runs.

Not verified: replacing or rolling a master (needs a destructive plan and the join contract that
is deliberately out of scope), and a classic bashible cluster (no such stand was up).

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Converge builds the control plane hook or fails loudly instead of silently replacing a master VM with no hook, treats an etcd learner as not yet returned, and detects a VM destroyed by a pure delete.
impact_level: low
```

```changes
section: dhctl
type: feature
summary: Converge refuses the node phases for immutable NodeGroups and skips the bashible-delivered NodeUser when it runs over the Kubernetes API without SSH.
impact_level: low
```


borg-z:
Superseded by #22212. All eight commits from this branch were cherry-picked onto `immutable-static-bootstrap` after #21557 merged, and the description of #22212 now covers both halves. Verified before the move: the diff of the eight commits is byte-identical to the one here (31 files, same per-file numstat), and `dhctl` plus `node-controller` tests add no failure beyond the ones already red on `main`.

Closing in favour of #22212. The branch `immutable-converge` is left in place for reference.

PR #22362, opened 2026-08-21
[user-authn] Use the release branch values path in the basic-auth-proxy test fixture

Jabejixo:
## Description

The backport of #22214 to this branch (#22355) brought a test fixture that sets `userAuthn.internal.publishAPI.enabled` and `userAuthn.internal.publishAPI.addKubeconfigGeneratorEntry`. That values path exists from 1.77 onwards; on `release-1.76` the setting lives at `userAuthn.publishAPI`, so values validation rejects the fixture and every spec in the `basic auth proxy` context fails.

Ten specs fail on the branch as it stands, `Summarizing 10 Failures` on a clean checkout of `release-1.76`. The fix points the two lines at the path this branch serves; nothing else changes, and the assertions the backport added keep their meaning — they verify that a multiline value cannot break out of a single argument of the proxy container.

Verified on the branch: `go test ./modules/150-user-authn/template_tests/ -count=1` fails before and passes after.

Test-only change; no rendered manifest and no cluster component is affected.

The branch is red without it, which blocks every subsequent backport into 1.76 — each one has to be green on the same package before it can merge.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: fix
summary: Point the basic-auth-proxy test fixture at the publishAPI values path the 1.76 branch serves.
impact_level: low
```

### DIFF
--- a/modules/150-user-authn/template_tests/basic_auth_proxy_test.go
+++ b/modules/150-user-authn/template_tests/basic_auth_proxy_test.go
@@ -91,8 +91,8 @@ var _ = Describe("Module :: user-authn :: helm template :: basic auth proxy", fu
 		hec.ValuesSet("userAuthn.internal.selfSignedCA.cert", "test")
 		hec.ValuesSet("userAuthn.internal.selfSignedCA.key", "test")
 
-		hec.ValuesSet("userAuthn.publishAPI.enabled", true)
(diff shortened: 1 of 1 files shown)


Jabejixo:
Folded into #22366. The combined backport of #22222 cannot be green on this branch without this fixture fix, so it carries it directly. This branch is emptied first, so it proposes no changes.

PR #22363, opened 2026-08-21
Backport: [testing] Stop sharing Helm render capabilities between tests

Jabejixo:
> Backport of #22222 to `release-1.76`. The automatic cherry-pick failed because the squash commit of #22222 on `main` also carries #22217 and #22234 — both were merged into its branch before it merged — so picking it whole brings three changes and their conflicts at once.
>
> This PR carries only the two commits that belong to #22222 itself, cherry-picked individually. Both applied without conflict. #22217 and #22234 are backported separately.
>
> Verified on the branch: `go build ./testing/...` succeeds and `go test ./modules/002-deckhouse/... -count=1` passes. The `150-user-authn/template_tests` package is red on this branch for an unrelated reason, fixed in #22362.

## Description

`testing/library/helm.Renderer` built a capability set for each render but never used it: `chartutil.ToRenderValues` was called with `nil` capabilities. It only appeared to work because `caps := chartutil.DefaultCapabilities` copies a package-level **pointer**, so `caps.APIVersions = vers` mutated the process-wide default that Helm then fell back to. Every render extended the same list for the rest of the test binary.

This copies the capabilities per render and passes them explicitly, and adds an opt-in `WithAPIVersions` render option to `testing/helm`.

The option makes a whole class of templates testable for the first time. Helm's default capabilities carry bare `group/version` strings, so any template guarded by `helm_lib_kind_exists` rendered nothing in template tests regardless of the values supplied.

Does not restart any cluster component; test-only change.

Two defects in shared test infrastructure:

- **Cross-test leakage.** Mutating `chartutil.DefaultCapabilities` means one package's render can change what a later, unrelated render sees. Copying per render removes the shared mutable state.
- **Dead code path.** The assembled capabilities were discarded at the call site, so the intent of the existing `autoscaling.k8s.io/v1/VerticalPodAutoscaler` handling was only satisfied as a side effect of the leak. The pre-existing loop guarding that append was also ineffective — it reassigned its result on every iteration instead of accumulating, so only the last comparison survived.

`WithAPIVersions` is additive and opt-in: renders that do not pass it keep Helm's defaults, so existing tests are unaffected.

Not necessarily. It is test-only, but it cherry-picks cleanly and is a prerequisite for backporting admission-policy tests to 1.77 and 1.76.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: testing
type: fix
summary: Copy Helm render capabilities per render instead of mutating the shared defaults, and allow tests to declare available API kinds.
impact_level: low
```

### DIFF
--- a/testing/helm/init.go
+++ b/testing/helm/init.go
@@ -286,8 +286,11 @@ func WithFilteredRenderOutput(m map[string]string, filters []string) Option {
 }
 
 // WithAPIVersions presents the given group/version/Kind strings to the chart as available in the
-// cluster. Templates guarded by helm_lib_kind_exists render nothing without it, because Helm's
-// default capabilities list group/version pairs only.
+// cluster. Helm's default capabilities are group/version pairs with no kind in them, so a guard
+// that asks after a kind answers no without this. helm_lib_kind_exists takes a bare kind and looks
+// for it as a case-insensitive "/<Kind>" suffix in the list, while helm_lib_api_version_exists and
+// inline .Capabilities.APIVersions.Has compare a whole "group/version/Kind" string; spelling an
+// entry that way satisfies both.
 func WithAPIVersions(versions ...string) Option {
 	return func(options *configOptions) {
 		options.apiVersions = append(options.apiVersions, versions...)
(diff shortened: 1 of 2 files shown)


Jabejixo:
Superseded by #22366, which carries #22222 in the shape it has on main and release-1.77 — one commit that already includes #22217 and #22234. This branch is emptied first, so it proposes no changes.

PR #22368, opened 2026-08-21
[node-manager] add the missing Russian CRD docs

borg-z:
Two CRDs shipped by #21557 have no Russian counterpart, so `Docs Validation (base)` is red on that PR:

```
* modules/040-node-manager/crds/nodebootstrapconfig.yaml ... ERROR: related is absent
* modules/040-node-manager/crds/nodebootstrapconfigtemplate.yaml ... ERROR: related is absent
```

`testing/validate_doc_changes.sh` requires a `doc-ru-<name>.yaml` beside every changed resource. This adds the two missing files and nothing else.

Reproduced and verified locally with the repo's own script against the PR base: red before, `Validation successful.` after.

## Base

Targets `immutable-on-21372` (#21557), not `main` — the two CRDs it documents live there.

## Changelog entries

```changes
section: node-manager
type: chore
summary: Add the Russian documentation for the NodeBootstrapConfig and NodeBootstrapConfigTemplate CRDs.
impact_level: low
```

### DIFF
--- a/modules/040-node-manager/crds/doc-ru-nodebootstrapconfig.yaml
+++ b/modules/040-node-manager/crds/doc-ru-nodebootstrapconfig.yaml
@@ -1,0 +1,30 @@
+spec:
+  versions:
+    - name: v1alpha1
+      schema:
+        openAPIV3Schema:
+          description: |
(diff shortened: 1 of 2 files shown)


borg-z:
Not needed: these are internal CRDs and are not translated.

PR #22372, opened 2026-08-21
[deckhouse] wait until nelm deletes resources

ipaqsa:
## Description

`RemoveApp` / `RemoveModule` were fire-and-forget: the reconciler handed the package to the runtime and released the finalizer in the same reconcile, so the CR disappeared while the teardown was still running. Both removers now report whether the teardown has finished, and the reconciler holds the finalizer until they do.

- `lifecycle.Store.RemovalState` exposes explicit teardown state: no entry → nothing left to tear down; entry with `removing=true` → teardown in flight; any other entry → teardown not issued yet. `HandleEvent(EventRemove)` sets the marker, `Update` clears it on recreation, and `Store.Delete` guards on it. Removal state no longer depends on an empty package version.
- `RemoveApp`, `RemoveModule` and `RemoveEmbeddedModule` return `bool` and are idempotent by contract. A call made while a teardown runs returns `false` **without** re-issuing `EventRemove`: that event cancels the whole context tree, so re-issuing it would abort the very uninstall the caller is waiting for.
- A failed `Load` does **not** delete the previous release: the update path keeps it with `Disable(keep=true)`, so the user's workload stays running. Only if the `Application` / `Module` resource is later deleted and its instance is already absent from `r.apps` / `r.modules`, the remove path enqueues `Uninstall` by namespace and release name instead of skipping Helm deletion. The normal loaded-instance remove path still uses the unchanged `Disable` task and runs `beforeDeleteHelm` / `afterDeleteHelm` hooks.
- `HandleEvent(EventSchedule)` returns nil while removal is in flight. This prevents a scheduler event already read before `RemoveNode` from appending `Configure` / `Enable` / `Run` work after teardown has started.
- `handleDelete` polls the remover (`RequeueAfter: 5s`) and keeps the finalizer until it reports done. The teardown finishes in a queue callback the API server never sees, so asking again is the only way to observe it.
- Detaching the `ApplicationPackageVersion` / `ModulePackageVersion` moved **after** the teardown. Releasing it earlier let it be garbage collected — together with the package files on disk — while the uninstall still needed them.
- `preflight` no longer preserves an Application that is already terminating. The runtime forgets a teardown across a restart, and a terminating Application never reaches `handleCreateOrUpdate`, so it is never registered and `RemoveApp` would report that nothing is tracked; the boot cleanup is the only owner left of its release.

`kubectl delete application` returned immediately while the Helm release, its workloads and its hooks were still being torn down. Three consequences:

1. The object was gone from the API server while its resources were still running, so nothing downstream could tell deletion from completion.
2. A failing uninstall was invisible — the status had been deleted along with the CR, and the release stayed orphaned until some later restart's cleanup happened to pick it up.
3. The package version could be released, and its files removed from disk, in the middle of the uninstall.

When an `Application` / `Module` resource is deleted, its Helm release is normally uninstalled by the `Disable` task (`beforeDeleteHelm` → `nelm.Delete` → `afterDeleteHelm`). If no loaded package instance remains at that moment, the fallback `Uninstall` task stops the release monitor and calls `nelm.Delete` directly by namespace and name. The files are then taken off disk by `Undeploy` (or `Dummy` anchors cleanup for embedded modules), and the runtime state is dropped by the cleanup riding the last task's `OnDone` — which only fires on success. That callback is therefore the honest "fully removed" signal, and it is what the finalizer now waits for.

**Behaviour change to be aware of:** the wait is unbounded, deliberately. The queue retries a failed teardown with exponential backoff, so a stuck uninstall now keeps the object in `Terminating` instead of orphaning its resources. Deleting such an object means fixing the teardown or removing the finalizer by hand.

**Known limitation, unchanged by this PR:** module releases live in the namespace the boot cleanup deliberately skips, so a Module deleted while the controller was down still has its finalizer released without an uninstall. Applications are covered by the `preflight` change above.

## Verification

- `go test ./deckhouse-controller/internal/packages/runtime/... ./deckhouse-controller/pkg/controller/packages/application/... ./deckhouse-controller/pkg/controller/packages/module/... -race -count=1`
- `golangci-lint run ./deckhouse-controller/internal/packages/runtime/... ./deckhouse-controller/pkg/controller/packages/application/... ./deckhouse-controller/pkg/controller/packages/module/...` — 0 issues

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: chore
summary: Unset a package finalizer when the package is fully deleted.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/internal/packages/runtime/apps.go
+++ b/deckhouse-controller/internal/packages/runtime/apps.go
@@ -164,7 +164,13 @@ func (r *Runtime) loadApp(ctx context.Context, repo registry.Remote, packagePath
 	return app.GetVersion().String(), nil
 }
 
-// RemoveApp removes an application and cancels all its running operations.
+// RemoveApp removes an application, cancels all its running operations and reports whether the
+// teardown has finished. The caller polls it and holds the Application's finalizer until it
+// returns true, so the CR outlives the Helm release it owns.
+//
+// It is idempotent by contract: a call made while the teardown runs must not re-issue
+// EventRemove, because that cancels the whole context tree (lifecycle.Package.newContext) and
+// would restart the very uninstall the caller is waiting for.
 //
 // After the undeploy task succeeds, a cleanup goroutine removes the
 // Store entry and stops the queue. The goroutine is necessary because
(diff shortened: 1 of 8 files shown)


ldmonster (approved):

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit fa325ac81e52a9b3ff2008001d520e4bc2498339 to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/33165231413) for details.

PR #22388, opened 2026-08-21
[user-authn] Prove authority over Kubernetes API access with a dedicated permission

Jabejixo:
## Description

Reworks the authority check of the `dex-allow-access-to-kubernetes.deckhouse.io` policy introduced in #22219.

- Authority is now the `create` permission on an `allow-access-to-kubernetes` subresource of `dexclients` and `dexauthenticators`. No controller serves it; it exists solely so that the authority can be delegated on its own, the way `pods/exec` or the `escalate` verb are. The check is scoped to the request's namespace, so a grant can be limited to one namespace with a RoleBinding or made cluster-wide with a ClusterRoleBinding.
- The platform's core namespaces are excluded by match conditions, reusing the exclusion set of the reserved public hosts policy minus its namespace-prefix condition: system groups, infrastructure users, and thereby the service accounts of `d8-system` and `kube-system`. That covers the module charts which ship an annotated DexAuthenticator, kiali in istio and the UI in cilium-hubble, since those are applied by the deckhouse service account, and it makes the hardcoded bypass of that account redundant. Excluding the whole `d8-` prefix was deliberately not done: a module from any ModuleSource gets a `d8-<module>` namespace, and it should declare this authority in its own RBAC rather than inherit it from where it runs. Provisioners do exactly that already — the Commander chart's `cluster-manager` Role enumerates both subresources.
- The permission rides in the roles that already conveyed this authority, `d8:manage:permission:module:user-authn:edit` and `d8:user-authz:user-authn:cluster-admin`, so module administrators keep working as before. Delegating it to anyone else is a Role or ClusterRole of the admin's own making, bound where it is needed; no extra role is shipped for that.

The policy had no match conditions and derived authority from write access to the `user-authn` ModuleConfig, so it denied every writer that was neither `system:serviceaccount:d8-system:deckhouse` nor an administrator of the module. Cluster provisioning broke on it: an agent patching a DexClient in its own namespace during bootstrap is denied, and the bootstrap task fails with

```
dexclients.deckhouse.io "cluster-..." is forbidden:
ValidatingAdmissionPolicy 'dex-allow-access-to-kubernetes.deckhouse.io' denied request
```

Granting such a provisioner write access to the `user-authn` ModuleConfig would work around it, but that permission is effectively ownership of the whole authentication stack: the holder can add an OIDC provider, enable `publishAPI`, or change the static user policy. The platform's own role model already keeps the two apart, which is why the ModuleConfig was the wrong lever to begin with: `dexclients` live in use roles (`modules/150-user-authn/templates/rbacv2/use/edit.yaml`), `moduleconfigs` in manage roles (`modules/140-user-authz/templates/rbacv2/manage/edit.yaml`). A dedicated permission keeps the original intent — only a cluster administrator decides who may reach the Kubernetes API from a namespace — without coupling it to the module's configuration.

#22219 is already merged into `main`, `release-1.77` and `release-1.76` under milestone v1.76.11, and it denies a provisioning flow that works today. The tags are not cut yet, so this has to land in the same patch releases for users never to see the regression.

## Checklist
- [ ] e2e tests passed.

Documentation for the annotation is not in `main` yet: it is part of the open #22250, which is updated alongside this PR.

The test harness now evaluates match conditions before the validations, the way kube-apiserver does, so the exclusions are under test rather than assumed. Removing the match conditions from the template fails two of the new cases. The authorizer mock narrows the permission to a single resource and a single namespace, so a grant on DexClients is shown not to carry over to DexAuthenticators, nor to another namespace.

Verified on a live cluster that RBAC on a subresource no controller serves is evaluable from a policy: with the policy bound and the permission absent, the request is denied; after a namespaced RoleBinding granting `dexclients/allow-access-to-kubernetes`, the same request is admitted.

## Changelog entries

```changes
section: user-authn
type: fix
summary: The policy gating the allow-access-to-kubernetes annotation no longer applies to platform components, and proves authority with a create permission on the dexclients/allow-access-to-kubernetes and dexauthenticators/allow-access-to-kubernetes subresources instead of write access to the user-authn ModuleConfig.
impact: |
  Setting the annotation no longer requires write access to the user-authn ModuleConfig. Module administrators are unaffected, as their roles already carry the new permission, and so are the platform's own components in `d8-system` and `kube-system`. Any other subject that sets the annotation, a cluster provisioner in particular, has to be granted `create` on the subresource, which for a single namespace is:

      apiVersion: rbac.authorization.k8s.io/v1
      kind: Role
      metadata:
        name: allow-access-to-kubernetes
        namespace: <namespace>
      rules:
      - apiGroups: ["deckhouse.io"]
        resources: ["dexclients/allow-access-to-kubernetes", "dexauthenticators/allow-access-to-kubernetes"]
        verbs: ["create"]

  bound to the subject with a RoleBinding in the same namespace, or with a ClusterRole and a ClusterRoleBinding to grant it cluster-wide.
impact_level: high
```

### DIFF
--- a/modules/150-user-authn/template_tests/validation_test.go
+++ b/modules/150-user-authn/template_tests/validation_test.go
@@ -248,10 +248,29 @@ var _ = Describe("Module :: user-authn :: helm template :: validation", func() {
 					allowed:     true,
 				},
 				{
+					// The other half of the dynamic resource resolution: the grant has to admit the
+					// kind it was issued for, not just deny the kind it was not.
+					name:        "subject holding the allow-access subresource grants access on a dex authenticator",
+					resource:    "dexauthenticators",
+					kind:        "DexAuthenticator",
+					canGrant:    true,
+					annotations: map[string]string{dexAuthenticatorAnnotation: "true"},
+					allowed:     true,
+				},
+				{
 					name:            "subresource granted on the other kind does not carry over",
 					canGrant:        true,
 					grantedResource: "dexauthenticators",
 					annotations:     map[string]string{dexClientAnnotation: "true"},
+					allowed:         false,
+				},
+				{
+					name:            "subresource granted on dex clients does not admit a dex authenticator",
+					resource:        "dexauthenticators",
+					kind:            "DexAuthenticator",
+					canGrant:        true,
+					grantedResource: "dexclients",
+					annotations:     map[string]string{dexAuthenticatorAnnotation: "true"},
 					allowed:         false,
 				},
 				{
@@ -514,10 +533,19 @@ func compileRenderedPolicy(renderedPolicy string) compiledPolicy {
 		Expect(result.Error).To(BeNil(), "variable %q should compile", variable.Name)
 	}
 
+	// Match conditions are compiled by a compiler that holds no variables, because kube-apiserver
+	// rejects a match condition referring to them: they are evaluated before the rest of the policy.
+	// The authorizer, in contrast, is declared there as well. Sharing the compiler above would let a
+	// match condition using variables pass here and be refused by a real API server.
+	matchCompiler, err := admissioncel.NewCompositedCompiler(
+		environment.MustBaseEnvSet(minimalSupportedKubernetesVersion(), true),
+	)
+	Expect(err).ToNot(HaveOccurred())
+
 	matchConditions := make([]admissioncel.ExpressionAccessor, 0, len(policy.Spec.MatchConditions))
 	for _, condition := range policy.Spec.MatchConditions {
 		expression := boolExpression{expression: condition.Expression}
-		Expect(compiler.CompileCELExpression(expression, validationVars, environment.NewExpressions).Error).To(BeNil(),
+		Expect(matchCompiler.CompileCELExpression(expression, validationVars, environment.NewExpressions).Error).To(BeNil(),
 			"match condition %q should compile", condition.Name)
 
 		matchConditions = append(matchConditions, expression)
@@ -537,7 +565,7 @@ func compileRenderedPolicy(renderedPolicy string) compiledPolicy {
 	}
 
 	return compiledPolicy{
-		matchConditions: compiler.CompileCondition(matchConditions, validationVars, environment.NewExpressions),
+		matchConditions: matchCompiler.CompileCondition(matchConditions, validationVars, environment.NewExpressions),
 		validations:     compiler.CompileCondition(validations, validationVars, environment.NewExpressions),
 	}
 }
--- a/modules/150-user-authn/template_tests/validation_test.go
+++ b/modules/150-user-authn/template_tests/validation_test.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	celgo "github.com/google/cel-go/cel"
 	celtypes "github.com/google/cel-go/common/types"
@@ -56,9 +57,9 @@ const (
 	dexAuthenticatorAnnotation = "dexauthenticator.deckhouse.io/allow-access-to-kubernetes"
 
 	deckhouseServiceAccount = "system:serviceaccount:d8-system:deckhouse"
-	// A cluster provisioner running as a platform component. Deckhouse Commander patches DexClients
-	// in its own namespace during cluster bootstrap, which is what makes this case load-bearing.
-	commanderServiceAccount  = "system:serviceaccount:d8-commander:commander"
+	// The identity that patches DexClients during cluster bootstrap: a namespace-scoped service
+	// account outside the core namespaces, holding an enumerated Role rather than broad RBAC.
+	commanderServiceAccount  = "system:serviceaccount:d8-commander:cluster-manager"
 	kubeSystemServiceAccount = "system:serviceaccount:kube-system:some-controller"
 	tenantServiceAccount     = "system:serviceaccount:attacker-ns:tenant"
 	namespaceEditor          = "editor@example.com"
@@ -130,11 +131,13 @@ var _ = Describe("Module :: user-authn :: helm template :: validation", func() {
 				ContainSubstring("allow-access-to-kubernetes subresource"),
 			)
 
-			// Platform components have to be excluded, otherwise the gate denies the provisioning
-			// flows that create these objects on the cluster's behalf.
+			// The core namespaces have to be excluded, otherwise the gate denies the module charts
+			// that ship an annotated DexAuthenticator. Nothing wider: a namespace prefix must not
+			// confer the authority on whatever runs under it.
 			matchConditions := policy.Field("spec.matchConditions").String()
-			Expect(matchConditions).To(ContainSubstring("system:serviceaccount:d8-"))
-			Expect(matchConditions).To(ContainSubstring("system:serviceaccount:kube-"))
+			Expect(matchConditions).To(ContainSubstring("system:serviceaccounts:d8-system"))
+			Expect(matchConditions).To(ContainSubstring("system:serviceaccounts:kube-system"))
+			Expect(matchConditions).ToNot(ContainSubstring("startsWith"))
 
 			binding := hec.KubernetesGlobalResource("ValidatingAdmissionPolicyBinding", allowAccessPolicyName)
 			Expect(binding.Exists()).To(BeTrue())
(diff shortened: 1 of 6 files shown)


z9r5 (approved):

modules/150-user-authn/templates/validation.yaml
```diff
@@ -42,6 +42,16 @@ spec:
...
         resources:   ["dexclients", "dexauthenticators"]
+  # Platform components provision these objects on the cluster's behalf and are already privileged
+  # by virtue of running in a system namespace, so gating them buys nothing and only breaks
+  # provisioning flows. Tenants cannot place a workload in these namespaces to begin with.
+  matchConditions:
+    - name: 'exclude-groups'
+      expression: '!(["system:nodes", "system:serviceaccounts:kube-system", "system:serviceaccounts:d8-system"].exists(e, (e in request.userInfo.groups)))'
+    - name: 'exclude-users'
+      expression: '!(["system:sudouser", "system:apiserver", "system:kube-controller-manager", "system:kube-scheduler", "system:volume-scheduler", "dhctl", "observability"].exists(e, (e == request.userInfo.username)))'
+    - name: 'exclude-system-serviceaccounts'
```
AlwxSin: 🟡 **2. The exclusion set trusts every service account in any `d8-`-prefixed namespace**

The old check exempted exactly one identity, `system:serviceaccount:d8-system:deckhouse`. This condition exempts every SA whose namespace starts with `d8-` or `kube-`.

The comment above justifies it with "tenants cannot place a workload in these namespaces to begin with", and that part holds: `system-ns.deckhouse.io` (`modules/002-deckhouse/templates/validation.yaml:18`) blocks creating `d8-`/`kube-` namespaces for anyone but the deckhouse SA and `dhctl`. What it does not cover is that any module installed from any ModuleSource gets a `d8-<module>` namespace created by the deckhouse SA, so a third-party module's service account now unconditionally holds the authority to mint an `aud=kubernetes` grant.

Is that the intent? If not, the narrower exclusion is the deckhouse SA plus the `d8-commander` / `d8-commander-agent` SAs that the broken provisioning flow actually needs.

For the record, the two SA-group entries in `exclude-groups` above are strictly subsumed by this condition, but forging those groups from an IdP is not possible — `modules/040-control-plane-manager/templates/_authentication_configuration.tpl:31-35` rejects `system:`-prefixed groups and `basic-auth-proxy` strips them in `filterReservedGroups` — so keeping the set byte-identical to `reserved-public-hosts.yaml` looks like the right call.

modules/150-user-authn/templates/validation.yaml
```diff
@@ -70,15 +80,19 @@ spec:
...
+    # model keeps these apart: DexClient lives in use roles, ModuleConfig in manage roles.
+    - expression: '!variables.annotated
         || (variables.annotatedBefore && (!variables.grantsAccess || variables.grantedBefore))
-        || authorizer.group("deckhouse.io").resource("moduleconfigs").name("user-authn").check("update").allowed()'
+        || authorizer.group("deckhouse.io").resource(request.resource.resource).subresource("allow-access-to-kubernetes").namespace(request.namespace).check("create").allowed()'
       reason: Forbidden
       messageExpression: '"the " + variables.annotation + " annotation controls access to the Kubernetes API
-        and may only be introduced or widened by a cluster administrator who is allowed to manage the user-authn module"'
+        and may only be introduced or widened by a subject allowed to create the "
+        + request.resource.resource + "/allow-access-to-kubernetes subresource"'
```
AlwxSin, line 98: 🔵 **3. The deny message points at a permission that has no documentation in the target releases**

The message tells the subject to obtain `create` on `<resource>/allow-access-to-kubernetes`, and the annotation's documentation is deferred to the open #22250. If this lands in `release-1.76` / `release-1.77` while #22250 goes to 1.78, patch-release users who hit the deny have nothing to read: no shipped role to bind, no example Role or RoleBinding in the docs.

Naming the exact subresource in the message mitigates it. A two-line Role/RoleBinding example in the changelog `impact` would close it entirely.

modules/150-user-authn/template_tests/validation_test.go
```diff
@@ -197,11 +214,53 @@ var _ = Describe("Module :: user-authn :: helm template :: validation", func() {
...
+				{
+					// A tenant's own service account is not a platform component: the namespace
+					// prefix is what separates them, and this is the case the policy exists for.
+					name:        "tenant service account grants access",
+					username:    tenantServiceAccount,
+					annotations: map[string]string{dexClientAnnotation: "true"},
+					allowed:     false,
+				},
+				{
+					name:        "subject holding the allow-access subresource grants access",
```
AlwxSin, line 260: 🔵 **4. No positive test for the DexAuthenticator side of the dynamic resource check**

The authority expression resolves the resource dynamically — `.resource(request.resource.resource)`. This case proves a grant admits a `dexclients` write, and the case below proves a grant on `dexauthenticators` does not carry over to `dexclients`, but nothing proves a grant on `dexauthenticators` admits a `dexauthenticators` write.

One more case — `resource: "dexauthenticators"`, `kind: "DexAuthenticator"`, `canGrant: true`, `allowed: true` — would pin both directions of the dynamic resolution.

modules/150-user-authn/template_tests/validation_test.go
```diff
@@ -450,6 +514,15 @@ func compileRenderedPolicy(renderedPolicy string) compiledPolicy {
 		Expect(result.Error).To(BeNil(), "variable %q should compile", variable.Name)
 	}
 
+	matchConditions := make([]admissioncel.ExpressionAccessor, 0, len(policy.Spec.MatchConditions))
+	for _, condition := range policy.Spec.MatchConditions {
+		expression := boolExpression{expression: condition.Expression}
+		Expect(compiler.CompileCELExpression(expression, validationVars, environment.NewExpressions).Error).To(BeNil(),
```
AlwxSin: 🔵 **5. The harness compiles match conditions with the validation variable set**

`validationVars` includes the policy's `spec.variables`. kube-apiserver does not make them available in match conditions: those are evaluated before the rest of the policy, which is why `variables` is rejected there.

The shipped expressions use none, so nothing is wrong today. But a future match condition referencing, say, `variables.annotated` would compile and pass here and be rejected by a real API server. Compiling match conditions against a variable-free environment would make the harness fail where the cluster fails.

modules/150-user-authn/templates/validation.yaml
Jabejixo: Not intended, and the chart settles it: the identity that patches DexClients during bootstrap is a namespace-scoped service account whose Role already enumerates both `allow-access-to-kubernetes` subresources, so the prefix bought the provisioning flow nothing and only handed every `d8-<module>` account the same authority for free. `exclude-system-serviceaccounts` is gone; what remains covers `d8-system` and `kube-system`, which is needed because the two charts shipping an annotated DexAuthenticator, kiali in istio and the cilium-hubble UI, are applied by the deckhouse service account. I checked that nothing else in the tree writes these objects: no controllers, no Go or Python hooks, just those two templates.

modules/150-user-authn/templates/validation.yaml
Jabejixo: Added a `Role` with both subresources and the RoleBinding sentence to the changelog `impact`, so a patch-release user who hits the deny has something to act on without waiting for #22250. No role is shipped to bind: the standalone ClusterRole was dropped because the linter wants it in `user-authz-cluster-roles.yaml`, and delegating this single authority is a Role of the admin's own making anyway.

modules/150-user-authn/template_tests/validation_test.go
Jabejixo: Added, together with the mirror case: a grant on `dexauthenticators` admits a DexAuthenticator write, and a grant on `dexclients` does not admit one. Both directions of the dynamic resolution are now pinned for both kinds.

modules/150-user-authn/template_tests/validation_test.go
Jabejixo: You are right, and a live cluster confirms it: a match condition referring to `variables` is rejected on submission with `undeclared reference to 'variables'`, while one calling `authorizer` is accepted, so that is the exact declaration set. Match conditions now compile on a compiler of their own that holds no variables; putting such an expression back into the template fails the harness with the same error the API server returns.

AlwxSin (approved):

PR #22411, opened 2026-08-23
[deckhouse] Reserve the d8a5e- object name prefix for the ansible module

fl64:
## Description
The `ansible` module creates two objects for every run — a runner Pod and a Secret with its input — in the namespace of the project the run belongs to. Their names start with `d8a5e-`, and this reserves that prefix the way `d8ms-` and `d8a-` are already reserved: nobody except the module's controller creates, updates or deletes an object whose name starts with it.

The new `ValidatingAdmissionPolicy` mirrors the `d8a-` one:

- service accounts of `d8-` namespaces are excluded — the controller runs in `d8-ansible` and creates these objects from there;
- deleting a Pod stays allowed, so the owner of a project can stop a run that is in the way;
- the exclusion for service accounts of `d8a-` namespaces is left out, because no namespace carries the `d8a5e-` prefix.

No behaviour changes for existing objects: nothing in the platform is named `d8a5e-*` today except the objects of this module.

Two reasons.

The module already names its objects this way, and without the policy anyone with write access to the namespace can rewrite or delete the input of a playbook that is currently running — the Secret holds the generated inventory and the playbook itself.

And prefixes need a single place where they are taken, so that two modules do not pick the same one. `d8a` turned out to be taken exactly this way, which is why the module moved to `d8a5e`.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

The policy is a manifest, so there is no code to unit-test; the repository has no
tests over the policies of this module. Nothing user-facing is documented by it —
the prefix belongs to the `ansible` module, whose own documentation describes the
names of its objects. Cluster testing is described in a comment below.

## Changelog entries
```changes
section: deckhouse
type: feature
summary: The `d8a5e-` object name prefix is reserved for the ansible module.
```

### DIFF
--- a/modules/002-deckhouse/templates/validation.yaml
+++ b/modules/002-deckhouse/templates/validation.yaml
@@ -229,6 +229,58 @@ spec:
   - Deny
   - Audit
 
+---
+{{- $policyName := "d8a5e-prefix.deckhouse.io" }}
+apiVersion: {{ include "helm_lib_get_api_version_by_kind" (list . "ValidatingAdmissionPolicy") }}
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: {{ $policyName }}
+  {{- include "helm_lib_module_labels" (list . (dict "app" "deckhouse") ) | nindent 2 }}
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["*"]
+        apiVersions: ["*"]
+        operations:  ["CREATE", "UPDATE", "DELETE"]
+        resources:   ["*"]
+  matchConditions:
+    - name: 'exclude-groups'
+      expression: '!(["system:nodes", "system:serviceaccounts:kube-system", "system:serviceaccounts:d8-system", "system:serviceaccounts:d8-cni-cilium"].exists(e, (e in request.userInfo.groups)))'
+    - name: 'exclude-users'
+      expression: '!(["system:sudouser", "system:apiserver", "system:kube-controller-manager", "system:kube-scheduler", "system:volume-scheduler", "dhctl"].exists(e, (e == request.userInfo.username)))'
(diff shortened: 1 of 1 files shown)


fl64:
Tested on a cluster running this branch's dev build (`dev-registry.deckhouse.io/sys/deckhouse-oss:pr22411`). The policy and its binding arrive as expected, next to `d8a-prefix.deckhouse.io` and `d8ms-prefix.deckhouse.io`.

What the module does over its own objects — all allowed, nothing denied:

- the controller (`system:serviceaccount:d8-ansible:ansible`) creates the runner Pod `d8a5e-<run>-<suffix>` and the Secret `d8a5e-<run>-input`;
- it then repoints the Secret's owner from the AnsibleRun to the Pod — the Secret ended up with `ownerReferences: Pod/d8a5e-owner-check-6mt8t`, so a PATCH over a prefixed object goes through;
- a run reached `PlaybookSucceeded`, after which the Pod and its Secret were collected, leaving no prefixed objects behind;
- the controller's log carries zero `forbidden`/`denied` lines over the run.

What a user cannot do — checked from a service account outside any excluded group, namespace or user (`system:serviceaccount:default:admin-sa`):

- updating the Secret of a run → `Forbidden`, `Creating, updating and deleting objects with the ansible module prefix is forbidden`;
- deleting that Secret → the same;
- creating an object of their own with this prefix (`d8a5e-squatter`) → the same.

What stays allowed on purpose: deleting the runner Pod. `kubectl delete pod d8a5e-...` succeeded and the run moved to `PlaybookFailed`, so the owner of a project can stop a run that is in the way.

Not covered by this testing: the exclusions for `system:sudouser` and for the `kube-system`, `d8-system`, `d8-cni-cilium` groups; the audit annotation; editions other than FE; and the platform e2e suite.

ldmonster (approved):

PR #22444, opened 2026-08-24
[deckhouse] Compile the reserved public hosts into the policies instead of a paramKind

Jabejixo:
## Description

The `reserved-public-hosts-*` ValidatingAdmissionPolicies no longer declare `paramKind: v1/ConfigMap`, and their bindings no longer carry a `paramRef`. The reserved hostnames, the pattern and the allowlist are rendered into `spec.variables` as CEL literals instead.

The `d8-reserved-public-hosts` ConfigMap is still rendered from the same values, so what an operator reads there and what the apiserver enforces cannot drift apart, and it still carries the grandfather record `hooks/snapshot_reserved_public_hosts.go` reads back. What is reserved does not change on any mode or setting.

No user-facing behaviour changes for tenants: the same hostnames are denied on the same six kinds.

**An apiserver already stuck in this state is not released by the manifest change alone and has to be restarted.** `informer-sync` asks the shared informer factory, and the factory only ever adds to `startedInformers` — dropping the last `paramKind` cancels the informer (`admission/plugin/policy/generic/policy_source.go`) but leaves it registered as started with `HasSynced` false, so the check stays red for the life of the process. An upgrade does the restart on its own, because the pod template's image is a per-build digest and the Deployment rolls. Both halves of this were verified in a cluster: correcting the policies under a running `permission-browser-apiserver` left it `0/1` with no change for two minutes, and deleting the pod brought it back `1/1` with the APIService `Available=True`.

A `paramKind` makes the apiserver that runs the policy open a cluster-wide informer on that type. `paramRef` narrows nothing at that stage — it is applied later, when the policy is evaluated. And the `validatingadmissionpolicy` plugin is part of the default `RecommendedOptions` admission chain, so **every aggregated apiserver** runs these policies too, not just kube-apiserver.

`permission-browser-apiserver` in `d8-user-authz` holds no read on `configmaps`, so in a cluster with multitenancy enabled:

- `failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:d8-user-authz:permission-browser-apiserver" cannot list resource "configmaps" in API group "" at the cluster scope`
- `[-]informer-sync failed: 1 informers not started yet: [*v1.ConfigMap]`, so the pod stays `0/1`. It is a readiness check, so nothing restarts the container, and while the policies are in place a restart would not help either — the new process opens the same informer.
- `v1alpha1.authorization.deckhouse.io` goes `Available=False` on missing endpoints
- the namespace controller stops deleting namespaces **cluster-wide**: `NamespaceDeletionDiscoveryFailure=True`, `stale GroupVersion discovery: authorization.deckhouse.io/v1alpha1`
- module installs then time out against namespaces stuck in `Terminating`

None of it is doing any work: the `matchConstraints` of these policies name ingresses, gateways and routes, which `permission-browser-apiserver` does not serve. Granting it `configmaps` would fix the symptom by making that informer succeed — a permanent cluster-wide ConfigMap cache and watch in an apiserver that answers RBAC questions — and would leave the same mine for the next aggregated apiserver. `bashible-apiserver` is unaffected only because it is built on `k8s.io/apiserver` v0.29, where the feature gate is off by default.

The parameters are Helm-rendered and static within a converge, so reading them through the API bought nothing in the first place. `reserved-public-hosts.yaml` was the only `paramKind` in the repository.

It stops namespace garbage collection cluster-wide and stalls converge, on any release that ships both these policies and an aggregated apiserver built on `k8s.io/apiserver` >= 1.30, where the plugin is GA and always on. Cherry-picks cleanly onto `release-1.77` and `release-1.76`; the touched files are identical on all three branches and the module tests pass on each.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: The reserved public hosts policies no longer make every aggregated apiserver in the cluster watch all ConfigMaps, which left permission-browser-apiserver permanently unready and stopped namespace deletion cluster-wide.
```

### DIFF
--- a/modules/002-deckhouse/template_tests/reserved_public_hosts_test.go
+++ b/modules/002-deckhouse/template_tests/reserved_public_hosts_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package template_tests
 
 import (
+	"encoding/json"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -155,6 +156,27 @@ func matchConditionNames(policy object_store.KubeObject) []string {
 		names = append(names, condition.Get("name").String())
 	}
 	return names
+}
+
+// The reserved set reaches the policies as CEL literals rendered by toJson, so a literal parses back
+// as the JSON it was rendered from. Reading it that way rather than by substring keeps the
+// assertions about what is reserved, not about how the template spells it.
+func celVariable(policy object_store.KubeObject, name string, into any) {
+	expression := policy.Field(`spec.variables.#(name=="` + name + `").expression`).String()
+	Expect(json.Unmarshal([]byte(expression), into)).ShouldNot(HaveOccurred(),
+		"variable %s should be a JSON-compatible CEL literal, got %q", name, expression)
+}
+
+func celStringLiteral(policy object_store.KubeObject, name string) string {
+	var value string
+	celVariable(policy, name, &value)
+	return value
+}
+
+func celStringList(policy object_store.KubeObject, name string) []string {
+	values := []string{}
+	celVariable(policy, name, &values)
+	return values
 }
 
 // gatewayValues is the shape both places that name a Gateway take, global.modules.gatewayAPIGateway
@@ -426,7 +448,6 @@ var _ = Describe("Module :: deckhouse :: reserved public hosts ::", func() {
 			vap := f.KubernetesGlobalResource("ValidatingAdmissionPolicy", reservedHostsIngressPolicy)
 			Expect(vap.Exists()).To(BeTrue())
 			Expect(vap.Field("spec.failurePolicy").String()).To(Equal("Fail"))
-			Expect(vap.Field("spec.paramKind.kind").String()).To(Equal("ConfigMap"))
 			Expect(vap.Field("spec.matchConstraints.resourceRules.0.apiGroups").String()).To(MatchJSON(`["networking.k8s.io"]`))
 			Expect(vap.Field("spec.matchConstraints.resourceRules.0.resources").String()).To(MatchJSON(`["ingresses"]`))
 			// DELETE is deliberately absent: an Ingress that predates the policy stays routable
@@ -444,16 +465,29 @@ var _ = Describe("Module :: deckhouse :: reserved public hosts ::", func() {
 				To(ContainSubstring("system:serviceaccount:d8-"))
 		})
 
-		It("binds the policy to the parameters and skips the platform's own namespaces", func() {
+		// The reserved set is compiled into the policy instead of being read through paramKind, so
+		// that no apiserver has to hold a cluster-wide read on the parameter type to enforce it. Any
+		// aggregated apiserver runs the ValidatingAdmissionPolicy plugin out of the default admission
+		// chain, and one that cannot list the type fails its own informer-sync readiness check for
+		// good -- which is how a policy over Ingress hostnames stopped namespace deletion cluster-wide.
+		It("carries the reserved set as literals rather than reading it from the ConfigMap", func() {
+			vap := f.KubernetesGlobalResource("ValidatingAdmissionPolicy", reservedHostsIngressPolicy)
+			cm := configMap()
+
+			Expect(vap.Field("spec.paramKind").Exists()).To(BeFalse())
+			// The same values the ConfigMap publishes, so what an operator reads there and what the
+			// apiserver enforces cannot drift apart.
+			Expect(celStringLiteral(vap, "reservedPattern")).To(Equal(cm.Field("data.hostPattern").String()))
+			Expect(celStringList(vap, "reservedHosts")).To(Equal(strings.Fields(cm.Field("data.hosts").String())))
+			Expect(celStringList(vap, "allowedHosts")).To(Equal(strings.Fields(cm.Field("data.allowedHosts").String())))
+		})
+
+		It("skips the platform's own namespaces and references no parameters", func() {
 			binding := f.KubernetesGlobalResource("ValidatingAdmissionPolicyBinding", reservedHostsIngressPolicy)
 			Expect(binding.Exists()).To(BeTrue())
 			Expect(binding.Field("spec.policyName").String()).To(Equal(reservedHostsIngressPolicy))
 			Expect(binding.Field("spec.validationActions").String()).To(MatchJSON(`["Deny","Audit"]`))
-			Expect(binding.Field("spec.paramRef.name").String()).To(Equal(reservedHostsConfigMapName))
-			Expect(binding.Field("spec.paramRef.namespace").String()).To(Equal("d8-system"))
-			// Without this, losing the ConfigMap would take every Ingress write in the cluster
-			// down with it.
-			Expect(binding.Field("spec.paramRef.parameterNotFoundAction").String()).To(Equal("Allow"))
+			Expect(binding.Field("spec.paramRef").Exists()).To(BeFalse())
 			Expect(binding.Field("spec.matchResources.namespaceSelector.matchExpressions").String()).To(MatchJSON(`[
 				{"key": "heritage", "operator": "NotIn", "values": ["deckhouse"]},
 				{"key": "security.deckhouse.io/reserved-hosts-bypass", "operator": "NotIn", "values": ["true"]}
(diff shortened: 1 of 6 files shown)


modules/002-deckhouse/templates/reserved-public-hosts.yaml
```diff
@@ -389,9 +412,6 @@ metadata:
   {{- include "helm_lib_module_labels" (list $context (dict "app" "deckhouse")) | nindent 2 }}
 spec:
   failurePolicy: Fail
-  paramKind:
```
AlwxSin, line 392: 🟡 **Removing the `paramKind` does not unstick an apiserver process that is already broken**

The changelog says the fix stops `permission-browser-apiserver` being permanently unready. It stops it for a *new* process, not for the running one.

`SharedInformerFactory.WaitForCacheSync` iterates `f.startedInformers` (client-go `informers/factory.go:180-190`), and `startedInformers[t] = true` (`:162`) is never deleted — there is no removal path in the factory. When the last policy loses its `paramKind`, `k8s.io/apiserver/pkg/admission/plugin/policy/generic/policy_source.go:367-372` only calls `info.cancelFunc()` and drops the entry from `paramsCRDControllers`; the ConfigMap informer stops, its `HasSynced` stays false, and the factory still reports it as started. `/readyz` `informer-sync` therefore stays red for the lifetime of the process — so the pod stays `0/1`, the APIService stays unavailable, and namespace deletion stays stuck, with the corrected policies already in the cluster.

In practice a release upgrade papers over this: the pod template's image is `helm_lib_module_image` → a per-build digest (`ee/be/modules/140-user-authz/templates/permission-browser-apiserver/deployment.yaml:78`), so the Deployment rolls and the new pods never create the informer. But that is incidental rather than part of the fix, which makes two things worth doing: state in the PR body that the affected aggregated apiservers have to be restarted and that the upgrade does it, so nobody hand-applies the manifest change into a broken cluster and concludes from the unchanged symptom that the diagnosis was wrong.

modules/002-deckhouse/template_tests/reserved_public_hosts_test.go
```diff
@@ -444,16 +465,29 @@ var _ = Describe("Module :: deckhouse :: reserved public hosts ::", func() {
...
+		It("carries the reserved set as literals rather than reading it from the ConfigMap", func() {
+			vap := f.KubernetesGlobalResource("ValidatingAdmissionPolicy", reservedHostsIngressPolicy)
+			cm := configMap()
+
+			Expect(vap.Field("spec.paramKind").Exists()).To(BeFalse())
+			// The same values the ConfigMap publishes, so what an operator reads there and what the
+			// apiserver enforces cannot drift apart.
+			Expect(celStringLiteral(vap, "reservedPattern")).To(Equal(cm.Field("data.hostPattern").String()))
+			Expect(celStringList(vap, "reservedHosts")).To(Equal(strings.Fields(cm.Field("data.hosts").String())))
+			Expect(celStringList(vap, "allowedHosts")).To(Equal(strings.Fields(cm.Field("data.allowedHosts").String())))
```
AlwxSin: 🔵 **The ConfigMap↔policy equality is asserted only where two of the three values are empty**

This is the test that pins the PR's headline invariant — "what an operator reads there and what the apiserver enforces cannot drift apart" — but it lives in `Context("The reservation follows publicDomainTemplate by default")` (`:338`), which renders `%s.example.com` with no `additionalHosts`, no `excludedServices` and no snapshot. So at the moment the equality is checked, `reservedHosts` is a single wildcard and `allowedHosts` is `[]`. The contexts that populate all three — `:791` (`additionalHosts`), `:880` (`excludedServices`), `:892` (snapshot), `:910` (List) — never re-run it; they assert the ConfigMap keys and, separately, CEL behaviour.

Low risk, because a real drift would have to satisfy both independent sets of assertions, and `expectSameOnEveryKind` in the CEL suite exercises non-empty sets against real cel-go on all six kinds. Cheap to close all the same: hoist these three lines into a helper and call it from those four contexts.

modules/002-deckhouse/templates/reserved-public-hosts.yaml
```diff
@@ -416,22 +436,15 @@ spec:
...
-      # (parameterNotFoundAction: Allow); the guards are for a ConfigMap that lost a key.
-      expression: |
-        params != null && has(params.data) && 'hosts' in params.data
-          ? params.data['hosts'].split('\n').filter(h, h != '')
-          : []
+      # Two layers of quoting, and both are load-bearing. toJson renders the CEL literal, so a
+      # hostname is data and never source -- a backslash in the pattern survives as one, and nothing
+      # an operator writes in additionalHosts can be read as an expression. quote then wraps that
+      # literal for YAML, without which the list would parse as a YAML sequence instead of a string.
+      expression: {{ $params.hosts | toJson | quote }}
```
AlwxSin: 🔵 **The reserved set now lives in objects every aggregated apiserver watches and compiles**

Before, the data sat in one ConfigMap that only kube-apiserver read. Now it is inlined into up to six ValidatingAdmissionPolicies, each of which every apiserver in the cluster watches and CEL-compiles — which is the same blast radius the change is trying to shrink, just with a much smaller payload.

`allowedHosts` is fed by `deckhouse.internal.reservedPublicHosts.hosts`, the grandfather snapshot, and that has no `maxItems` in `modules/002-deckhouse/openapi/values.yaml:36-51`. In a large multi-tenant cluster upgraded into Template mode the record is whatever tenants happened to be serving under the domain, multiplied by six. Bounded enough in practice; worth a `maxItems` on the snapshot, or a sentence saying why none is needed.

modules/002-deckhouse/templates/reserved-public-hosts.yaml
```diff
@@ -370,7 +393,7 @@ data:
 
 {{- define "reserved_public_hosts_policy" }}
   {{- $context      := index . 0 }}
-  {{- $paramsName   := index . 1 }}
+  {{- $params       := index . 1 }}
```
AlwxSin: 🔵 **Naming leftovers from the thing that was removed**

`$params` and `$policyParams` (`:344`) still say "params" in a file whose entire point is that there are no params any more, and `$paramsName` (`:94`, outside this diff) now only names the ConfigMap.

There is a small real cost, not just aesthetics: the snapshot test's regex had to be loosened from `\$paramsName` to `\$\w+` to keep matching the include call (`modules/002-deckhouse/hooks/snapshot_reserved_public_hosts_test.go:38`), so it now matches any variable in that position. Renaming to `$reserved` and `$configMapName` lets that regex stay exact.

modules/002-deckhouse/templates/reserved-public-hosts.yaml
Jabejixo: Confirmed in a cluster: correcting the policies under a running `permission-browser-apiserver` left it `0/1` unchanged for two minutes, same process (`restarts=0`), and only deleting the pod brought it back `1/1`. Added to the PR description that the affected apiservers have to be restarted and that an upgrade does it via the per-build image digest.

modules/002-deckhouse/template_tests/reserved_public_hosts_test.go
Jabejixo: Fixed: hoisted into `expectPolicyCarriesConfigMap`, called from the `additionalHosts`, `excludedServices`, snapshot and `mode: List` contexts, the last being the only render with an empty pattern. Mutating the template to drop `allowedHosts` from the policy fails only the two new specs whose contexts populate it, so the gap was real rather than theoretical.

modules/002-deckhouse/templates/reserved-public-hosts.yaml
Jabejixo: Added the bound: `maxItems: 1000` on the snapshot, with the hook truncating to the same number before it writes, so a record over the cap un-grandfathers its tail with a warning instead of failing values validation and stopping the module from converging. A test pins the constant to the schema so the two cannot drift.

modules/002-deckhouse/templates/reserved-public-hosts.yaml
Jabejixo: Renamed to `$reserved` and `$configMapName`, and tightened the snapshot regex back to `\$reserved` so it pins the include call exactly again.

AlwxSin (approved):

z9r5 (approved):

ldmonster (approved):

deckhouse-BOaTswain:
Cherry pick PR [22462](https://github.com/deckhouse/deckhouse/pull/22462) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [22463](https://github.com/deckhouse/deckhouse/pull/22463) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #22456, opened 2026-08-24
Fencing 2 a

W1seKappa:
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


z9r5 (approved):

ergoz:
<p>I ran a micro-benchmark of the domain layer of this branch to see how the new per-second reconcile behaves on large NodeGroups. The design itself holds up well — the idle cost is negligible up to a few thousand nodes, and <code>WriterRank</code> is allocation-free. Two things are worth addressing before this is exercised on groups of thousands.</p>
<p><strong>Environment:</strong> Go 1.27.1, darwin/arm64 (Apple M1 Pro). Server-class CPUs are typically 1.5–2x slower, so scale accordingly. "Idle pass" is the domain part of <code>Writer.reconcile()</code> with nothing failed: <code>NewView</code> + <code>SeenAlive.Observe</code> + <code>SeenAlive.Retain</code> + <code>HasQuorum</code> + <code>View.Failed</code>. Names are synthetic (<code>worker-N</code>), which is the hard case for the hash, as <code>TestDesignatedWriterSpreadsTheLoad</code> already notes.</p>
<pre><code>idle pass (everything alive, nothing failed)
  1 000 nodes      340 µs      218 KB/pass    12 allocs
  2 000 nodes      599 µs      437 KB/pass    20 allocs
  5 000 nodes    1 575 µs      890 KB/pass    36 allocs
 10 000 nodes    3 457 µs    1 781 KB/pass    68 allocs

WriterRank, one failed peer, 0 allocs
  1 000 alive       35 µs
  5 000 alive      132 µs

mass failure at the quorum edge (rank computed for every failed peer)
  1 000 alive /   499 failed     13 ms
  5 000 alive / 2 499 failed    350 ms
</code></pre>
<h3 data-heading="1. &#x60;SetTransform&#x60; on the Node informer (pre-existing, not introduced here)">1. <code>SetTransform</code> on the Node informer (pre-existing, not introduced here)</h3>
<p><strong>What:</strong> pass <code>informers.WithTransform(...)</code> in <code>NewNodeWatcher</code> (<code>internal/adapters/kubeclient/nodewatch.go</code>) and trim each <code>Node</code> down to the three fields the agent actually consumes.</p>
<p><strong>Why:</strong> the factory currently caches full <code>Node</code> objects while the agent only ever reads <code>{Name, InternalIP, UID}</code> into <code>domain.Peer</code>. A real <code>Node</code> carries <code>managedFields</code>, <code>status.images</code>, conditions and allocatable — commonly 5–20 KB per object. At 5 000 nodes that is an estimated 50–150 MB resident in a pod whose VPA <code>maxAllowed</code> is <code>memory: 70Mi</code> (<code>templates/fencing-agent/daemonset.yaml</code>), i.e. OOMKill long before CPU matters. A trimmed object is a few hundred bytes, so the same cache lands in single-digit MB. I did not measure the resident size on a live cluster — the estimate is per-object size times group size, and it is worth confirming with pprof.</p>
<p>This predates the PR (only the <code>UID</code> line changed here), but the new writer calls <code>Expected.Snapshot()</code> every second, which makes group size a first-class scaling axis for the agent, so it belongs in the same conversation.</p>
<pre><code class="language-go">informers.WithTransform(func(obj any) (any, error) {
	node, ok := obj.(*corev1.Node)
	if !ok {
		return obj, nil // tombstones and already-trimmed objects pass through
	}

	return &#x26;corev1.Node{
		ObjectMeta: metav1.ObjectMeta{Name: node.Name, UID: node.UID, Labels: node.Labels},
		Status:     corev1.NodeStatus{Addresses: node.Status.Addresses},
	}, nil
}),
</code></pre>
<p>The transform must stay idempotent, since it is also applied on resync updates of objects already in the store — building a fresh object from the retained fields satisfies that.</p>
<h3 data-heading="2. Derive &#x60;View&#x60; and the candidate order only when the inputs change">2. Derive <code>View</code> and the candidate order only when the inputs change</h3>
<p><strong>What:</strong> cache the <code>View</code> (and the per-incident candidate ordering) against a version of its inputs — e.g. a generation counter bumped in <code>Membership.Upsert</code>/<code>Delete</code> plus the alive set — and rebuild only on change.</p>
<p><strong>Why, part one — idle waste.</strong> Every pass rebuilds <code>map[string]Peer</code> over the whole group in <code>NewView</code>, plus a second full map in <code>SeenAlive.Retain</code>, plus two sorts. That is the 890 KB/pass at 5 000 nodes, ~53 MB/min of garbage, for inputs that change maybe hourly. Not fatal (0.16 % of a core), but it is pure churn in a DaemonSet that runs on every node.</p>
<p><strong>Why, part two — the quadratic peak.</strong> <code>WriterRank</code> is O(alive) per failed peer, so a pass is <strong>O(failed × alive)</strong> ≈ n²/2 at the quorum edge. The measurements confirm it exactly: 5x the group is 26x the time. Per-pair cost is ~26 ns, so:</p>
<pre><code>per-pass cost ≈ failed × alive × 26 ns
</code></pre>
<p>At 5 000 nodes:</p>

failure scope | per pass | share of one core at 1 s tick
-- | -- | --
rack, 100 nodes | 13 ms | 1.3 %
zone, 250 nodes | 33 ms | 3.3 %
1 000 nodes | 132 ms | 13 %
2 499 nodes (quorum edge) | 350 ms | 35 %


<p>350 ms of CPU every second is ~35x the VPA <code>maxAllowed cpu: 20m</code>, and it lands exactly when the node is already under stress and the fencing path is time-critical. There are no CPU limits, so this will not throttle hard, but under CFS contention the reconcile stretches out. The same peak appears on <strong>mass recovery</strong>: once the peers are back in gossip, <code>WriterRank</code> is called from <code>clear()</code> for every record. (During the failure itself <code>clear()</code> returns in the <code>default</code> branch before the call, so there is no double count.)</p>
<p><strong>Why caching rather than a cheaper election.</strong> The obvious O(1) alternative — <code>sortedAlive[hash(failedNode) % len(alive)]</code> — would break the property <code>TestDesignatedWriterSurvivesLosingOtherPeers</code> protects: an unrelated node leaving would hand the incident to a different agent. Memoizing the rank per <code>(aliveVersion, failedNode)</code> keeps that property intact and, because the alive set stabilizes within a pass or two of a mass failure, turns the cost into "pay once per membership change" instead of "pay every second". From the second pass onward the hashing drops to zero.</p>
<p><strong>Optional micro-optimization,</strong> if you want the first pass cheaper too: hash <code>failedNode + sep + candidate</code> instead of <code>candidate + sep + failedNode</code>, so the FNV state after the now-common prefix can be computed once per incident and reused for all candidates — roughly halves the hashing work. Note it changes which agent gets elected (not correctness, and the separator and the avalanche finalizer are still needed for the same reasons documented in <code>writer.go</code>).</p>
<details> <summary>Benchmark used</summary>
<p>Dropped into <code>internal/domain/</code> on this branch; <code>benchIdle</code> mirrors the domain part of one reconcile pass, <code>benchMassFailure</code> calls <code>WriterRank</code> for <code>n/2-1</code> failed peers against <code>n</code> alive.</p>
<pre><code class="language-go">func benchIdle(b *testing.B, n int) {
	expected := benchPeers(n)   // Peer{Name: worker-i, IP: 10.0.x.y, UID: &#x3C;uuid>}
	members := benchNames(n)
	var seen SeenAlive
	seen.Observe(members)

	b.ReportAllocs()
	b.ResetTimer()

	for range b.N {
		view := NewView(expected, members)
		seen.Observe(view.Alive())
		seen.Retain(expected)
		_ = view.HasQuorum()
		_ = view.Failed(&#x26;seen)
	}
}

func benchMassFailure(b *testing.B, n int) {
	alive := benchNames(n)
	failed := make([]string, 0, n/2-1)
	for i := range n/2 - 1 {
		failed = append(failed, fmt.Sprintf("dead-%d", i))
	}

	b.ReportAllocs()
	b.ResetTimer()

	for range b.N {
		for _, name := range failed {
			_ = WriterRank(alive, name, "worker-7")
		}
	}
}
</code></pre>
</details>

PR #22470, opened 2026-08-24
[user-authz] Add RBACv1 core missing rules

magna-z:
## Description
Adds RBAC v1 rules for 3 previously-uncovered resources/subresources in
`modules/140-user-authz/templates/cluster-roles.yaml` (down from 7 in the
original scope — 4 were pulled after review, see below).

- `pods/resize`: `get,patch` — Editor+
- `serviceaccounts/token`: `create` — Admin+
- `namespaces/finalize`: `update` — ClusterAdmin

Removed from this PR after review, tracked separately:
- `pods/ephemeralcontainers` — gap in admission-policy-engine's `deny-exec-heritage`
  webhook (doesn't cover this subresource in `d8-*`/`kube-system`); needs a fix on
  that module's side, out of scope here.
- `pods/proxy`, `services/proxy` — routes around the explicit `portForwarding`
  opt-in pattern used elsewhere in this module; needs a deliberate design
  decision, not a silent default grant.
- `pods/binding` — bypasses the scheduler entirely (taints, affinity, resource
  fit are enforced by kube-scheduler only); upstream `edit`/`admin` don't grant
  this either.

Also in the diff, unrelated to the permission change:
- `cluster-role-port-forwad.yaml` renamed to `cluster-role-port-forward.yaml` (typo, not referenced by name).
- YAML fixture constants in `template_tests/module_test.go` re-indented — semantically identical.
- Doc-only catch-up: 7 `nodes/*` entries added to the ClusterAdmin listing in `RBAC_CURRENT*.md` / `ROLE_MODEL*.md`; the templates already granted them.

Part of an RBAC v1 resource-coverage audit: users at standard access levels had
no RBAC path to features tied to these subresources, even though their
`accessLevel` should logically cover them — "no rights = no feature" for
otherwise-supported platform functionality. Several other candidate resources
turned out to intersect with a different module's admission controls or bypass
a safety mechanism (scheduler, opt-in gating); those were deliberately excluded
rather than shipped with a workaround, and are tracked as separate follow-ups.

Not necessarily.

## Checklist
- [x] The code is covered by unit tests. (`template_tests`: `namespaces/finalize` present on `user-authz:cluster-admin`, absent on `user-authz:cluster-editor`)
- [ ] e2e tests passed.

## Changelog entries
```changes
section: user-authz
type: fix
summary: "Added missing RBAC v1 rules for pods/resize (Editor+), serviceaccounts/token (Admin+) and namespaces/finalize (ClusterAdmin)."
impact: |
  Permission expansion for existing bindings.

  - Users already bound to Editor (and above) will gain in-place pod resource resize (pods/resize: get,patch), which changes container requests/limits on running pods without recreating them.
  - Users already bound to Admin (and above) will gain the ability to mint ServiceAccount tokens (serviceaccounts/token: create) in their namespaces."
impact_level: high
```


z9r5 (approved):

deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:fix/rbac-v1-core-missing-rules` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/32823010208) in 26m43s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.3.169`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 25m26s.

</details>

modules/140-user-authz/templates/cluster-roles.yaml
```diff
@@ -256,6 +258,13 @@
   - pods/eviction
   verbs:
   - create
+- apiGroups:
+  - ""
+  resources:
+  - pods/ephemeralcontainers
```
AlwxSin: 🔴 **1. `pods/ephemeralcontainers` at PrivilegedUser is not covered by admission control in system namespaces**

An ephemeral container is a shell in an existing pod — `kubectl debug --target` shares the process namespace, so the target container's `/var/run/secrets/...` is reachable via `/proc/1/root`. In `d8-*` / `kube-*` namespaces nothing validates it:

- `admission-policy-engine.deckhouse.io` has `namespaceSelector: heritage NotIn [deckhouse]`, and every `d8-*` namespace plus `kube-system` carries `heritage: deckhouse` — SecurityPolicy never sees those namespaces.
- `deny-exec-heritage.admission-policy-engine.deckhouse.io` is the only webhook that does cover them, and its rules are `pods/exec|pods/attach: CONNECT` only.
- the `D8DenyExecHeritage` constraint matches `kinds: ["PodExecOptions","PodAttachOptions"]`, and its rego fires only on `subresource_is("exec"|"attach")`.

A `ClusterAuthorizationRule` always renders a ClusterRoleBinding (`templates/cluster-role-bindings.yaml:41`), and `limitNamespaces` / `allowAccessToSystemNamespaces` require `enableMultiTenancy: true` — off by default, the render hard-fails otherwise. So with the default configuration this grant is cluster-wide, system namespaces included.

To be fair on scope: `user_authz_privileged_user_rules` already grants `secrets: get,list,watch` under that same default, so this is not a new path to cluster-admin. The point is narrower — the grant makes `deny-exec-heritage` advisory. The control written to keep these access levels out of Deckhouse system pods no longer covers the equivalent primitive.

Closing it on the APE side is three coordinated changes, not one:

1. `pods/ephemeralcontainers: UPDATE` in the `deny-exec-heritage` webhook rules;
2. `Pod` in `D8DenyExecHeritage.spec.match.kinds`;
3. the rego — `is_connect` accepts only `CONNECT` / `CREATE` / `""`, so an `UPDATE` review is silently allowed, and `is_pod_exec_or_attach` needs an `ephemeralcontainers` clause. Mind the `connect_to_pod_exec_or_attach` clause that matches an *empty* subresource when widening `is_connect`.

Step 2 does not over-match: the constraint is scoped to `namespaces: ["d8-*","kube-*"]` and the main webhook excludes exactly those, so no plain Pod `CREATE` can reach it.

Alternative, and probably the better trade for this PR: keep the grant out of `privileged_user_rules` and put it behind an explicit opt-in, the way `spec.portForwarding` and `spec.allowScale` already work. Upstream `edit` and `admin` do not grant `pods/ephemeralcontainers` either.

modules/140-user-authz/templates/cluster-roles.yaml
```diff
@@ -265,6 +274,24 @@
 {{- end -}}
 
 {{- define "user_authz_editor_rules" }}
+- apiGroups:
+  - ""
+  resources:
+  - pods/bindibg
```
AlwxSin: 🔴 **2. Typo in the resource name: `pods/bindibg`**

Should be `pods/binding`. RBAC does not validate resource names, so the rule renders and applies cleanly while granting nothing — one of the seven additions in this PR is a no-op.

It propagated into all four doc copies: `modules/140-user-authz/docs/README.md:363`, `modules/140-user-authz/docs/README_RU.md:369`, `docs/documentation/pages/admin/configuration/access/authorization/RBAC_CURRENT.md:126`, `RBAC_CURRENT_RU.md:122`.

modules/140-user-authz/templates/cluster-roles.yaml
```diff
@@ -265,6 +274,24 @@
...
+  verbs:
+  - create
```
AlwxSin: 🟡 **5. `pods/binding` at Editor bypasses the scheduler** — currently inert only because of the typo above

`POST .../pods/{name}/binding` assigns a pod to a node directly. Taints, `nodeSelector`, affinity and topology spread are all scheduler-side, so none of them apply — `node-role.kubernetes.io/control-plane:NoSchedule` does not stop a direct binding.

Editor can create Deployments (`user_authz_common_rules` with `$mode = rw`), the controller produces Pending pods, and Editor can then place them on any node, control-plane included. Upstream `edit` / `admin` do not grant `pods/binding` — they have `bindings: get,list,watch` only.

Worth settling before the typo on the line above is fixed, since that is the only thing keeping the rule inert today.

modules/140-user-authz/templates/cluster-roles.yaml
```diff
@@ -265,6 +274,24 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - pods/resize
+  verbs:
+  - update
```
AlwxSin: 🔴 **3. Granted verbs do not match what the API serves — `pods/resize` and `pods/ephemeralcontainers` stay unusable**

`kubectl get --raw /api/v1` on 1.34:

```
pods/ephemeralcontainers   get,patch,update
pods/resize                get,patch,update
```

Two distinct problems:

- `pods/resize` is granted `update`. `update` *is* served — it just isn't what the client uses: `kubectl patch pod … --subresource resize` is a `PATCH`. Without `patch` (and `get` to read the result back) in-place resize does not work from kubectl.
- `pods/ephemeralcontainers` (line 266) is granted `create,update`, but `create` is not a verb on that subresource at all, so half of that rule is dead regardless of client. `kubectl debug` issues a strategic-merge `PATCH`, so `update` alone does not authorize it either.

Net effect: the two features named in the PR description — in-place pod resize and ephemeral debug containers — remain unavailable at every access level after this change.

modules/140-user-authz/templates/cluster-roles.yaml
```diff
@@ -265,6 +274,24 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts/token
+  verbs:
+  {{- include "user_authz_verbs" "rw" }}
```
AlwxSin: 🟡 **4. `serviceaccounts/token` granted with the full `rw` verb set**

`{{- include "user_authz_verbs" "rw" }}` expands to `get,list,watch,create,delete,deletecollection,patch,update`. The subresource serves only `create` — seven of the eight verbs do not exist. Upstream `edit` / `admin` grant it as `serviceaccounts/token: create`, which is the right shape here too.

Beyond the dead verbs, the `rw` alias makes the generated role reference advertise read-write on a token-minting endpoint (`modules/140-user-authz/docs/README.md`, Editor section), which reads considerably worse than what is actually granted.

On the grant itself: with the default `enableMultiTenancy: false`, a CAR with `accessLevel: Editor` puts this cluster-wide — minting a token for any ServiceAccount in any namespace. Editor already holds cluster-wide `secrets` rw (`r` from `privileged_user_rules`, `w` from `editor_rules`), so a legacy SA-token secret gets there anyway — but that is an argument for scoping Editor, not for landing this without a word in the changelog.

modules/140-user-authz/templates/cluster-roles.yaml
```diff
@@ -247,6 +247,8 @@
   resources:
   - pods/attach
   - pods/exec
+  - pods/proxy
```
AlwxSin: 🟡 **6. `pods/proxy` routes around the explicit `spec.portForwarding` opt-in**

`pods/portforward` is deliberately not part of any access level: it lives in a separate `user-authz:port-forward` ClusterRole bound only when `spec.portForwarding: true` (`templates/cluster-role-port-forwad.yaml`, `templates/cluster-role-bindings.yaml:62`). `GET /api/v1/namespaces/{ns}/pods/{name}:{port}/proxy/{path}` reaches an arbitrary port inside an arbitrary pod over HTTP — the same capability minus the protocol generality — and this grants it to PrivilegedUser unconditionally.

`services/proxy` additionally turns the apiserver into an SSRF hop that ignores NetworkPolicy, and under the default non-multitenancy configuration both reach `d8-*`, where nothing validates them (see the `pods/ephemeralcontainers` comment).

The grant is not wrong in itself — upstream `edit` has both, with the full `create,delete,get,patch,update` set, so `create,get` here is also incomplete for anything other than GET. The objection is that it silently sidesteps a flag the platform makes users set explicitly.

docs/documentation/pages/admin/configuration/access/authorization/RBAC_CURRENT.md
```diff
@@ -169,8 +178,17 @@ write:
...
 ```text
+update:
+    - namespaces/finalize
 read-write:
     - deckhouse.io/clusterauthorizationrules
+    - nodes/log
+    - nodes/metrics
+    - nodes/proxy
+    - nodes/stats
+    - nodes/configz
```
AlwxSin: 🔵 **8. Hand-synced doc block does not match generator output ordering**

These seven `nodes/*` entries are a legitimate catch-up: `modules/140-user-authz/docs/README.md:540-546` — the file the generator actually writes — already lists all of them under ClusterAdmin `read-write`. But the ordering here differs from the generated original: `nodes/configz` comes after `nodes/stats`, and the `update:` group above was inserted before `read-write:` rather than after it.

Worth noting that `RBAC_CURRENT*.md` carries the `<!-- start user-authz roles placeholder -->` markers while the generator only writes `modules/140-user-authz/docs/README.md` and `README_RU.md` (`tools/helm_generate/runners/authz_generate_rules_for_roles/authz-generate.go:227-228`), so these copies are maintained by hand and will keep drifting.

docs/documentation/pages/admin/configuration/access/authorization/RBAC_CURRENT.md
```diff
@@ -169,8 +178,17 @@ write:
 {{site.data.i18n.common.role[page.lang] | capitalize }} `ClusterAdmin` ({{site.data.i18n.common.includes_rules_from[page.lang]}} `User`, `PrivilegedUser`, `Editor`, `Admin`, `ClusterEditor`):
 
 ```text
+update:
+    - namespaces/finalize
```
AlwxSin: 🔵 **10. The same role listing in `ROLE_MODEL*.md` was not updated**

`docs/site/pages/virtualization-platform/documentation/admin/platform-management/access-control/ROLE_MODEL.md:199-324` and `ROLE_MODEL_RU.md:206-331` carry the same `<!-- start user-authz roles placeholder -->` block and still show the pre-PR content — PrivilegedUser as `create,get: pods/attach, pods/exec` / `delete,deletecollection: pods` / `read: secrets`.

After this PR there are four hand-maintained copies of the role reference in three different states.

modules/140-user-authz/template_tests/module_test.go
```diff
@@ -662,6 +662,24 @@ var _ = Describe("Module :: user-authz :: helm template ::", func() {
 		})
 	})
 
+	Context("namespaces/finalize access", func() {
```
AlwxSin, line 665: 🟡 **9. Test coverage lands on the one addition that carries no risk**

Both new `It`s cover `namespaces/finalize` — the only addition in this PR that is neither a permission expansion for existing bindings nor security-relevant. The other six get nothing.

An assertion on the rendered resource *names* for the PrivilegedUser and Editor roles would have caught the `pods/bindibg` typo. That is the same coverage gap that let the `SecurityPolicyException` grant sit unnoticed until it was moved to ClusterAdmin.

Minor, same place: neither new `It` asserts `f.RenderError`, unlike the surrounding contexts.

modules/140-user-authz/template_tests/module_test.go
```diff
@@ -35,7 +35,7 @@ func Test(t *testing.T) {
 const (
 	customClusterRolesFlat = `---
 admin:
-  - cert-manager:user-authz:user
+- cert-manager:user-authz:user
```
AlwxSin, line 38: 🔵 **12. Unrelated churn in the diff**

43 of the 68 changed lines in this file are YAML re-indentation of existing fixture constants — semantically identical and unrelated to the change. The `cluster-role-port-forwad.yaml` → `cluster-role-port-forward.yaml` rename is safe (nothing references it by name) but likewise unrelated.

Both are fine on their own; inside an RBAC permission change they make the security-relevant part harder to find.

AlwxSin (commented):
Two findings with no line to attach to, both about the PR description itself. Everything else is inline.

🟡 **7. `impact_level: default` understates the change**

Six of the seven additions move a resource from SuperAdmin-only down to PrivilegedUser / Editor; the seventh (`namespaces/finalize`) goes from SuperAdmin to ClusterAdmin. Existing `ClusterAuthorizationRule`s pick up `pods/proxy`, `services/proxy` and `pods/ephemeralcontainers` on upgrade — no flag, no migration, no opt-in. The `impact` text already says it out loud ("a real permission expansion, not just a bugfix in behavior"), which is `impact_level: high`, not `default`.

🔵 **11. The description contradicts the code**

The description lists `pods/resize`: `create`; the template grants `update`; the API serves `get,patch,update` — three different answers for one rule. `pods/binding` is listed as granted at Editor+ while the rule is typo'd (see inline). Worth fixing the description together with the code, since the changelog is generated from that block.

modules/140-user-authz/template_tests/module_test.go
magna-z: The re-indentation is a side effect of gofmt running after I added the new test — the fixture block was misaligned before this PR and gofmt reformatted the whole block, not just my addition.
Same for the file rename: `cluster-role-port-forwad.yaml` had a typo (`forwad` → `forward`), fixed it while I was in the area since nothing references it by name (confirmed — no matches for the old filename anywhere in the module).

Both are drive-by fixes, not part of the functional change; happy to split them into a separate commit if that's preferred for review clarity.

modules/140-user-authz/template_tests/module_test.go
magna-z: Fair point on scope — content-level tests for the static `cluster-roles.yaml` rules (resource/verb per role) aren't really a thing in this suite today, and adding that kind of coverage across all core resources wasn't the goal of this PR.
The two `It`s on `namespaces/finalize` specifically are there to guard against a privilege leak — to make sure that rule can't accidentally end up available to anything below `cluster-admin` in a future refactor, not to establish general coverage for the other additions.

modules/140-user-authz/templates/cluster-roles.yaml
magna-z: Agreed this deserves more thought than fits in this PR's scope — pulling both out for now. Will look for a more deliberate solution separately (likely the same opt-in-flag pattern as `allowScale`/`portForwarding`, but want to weigh it against upstream's posture before committing to that shape).

modules/140-user-authz/templates/cluster-roles.yaml
magna-z: Agreed, pulling this one out of the PR as well — the gap in `deny-exec-heritage` is real and deserves a proper fix on the admission-policy-engine side rather than a workaround here. Will track it separately and coordinate with that module's owners.

modules/140-user-authz/templates/cluster-roles.yaml
magna-z: Agreed, this bypasses taint/affinity enforcement entirely and upstream doesn't grant it to `edit`/`admin` for exactly that reason. Will remove it from this PR and look at whether it's needed at all, or belongs behind an explicit opt-in like the others.

modules/140-user-authz/templates/cluster-roles.yaml
magna-z: Fixed — down to `create` only, matching upstream `edit`/`admin`.

docs/documentation/pages/admin/configuration/access/authorization/RBAC_CURRENT.md
magna-z: Fixed manually — reordered to match `README.md`'s generated output (`nodes/configz` after `nodes/stats`, `update:` after `read-write:`).
Also ran `make generate` locally to confirm it produces no diff here — the generator only writes `README.md`/`README_RU.md`, so `RBAC_CURRENT*.md` stays a manually-synced copy and will need the same care going forward.

docs/documentation/pages/admin/configuration/access/authorization/RBAC_CURRENT.md
magna-z: Fixed — updated `ROLE_MODEL.md` and `ROLE_MODEL_RU.md` to match the current state of the other three copies.

modules/140-user-authz/templates/cluster-roles.yaml
magna-z: Fixed

modules/140-user-authz/templates/cluster-roles.yaml
magna-z: `pods/binding` removed from this PR

z9r5 (approved):

AlwxSin (approved):

PR #22483, opened 2026-08-25
[user-authn] Honor administrator lock without passwordPolicy.lockout

Jabejixo:
## Description

`UserOperation` / `d8 iam user lock` now honors `LockedUntil` on the password login form, on `grant_type=password` (ROPC / basic-auth-proxy), on refresh-token requests, and on token exchange. `passwordPolicy.lockout` stays only for automatic lockout after a configured number of failed attempts.

For a non-local connector (LDAP, Crowd) the login-form path reads `OfflineSessions` on every attempt so an administrator lock is visible even when no password policy is configured (one extra storage GET on that path).

Already-issued LDAP refresh tokens stay in storage for their TTL: lock sets `lockedUntil` and refresh now rejects while it is active; it does not delete the token. After the lock expires, that refresh can work again.

Administrator lock did nothing unless `passwordPolicy.lockout` was set. The default module config leaves that section unset, so a documented lock was a no-op. The first cut only blocked the login form; `grant_type=password` (used by basic-auth-proxy for LDAP with `enableBasicAuth`) and refresh still issued tokens.

The lock is already shipped and documented. Without this, `d8 iam user lock` does not stop a new login or basic-auth on supported releases.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Administrator lock now blocks login, password grant, and refresh without passwordPolicy.lockout.
impact_level: default
```


modules/150-user-authn/images/dex/patches/018-admin-lock-without-lockout-policy.patch
```diff
@@ -0,0 +1,267 @@
...
++		if localConnector {
+ 			p, err = s.storage.GetPassword(ctx, username)
+ 			if err != nil {
+ 				if err != storage.ErrNotFound {
+@@ -457,7 +459,7 @@ func (s *Server) handlePasswordLogin(w http.ResponseWriter, r *http.Request) {
+ 				// User not found - will be handled as invalid credentials by pwConn.Login
+ 			} else {
+ 				passwordFound = true
+-				if s.passwordPolicy.IsPasswordLocked(p) {
++				if passwordHasActiveLock(p.LockedUntil) {
```
AlwxSin: 🔴 **`grant_type=password` still bypasses the administrator lock**

This patch fixes `handlePasswordLogin`. The ROPC handler `handlePasswordGrant` (upstream `server/handlers.go:1119`) checks neither `LockedUntil` nor the failed-attempt lockout — patch 015 only added log lines there, and 018 does not touch it.

Dex enables `grant_type=password` whenever `oauth2.passwordConnector` is set, and `modules/150-user-authn/templates/dex/config.yaml:88-102` sets it for any LDAP provider with `enableBasicAuth: true`. On such a cluster the reported symptom survives this fix: the correct password of a locked account still mints tokens, over `/token` instead of the login form.

The PR description states the lock "now blocks login whenever `LockedUntil` is in the future", which is true only for the login form.

modules/150-user-authn/images/dex/patches/018-admin-lock-without-lockout-policy.patch
```diff
@@ -0,0 +1,267 @@
...
+ 		// We need identity.UserID to index OfflineSessions, which is supplied
+ 		// by the LDAP / Crowd connectors as a partial Identity even on
+ 		// validPass=false (see their Login implementations).
+-		if nonLocalLockoutEnabled && identity.UserID != "" {
++		if !localConnector && identity.UserID != "" {
+ 			session, sErr := s.storage.GetOfflineSessions(ctx, identity.UserID, authReq.ConnectorID)
+ 			switch {
+ 			case sErr == nil:
+-				if s.passwordPolicy.IsOfflineSessionLocked(session) {
++				if passwordHasActiveLock(session.LockedUntil) {
```
AlwxSin: 🟡 **Locking a non-local session leaves its refresh token usable**

For a local user `executeLock` calls `invalidateLocalUserSessions` and deletes the RefreshToken / OfflineSessions objects. For a targeted session it returns through `lockOfflineSession` *before* that call — `modules/150-user-authn/images/user-authn-controller/src/internal/controller/useroperation/execute.go:74-76` vs `:100` — so nothing is revoked.

And no refresh path reads `LockedUntil`: not here, and not in #22484, which adds allow-list checks only. A locked LDAP account keeps exchanging its refresh token for fresh ID tokens until that token expires on its own.

This is item 4 of the reporter's fix list ("check lock state on every refresh / token exchange"), and it stays open after both PRs.

modules/150-user-authn/images/dex/patches/018-admin-lock-without-lockout-policy.patch
```diff
@@ -0,0 +1,267 @@
...
++		// Honor LockedUntil on OfflineSessions for non-local connectors even
++		// when passwordPolicy.lockout is off (administrator lock via
++		// UserOperation). Failed-attempt lockout still requires
++		// nonLocalLockoutEnabled and is applied after a failed Login.
+ 		//
+ 		// We need identity.UserID to index OfflineSessions, which is supplied
+ 		// by the LDAP / Crowd connectors as a partial Identity even on
+ 		// validPass=false (see their Login implementations).
+-		if nonLocalLockoutEnabled && identity.UserID != "" {
++		if !localConnector && identity.UserID != "" {
```
AlwxSin, line 43: 🟡 **Offline-session read now runs on every non-local login**

The gate changed from "lockout configured for this connector type" (`nonLocalLockoutEnabled`) to "any non-local connector", so `GetOfflineSessions` is now issued for every LDAP / Crowd login attempt, including on clusters that never configured a password policy.

Correct for the lock semantics, but it is an unadvertised load change on the authentication hot path — worth a line in the description at minimum.

(`GetPassword` on the local path is unconditional now too, but patch 011 already fetched it later in the same handler, so that one nets out.)

modules/150-user-authn/images/dex/patches/018-admin-lock-without-lockout-policy.patch
```diff
@@ -0,0 +1,267 @@
...
+-// the failed-attempt state is stored in OfflineSessions (LDAP, Crowd, ...).
++// currently within an active lock window. Used for connector types where
++// lock state is stored in OfflineSessions (LDAP, Crowd, ...), including
++// an administrator lock that does not require passwordPolicy.lockout.
+ func (pp PasswordPolicy) IsOfflineSessionLocked(o storage.OfflineSessions) bool {
+-	return pp.lockout != nil && o.LockedUntil != nil && o.LockedUntil.After(time.Now())
++	return passwordHasActiveLock(o.LockedUntil)
++}
++
++func passwordHasActiveLock(until *time.Time) bool {
```
AlwxSin: 🔵 **Naming, and receivers that are now unused**

`passwordHasActiveLock` is named for `Password` but is also the `OfflineSessions` check — `lockedUntilIsActive` would read straighter.

`IsPasswordLocked` and `IsOfflineSessionLocked` no longer touch their `pp` receiver, which some linters flag.

Moving the check to a free function rather than a method on a possibly-nil `*PasswordPolicy` is the right call, though: the nil-guard the old code depended on is gone by construction.

AlwxSin (approved):

docs/documentation/pages/admin/configuration/access/authentication/LOCAL_RU.md
```diff
@@ -145,7 +145,9 @@ spec:
 
 ## Настройка парольной политики
 
-Парольная политика позволяет контролировать сложность пароля, ротацию и блокировку пользователей.
+Парольная политика позволяет контролировать сложность пароля, ротацию и автоматическую блокировку после нескольких неуспешных попыток входа.
+
+Администратор может заблокировать или разблокировать пользователя командами [`d8 iam user lock` / `unlock`](/products/kubernetes-platform/documentation/v1/cli/d8/reference/#d8-iam) (или через UserOperation) без включения `passwordPolicy.lockout`. Секция `lockout` включает только автоматическую блокировку после заданного числа неудачных попыток.
```
z9r5:
```suggestion
Администратор может заблокировать или разблокировать пользователя командами [`d8 iam user lock` / `unlock`](../../../../cli/d8/reference/#d8-iam) (или через UserOperation) без включения `passwordPolicy.lockout`. Секция `lockout` включает только автоматическую блокировку после заданного числа неудачных попыток.
```

docs/documentation/pages/admin/configuration/access/authentication/LOCAL.md
```diff
@@ -144,7 +144,9 @@ Once the group is created and includes all necessary users, proceed by configuri
 
 ## Configuring password policy
 
-Password policy allows controlling password complexity, rotation, and user lockout.
+Password policy allows controlling password complexity, rotation, and automatic lockout after consecutive failed login attempts.
+
+An administrator can lock or unlock a user with [`d8 iam user lock` / `unlock`](/products/kubernetes-platform/documentation/v1/cli/d8/reference/#d8-iam) (or a UserOperation) without enabling `passwordPolicy.lockout`. The `lockout` section only turns on automatic lockout after the configured number of failed attempts.
```
z9r5:
```suggestion
An administrator can lock or unlock a user with [`d8 iam user lock` / `unlock`](../../../../cli/d8/reference/#d8-iam) (or a UserOperation) without enabling `passwordPolicy.lockout`. The `lockout` section only turns on automatic lockout after the configured number of failed attempts.
```

z9r5 (approved):

PR #22484, opened 2026-08-25
[user-authn] Enforce DexClient allow-lists on token refresh

Jabejixo:
> Based on #22483. Merge that PR first; this one targets its branch and should be retargeted to `main` once it lands.

## Description

`allowedEmails` and `allowedGroups` on DexClient / DexAuthenticator are now applied on login, on `grant_type=password` (ROPC / basic-auth-proxy), and on every refresh-token request. A session that no longer matches is rejected with `access_denied` and the refresh token is deleted.

Token exchange also applies the same filters. The CRD still documents login and refresh-token requests; we do not claim every grant type.

An administrator lock on refresh still returns `access_denied` without deleting the token (see #22483).

The client filter ran only in `finalizeLogin`. Changing `allowedGroups` after a user had already signed in left their refresh token able to mint new ID tokens with the old groups. `grant_type=password` never checked the allow-list at all.

Directory group membership is re-read only when Dex actually contacts the connector (LDAP already does this in `Refresh()`). A refresh served from `reuseInterval` still uses the groups stored on the token.

Revoking access by editing the allow-list did not take effect until the next interactive login. Existing refresh tokens that no longer match will start failing, those tokens are removed, and those users must sign in again.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: DexClient and DexAuthenticator allow-lists are now enforced on password grant and token refresh.
impact: |
  Changing allowedGroups or allowedEmails deletes refresh tokens that no longer match. Those users must sign in again.
impact_level: high
```


modules/150-user-authn/images/dex/patches/019-refresh-client-filters.patch
```diff
@@ -0,0 +1,476 @@
...
+--- a/server/handlers.go
++++ b/server/handlers.go
+@@ -770,6 +770,22 @@ func (e *NotAllowedError) Error() string {
+ 	return "not allowed: " + e.Reason
+ }
+ 
++// applyClientAccessFilters enforces DexClient allowedEmails / allowedGroups on
++// the claims that will be issued. Groups are reduced to the intersection with
++// allowedGroups. An empty allow-list means no restriction.
++func applyClientAccessFilters(client storage.Client, claims *storage.Claims) error {
```
AlwxSin, line 479: 🟡 **Password grant and token exchange still skip the allow-lists**

`applyClientAccessFilters` is called from `finalizeLogin` and `handleRefreshToken`. Two other token-issuing handlers never see it: `handleTokenExchange` (upstream `server/handlers.go:1350`) and `handlePasswordGrant` (`:1119`). Both build claims straight from the connector identity and issue an ID token with no filtering.

`grantTypeTokenExchange` is in dex's default grant set (`server/server.go:236-241`) and the module sets no `AllowedGrantTypes`, so it is live; `grantTypePassword` goes live together with `enableBasicAuth`.

Token exchange is the narrower of the two — it needs a connector implementing `TokenIdentityConnector` plus client credentials — but the CRD text this PR adds ("Checked on login and on every refresh-token request", `modules/150-user-authn/crds/dex-client.yaml:43`) will be read as "always".

modules/150-user-authn/images/dex/patches/019-refresh-client-filters.patch
```diff
@@ -0,0 +1,476 @@
...
++	// the refresh token. Login already stored the intersection of groups; a
++	// later change of allowedGroups must revoke that session.
++	storedClaims := rCtx.storageToken.Claims
++	if err := applyClientAccessFilters(client, &storedClaims); err != nil {
++		s.logger.WarnContext(r.Context(), "refresh denied by client access filter",
++			"client_id", client.ID,
++			"email", rCtx.storageToken.Claims.Email,
++			"err", err,
++		)
++		s.refreshTokenErrHelper(w, newAccessDeniedRefreshError())
```
AlwxSin, line 614: 🟡 **A denied refresh token stays valid in storage**

The filter returns `access_denied` and leaves the refresh token untouched, so it sits in storage until its own expiry. Put the group back and the old session resumes; nothing records that the token was ever rejected.

The patch README says this "revokes the session immediately" (`README.md:205`), which is stronger than what the code does — it declines to extend the session. Deleting the refresh token on denial would match both the wording and the containment intent.

modules/150-user-authn/images/dex/patches/019-refresh-client-filters.patch
```diff
@@ -0,0 +1,476 @@
...
+ 	invalidErr = newBadRequestError("Refresh token is invalid or has already been claimed by another client.")
+ 	expiredErr = newBadRequestError("Refresh token expired.")
+@@ -682,6 +686,20 @@ func (s *Server) handleRefreshToken(w http.ResponseWriter, r *http.Request, clie
+ 		return
+ 	}
+ 
++	// Fail closed against the current DexClient allow-lists before rotating
++	// the refresh token. Login already stored the intersection of groups; a
++	// later change of allowedGroups must revoke that session.
++	storedClaims := rCtx.storageToken.Claims
```
AlwxSin, line 606: 🔵 **`storedClaims` is filtered and then discarded**

Only the error is used; the filtered copy goes nowhere. It is harmless — `groups.Filter` allocates a fresh slice, so `rCtx.storageToken.Claims.Groups` is not mutated — but confirming that means going to read `pkg/groups`. A one-line comment saying the call is a check, not a transformation, would save the trip.

AlwxSin (approved):

z9r5 (approved):

PR #22490, opened 2026-08-25
[node-manager] Fencing agent 2.0/fencing 1e

ortherion:
## Description

Implements the fencing-controller FSM and SLA profile loading for `FencingFailedNodeState` incidents.

This change adds:
- `NewFSMFromCR` that restores the machine from `status.phase`
- ADR-aligned states and events (`Healthy`, `Suspected`, `FallbackAlive`, `ReadyToEvict`, `Evicting`, `Done`, `Error`) without extra transitions
- Resolve of built-in profile timings (`critical`, `medium`, `moderate`, `slow`) from `spec.profileRef.name`
- Snapshot of `FallbackTTL` and `EvacuationDelay` for the active incident reconcile loop
- `ConfigurationError` condition when the profile is missing or unusable, blocking the fast eviction path

Pod cleanup / eviction is intentionally not implemented yet. The reconciler stops at `ReadyToEvict` and prepares timings for follow-up work.

No restarts of ingress-controllers, control-plane, Prometheus, or other critical cluster components.

`fencing-controller` needs a deterministic decision path for fencing incidents based on agent signals (`status.failed` / `status.fallback`) and NodeGroup SLA profiles.

Without an FSM restored from `status.phase`, the controller cannot safely resume after restart. Without profile-based `FallbackTTL` / `EvacuationDelay`, it cannot apply profile-specific reaction windows or protect nodes that still send a fresh fallback heartbeat.

This PR introduces that decision core: restore FSM state, resolve profile timings once per incident, and advance only according to ADR transitions.

Not necessarily

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: feature
summary: Add fencing-controller FSM and built-in SLA profile loading for FencingFailedNodeState.
impact_level: low
```


ortherion:
Superseded by #22519, which carries the same work as a single commit from a clean branch. The build here failed because `metrics.go` never made it into the commit, so `reconciler.go` called `reportConfigurationError`/`clearConfigurationError` that did not exist.

PR #22513, opened 2026-08-25
[d8-shutdown-inhibitor] Rename drop-in to `zz-` for priority, improve log messages, and clean up old config files.

090809:
## Description

Renames the logind drop-in from `99-node-d8-shutdown-inhibitor.conf` to `zz-d8-shutdown-inhibitor.conf`.

systemd sorts `*.conf.d` drop-ins by file name across all search directories and the last assignment wins; the directory only breaks ties between identical names. `unattended-upgrades` ships `/usr/lib/systemd/logind.conf.d/unattended-upgrades-logind-maxdelay.conf` with `InhibitDelayMaxSec=30`, and `'u'` sorts after `'9'`, so it overrode the Deckhouse value. `'z'` sorts after every drop-in shipped into `logind.conf.d` by any Ubuntu noble or Debian bookworm/trixie package, and after kubelet's `99-kubelet.conf`.

Also stops `pkg/app/app.go` from treating a lower-than-requested `InhibitDelayMaxSec` as fatal. That check ran as the first statement of `App.Start()`, so the process died before wiring any task and the unit (`Restart=always`, `RestartSec=10`, `StartLimitInterval=0`) restarted it forever. A lost sort only caps the `mode=delay` shutdown lock — it does not affect the `mode=block` lock on `handle-power-key`, node cordoning, the node condition or `wall` broadcasts — and a restart cannot change the outcome, so it is now a warning and startup continues. Transient failures in the same function (D-Bus connect, property read, file write, logind reload) stay fatal.

`076_install_d8_shutdown_inhibitor.sh.tpl` and `cleanup_static_node.sh.tpl` now remove the drop-in by the glob `*d8-shutdown-inhibitor.conf` so cleanup catches both the new and the legacy name. The legacy file is left in place on running nodes: it sorts before `zz-` and carries the same value.

No impact on running components. The only unit affected is `d8-shutdown-inhibitor.service` on nodes, restarted by the existing `inhibitor-need-restart` bashible flag on image digest change.

On Ubuntu 24.04 and Debian 12/13 with `unattended-upgrades` installed, `d8-shutdown-inhibitor` never started: logind reported 30 s against the requested 72 h and the process exited every 10 seconds. Pods labelled `pod.deckhouse.io/inhibit-node-shutdown` were not protected, the node was not cordoned and the graceful-shutdown node condition was never set. `StartLimitInterval=0` kept the unit out of the failed state, so it did not surface in degraded-unit alerting.

The same collision defeats kubelet's own graceful shutdown on these distributions (`99-kubelet.conf` also sorts before `unattended-upgrades-…`, cf. kubernetes/kubernetes#102818); the rename fixes that as a side effect.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: fix
summary: Fix d8-shutdown-inhibitor failing to start on nodes where another package overrides the logind inhibit delay.
impact_level: default
```

```changes
section: candi
type: fix
summary: Node cleanup now removes the d8-shutdown-inhibitor logind drop-in under both its current and legacy names.
impact_level: low
```

### DIFF
--- a/candi/bashible/cleanup_static_node.sh.tpl
+++ b/candi/bashible/cleanup_static_node.sh.tpl
@@ -68,7 +68,7 @@ SYSTEMD_FILES=(
   /lib/systemd/system/containerd-deckhouse*
   /etc/systemd/system/d8-shutdown-inhibitor*
   /lib/systemd/system/d8-shutdown-inhibitor*
-  /etc/systemd/logind.conf.d/99-node-d8-shutdown-inhibitor.conf
+  /etc/systemd/logind.conf.d/*d8-shutdown-inhibitor.conf
   /etc/systemd/system/kubelet*
   /lib/systemd/system/kubelet*
 )
--- a/candi/bashible/common-steps/all/076_install_d8_shutdown_inhibitor.sh.tpl
+++ b/candi/bashible/common-steps/all/076_install_d8_shutdown_inhibitor.sh.tpl
@@ -30,7 +30,8 @@
 bb-log-debug "shutdown inhibitor package={{ $inhibitorPackage }} present={{ $imagePresent }}"
 
 inhibitor_service_name="d8-shutdown-inhibitor.service"
-extra_logind_conf="/etc/systemd/logind.conf.d/99-node-d8-shutdown-inhibitor.conf"
+# Glob matches the current zz- drop-in and the legacy 99-node- one left by older releases.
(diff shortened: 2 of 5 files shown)


borg-z (approved):

deckhouse-BOaTswain:
Cherry pick PR [22530](https://github.com/deckhouse/deckhouse/pull/22530) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22520, opened 2026-08-26
[control-plane-manager] Virtual control plane: bootstrap vcp nodes via the stock node-manager path

trofimovdals:
## Description

- Tenant nodes now join through the stock `node-manager` manual-bootstrap flow instead of a bespoke host-rendered `join.sh`.
- The tenant owns its `NodeGroup`s and per-NG bootstrap tokens (`order_bootstrap_token`); the host stops publishing them and only supplies facts a tenant can't derive (kubernetes CA, api-proxy certs, ALB VIP).
- A default static `worker` `NodeGroup` is seeded in fresh tenants (idempotent, never patched).
- Nodes map control-plane SNI hostnames to the ALB VIP in `/etc/hosts`; the `:80` tokenless rpp-get bootstrap route is scoped by tenant UID and rewritten to the parent UID for `registry-packages-proxy`.
- Removes `reconcile_joinscript.go`, `reconcile_kubelet_csr.go`, `bootstraptoken` and the host-side join-script secret; kubelet-serving CSRs are approved by the tenant's own `node-manager`.
- DKP pod reaches the tenant apiserver by ClusterIP, not the ALB
- Restart the tenant deckhouse pod once the cluster is bootstrapped (workaround)
Does not affect self-hosted clusters (all new logic is gated on `global.deckhouseSelfHosted`).
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: node-manager
type: feature
summary: Virtual control plane tenants bootstrap nodes through the stock node-manager path and own their NodeGroups and bootstrap tokens.
impact_level: low
```


modules/040-node-manager/templates/vcp-bootstrap-rbac.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if not .Values.global.deckhouseSelfHosted }}
```
sprait: is it not the same as in https://github.com/deckhouse/deckhouse/blob/d6ae52baf0540a0e174fb3f540748e62aaf6b385/modules/040-node-manager/templates/rbac-for-us.yaml#L39-L82 ?

modules/040-node-manager/templates/node-group/_static_or_hybrid_script.tpl
```diff
@@ -32,6 +32,21 @@ EOF
 chmod 0600 /var/lib/bashible/bootstrap-token
```
sprait, line 32: pls check https://github.com/deckhouse/deckhouse/blob/d6ae52baf0540a0e174fb3f540748e62aaf6b385/modules/040-node-manager/templates/_helpers.yaml#L151-L155 

It looks like there is data here that isn't currently being used

z9r5 (approved):

modules/040-node-manager/hooks/bashible_context_vcp.go
```diff
@@ -47,8 +48,10 @@ const bashibleExternalInputsVersion = 1
 const bashibleExternalInputsSecretName = "bashible-external-inputs"
 
 var _ = sdk.RegisterFunc(&go_hook.HookConfig{
-	Queue:        "/modules/node-manager/bashible_context_vcp",
-	OnBeforeHelm: &go_hook.OrderedConfig{Order: 20},
+	Queue: "/modules/node-manager/bashible_context_vcp",
+	// After get_crds (10) and order_bootstrap_token (20): the tenant nodeGroups and per-NG tokens
+	// this hook folds into the context are set by those.
+	OnBeforeHelm: &go_hook.OrderedConfig{Order: 30},
```
sprait, line 61: adopt hook with same hook in one order?
https://github.com/deckhouse/deckhouse/blob/8f711a418c0c16f5b1a9795fcd37c85bf2684847/modules/040-node-manager/hooks/bashible_context_vcp_adopt.go#L67-L69

sprait (approved):

PR #22532, opened 2026-08-26
[deckhouse] Fall back to List when publicDomainTemplate shares the first label with a suffix

Jabejixo:
## Description

`settings.reservedPublicHosts.mode` now defaults to `List`: only the hostnames Deckhouse publishes are reserved. Set `Template` for the wider reservation.

`mode: Template` still reserves every hostname `publicDomainTemplate` can render. When `%s` is only a prefix of the first label (`%s-cluster.example.com`), that set is the same as tenant Ingresses named with the cluster suffix, so Template falls back to List even if it was asked for. `kube-%s.example.com` is unchanged: the literal is a prefix, so only `kube-*` is reserved. `%s.example.com` stays `Template` when Template is set.

Helm and the snapshot hook use the same decision. The ModuleConfig field is not rewritten; the ConfigMap `mode` key shows the effective value.

A cluster whose `publicDomainTemplate` embeds a cluster id in the first label could not create or update tenant Ingresses on that same suffix after the reservation shipped. Defaulting to List keeps clusters that never set the field on the narrower reservation.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Default reservedPublicHosts.mode to List; Template no longer covers publicDomainTemplate values where %s is only a prefix of the first label.
impact: Clusters that never set mode now reserve only the platform service hostnames. Set mode Template for the wider reservation. A publicDomainTemplate like %s-cluster.example.com stays on List even if Template is asked for. kube-%s.example.com stays Template when Template is set.
impact_level: high
```


deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:fix/reserved-hosts-suffix-template` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/33504731358) in 22m30s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.2.237`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 20m58s.

</details>

modules/002-deckhouse/openapi/config-values.yaml
```diff
@@ -345,22 +345,26 @@ properties:
...
-      - mode: Template
+      - mode: List
         additionalHosts: ["admin.example.com"]
         excludedServices: ["grafana"]
     properties:
       mode:
         type: string
         enum: ["Template", "List"]
-        default: Template
+        default: List
```
AlwxSin, line 355: 🔵 **10. `impact_level: default` understates a cluster-wide narrowing of a security control**

The change itself is right, and the Go and Helm sides of the decision do agree on every shape the global schema admits (checked `%s.example.com`, `kube-%s.example.com`, `%s-cluster.example.com`, two `%s`, none; `foo.%s.example.com` is unreachable because `global-hooks/openapi/config-values.yaml:96` only admits `%s` inside the first label).

But every cluster that never set `mode` is on the effective `Template` today and silently drops to `List` on upgrade — a reservation that shrinks without anyone touching a ModuleConfig. The changelog text says so; the level arguably should too.

Jabejixo: Raised `impact_level` to `high` so the default Template→List change is copied into Know before update.

ldmonster (approved):

modules/002-deckhouse/openapi/config-values.yaml
```diff
@@ -345,22 +345,26 @@ properties:
...
         description: |
           How much of the platform's domain is reserved.
 
+          Default is `List`. Set `Template` to reserve every hostname the template can render.
```
z9r5:
```suggestion
          Set `Template` to reserve every hostname the template can render.
```

Default is rendered automatically from the spec

modules/002-deckhouse/openapi/doc-ru-config-values.yaml
```diff
@@ -220,11 +220,15 @@ properties:
         description: |
           Какая часть домена платформы резервируется.
 
+          По умолчанию — `List`. Чтобы резервировать каждое имя, которое способен сформировать шаблон, задайте `Template`.
```
z9r5:
```suggestion
          Чтобы резервировать каждое имя, которое способен сформировать шаблон, задайте `Template`.
```

z9r5 (approved):

AlwxSin (approved):

PR #22566, opened 2026-08-26
[istio] Refactor proxyv2 image builds

yshch:
Logical backport of https://github.com/deckhouse/deckhouse/pull/22584

## Description

This PR reworks how the `istio` module builds the `proxyv2` (envoy) images and documents the procedure for regenerating their bazel inputs.

Build environment:
- moves the Istio 1.21.6 build image from `builder/alt` (ALT Linux P11) to `builder/golang-bookworm`, with clang 14, lld 14 and libc++ 14 taken from Debian packages
- drops the `build-libcxx-artifact` stage and the LLVM source checkout that Istio 1.21.6 needed to build `libcxxabi`/`libcxx` by hand, along with the envoy patches that pointed the build at those self-built libraries
- renames the shared `golang-artifact` image to a per-image `-golang-artifact` so it no longer collides between proxyv2 versions
- bumps the 1.21.6 bazel deps/cache revisions to the new `-bookworm-v1` branches

glibc in final images:
- replaces the `libs/glibc-v2.41` imports in `proxyv2-v1x21x6`, `proxyv2-v1x25x2`, `proxyv2-v1x27x9`, `ztunnel-v1x25x2` and `ztunnel-v1x27x9` with a per-image `-glibc-artifact` stage that installs `libc@2.43` from `builder/distroless` and exports only `libc.so.6`, `libm.so.6` and `ld-linux-x86-64.so.2`

Bazel deps and cache handling, unified across all three proxyv2 versions:
- deps (`external.tar.gz`) are always prepared and unpacked, and `build --nofetch` is always set, so the envoy build never touches the network
- the `$buildWithPreparedCacheAndDeps` toggle becomes `$buildWithPreparedCache` and now only controls the disk cache: `true` clones the prepared cache into a `tmp_dir` mount so it never lands in an image layer, `false` starts from an empty cache and keeps `/tmp/bazel-cache` in the artifact image for export as a new cache revision
- build flags move from ad-hoc `bazel build` arguments and `envoy.bazelrc` `sed` edits into a generated `user.bazelrc`
- the envoy binary is moved out of `bazel-bin` in the same shell that computed `BAZEL_OUTPUT_BASE`, since each werf stage runs its own shell

Tooling and docs:
- adds `modules/110-istio/tools/build-envoy-artifacts`, a Go helper that reproduces the CI build environment in a container and produces either the deps tarball (`bazel build --nobuild`, analysis only) or a filled bazel disk cache, with per-version profiles for 1.21.6, 1.25.2 and 1.27.9 and support for LLVM-from-source versions via `-source-repo`/`-ssh-key`
- adds `modules/110-istio/tools/go.mod`, registers that directory with Dependabot, and excludes `tools` from the module Helm chart via `.helmignore`
- rewrites `modules/110-istio/images/README.md` to describe the new 1.21.6 build, the difference between deps and cache, and step-by-step regeneration procedures for both

The Istio 1.21.6 proxyv2 image was the last one still built on ALT Linux, with a bespoke chain that compiled `libcxxabi`/`libcxx` from LLVM sources just to get an envoy toolchain the distribution packages could not provide. That stage was slow, fragile and unlike how the 1.25 and 1.27 images are built, so every change to the envoy build had to be made twice in two different shapes. Moving 1.21.6 onto the same Debian-based toolchain removes the custom libc++ build entirely and leaves one build pattern across all supported proxyv2 versions.

The deps/cache handling was also inconsistent: one version fetched during the build, another disabled fetching, and a single flag conflated "use a prepared cache" with "use prepared deps". Separating them makes the guarantee explicit — the build is always offline, and the cache only affects how long it takes — and prevents a stale cache from silently turning into a build failure.

Regenerating those bazel inputs was previously undocumented tribal knowledge that required assembling the build container by hand. The new helper and README make it a reproducible command and describe when each artifact has to be rebuilt.

The per-image `-glibc-artifact` stage takes the runtime glibc files from the same distroless package source the rest of the images use, instead of depending on a separate `libs/glibc-v2.41` image.

There are no functional changes to the resulting images: same Istio versions, same binaries, same contents.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: istio
type: chore
summary: Rework proxyv2 image builds onto a common Debian-based toolchain and document envoy bazel artifact regeneration.
impact_level: low
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -808,6 +808,15 @@ updates:
   open-pull-requests-limit: 0
 
 - package-ecosystem: "gomod"
+  directory: "/modules/110-istio/tools"
+  labels:
+  - "type/dependencies"
+  - "status/ok-to-test"
+  schedule:
+    interval: "daily"
+  open-pull-requests-limit: 0
+
+- package-ecosystem: "gomod"
   directory: "/modules/150-user-authn/images/basic-auth-proxy/app"
   labels:
   - "type/dependencies"
--- a/ee/modules/110-istio/images/ztunnel-v1x25x2/werf.inc.yaml
+++ b/ee/modules/110-istio/images/ztunnel-v1x25x2/werf.inc.yaml
@@ -9,16 +9,16 @@ import:
     owner: 1337
     group: 1337
     before: install
-  - from: {{ index $.Images "libs/glibc-v2.41" }}
-    add: /lib64/libc.so.6
+  - image: {{ $.ModuleName }}/{{ $.ImageName }}-glibc-artifact
+    add: /out/lib64/libc.so.6
     to: /lib64/libc.so.6
     before: install
-  - from: {{ index $.Images "libs/glibc-v2.41" }}
-    add: /lib64/libm.so.6
+  - image: {{ $.ModuleName }}/{{ $.ImageName }}-glibc-artifact
+    add: /out/lib64/libm.so.6
     to: /lib64/libm.so.6
     before: install
-  - from: {{ index $.Images "libs/glibc-v2.41" }}
-    add: /lib64/ld-linux-x86-64.so.2
+  - image: {{ $.ModuleName }}/{{ $.ImageName }}-glibc-artifact
+    add: /out/lib64/ld-linux-x86-64.so.2
     to: /lib64/ld-linux-x86-64.so.2
     before: install
   - from: {{ index $.Images "libs/xz-v5.8.1" }}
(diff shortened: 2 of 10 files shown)


deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:istio-refactor-proxyv2-image-builds-1.77` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/33065480583) in 39m33s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@93.77.187.176`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 38m10s.

</details>

skurbatov (approved):

Nikolay1224 (approved):

himax1991 (approved):

PR #22584, opened 2026-08-27
[istio] Refactor proxyv2 image builds

yshch:
## Description

This PR unifies how the `istio` module builds the `proxyv2` (envoy) images and documents the procedure for regenerating their bazel inputs.

glibc in final images:
- replaces the `libs/glibc-v2.41` imports in `proxyv2-v1x25x2`, `proxyv2-v1x27x9`, `proxyv2-v1x29x6`, `ztunnel-v1x25x2`, `ztunnel-v1x27x9` and `ztunnel-v1x29x6` with a per-image `-glibc-artifact` stage that installs `libc@2.43` from `builder/distroless` and exports only the runtime files each image actually needs

Bazel deps and cache handling, made the same across all three proxyv2 versions:
- deps (`external.tar.gz`) are always prepared and unpacked, and `build --nofetch` is always set, so the envoy build never touches the network
- the `$buildWithPreparedCacheAndDeps` toggle becomes `$buildWithPreparedCache` and now only controls the disk cache: `true` clones the prepared cache into a `tmp_dir` mount so it never lands in an image layer, `false` starts from an empty cache and keeps `/tmp/bazel-cache` in the artifact image for export as a new cache revision
- build flags move from ad-hoc `bazel build` arguments into a generated `user.bazelrc`
- the envoy binary is moved out of `bazel-bin` in the same shell that computed `BAZEL_OUTPUT_BASE`, since each werf stage runs its own shell
- drops unused `$llvmRev` and `$goVersion` variables from the 1.25.2 build definition

Tooling and docs:
- adds `modules/110-istio/tools/build-envoy-artifacts`, a Go helper that reproduces the CI build environment in a container and produces either the deps tarball (`bazel build --nobuild`, analysis only) or a filled bazel disk cache, with per-version profiles for 1.25.2, 1.27.9 and 1.29.6 and support for LLVM-from-source versions via `-source-repo`/`-ssh-key`
- adds `modules/110-istio/tools/go.mod`, registers that directory with Dependabot, and excludes `tools` from the module Helm chart via `.helmignore`
- extends `modules/110-istio/images/README.md` with the difference between deps and cache and step-by-step regeneration procedures for both

The deps/cache handling differed between proxyv2 versions: 1.25.2 fetched during the build, 1.27.9 and 1.29.6 disabled fetching, and a single flag conflated "use a prepared cache" with "use prepared deps". Separating them makes the guarantee explicit — the build is always offline, and the cache only affects how long it takes — and prevents a stale cache from silently turning into a build failure. It also means a change to the envoy build no longer has to be written twice in two different shapes.

Regenerating those bazel inputs was previously undocumented tribal knowledge that required assembling the build container by hand. The new helper and README make it a reproducible command and describe when each artifact has to be rebuilt.

The per-image `-glibc-artifact` stage takes the runtime glibc files from the same distroless package source the rest of the images use, instead of depending on a separate `libs/glibc-v2.41` image.

There are no functional changes to the resulting images: same Istio versions, same binaries, same contents.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: istio
type: chore
summary: Unify proxyv2 envoy image builds and document envoy bazel artifact regeneration.
impact_level: low
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -808,6 +808,15 @@ updates:
   open-pull-requests-limit: 0
 
 - package-ecosystem: "gomod"
+  directory: "/modules/110-istio/tools"
+  labels:
+  - "type/dependencies"
+  - "status/ok-to-test"
+  schedule:
+    interval: "daily"
+  open-pull-requests-limit: 0
+
+- package-ecosystem: "gomod"
   directory: "/modules/150-user-authn/images/basic-auth-proxy/app"
   labels:
   - "type/dependencies"
--- a/ee/modules/110-istio/images/ztunnel-v1x25x2/werf.inc.yaml
+++ b/ee/modules/110-istio/images/ztunnel-v1x25x2/werf.inc.yaml
@@ -9,16 +9,16 @@ import:
     owner: 1337
     group: 1337
     before: install
-  - from: {{ index $.Images "libs/glibc-v2.41" }}
-    add: /lib64/libc.so.6
+  - image: {{ $.ModuleName }}/{{ $.ImageName }}-glibc-artifact
+    add: /out/lib64/libc.so.6
     to: /lib64/libc.so.6
     before: install
-  - from: {{ index $.Images "libs/glibc-v2.41" }}
-    add: /lib64/libm.so.6
+  - image: {{ $.ModuleName }}/{{ $.ImageName }}-glibc-artifact
+    add: /out/lib64/libm.so.6
     to: /lib64/libm.so.6
     before: install
-  - from: {{ index $.Images "libs/glibc-v2.41" }}
-    add: /lib64/ld-linux-x86-64.so.2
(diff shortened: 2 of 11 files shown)


deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:istio-refactor-proxyv2-image-builds` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/33065734419) in 41m9s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@51.250.4.196`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 39m46s.

</details>

skurbatov (approved):

Nikolay1224 (approved):

himax1991 (approved):

PR #22609, opened 2026-08-27
[multitenancy-manager] Adopt every namespace outside a project into a full project of its own

Jabejixo:
## Description

Every namespace that belongs to no project now becomes a full-fledged `Project` with a template, and the "project managed by its namespace" model is gone.

- The template is inferred from what the namespace already carries: `secure` when it has `security-scanning.deckhouse.io/enabled`, `default` when it has `security.deckhouse.io/pod-policy` or `extended-monitoring.deckhouse.io/enabled`, and `simple` otherwise.
- `spec.parameters` are seeded from the current state of the namespace so the first render is a no-op. This matters because `default` and `secure` default `networkPolicy` to `Isolated` and `podSecurityProfile` to `Baseline`; adopting an existing namespace on those defaults would drop an isolating NetworkPolicy into it and tighten the Pod Security Standard on workloads running there today. The seeded values are `networkPolicy: NotRestricted` and the profile read back from the existing `pod-policy` label (`Privileged` when the label is absent). Namespace admins are **not** copied into `.spec.administrators`.
- Adopt stamps Helm ownership on the existing namespace (`app.kubernetes.io/managed-by=Helm` and `meta.helm.sh/release-name`). The release name is the project name when it fits Helm's 53-character limit, otherwise a deterministic truncation (prefix + 8 hex of md5).
- A leftover wrap (`multitenancy.deckhouse.io/project-managed-by-namespace=true`) is finished on controller start and on every reconcile: a live namespace is inferred into a real template; a Terminating or missing namespace deletes the leftover `Project` so Helm never recreates it. Handmade template-less projects (no wrap label) still fall back to `simple`. Virtual projects (`deckhouse`, `default`) are skipped.
- The concept of a project without a template is removed: `spec.projectTemplateName` now defaults to `simple` in the CRD schema (the field stays optional, so existing manifests keep applying). Defaulting is done in the schema rather than a mutating webhook to avoid adding another admission hop with its own `failurePolicy` and bootstrap deadlock risk. `ensureTemplateName` does not pin `simple` on a leftover wrap.
- The `projects.deckhouse.io/adopt` annotation and the namespace-creation-deny webhook are removed. `allowNamespacesWithoutProjects` is kept as a deprecated no-op for one release so existing ModuleConfigs keep applying.

These namespaces are **not** adopted; they stay on a virtual inventory project, and virtual projects never create or recreate namespaces:

- virtual `deckhouse`: `d8-*`, `kube-*`, `upmeter-*`, `IsSystem`, `heritage: deckhouse|upmeter`
- virtual `default`: the `default` namespace and any unowned name longer than 61 characters (Project name limit)

Namespaces already labelled `projects.deckhouse.io/project` are owned and are not adopted. After adopt, virtual inventory is refreshed from both the old and the new namespace so the adopted name drops off `default` / `deckhouse`.

The namespace-managed project was a half-citizen: its `spec.parameters` were silently ignored (the template-less path never reads them), which is why a separate mirroring mechanism had to keep the namespace labels in sync with parameters that did nothing. Its spec could not be edited, it was exempt from the grant allow-list, and it needed a bespoke exemption in the admission policy. Making every namespace a real project removes all four special cases and leaves one model to reason about.

Not necessarily.

## Checklist

- [ ] e2e tests passed.

### Manual cluster verification

Re-checked on `v1.78.0-pr22609` after leftover-wrap / virtual-inventory / Helm-stamp work.

- A fresh plain namespace is adopted as `simple` / `Deployed`, Helm-owned, with empty `.spec.administrators`; virtual `default` drops the name.
- A 62-character namespace is not adopted and stays on virtual `default`.
- `heritage: deckhouse` is not adopted, is inventoried on virtual `deckhouse`, and deleting it is not recreated.
- A 61-character adopted namespace is stamped with a Helm release name of length 53.
- A leftover wrap with no namespace is deleted; Helm does not create the namespace.
- A leftover wrap whose namespace is disappearing is deleted; Helm does not recreate it.
- Virtual projects `deckhouse` and `default` keep the `virtual` template; no project is created for `kube-system` or `d8-system`.
- A namespace with `security.deckhouse.io/pod-policy: restricted` gets `default` and `podSecurityProfile: Restricted`; one with `security-scanning.deckhouse.io/enabled` gets `secure` and `securityScanningEnabled: true`. In both cases `networkPolicy: NotRestricted` is seeded and no NetworkPolicy is rendered.

## Changelog entries

```changes
section: multitenancy-manager
type: feature
summary: Every namespace outside a project is adopted into a full project with a template inferred from the namespace, and the project managed by its namespace is gone.
impact: |
  Breaking changes for clusters that relied on the previous model.

  - An auto-wrapped namespace used to be freely deletable, and deleting it deleted its project. Its namespace is an ordinary project namespace now, so the admission policy refuses to delete or edit it; delete the `Project` instead.
  - Namespaces that used to be orphans are project namespaces now, so the cluster-resource grant allow-list applies to them. RoleBindings to non-delegatable ClusterRoles in those namespaces start being denied.
  - The spec of a previously namespace-managed project is no longer controller-owned and can be edited directly; its labels and annotations are no longer mirrored from the namespace.
  - The `projects.deckhouse.io/adopt` annotation no longer does anything; adoption is automatic.
  - `allowNamespacesWithoutProjects` is deprecated and ignored, and will be removed in the next release.
  - A leftover wrap whose namespace is already gone or Terminating is deleted instead of Helm-recreating the namespace.
```


deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:feat/mtm-namespaces-as-full-projects` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/33504772140) in 27m46s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.5.126`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 26m5s.

</details>

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/namespace/adopt.go
```diff
@@ -0,0 +1,197 @@
...
+
+	if template == TemplateSimple {
+		if len(params) == 0 {
+			return nil
+		}
+		return params
+	}
+
+	labels := namespace.GetLabels()
+	params["networkPolicy"] = networkPolicyNotRestricted
```
AlwxSin, line 91: 🔴 **8. Adoption is not a no-op: `default` and `secure` drop a Deny-mode OperationPolicy on live namespaces**

Real location is the two template files this PR does not touch — `src/templates/default.yaml` and `src/templates/secure.yaml` — so the comment sits on the seeding that is supposed to make adoption inert.

`ParametersFor` carefully neutralises the two *parameters* that tighten (`networkPolicy` → `NotRestricted`, `podSecurityProfile` read back from the label). But both templates render an **unconditional** `OperationPolicy`, and no parameter turns it off:

```yaml
kind: OperationPolicy
metadata:
  name: required-requests-<hash>
spec:
  policies:
    requiredResources:
      requests: [cpu, memory]
  match:
    namespaceSelector:
      labelSelector:
        matchLabels:
          kubernetes.io/metadata.name: "<project>"
```

`enforcementAction` defaults to `Deny` (`modules/015-admission-policy-engine/crds/operation-policy.yaml:34-36`), and nothing exempts an adopted namespace from gatekeeper: the admission-policy-engine webhook's `namespaceSelector` (`modules/015-admission-policy-engine/templates/validatingwebhookconfiguration.yaml:69-74`) excludes only `heritage: deckhouse`, while a project namespace is stamped `heritage: multitenancy-manager` (`internal/manager/project/standard_fields.go:37`). There is no project-wide exclusion in `templates/_helpers.tpl` either.

So every existing namespace carrying `security.deckhouse.io/pod-policy` or `extended-monitoring.deckhouse.io/enabled` (→ `default`), or `security-scanning.deckhouse.io/enabled` (→ `secure`), starts refusing workloads without CPU and memory requests the moment it is adopted — deployments in those namespaces stop rolling. The manual-verification list in the PR body covers the NetworkPolicy and the PSS label, but not this.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/helm/ownership.go
```diff
@@ -0,0 +1,78 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ManagedByHelm is the value Helm writes on objects it created (or was told it created).
+const ManagedByHelm = "Helm"
+
+// ApplyReleaseOwnership writes Helm ownership metadata onto ns in memory. It
+// returns true when the object changed and must be persisted. A terminating
+// namespace is left alone: there is nothing to stamp.
+func ApplyReleaseOwnership(ns *corev1.Namespace, releaseName string) bool {
```
AlwxSin, line 126: 🟡 **9. Adoption overwrites Helm ownership that a foreign release already holds**

This writes `app.kubernetes.io/managed-by=Helm`, `meta.helm.sh/release-name=<project>` and `meta.helm.sh/release-namespace=""` onto any namespace being adopted, without checking what is already there.

A namespace created by a user's own Helm chart, or by Argo CD's Helm renderer, carries `release-name: <their release>` and `release-namespace: <ns>`. Adoption silently rewrites both, and their next `helm upgrade` fails with an ownership conflict on a resource it legitimately owns — for a namespace whose only "fault" was not having a `projects.deckhouse.io/project` label yet.

Worth refusing to take over a namespace that names a different release, and surfacing that as a project condition instead of stamping over it.

z9r5 (approved):

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/namespace/adopt.go
Jabejixo: Agreed — seeding NP/PSS left the required-requests OperationPolicy on. We added a `requiredRequests` param (schema default stays `true`) and seed `false` on adopt; a missing key still renders the policy.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/helm/ownership.go
Jabejixo: Agreed — we no longer overwrite a foreign `(release-name, release-namespace)` pair. The Project goes to Error with `HelmOwnership`, and a brand-new unstamped namespace is left alone for 15s so Helm/Argo can stamp first.

AlwxSin (approved):

Lada7878 (approved):

z9r5 (approved):

PR #22626, opened 2026-08-28
[deckhouse] fix nelm tracking

ipaqsa:
## Description

Nelm tracking options were hardcoded inside `nelm.Client.Install`, so every release — module or
application — was installed with:

```go
TrackingOptions{
    NoPodLogs:                    true,
    NoFinalTracking:              true,
    LegacyHelmCompatibleTracking: true,
}
```

`NoFinalTracking: true` disables the phase that waits for resources to reach readiness, and
`LegacyHelmCompatibleTracking: true` narrows tracking to Job hooks only. That is the behaviour
modules inherited from the addon-operator era and need to keep, but it also meant applications
produced almost no tracking events — so the progress that `status.UpdateTracking` writes into the
`ManifestsApplied` condition never actually materialised for them.

This PR makes tracking a decision of the caller instead of a constant:

* `nelm.InstallOptions` gains a `TrackingOptions` field, and `Client.Install` passes it through
  instead of building its own.
* `packages/nelm.Service.Upgrade` takes a new `UpgradeOptions` struct (`TrackingOptions`,
  `ExtraLabels`) and forwards both — the labels merged into the release's `ResourcesLabels`.
* The `run` task builds those options per package kind and passes the same value to both of its
  `Upgrade` call sites (the initial upgrade and the re-upgrade after hooks change values):
  * **applications** — `NoPodLogs: true` only, so final tracking runs over all resources and real
    readiness progress reaches the `ManifestsApplied` condition;
  * **modules** — the previous legacy triple, so their behaviour is unchanged.

The kind is determined by a type assertion against a local one-method `application` interface
(`GetInstance() string`), which today only `*Application` satisfies.

Additionally, release resources are now stamped with two labels:

* `packages.deckhouse.io/package` — the package name, for every package;
* `packages.deckhouse.io/instance` — the application instance name, applications only.

To supply them, `GetPackage()` is added to both `Application` and `Module` (and to the `run` task's
`packageI` interface), and `Application.GetInstance()` is exported.

### Impact

Labels are applied via nelm's `ExtraLabels` (`NewExtraMetadataPatcher`), which patches top-level
resource metadata only — pod templates are untouched, so **no workload restarts**. However, the
labels are part of the rendered manifests that produce the upgrade checksum, so rolling out this
change re-applies **every** module and application release exactly once. No critical component is
restarted by it.

Applications had no usable deployment progress or readiness signal: with final tracking disabled,
`OnTrackingEvent` fired for practically nothing, so an application's `ManifestsApplied` condition
never reflected what was happening in the cluster. Applications are the flow where that feedback
matters — they are user-installed workloads whose rollout the user watches — while modules still
depend on Helm-compatible tracking semantics.

Hardcoding one tracking policy for both flows made it impossible to serve them differently. Moving
the options to the call site fixes application tracking without changing module behaviour, and the
new resource labels make a release's resources attributable to the package and application instance
that own them.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Fix nelm tracking for applications and modules.
impact_level: low
```

### DIFF
--- a/deckhouse-controller/internal/nelm/client.go
+++ b/deckhouse-controller/internal/nelm/client.go
@@ -294,6 +294,8 @@ type InstallOptions struct {
 
 	ResourcesLabels map[string]string // Labels to apply to all resources
 
+	TrackingOptions common.TrackingOptions
+
 	// OnTrackingEvent is an optional callback invoked with progress updates
 	// as Kubernetes resources are being tracked for readiness during install.
 	OnTrackingEvent func(name string, report progrep.ProgressReport)
@@ -362,11 +364,7 @@ func (c *Client) Install(ctx context.Context, namespace, releaseName string, opt
 			ValuesFiles: opts.ValuesPaths,
 			RootSetJSON: valuesSet,
 		},
-		TrackingOptions: common.TrackingOptions{
-			NoPodLogs:                    true,
-			NoFinalTracking:              true,
-			LegacyHelmCompatibleTracking: true,
-		},
+		TrackingOptions:        opts.TrackingOptions,
 		Chart:                  opts.Path,
 		DefaultChartName:       releaseName,
 		DefaultChartVersion:    "0.2.0",
--- a/deckhouse-controller/internal/packages/apps/app.go
+++ b/deckhouse-controller/internal/packages/apps/app.go
@@ -287,6 +287,11 @@ func (a *Application) GetInstance() string {
 	return a.instance
 }
 
+// GetPackage returns the application package name.
+func (a *Application) GetPackage() string {
+	return a.definition.Name
+}
+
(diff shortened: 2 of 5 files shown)


ldmonster (approved):

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 2963f2afc30fcfa1fcc027fe6f65456da35168f1 to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/33161626202) for details.

ldmonster:
not need to backport, there is no changes for it

PR #21888, opened 2026-08-05
[go-lib] Fix validation for optional module not found

riptide-01:
## Description
Fixes module release validation so that an `!optional` dependency on a module is correctly skipped when the required module does not exist at all (its `Module` resource is missing), not just when it exists but is disabled.

## Why do we need it, and what problem does it solve?
The purpose of marking a dependency `!optional` is to allow a module release to pass validation even when that dependency isn't present/enabled. The "not found" case was not covered by this check, so module releases with optional dependencies on modules that aren't installed at all were incorrectly rejected.

## Checklist
- [x] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: tools
type: fix
summary: Fixed module release validation incorrectly failing for `!optional` dependencies when the required module doesn't exist in the cluster.
impact_level: default
```

PR #22640, opened 2026-08-28
[deckhouse] Fix ModuleConfig admission rejecting a module with a disabled conditional dependency

Fkuloff:
## Description

The ModuleConfig admission webhook read only the keys of `properties.requirements.modules`, so the
`!optional` marker in the map value was invisible and every parent was treated as mandatory. The
loop now binds the constraint and skips the parents carrying the marker.

`CheckEnabling` was optional-blind the same way for a parent whose Module resource is absent, and
now mirrors the fix applied to `ValidateRelease` in #21888.

A module could not be enabled while its conditional dependency was disabled or not installed, the
case `!optional` exists for and where the requirement must be skipped. Added in #20827, shipped in
v1.77.0.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Fix ModuleConfig admission rejecting a module with a disabled conditional dependency
impact_level: default
```

### DIFF
--- a/deckhouse-controller/pkg/apis/deckhouse.io/v1alpha1/module.go
+++ b/deckhouse-controller/pkg/apis/deckhouse.io/v1alpha1/module.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


ldmonster (approved):

deckhouse-BOaTswain:
Cherry pick PR [22642](https://github.com/deckhouse/deckhouse/pull/22642) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22667, opened 2026-08-31
[user-authz] Gate identity assignment with can-assign cover or range

Jabejixo:
## Description

Replaces the privilege-rank webhook with a can-assign kernel.

Deckhouse treats the Kubernetes username as the Dex email claim. Grants live on that string (ClusterAuthorizationRule, AuthorizationRule, ClusterRoleBinding), not on the User object. Creating a credential for a free string remains ordinary CRUD. Occupying an email or group that already carries a grant, or writing a ClusterAuthorizationRule, is user-role assignment.

Assignment is allowed if either:

- the requester already covers the target role rules (conservative port of Kubernetes `rbac.Covers`), or
- every target role sits inside a labeled can-assign range on a **platform-owned** ClusterRole the requester holds (`cluster-admin`, `user-authz:*`, `d8:*` except `d8:custom:*`).

Custom ClusterRoles are cover-only. Forged `can-assign-*` labels on custom names are ignored. The same hook denies CREATE/UPDATE that change `user-authz.deckhouse.io/can-assign-*` unless the requester is an exempt controller (`system:apiserver`, `d8-system:deckhouse`, aggregation controller, `dhctl`, `system:masters`, and the rest of the matchConditions list). Native `escalate` does not protect labels.

This PR admits:

- `users` CREATE/UPDATE/DELETE
- `groups` CREATE/UPDATE/DELETE
- `clusterauthorizationrules` CREATE/UPDATE
- `clusterroles` CREATE/UPDATE (label integrity only)

Range labels on this release (dual-model: basic lattice plus rbacv2):

| ClusterRole | basic-max | rbacv2 range |
|---|---|---|
| `d8:manage:security:manager` | ClusterAdmin | subsystem / security / admin |
| `d8:manage:all:manager` | ClusterAdmin | system / admin |
| `d8:manage:permission:module:user-authz:edit` | ClusterAdmin | — |
| `user-authz:cluster-admin` | ClusterAdmin | — |
| `user-authz:super-admin` | SuperAdmin | system / superadmin |

`d8:manage:permission:module:user-authn:edit` has no can-assign labels.

The webhook-handler ClusterRole now lists and watches `clusterrolebindings` so the kernel can see CRB-only grants (a security manager bound only by ClusterRoleBinding is visible).

Authorization rules bind to the email or group string, not to the User or Group object. Rank-by-role-name missed ClusterRoleBinding-only security managers and could not express a bind range without a hardcoded dictionary.

This is the same problem Kubernetes already solves for Role / ClusterRole / RoleBinding with `escalate` and `bind` (`Covers` on PolicyRules). Those checks never run on `deckhouse.io` CRDs, and they ignore ClusterRole labels. Writing a User for an occupied email, or a ClusterAuthorizationRule that Helm later turns into ClusterRoleBindings as the Deckhouse service account, is not a binding update, so native `bind` never evaluates the human requester.

The kernel is a URA-style can-assign relation (Sandhu–Bhamidipati–Munawer ARBAC97) plus the Kubernetes cover path: a security subsystem manager can assign basic access up to ClusterAdmin and security roles up to the named admin level, and cannot assign SuperAdmin, `cluster-admin`, or an rbacv2 `superadmin` role. The `user-authn` edit permission alone cannot take over an occupied SuperAdmin identity. The collision annotation from #22218 does not bypass can-assign.

Permission-role and role-role administration stay platform-owned.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: feature
summary: Occupying an identity that already carries a grant, or writing a ClusterAuthorizationRule, now requires covering permissions or an explicit can-assign range.
impact: |
  A subject who can manage Users or Groups cannot delete or recreate an identity
  whose email or group name already carries roles they cannot assign. The same
  subject cannot create or update a ClusterAuthorizationRule whose accessLevel
  or additionalRoles sit outside their can-assign range. A security subsystem
  manager can still assign ClusterAdmin and cannot assign SuperAdmin.
  The collision annotation does not bypass this check. GitOps ServiceAccounts
  that are not exempt cannot create a User after a SuperAdmin CAR for that email.
impact_level: high
```


modules/140-user-authz/webhooks/validating/identity_assign.py
```diff
@@ -0,0 +1,741 @@
...
+    if not targets:
+        return None
+
+    rng = actor_range(actor_role_names, catalog)
+    missing = [n for n in targets if n not in catalog]
+
+    if not missing:
+        actor_rules = union_rules(actor_role_names, catalog)
+        target_rules = union_rules(targets, catalog)
+        if covers(actor_rules, target_rules):
```
AlwxSin: 🔴 **1. An emptied target ClusterRole satisfies `covers()`, so a security manager can assign SuperAdmin**

`covers()` itself is a faithful port — an empty servant rule set being covered is exactly upstream `rbac.Covers` semantics. The defect is the ordering here: `can_assign` reaches this cover path and returns `None` (allow) on the next line, before `role_in_range` runs, so `is_disaster` — the check that refuses `SuperAdmin` and `cluster-admin` — is never consulted.

The rules being compared are read from the live ClusterRole snapshot, which the actor this webhook polices can write:

- `d8:manage:permission:subsystem:security:manage_resources` (`modules/140-user-authz/templates/rbacv2/global/manage/permissions/manage_security_resources.yaml:24-34`) grants `create/update/patch/delete` on `clusterroles`, and aggregates into `d8:manage:security:manager`.
- Removing rules never trips the native escalation check — it only confirms the *new* rules are covered.
- The `system-roles.deckhouse.io` webhook guards `heritage: deckhouse` ClusterRoles on CREATE only.
- The new `clusterroles` rule in this PR guards `can-assign-*` labels, not `.rules`.

So a `security` subsystem manager can run

```
kubectl patch clusterrole user-authz:super-admin --type=json \
  -p '[{"op":"replace","path":"/rules","value":[]}]'
```

and then create a `ClusterAuthorizationRule` with `accessLevel: SuperAdmin` for their own email. It is admitted, and Deckhouse reconciles the ClusterRole back to the real SuperAdmin rules. Same trick works for `cluster-admin`.

Verified against this head: `can_assign(["d8:manage:security:manager"], ["user-authz:super-admin"], catalog)` returns `["user-authz:super-admin"]` normally, and `None` once that catalog entry's rules are `[]`. `identity_assign_test.py:87` (`test_empty_target_is_covered`) currently locks the behaviour in as expected.

Related, same root cause: `basic_level_of` (line 394) maps a role into the basic lattice from the `user-authz.deckhouse.io/access-level` annotation on any ClusterRole. Unlike the `can-assign-*` labels, that annotation is not integrity-protected by `validate_clusterrole`.

Suggested direction: evaluate the cover path per target, refuse it for a target whose rule set is empty, and run the `is_disaster` check before the cover short-circuit rather than only inside `role_in_range`.

modules/140-user-authz/webhooks/validating/identity_assign.py
```diff
@@ -0,0 +1,741 @@
...
+
+    found: List[str] = []
+    seen: Set[str] = set()
+
+    def add(role: str) -> None:
+        if role and role not in seen:
+            seen.add(role)
+            found.append(role)
+
+    for snap_name, namespaced in ((CAR_SNAP, False), (AR_SNAP, True)):
```
AlwxSin: 🟡 **2. A namespaced AuthorizationRule counts as a cluster-wide role for the requester**

`AuthorizationRule` is `scope: Namespaced` (`modules/140-user-authz/crds/authorizationrule.yaml:11`), but this loop folds AR-derived role names into the same flat list as the CAR and ClusterRoleBinding ones, with no namespace attached. `actor_roles` (line 291) then hands that list to both `covers()` and `actor_range`, so a user who is `Admin` in exactly one namespace via an AR is treated as holding `user-authz:admin` cluster-wide.

`AR_ACCESS_LEVEL_CAP = "Admin"` caps how high this reaches, but not the scope inflation itself. Verified: `actor_roles({"username": "eve@corp"}, {AR_SNAP: [<Admin in team-foo>]})` returns `["user-authz:admin"]`.

The `additionalRoles` branch in `_fr_role_names` is dead for AR today — the CRD has no such field — so `Admin` is the ceiling. It stops being one the moment AR grows `additionalRoles`.

modules/140-user-authz/docs/README.md
```diff
@@ -75,9 +75,7 @@ The manage role grants access only to system namespaces (starting with `d8-` or
...
-
-A subject that can manage the `user-authn` module can register an identity provider or reset the credentials of any local user. A subject that can manage the `user-authz` module can modify authorization rules. In both cases, they can obtain an identity with any privileges, including cluster administrator, so treat a `security` subsystem manage role as a cluster administrator role when planning access.
+A manage role limits which modules and namespaces a subject can address. Creating a local user or group whose name is not already a grant subject remains ordinary CRUD. Occupying an email or group that already carries a grant, or writing a ClusterAuthorizationRule, is an assignment: the requester must already hold covering permissions or an explicit can-assign range on the target roles. A `security` subsystem manager can assign basic access up to `ClusterAdmin` and security roles up to the named admin level, and cannot assign SuperAdmin or `cluster-admin`. The `user-authn` edit permission alone does not assign occupied SuperAdmin identities.
```
AlwxSin: 🟡 **3. The "treat a security manage role as cluster-admin" warning is removed while the gate is still bypassable**

The removed alert was the operator-facing statement of exactly the property finding 1 shows is still true (same edit in `README_RU.md:77`). Until the cover path in `identity_assign.py` is closed, dropping it overstates what the gate delivers — worth restoring alongside the fix rather than ahead of it.

modules/140-user-authz/docs/README_RU.md
```diff
@@ -74,9 +74,7 @@ Manage-роль определяет доступ только к системн
...
-
-Субъект, который может управлять модулем `user-authn`, способен зарегистрировать провайдер идентификации или сбросить учётные данные любого локального пользователя. Субъект, который может управлять модулем `user-authz`, способен изменить правила авторизации. В обоих случаях он может получить идентичность с любыми привилегиями, включая администратора кластера, поэтому при планировании доступа считайте manage-роль подсистемы `security` равной роли администратора кластера.
+Manage-роль ограничивает, к каким модулям и неймспейсам обращается субъект. Создать локального пользователя или группу с именем, которое ещё не является субъектом гранта, — обычный CRUD. Занять email или группу, на которых уже висит грант, или записать ClusterAuthorizationRule — это назначение: у запрашивающего должны быть покрывающие права либо явный диапазон can-assign на целевые роли. Менеджер подсистемы `security` может назначать basic-доступ до `ClusterAdmin` и роли security до названого admin и не может назначить SuperAdmin или `cluster-admin`. Одного права `user-authn` edit недостаточно, чтобы занять уже существующую SuperAdmin-идентичность.
```
z9r5: basic-доступ не раскрыт вроде нигде тут выше. Можем пояснить?

z9r5:
```suggestion
Manage-роль ограничивает, к каким модулям и неймспейсам обращается субъект. Создать локального пользователя или группу с именем, которое ещё не является субъектом гранта, — обычный CRUD. Занять email или группу, на которых уже висит грант, или записать ClusterAuthorizationRule — это назначение: у запрашивающего должны быть покрывающие права либо явный диапазон can-assign на целевые роли. Менеджер подсистемы `security` может назначать basic-доступ до `ClusterAdmin` и роли security до названого admin и не может назначить  уровень доступа SuperAdmin или ClusterAdmin. Одного права `user-authn` edit недостаточно, чтобы занять уже существующую SuperAdmin-идентичность.
```
basic-доступ не раскрыт вроде нигде тут выше. Можем пояснить?

> Одного права `user-authn` edit недостаточно
>
> занять уже существующую SuperAdmin-идентичность

Не понятно. Можем переформулировать?

z9r5 (commented):
Comments

modules/140-user-authz/webhooks/validating/identity_assign.py
Jabejixo: `can_assign` now evaluates cover per target and only if that role still has a non-empty rule set; an emptied SuperAdmin/`cluster-admin` falls through to range and is refused. The `access-level` annotation is honored only on platform-owned ClusterRole names.

modules/140-user-authz/webhooks/validating/identity_assign.py
Jabejixo: The requester’s roles no longer include namespaced AuthorizationRules, so an Admin AR in one namespace is not treated as cluster-wide `user-authz:admin`. Forged `additionalRoles` on an AR snapshot are ignored; AR still counts when the target identity already holds that grant.

modules/140-user-authz/docs/README.md
Jabejixo: Finding 1 is closed, so the old “security is cluster-admin” warning stayed gone. The alert now states the actual ceiling: CAR `accessLevel` up to ClusterAdmin, not SuperAdmin or Kubernetes `cluster-admin`.

modules/140-user-authz/docs/README_RU.md
Jabejixo: basic — это `spec.accessLevel` у ClusterAuthorizationRule (`User` … `SuperAdmin`), не уровни manage-ролей `viewer`/`manager`. В warning это теперь написано отдельно.

Jabejixo: Абзац про user-authn переформулировал: создать User/Group недостаточно, если на этом email уже висят роли — это раздача ролей, а не обычный CRUD. Security по-прежнему может CAR до ClusterAdmin и не может SuperAdmin и `cluster-admin`.

z9r5: ИМХО чуть более понятно стало, но для нас. Для пользователя сложновато. Аппрув чтобы не блочить. Но хорошо бы потом порефлексировать в комплексе.

z9r5 (approved):

modules/140-user-authz/webhooks/validating/identity_assign.py
AlwxSin: Re-checked at `20c813eff6` — the reported instance is closed, the class is not.

`rules: []` now fails `_has_coverable_rules` and falls through to `role_in_range`, so `is_disaster` refuses it, and both new tests pin exactly `rules=[]`. But the ordering is unchanged: when cover passes, `continue` skips `role_in_range`, so `is_disaster` is still never consulted. Emptying the role was only the cheapest way to get there — it is enough to rewrite its rules to something the requester already holds:

```
kubectl patch clusterrole user-authz:super-admin --type=json -p '[{"op":"replace","path":"/rules","value":[
  {"apiGroups":["rbac.authorization.k8s.io"],"resources":["clusterroles"],"verbs":["update"]}]}]'
```

That rule comes from `manage_security_resources.yaml`, so a `security` subsystem manager holds it by definition of the role. `_has_coverable_rules` sees one atom, `covers(actor_rules, entry.rules)` passes, `SuperAdmin` becomes assignable, and Deckhouse reconciles the real rules back afterwards.

Verified at this head, actor `d8:manage:security:manager`:

| target `user-authz:super-admin` rules | verdict |
|---|---|
| real (`STAR_ALL`) | denied |
| `[]` | denied (this fix) |
| one rule the actor holds | **allowed** |
| `cluster-admin` with one rule the actor holds | **allowed** |

Closing the class rather than the instance means running the disaster check before the cover short-circuit: for a name in `DISASTER_NAMES`, or an rbacv2 `superadmin` level, require the range whatever the live rule set says. Then no ClusterRole write the requester is allowed to make can turn `SuperAdmin` or `cluster-admin` into something assignable.

The `access-level` part of the original comment is genuinely fixed — a custom `pwn` carrying `access-level: User` with star rules is denied now. A `d8:`-prefixed name with a forged annotation still maps into the basic lattice, but writing one needs `escalate`, so it is not a separate door.

modules/140-user-authz/webhooks/validating/identity_assign.py
```diff
@@ -0,0 +1,763 @@
...
+    targets = [n for n in target_role_names if isinstance(n, str) and n]
+    if not targets:
+        return None
+
+    rng = actor_range(actor_role_names, catalog)
+    actor_rules = union_rules(actor_role_names, catalog)
+    leftover = []
+    for name in targets:
+        entry = catalog.get(name)
+        if entry is None:
```
AlwxSin, line 622: 🔵 **12. A target role missing from the ClusterRole snapshot is now unconditionally denied**

This replaced the old `role_in_range(name, None, rng)` fallback for a target that is not in the catalog. A `d8:`-named role whose ClusterRole is not present — module not enabled yet, or a ClusterAuthorizationRule applied ahead of the role it names — used to be admitted when it sat inside the requester's can-assign range, and is now always refused.

Verified at `20c813eff6`: `can_assign(["d8:manage:security:manager"], ["d8:manage:security:user"], <catalog without that entry>)` returns `["d8:manage:security:user"]`.

Fail-closed, and arguably the more correct answer — you cannot know what you are granting. But it is a behaviour change that the commit message does not mention and that no test covers, so worth confirming it is intended rather than a side effect of moving the cover check per target.

modules/140-user-authz/webhooks/validating/identity_assign.py
Jabejixo: Disaster names now require the SuperAdmin range before cover runs. Shrinking SuperAdmin or `cluster-admin` to a rule the security manager already holds is leftover. Empty rules stay leftover as before.

modules/140-user-authz/webhooks/validating/identity_assign.py
Jabejixo: Intended fail-closed: a role missing from the ClusterRole snapshot is not assigned by range on the name alone. Added a test for `d8:manage:security:user` absent from the catalog; the same name is admitted once the ClusterRole is present.

ldmonster (approved):

AlwxSin (approved):

PR #22672, opened 2026-08-31
fix(registrypackages): close CVEs in cfssl, docker-registry, d8-bcheck, rpp-get and yq

090809:
## Description

Four images carried vulnerable transitive Go dependencies:

| Image | Change | CVEs closed |
|---|---|---|
| `cfssl` | new `patches/001-go-mod.patch` | 34 |
| `docker-registry` (`auth_server`) | regenerated `patches/docker_auth/001-go-mod.patch` | 24 + GHSA-hrxh-6v49-42gf |
| `d8-bcheck` | `src/go.mod` | 24 |
| `rpp-get` | `src/go.mod` | 17 |

All four now resolve `golang.org/x/crypto v0.55.0`, `x/net v0.58.0`, `x/sys v0.47.0` and `x/text v0.41.0`, plus `go.mongodb.org/mongo-driver v1.17.9`, `google.golang.org/grpc v1.83.2`, `go-jose/v4 v4.1.4` and `google.golang.org/protobuf v1.36.12` where those apply. Versions sit above the minimum fix line where a higher release existed: `x/net v0.56.0` only requires `x/text v0.38.0`, one short of the `v0.39.0` that CVE-2026-56852 needs. The `go` directive stays at `1.25.0` in every module, within the `golang:1.25.13` builder.

`cfssl` needed two extra changes. It had no `patches/` directory, so `werf.inc.yaml` gains the git stanza and the `git apply` step, in the same shape `docker-registry` already uses. It also builds through a `Makefile` that sets `GOFLAGS=-mod=vendor` over a 32 MB vendored tree, so a go.mod-only patch would leave `make` compiling the old sources. The build now runs `go mod vendor` before `make`.

`yq` moves from a prebuilt digest to `pm`. It used to import the `tools/yq-v4.47.1` binary by digest, which meant waiting for a base-images rebuild to get a newer yq. It now runs `pm install yq@4.53.6`, matching how `xfsprogs` and other packages here are built. The `pm` index carries yq 4.53.6, which scans clean. The version pin is deliberate: `pm` holds a single version per package, so the build fails loudly if the index moves ahead of `oss.yaml`, rather than shipping an image whose label misstates its contents. This renames the image from `yq4471` to `yq4536`, so the three bashible references and the generated test fixture move with it.

`GO-2026-5932` cannot be closed by any bump. It is the standing advisory that `golang.org/x/crypto/openpgp` is unmaintained, and it has no fixed version. Three `known_vulnerabilities.vex` files mark it `not_affected` with justification `vulnerable_code_not_in_execute_path`. That claim was checked per component rather than assumed: `openpgp` is absent from `vendor/modules.txt` in cfssl, and absent from `go list -deps` output for `auth_server` and `d8-bcheck`. Each `impact_statement` names the x/crypto subpackages that binary actually resolves. This follows the existing waivers for the same advisory in `istio/api-proxy`, `admission-policy-engine/ratify` and others.

`candi/base_images.yml` moves from v0.5.80 to v0.5.85. The Go builders go from 1.25.10 to 1.25.13, which clears the Go stdlib CVEs in the prebuilt `cosign` and `yq` binaries. The update rewrites the whole file, so it also touches the python, ruby and shell-operator base images.

The root `go.sum` gains 240 lines. They are all `/go.mod` graph-metadata entries added by `make generate`; `go.mod` is unchanged and nothing was removed.

No impact on running components. These are node-level registry packages and build-time changes, with no control-plane, ingress or Prometheus restart. On update, nodes fetch the renamed `yq` package and reinstall a static binary into `/opt/deckhouse/bin`. No daemon restarts.

### Verification

Every component was scanned with Trivy before and after, gated on each targeted CVE being absent and on no new CVE appearing against the pre-fix baseline. All four now report zero findings with their VEX applied. `GO-2026-5932` was confirmed to go from present to suppressed on each of the three waived components.

Builds were run rather than assumed:

* `cfssl`: real `make` on the patched pristine clone produced all 8 binaries, and `go version -m bin/cfssl` confirms the bumped dependencies reached the artifact.
* `auth_server`: the `linux/amd64` build links, and the `mongo-driver` v1.10.2 to v1.17.9 jump compiles without source changes.
* `rpp-get`: `go build`, `go vet` and all three test packages pass.
* `d8-bcheck`: the full non-stdlib dependency closure compiles for `linux/amd64`. The host build on darwin fails, but identically before and after this change, because `delivery-kit-sdk` declares `Verify(ctx, rootCert string, path string)` in its `!linux || !cgo` stub and `Verify(ctx, rootCertRefs []string, path string)` in `file_linux.go`. werf builds this with `GOOS=linux CGO_ENABLED=1`, which selects the working variant.
* Regenerated patches were reapplied to freshly cloned upstream trees to confirm they still apply in order.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: registrypackages
type: fix
summary: Close CVEs in `cfssl`, `docker-registry`, `d8-bcheck` and `rpp-get` by bumping vulnerable Go dependencies, and build `yq` from `pm` at 4.53.6.
impact_level: default
```

```changes
section: candi
type: fix
summary: Bump base images to v0.5.85, moving the Go builders to 1.25.13 and clearing Go stdlib CVEs in prebuilt images.
impact_level: default
```


RomanenkoDenys (approved):

borg-z (approved):

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 35ea19e5ef433cab1e041072417f40cddb929a3d to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/33389790335) for details.

PR #22694, opened 2026-08-31
[user-authz] Admit UserOperation and DexProvider through can-assign

Jabejixo:
## Description

Stacked on #22667. Same can-assign kernel, more admission doors, and User membership from `Group.spec.members` including nested groups.

**UserOperation CREATE** (ResetPassword, Lock, Unlock, Reset2FA) on an occupied identity is allowed only if the requester can assign the roles that identity already carries. The target email comes from `spec.user` via the User snapshot, or from `spec.target.email`. Group membership is taken from `Group.spec.members` (User and Group members, ancestors included), not from the deprecated `User.spec.groups`. `initiatorType: self` and `initiatorType: system` do not bypass the check.

**DexProvider CREATE/UPDATE** is measured against the identity space the provider can assert. A provider asserts an email and a set of groups; the Kubernetes username is that email, so the provider reaches every grant already hanging on the identities it can assert.

- The space is bounded on two axes by the new `spec.allowedIdentities` block: `emails` / `emailDomains` bound the email, `groups` bounds the groups claim. Provider-specific group filters count as well where dex demonstrably narrows the claim: `oidc.allowedGroups`, `gitlab.groups`, `crowd.groups`, `bitbucketCloud.teams` without `includeTeamGroups`, `github.orgs[].teams` on every org, `saml.allowedGroups` with `filterGroups: true`.
- An open axis reaches every grant on subjects of that kind. SuperAdmin is granted to a User in every cluster, so a provider without an email limit needs a SuperAdmin-range requester. With both axes closed only the roles granted to the listed identities count, so a ClusterAdmin or a security-subsystem manager connects a provider for a contractor domain and group on their own.
- Group nesting from Group objects is not applied to the list: the platform expands it for local Users only, an external token carries exactly the groups the IdP asserted.
- Rotating `clientSecret`/`bindPW`, changing `displayName` or `enabled`, narrowing the limits and re-applying the same manifest (any list order) pass without a check. Anything else (IdP address, claim mapping, verification switches, widening) is measured as a fresh connection. DELETE is not gated.
- Dex enforces the same limits at sign-in (patch `020-connector-identity-filters`): email outside the lists is refused, the groups claim is reduced to the intersection, an empty intersection is refused. Applied at login, password grant, token exchange and refresh, for every connector type, on top of the client filters.

**Levels are trusted only on platform roles.** The kernel reads a level from a role name (`d8:manage:<subsystem>:admin`), from the access-level annotation and from can-assign labels. Anyone able to create a ClusterRole can pick such a name, and an empty role passes the apiserver's escalation check when created or bound. Levels, annotations and ranges are now read only from roles labelled `heritage: deckhouse`; a non-exempt write of that label on a platform-shaped name is refused, and `system-roles.deckhouse.io` covers UPDATE so the label cannot be patched on after creation.

**User CUD** follows the same membership walk, so deleting a SuperAdmin who is only a nested group member and has no `spec.groups` is no longer ordinary CRUD.

**ClusterAuthorizationRule DELETE** uses the same can-assign check as CREATE/UPDATE. A grant can be removed only if the requester could assign those roles. Platform hooks do not delete ClusterAuthorizationRule objects (they only drop the rendered ClusterRoleBinding). Exempt controllers (`d8-system:deckhouse`, `dhctl`, `system:masters`, and the rest of the matchConditions list) still pass.

Wave 1 closed delete+recreate of an occupied User and minting a SuperAdmin ClusterAuthorizationRule. Helpdesk could still reset that password through UserOperation, connect an IdP that asserts the SuperAdmin identity, delete a SuperAdmin ClusterAuthorizationRule they could not create, or delete a User who is SuperAdmin only through a parent Group. Those writes are the same user-role assignment: the username is the Dex email, so touching that string is assignment.

The first DexProvider gate in this PR compared the provider against every existing grant. Without an email limit that always resolved to SuperAdmin, so on every cluster only a SuperAdmin could create or reconfigure a provider and the user-authn subsystem stopped being manageable by its own admins. `spec.allowedIdentities` gives the provider a bounded space to be measured against and gives dex the same bound to enforce.

Native Kubernetes `escalate` / `bind` still do not see these CRDs. The follow-up uses the same cover-or-range relation as #22667.

This PR targets `fix/identity-privilege-gate` so it merges into #22667, not into `main` on its own.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: feature
summary: Resetting an occupied identity, deleting its ClusterAuthorizationRule, or connecting a DexProvider that can assert it now requires a can-assign range; levels are trusted only on platform-labelled roles.
impact: |
  A subject who can create UserOperations cannot reset, lock, or unlock an
  identity whose email or group membership already carries roles they cannot
  assign. Nested Group membership counts. Deleting a ClusterAuthorizationRule
  is allowed only if the requester could assign those roles. Creating or
  reconnecting a DexProvider without spec.allowedIdentities on a cluster that
  already has a SuperAdmin grant requires SuperAdmin; ClusterAdmin and the
  security subsystem manager keep rotating secrets, toggling, renaming,
  narrowing and re-applying existing providers, and connect providers whose
  allowedIdentities reach only roles they can assign. GitOps that adds an
  unlimited provider or removes a SuperAdmin rule must run as a SuperAdmin.
impact_level: high
```

```changes
section: user-authn
type: feature
summary: DexProvider gets spec.allowedIdentities (emails, emailDomains, groups); dex refuses sign-in outside the lists and reduces the groups claim to the intersection, for every provider type.
impact: |
  Opt-in. A provider without the block behaves as before. With the block set,
  users outside the lists cannot sign in through that provider and their
  existing refresh tokens are revoked on the next refresh.
impact_level: low
```


deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:fix/identity-assign-wave2` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/33496913999) in 36m42s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.4.41`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 35m7s.

</details>

modules/140-user-authz/webhooks/validating/identity_assign.py
```diff
@@ -310,6 +314,208 @@ def add_all(roles: Iterable[str]) -> None:
...
+        org = _dict(org)
+        org_name = org.get("name")
+        teams = [t for t in _list(org.get("teams")) if isinstance(t, str) and t]
+        if isinstance(org_name, str) and org_name and teams:
+            for team in teams:
+                add_group(f"{org_name}:{team}")
+    return emails, groups
+
+
+def dex_claims_closed(spec: Any) -> bool:
```
AlwxSin, line 500: 🔴 **4. "Closed claims" bound the groups claim only — a closed DexProvider can still assert any email**

For `SAML` with `filterGroups: true` + `allowedGroups`, and for `Github` with every org listing teams, this returns `True`, and `dex_target_roles` (line 498) then checks only the listed group strings. Nothing constrains the **email** claim — and the Kubernetes username *is* the email claim, which is the premise the whole PR rests on.

`dex_explicit_identities` (line 450) tries to collect `allowedEmails`, but `DexProvider` has no such field: `grep -rn allowedEmails modules/150-user-authn/crds/` hits only `dex-client.yaml` and `dex-authenticator.yaml`. So the email list is always empty and the closed path never looks at an email.

A requester with a `ClusterAdmin` range can therefore create

```yaml
type: SAML
saml:
  ssoURL: https://their-idp
  emailAttr: email
  filterGroups: true
  allowedGroups: ["harmless"]
```

`dex_target_roles` returns `[]`, the provider is admitted, and their IdP then asserts `email: root@corp` — which carries a SuperAdmin ClusterAuthorizationRule. The conservative OIDC/LDAP path denies exactly this case; the "closed" path does not.

Verified against this head with one SuperAdmin CAR in the snapshot: OIDC → `["user-authz:super-admin"]` → denied; SAML-closed and Github-closed → `[]` → allowed.

Suggested direction: closed claims should require the email claim to be bounded too, which the DexProvider API cannot express today — so either treat every connector as open claims, or check the occupied email grants regardless of group filtering. The docs (`modules/140-user-authz/docs/README.md`, `modules/150-user-authn/docs/USAGE.md`) currently promise the closed-claims behaviour and need the same edit.

modules/140-user-authz/webhooks/validating/identity_privilege.py
```diff
@@ -207,13 +237,14 @@ def validate_group(req, snapshots, actor: List[str], catalog: dict) -> Optional[
 
 
 def validate_car(req, actor: List[str], catalog: dict) -> Optional[str]:
-    spec = _spec(req.object)
+    obj = req.oldObject if req.operation == "DELETE" else req.object
```
AlwxSin: 🟡 **5. CAR `UPDATE` is checked against the new object only, so it bypasses the DELETE gate**

The PR body states "A grant can be removed only if the requester could assign those roles", and DELETE now honours that. UPDATE does not: `obj` is `req.oldObject` only for DELETE, so rewriting a SuperAdmin CAR down to `accessLevel: User` — or repointing its subjects — is evaluated solely against the new, in-range spec. That is the same grant removal the DELETE rule was added to gate.

Verified against this head with the same actor and the same CAR: `DELETE` → denied, `UPDATE accessLevel: SuperAdmin → User` → allowed.

Fix is small: on UPDATE, union `car_target_roles(req.oldObject)` with `car_target_roles(req.object)` instead of choosing one.

modules/140-user-authz/webhooks/validating/identity_privilege.py
```diff
@@ -222,5 +253,47 @@ def validate_clusterrole(req) -> Optional[str]:
...
+        return None
+    leftover = assign.can_assign(actor, targets, catalog)
+    if leftover is None:
+        return None
+    rng = assign.actor_range(actor, catalog)
+    display = email or user_name or _meta_name(req.object) or "obj"
+    return assign.deny_uo_message(display, leftover, rng)
+
+
+def validate_dexprovider(req, snapshots, actor: List[str], catalog: dict) -> Optional[str]:
```
AlwxSin, line 301: 🟡 **6. Every DexProvider UPDATE re-checks the full occupied-grant set, with no diff awareness**

`dexproviders` is registered for CREATE **and** UPDATE, and for open claims `dex_target_roles` returns `occupied_grant_roles(snapshots)` — every role hanging on any human subject in the cluster. On a cluster that has a single SuperAdmin grant, rotating a client secret or adding a redirect URI on a long-standing OIDC provider therefore needs a SuperAdmin-range requester.

Nothing distinguishes an update that widens the claim surface (adding a connector block, dropping `filterGroups`, extending `allowedGroups`) from one that does not. Gating on the delta of the asserted-identity surface would keep the property and leave routine provider maintenance to whoever owns the module.

modules/140-user-authz/webhooks/validating/identity_assign.py
```diff
@@ -310,6 +314,208 @@ def add_all(roles: Iterable[str]) -> None:
     return found
 
 
+def is_human_identity(name: str) -> bool:
+    if not isinstance(name, str) or not name:
+        return False
+    return not name.startswith("system:")
+
+
+def user_record(snapshots: Any, *, name: str = "", email: str = "") -> Optional[dict]:
```
AlwxSin, line 333: 🔵 **7. `user_record` returns the first record matching *either* key**

With both `spec.user` and `spec.target.email` set and disagreeing, this loop returns whichever record matches first on either key, so `membership_groups` can add a different user's groups. The effect is over-inclusive — more target roles means a denial — so it fails closed, but the lookup is not what the name says. Requiring the name match when a name is given, and only falling back to the email, would match the intent.

modules/140-user-authz/webhooks/validating/identity_assign.py
Jabejixo: A closed DexProvider still treats the email claim as unbounded, so we now also take every User-subject grant, not only the listed groups. A SuperAdmin grant that exists only on an unlisted group stays out of the target set.

modules/140-user-authz/webhooks/validating/identity_privilege.py
Jabejixo: CAR UPDATE now unions the old and new specs, so lowering SuperAdmin to User is the same check as DELETE.

modules/140-user-authz/webhooks/validating/identity_privilege.py
Jabejixo: Dex UPDATE is checked only on `new − old`; rotating a secret without new target roles no longer requires SuperAdmin.

modules/140-user-authz/webhooks/validating/identity_assign.py
Jabejixo: If a user name is set, `user_record` matches that name only and does not fall back to email.

modules/140-user-authz/webhooks/validating/identity_privilege.py
```diff
@@ -222,5 +266,51 @@ def validate_clusterrole(req) -> Optional[str]:
...
+    new_targets = assign.dex_target_roles(_spec(req.object), snapshots)
+    if req.operation == "UPDATE":
+        old_targets = set(assign.dex_target_roles(_spec(req.oldObject), snapshots))
+        targets = [name for name in new_targets if name not in old_targets]
```
AlwxSin: 🔴 **11. The UPDATE delta lets an existing DexProvider be repointed at another IdP**

Gating UPDATE on `new_targets − old_targets` fixes the secret-rotation case from finding 6. But for a provider that is *already* open claims, `old_targets` is `occupied_grant_roles(snapshots)` — every grant in the cluster — so the delta is always empty and **every** UPDATE is admitted.

Repointing `oidc.issuer` or `saml.ssoURL` at an attacker-controlled IdP does not grow the set of assertable identities; it changes who signs the assertions, and that is the escalation. Before this commit that UPDATE was denied.

Verified at `4fbf3623ed`, actor `d8:manage:security:manager`, one SuperAdmin ClusterAuthorizationRule in the snapshot:

| operation | verdict |
|---|---|
| OIDC CREATE | denied |
| OIDC UPDATE, `clientSecret` rotated | allowed (the intended fix) |
| OIDC UPDATE, `issuer` → attacker IdP | **allowed** |
| SAML `filterGroups: true` UPDATE, `ssoURL` → attacker IdP | **allowed** |

The delta needs to cover the trust anchor, not only the target set: treat a change to `type`, `issuer` / `ssoURL`, `clientID`, `clientSecret` or the CA bundle as a fresh CREATE and check `new_targets` in full, keeping the subtraction for edits that leave the anchor alone.

Jabejixo: UPDATE still uses `new − old` when the trust anchor is unchanged, so rotating `clientSecret` or renaming the provider stays allowed. Changing `type`, issuer / `ssoURL` / `ssoIssuer`, `clientID`, the CA bundle, or the LDAP/GitLab host is checked as CREATE. `clientSecret` is not part of the anchor — that would undo finding 6, and the table in this comment still wants secret rotation allowed.

modules/140-user-authz/webhooks/validating/identity_assign.py
```diff
@@ -320,6 +324,250 @@ def add_all(roles: Iterable[str]) -> None:
...
+        return all(bool(_list(o.get("teams"))) for o in orgs)
+    return False
+
+
+DEX_PROVIDER_BLOCKS = ("oidc", "saml", "ldap", "crowd", "gitlab", "github", "bitbucketCloud")
+
+# Who signs, or which app at that IdP is trusted. clientSecret and bindPW
+# stay off this list so rotating a credential at the same issuer is not a
+# new connection.
+DEX_TRUST_FIELDS = (
```
AlwxSin: 🔴 **13. The trust anchor covers who signs, not whether the signature is checked**

The issuer/`ssoURL`/`type` part of finding 11 is closed — verified. But this list is endpoints and credentials only. It does not carry the fields that decide **whether an assertion is verified at all**, or **which claim becomes the Kubernetes username** — and that username is the premise the whole gate rests on. Those UPDATEs leave `dex_trust_anchor` equal, fall back to `new − old` (empty for an already-open provider; for a closed one every User-subject grant is already in `old_targets`), and are admitted.

Verified at `66f3c97c3e`, actor `d8:manage:security:manager`, one SuperAdmin ClusterAuthorizationRule on `root@corp`:

| UPDATE on an existing provider | anchor | verdict |
|---|---|---|
| `saml.insecureSkipSignatureValidation` false → true | same | **allowed** |
| `oidc.claimMapping.email: nickname` + `claimMappingOverride: true` | same | **allowed** |
| `oidc.insecureSkipEmailVerified` → true | same | **allowed** |
| `oidc.userNameKey` → another claim | same | **allowed** |
| `saml.emailAttr` → another attribute | same | **allowed** |

The sharpest is `saml.insecureSkipSignatureValidation` — `modules/150-user-authn/crds/dex-provider.yaml:632-636`, "Skip signature validation of SAML responses. This is insecure and should only be used for testing". Flipping it on an already-connected SAML provider makes dex accept an **unsigned** response, so anyone able to POST to the ACS URL forges an assertion for `root@corp` and picks up the SuperAdmin grant. No new endpoint, no new credential — strictly stronger than the issuer swap that finding 11 just closed.

`claimMapping.email` with `claimMappingOverride: true` is the same shape one level down: the username stops coming from the verified `email` claim and starts coming from a claim the end user can usually edit at the IdP (`nickname`, `preferred_username`). `saml.emailAttr` and `oidc.userNameKey` are the same for their connectors.

Two ways out, both fine:

- extend this list with the verification switches (`insecureSkipSignatureValidation`, `insecureSkipEmailVerified`, `insecureSkipVerify`, `insecureNoSSL`, `startTLS`) and the identity-mapping fields (`claimMapping`, `claimMappingOverride`, `userNameKey`, `userIDKey`, `emailAttr`, `usernameAttr`, `groupsAttr`, `userSearch`, `groupSearch`);
- or invert it: treat any connector-block change except `clientSecret` / `bindPW` as a fresh CREATE. Same shape as today, but it fails closed on fields added to the CRD later instead of silently dropping them out of the anchor.

Also checked and correct, so no action: dropping `saml.filterGroups` (closed → open claims) is admitted only because the finding-4 fix already puts every User-subject grant in `old_targets`; a SuperAdmin grant on a *group* subject is still in the delta and refused.

Jabejixo: Took the second option. `dex_trust_anchor` is now the whole connector block with only `clientSecret` and `bindPW` stripped, so a field added to the CRD later is in the anchor by default instead of quietly dropping out of it.

`insecureSkipSignatureValidation`, `insecureSkipEmailVerified`, `claimMapping` / `claimMappingOverride`, `userNameKey`, `emailAttr`, and the search blocks all change the anchor now, so they are checked as CREATE. Credential rotation stays on the delta path for both OIDC and LDAP.

modules/140-user-authz/webhooks/validating/identity_assign.py
```diff
@@ -419,7 +789,7 @@ def is_disaster(name: str, entry: Optional[CatalogEntry]) -> bool:
 
 
 def _range_from_entry(entry: CatalogEntry) -> Optional[AssignRange]:
-    if not is_platform_range_role(entry.name):
+    if not is_platform_range_role(entry.name) or not is_platform_owned(entry):
```
AlwxSin, line 814: 🔴 **14. `role_in_range` reads a level out of a name and an annotation without checking who owns the role**

Real locations: `basic_level_of` (`identity_assign.py:774-786`) and the `describe_role` branch of `role_in_range` (`:844-864`). Commenting here because this is the line that does it right — the *actor's* range became heritage-only in `a0f09182f9`, and the target's level did not get the same treatment.

`system_roles` states the invariant this is supposed to hold to, in its own header comment:

> identity_assign.py reads a level out of a role's name and annotations only for roles carrying it (is_platform_owned), because a platform role name such as `d8:manage:<subsystem>:admin` is otherwise just a string anyone with `create` on clusterroles can pick.

It does not. `basic_level_of` accepts `entry.access_level` — the requester-settable `user-authz.deckhouse.io/access-level` annotation — for any name passing `is_platform_range_role`, and `role_in_range`'s rbacv2 branch derives the level from `describe_role(name, ...)`, i.e. from the name alone. Neither consults `is_platform_owned`. A failed `covers` does not deny either: it falls through to exactly this range check.

Measured at `919b3353e1`, with `apiGroups`/`resources`/`verbs` all `["*"]` on the target in every row:

| ladder | target role | target labels | verdict |
|---|---|---|---|
| rbacv2 | `d8:manage:networking:user` | none | **ALLOWED** |
| rbacv2 | `d8:manage:networking:user` | `heritage: deckhouse` | ALLOWED |
| basic | `d8:x`, annotation `access-level: User` | none | **ALLOWED** |
| basic | `d8:x`, annotation `access-level: User` | `heritage: deckhouse` | ALLOWED |
| control | `some-random-role` | none | denied |
| control | `d8:manage:all:superadmin` | none | denied (`is_disaster`) |
| control | `cluster-admin` | none | denied (`is_disaster`) |

Actor for the rbacv2 rows: one platform-owned role with `can-assign-max-level: admin`, `can-assign-scope: subsystem`, `can-assign-subsystem: networking`. For the basic rows: `can-assign-basic-max: Editor`. The verdict does not move when `heritage: deckhouse` is added or removed — ownership is not consulted on the target path at all. The controls confirm the earlier fixes still hold: a name outside the platform shapes is leftover, and a forged *high* level hits `is_disaster` first and requires the SuperAdmin range.

Why this matters here specifically: this gate exists because a ClusterAuthorizationRule is rendered into a binding by a controller holding `bind`, so the API server's own escalation prevention never sees the requester. On the target path the gate now trusts a name the requester picks and an annotation the requester writes, so any route that lands rules the requester does not hold inside a role the requester names defeats it. The candidate route is an `aggregationRule` with empty `rules` — the create-time escalation check has no rules to object to, and kube-controller-manager fills them in afterwards. I did not confirm on a stand that the API server admits that from a non-`escalate` actor, so treat the route as unconfirmed; the invariant violation above stands on its own.

Fix is the check that already exists on this line: require `is_platform_owned(entry)` before trusting `entry.access_level` in `basic_level_of`, and before the `describe_role` branch in `role_in_range`. Every platform role carries `heritage: deckhouse`, so no legitimate target changes verdict; a self-named role then falls to leftover for everyone below the SuperAdmin range, which is what the `can_assign` docstring already claims happens.

Worth noting no test catches this: `identity_assign_test.py` grew a lot in this push, but nothing asserts a *target* role's ownership.

modules/140-user-authz/webhooks/validating/identity_assign.py
```diff
@@ -341,6 +468,249 @@ def target_group_roles(snapshots: Any, group_name: str) -> List[str]:
...
+# where dex demonstrably narrows the claim to a list (see research R3.1). An
+# axis without a limiter is open: every grant on a human subject of that kind
+# is reachable. Group nesting from Group objects is not applied: the platform
+# expands it for local Users only, an external token carries exactly the
+# groups the IdP asserted (research R8).
+
+DEX_ALLOWED_IDENTITIES = "allowedIdentities"
+
+# Fields whose change neither moves the identity source nor widens the space.
+DEX_BENIGN_TOP_FIELDS = frozenset({"displayName", "enabled"})
```
AlwxSin: 🔵 **15. `enabled` is benign, so a disabled provider can be switched back on by an actor who could not create it**

`enabled: false → true` on a provider with no `allowedIdentities` is admitted with no target check (verified at `919b3353e1`). CREATE is always checked in full, so a wide provider cannot be introduced this way — but disabling a provider is the natural containment action for a suspect IdP, and undoing it is currently free for anyone who can write DexProviders at all.

Either drop `enabled` from the benign set (the re-check costs one snapshot pass), or treat `false → true` as non-benign and leave `true → false` free.

modules/150-user-authn/images/dex/patches/020-connector-identity-filters.patch
```diff
@@ -0,0 +1,403 @@
...
++// applyAccessFilters runs the DexClient filters and then the connector's own
++// limits. Both must pass; the connector is looked up in storage, where static
++// (config file) connectors are visible through WithStaticConnectors.
++func (s *Server) applyAccessFilters(ctx context.Context, client storage.Client, connectorID string, claims *storage.Claims) error {
++	if err := applyClientAccessFilters(client, claims); err != nil {
++		return err
++	}
++	conn, err := s.storage.GetConnector(ctx, connectorID)
++	if err != nil {
++		if errors.Is(err, storage.ErrNotFound) {
```
AlwxSin, line 270: 🔵 **16. `applyAccessFilters` treats a missing connector as "no limits"**

`GetConnector` returning `storage.ErrNotFound` falls through to `return nil`, so the identity is issued unfiltered. Not reachable today — config-file connectors are visible through `WithStaticConnectors`, and the `local` connector is registered with empty allow-lists — but it is the one fail-open in an otherwise fail-closed patch, and a login that reached `finalizeLogin` has a connector by definition, so `ErrNotFound` there means an inconsistency rather than an unrestricted provider. Denying would be the safer default.

modules/140-user-authz/webhooks/validating/identity_privilege.py
AlwxSin: Closed at `919b3353e1` — and closed by construction, not by extending the anchor.

The delta subtraction this finding was about is gone: `validate_dexprovider` now computes `dex_target_roles(new_spec)` in full and requires the actor to cover all of it. `dex_benign_update` replaced the anchor with an exception list — everything except `displayName`, `enabled`, `clientSecret`/`bindPW` and the group-limiter fields is compared for equality, and any difference means a full recompute. So repointing `issuer`/`ssoURL` is no longer admitted, and neither is anything else that moves the identity source.

What makes it hold rather than just cover today's fields: `DexProvider.spec.allowedIdentities` bounds what the connector may assert, dex enforces it on all four flows (`finalizeLogin`, `handlePasswordGrant`, `handleTokenExchange`, `handleRefreshToken`), and it filters `claims.Email` / `claims.Groups` — exactly the two claims `_authentication_configuration.tpl` maps to the Kubernetes username and groups. So who signs the assertion stops being the escalation: whoever signs, they can only assert identities the requester was already allowed to assign roles for.

Probed at `919b3353e1`:

| UPDATE | verdict |
|---|---|
| `clientSecret` rotated | benign, no check |
| `issuer` repointed | full re-check |
| `emailDomains` narrowed | benign, no check |
| `emailDomains` widened, or opened | full re-check |

Resolving. Two smaller things came out of reading the new model and are filed separately as 15 (`enabled` is in the benign set) and 16 (`ErrNotFound` from `GetConnector` skips the filter); neither belongs in this thread.

modules/140-user-authz/webhooks/validating/identity_assign.py
Jabejixo: Verified and fixed in c4b9117c9a. `basic_level_of` and `role_in_range` now require `is_platform_owned(entry)` before reading anything off the target: ladder names, the access-level annotation and the rbacv2 level from the name. A self-made role under a platform name is leftover for everyone below the SuperAdmin range; the disaster check stays name-based since it only ever tightens. Tests follow your table row by row (`TestTargetOwnership`), including the control where the forged high level still hits `is_disaster` first.

On the aggregationRule route: checked on a stand from an identity holding only `create` on clusterroles. kube-apiserver (v1.34) refuses it in the RBAC storage layer, before any admission webhook — `must have cluster-admin privileges to use the aggregationRule` (`pkg/registry/rbac/clusterrole/policybased/storage.go`) — while the same empty role without the aggregationRule is admitted. Independent of the Deckhouse image on the stand. So that route is closed by Kubernetes itself; the open one was the plain role plus a platform name, which this fix covers.

modules/140-user-authz/webhooks/validating/identity_assign.py
Jabejixo: Agreed, the containment argument settles it. `enabled` true→false stays free, false→true is measured as a fresh connection. For a closed provider within the actor's range it comes back on for its owner, so the day-to-day case is unaffected. Spec decision revised, docs and the e2e expectation updated in c4b9117c9a.

modules/150-user-authn/images/dex/patches/020-connector-identity-filters.patch
Jabejixo: Agreed, fail closed. A missing connector in `applyAccessFilters` now returns `NotAllowedError` instead of issuing unfiltered claims. One consequence worth naming: a refresh token whose connector was removed from the config now ends with `access_denied` on the next refresh, which is the right outcome for a removed IdP. Patch regenerated, README updated in c4b9117c9a.

AlwxSin (approved):

modules/140-user-authz/docs/README_RU.md
```diff
@@ -80,7 +80,9 @@ Manage-роль сама по себе не даёт права раздават
 
 Создать User или Group, которых ещё нет среди субъектов грантов, можно как обычный объект.
 
-Создать User на email, на котором грант уже есть, или записать ClusterAuthorizationRule — это раздача ролей. Запрос пройдёт, только если у запрашивающего уже есть покрывающие права или ему явно разрешено назначать эти роли.
+Создать User на email, на котором грант уже есть, записать или удалить ClusterAuthorizationRule либо сбросить такую идентичность через UserOperation — это раздача ролей. Запрос пройдёт, только если у запрашивающего уже есть покрывающие права или ему явно разрешено назначать эти роли.
+
+Подключение DexProvider проверяется так же, но относительно идентичностей, которые провайдер способен утвердить: `spec.allowedIdentities` (email, домены email, группы) и собственные фильтры групп провайдера ограничивают это множество, и учитываются только роли, уже выданные внутри него. Провайдер без ограничителя по email способен утвердить любой адрес, включая те, на которых висит SuperAdmin, поэтому его запись требует SuperAdmin. Подробности и список изменений, проходящих без проверки, — в документации модуля user-authn.
```
denmaloyreb, line 85:
```suggestion
Подключение DexProvider проверяется так же, но относительно идентичностей, которые провайдер способен утвердить: [`spec.allowedIdentities`](/modules/user-authn/cr.html#dexprovider-v1-spec-allowedidentities) (email, домены email, группы) и собственные фильтры групп провайдера ограничивают это множество, и учитываются только роли, уже выданные внутри него. Провайдер без ограничителя по email способен утвердить любой адрес, включая те, accessLevel у которых `SuperAdmin`, поэтому его запись требует `SuperAdmin`. Подробности и список изменений, проходящих без проверки, — в документации модуля user-authn.
```

modules/140-user-authz/docs/USAGE_RU.md
```diff
@@ -259,6 +259,8 @@ spec:
 
 Если совпадение сделано намеренно, например, если правило ClusterAuthorizationRule написано заранее, установите на User или Group аннотацию `user-authz.deckhouse.io/allow-authorization-rule-collision: "true"`. Аннотация только подтверждает совпадение имени. Она не позволяет назначить роли, которые запрашивающий не покрывает или которые вне его диапазона can-assign.
 
+Удаление ClusterAuthorizationRule проходит ту же проверку can-assign, что создание и обновление. Снять грант можно только если эти роли можно назначить. Обновление проверяется и по прежним, и по новым грантам: снизить SuperAdmin до User можно только если SuperAdmin можно назначить.
```
denmaloyreb, line 262:
```suggestion
Удаление ClusterAuthorizationRule проходит ту же проверку can-assign, что создание и обновление. Снять грант можно только если эти роли можно назначить. Обновление проверяется и по прежним, и по новым грантам: снизить `SuperAdmin` до `User` можно только если `SuperAdmin` можно назначить.
```

modules/140-user-authz/docs/README.md
```diff
@@ -81,7 +81,9 @@ A manage role does not, by itself, let you grant access to other people.
 
 Creating a User or Group that is not already a grant subject is ordinary object creation.
 
-Creating a User for an email that already carries a grant, or writing a ClusterAuthorizationRule, is granting roles. The request is admitted only if the requester already has covering permissions or is explicitly allowed to assign those roles.
+Creating a User for an email that already carries a grant, writing or deleting a ClusterAuthorizationRule, or resetting that identity through UserOperation, is granting roles. The request is admitted only if the requester already has covering permissions or is explicitly allowed to assign those roles.
+
+Connecting a DexProvider is measured the same way against the identities the provider can assert: `spec.allowedIdentities` (emails, email domains, groups) and the provider's own group filters bound that set, and only the roles already granted inside it count. A provider without an email limit can assert every email, including the ones that hold SuperAdmin, so writing it requires SuperAdmin. Details and the list of changes admitted without a check are in the user-authn module documentation.
```
denmaloyreb, line 86:
```suggestion
Connecting a DexProvider is measured the same way against the identities the provider can assert: [`spec.allowedIdentities`](/modules/user-authn/cr.html#dexprovider-v1-spec-allowedidentities) (emails, email domains, groups) and the provider's own group filters bound that set, and only the roles already granted inside it count. A provider without an email limit can assert every email, including the ones that hold SuperAdmin, so writing it requires SuperAdmin. Details and the list of changes admitted without a check are in the user-authn module documentation.
```

modules/140-user-authz/docs/USAGE.md
```diff
@@ -259,6 +259,8 @@ A `User` or `Group` subject is matched by name against the user identity in the
 
 If the match is intentional, for example, if the ClusterAuthorizationRule has been written in advance, set the `user-authz.deckhouse.io/allow-authorization-rule-collision: "true"` annotation on the User or Group. The annotation only acknowledges the name collision. It does not let a requester assign roles they cannot cover or that sit outside their can-assign range.
 
+Deleting a ClusterAuthorizationRule uses the same can-assign check as creating or updating it. You can remove a grant only if you could assign those roles. An update is checked against both the previous and the new grants, so lowering SuperAdmin to User still requires being able to assign SuperAdmin.
```
denmaloyreb, line 262:
```suggestion
Deleting a ClusterAuthorizationRule uses the same can-assign check as creating or updating it. You can remove a grant only if you could assign those roles. An update is checked against both the previous and the new grants, so lowering `SuperAdmin` to `User` still requires being able to assign `SuperAdmin`.
```

modules/150-user-authn/docs/USAGE_RU.md
```diff
@@ -671,7 +671,34 @@ d8 iam user delete anton --keep-memberships
...
+Удалить, пересоздать или выполнить UserOperation (`ResetPassword`, `Reset2FA`, `Lock`, `Unlock`) над локальным пользователем, чей email или членство в группе уже несёт грант, можно только если вы можете назначить эти роли (покрывающие права или явный диапазон can-assign). `initiatorType: self` эту проверку не обходит.
+
+Подключить DexProvider — то же назначение. Провайдер утверждает email и набор групп, а имя пользователя в Kubernetes — это и есть email, поэтому провайдер дотягивается до всех грантов, которые уже висят на идентичностях, которые он способен утвердить. Что он способен утвердить, ограничивается по двум осям блоком `spec.allowedIdentities`: `emails` и `emailDomains` ограничивают email, `groups` — claim групп. Фильтры групп, специфичные для типа провайдера, тоже учитываются (`oidc.allowedGroups`, `gitlab.groups`, `crowd.groups`, `bitbucketCloud.teams` без `includeTeamGroups`, `github.orgs[].teams` у каждой организации, `saml.allowedGroups` вместе с `filterGroups: true`). Ось без ограничителя открыта и дотягивается до всех грантов на субъектах этого вида.
```
denmaloyreb, line 676:
```suggestion
Подключить DexProvider — то же назначение. Провайдер утверждает email и набор групп, а имя пользователя в Kubernetes — это и есть email, поэтому провайдер дотягивается до всех грантов, имеющихся у идентичностей, которые он способен утвердить. Что он способен утвердить, ограничивается по двум осям блоком `spec.allowedIdentities`: `emails` и `emailDomains` ограничивают email, `groups` — claim групп. Фильтры групп, специфичные для типа провайдера, тоже учитываются (`oidc.allowedGroups`, `gitlab.groups`, `crowd.groups`, `bitbucketCloud.teams` без `includeTeamGroups`, `github.orgs[].teams` у каждой организации, `saml.allowedGroups` вместе с `filterGroups: true`). Ось без ограничителя открыта и дотягивается до всех грантов субъектов этого вида.
```

modules/150-user-authn/docs/USAGE_RU.md
```diff
@@ -671,7 +671,34 @@ d8 iam user delete anton --keep-memberships
...
+
+SuperAdmin в каждом кластере выдан пользователю (User), поэтому провайдер без ограничителя по email способен утвердить этот email, и создать его или подключить заново может только SuperAdmin. При ограничителях по обеим осям учитываются только роли, уже выданные перечисленным идентичностям: ClusterAdmin или менеджер подсистемы `security` подключает провайдер для `@contractor.example` и группы `contractors` без SuperAdmin, если ни у одной из этих идентичностей ещё нет роли, которую он не может назначить. Имена групп сравниваются точно; вхождение объекта Group в другие объекты Group список не расширяет: токен внешнего провайдера несёт только те группы, которые провайдер утвердил.
```
denmaloyreb, line 678:
```suggestion
`SuperAdmin` в каждом кластере выдан пользователю (User), поэтому провайдер без ограничителя по email способен утвердить этот email, и создать его или подключить заново может только `SuperAdmin`. При ограничителях по обеим осям учитываются только роли, уже выданные перечисленным идентичностям: `ClusterAdmin` или менеджер подсистемы `security` подключает провайдер для `@contractor.example` и группы `contractors` без `SuperAdmin`, если ни у одной из этих идентичностей ещё нет роли, которую он не может назначить. Имена групп сравниваются точно; вхождение объекта Group в другие объекты Group список не расширяет: токен внешнего провайдера несёт только те группы, которые провайдер утвердил.
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -23,6 +23,30 @@ spec:
                     Имя провайдера, которое будет отображено на странице выбора провайдера для аутентификации.
 
                     Если настроен всего один провайдер, страница выбора провайдера показываться не будет.
+                allowedIdentities:
+                  description: |
+                    Ограничивает идентичности, которые этот провайдер может утвердить, независимо от его типа.
+                    Dex отказывает во входе пользователю, чей email не входит в `emails` и `emailDomains`, если задан
```
denmaloyreb, line 29:
```suggestion
                    Ограничивает идентичности, которые этот провайдер может утвердить, независимо от его типа.

                    Dex отказывает во входе пользователю, чей email не входит в `emails` и `emailDomains`, если задан
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -23,6 +23,30 @@ spec:
...
+                    хотя бы один из списков, и сужает claim `groups` до пересечения со списком `groups`, если он задан
+                    (пустое пересечение — отказ во входе). Каждый список необязателен; отсутствующий или пустой список
+                    ничего не ограничивает.
+
+                    Ограничения также определяют, кто может создать или изменить провайдер. Провайдер, способный
+                    утвердить email или группу, на которых в кластере уже есть роли, может записать только тот, кому
```
denmaloyreb, line 35:
```suggestion
                    утвердить email или группу, которым в кластере уже назначены роли, может записать только тот, кому
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -23,6 +23,30 @@ spec:
...
+                    разрешено назначать эти роли. Без ограничения по email провайдер может утвердить любой адрес, в том
+                    числе с SuperAdmin, поэтому его запись требует SuperAdmin. С ограничениями по обоим признакам
```
denmaloyreb, line 37:
```suggestion
                    числе с `SuperAdmin`, поэтому его запись требует `SuperAdmin`. С ограничениями по обоим признакам
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -23,6 +23,30 @@ spec:
...
+                    учитываются только роли, выданные перечисленным идентичностям.
+                  properties:
+                    emails:
+                      description: 'Точные email-адреса, которые провайдер может утвердить. Сравнение без учёта регистра.'
+                    emailDomains:
+                      description: |
+                        Домены email, которые провайдер может утвердить, без `@`. Сравнивается точно с частью адреса
```
denmaloyreb, line 44:
```suggestion
                        Домены email, которые провайдер может утвердить, без `@`.

                        Сравнивается точно с частью адреса
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -23,6 +23,30 @@ spec:
...
+                        после последнего `@`; поддомены перечисляются явно.
+                    groups:
+                      description: |
+                        Имена групп, которые провайдер может утвердить. Применяется поверх типоспецифичных фильтров
```
denmaloyreb, line 48:
```suggestion
                        Имена групп, которые провайдер может утвердить.

                        Применяется поверх типоспецифичных фильтров
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -418,6 +442,30 @@ spec:
                     Имя провайдера, которое будет отображено на странице выбора провайдера для аутентификации.
 
                     Если настроен всего один провайдер, страница выбора провайдера показываться не будет.
+                allowedIdentities:
+                  description: |
+                    Ограничивает идентичности, которые этот провайдер может утвердить, независимо от его типа.
+                    Dex отказывает во входе пользователю, чей email не входит в `emails` и `emailDomains`, если задан
```
denmaloyreb, line 448:
```suggestion
                    Ограничивает идентичности, которые этот провайдер может утвердить, независимо от его типа.

                    Dex отказывает во входе пользователю, чей email не входит в `emails` и `emailDomains`, если задан
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -418,6 +442,30 @@ spec:
...
+                    хотя бы один из списков, и сужает claim `groups` до пересечения со списком `groups`, если он задан
+                    (пустое пересечение — отказ во входе). Каждый список необязателен; отсутствующий или пустой список
+                    ничего не ограничивает.
+
+                    Ограничения также определяют, кто может создать или изменить провайдер. Провайдер, способный
+                    утвердить email или группу, на которых в кластере уже есть роли, может записать только тот, кому
```
denmaloyreb, line 454:
```suggestion
                    утвердить email или группу, которым в кластере уже назначены роли, может записать только тот, кому
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -418,6 +442,30 @@ spec:
...
+                    разрешено назначать эти роли. Без ограничения по email провайдер может утвердить любой адрес, в том
+                    числе с SuperAdmin, поэтому его запись требует SuperAdmin. С ограничениями по обоим признакам
```
denmaloyreb, line 456:
```suggestion
                    числе с `SuperAdmin`, поэтому его запись требует `SuperAdmin`. С ограничениями по обоим признакам
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -418,6 +442,30 @@ spec:
...
+                    учитываются только роли, выданные перечисленным идентичностям.
+                  properties:
+                    emails:
+                      description: 'Точные email-адреса, которые провайдер может утвердить. Сравнение без учёта регистра.'
+                    emailDomains:
+                      description: |
+                        Домены email, которые провайдер может утвердить, без `@`. Сравнивается точно с частью адреса
```
denmaloyreb, line 463:
```suggestion
                        Домены email, которые провайдер может утвердить, без `@`.

                        Сравнивается точно с частью адреса
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -418,6 +442,30 @@ spec:
...
+                        после последнего `@`; поддомены перечисляются явно.
+                    groups:
+                      description: |
+                        Имена групп, которые провайдер может утвердить. Применяется поверх типоспецифичных фильтров
```
denmaloyreb, line 467:
```suggestion
                        Имена групп, которые провайдер может утвердить.

                        Применяется поверх типоспецифичных фильтров
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -50,6 +50,43 @@ spec:
                   type: string
                   description: |
                     The provider name to show on the authentication provider selection page. The selection page will not be displayed if there is only one provider configured.
+                allowedIdentities:
+                  type: object
+                  description: |
+                    Limits the identities this provider is allowed to assert, regardless of the provider type.
+                    Dex refuses to sign in a user whose email is outside `emails` and `emailDomains` when either is set,
```
denmaloyreb, line 57:
```suggestion
                    Limits the identities this provider is allowed to assert, regardless of the provider type.

                    Dex refuses to sign in a user whose email is outside `emails` and `emailDomains` when either is set,
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -50,6 +50,43 @@ spec:
...
+                    and narrows the `groups` claim to the intersection with `groups` when it is set (an empty
+                    intersection refuses the sign-in). Each list is optional; an omitted or empty list does not limit
+                    that attribute.
+
+                    The limits also decide who may create or change the provider. A provider that can assert an
+                    email or a group that already carries roles in the cluster may be written only by a requester
+                    who is allowed to assign those roles. Without an email limit the provider can assert any email,
+                    including the ones that hold SuperAdmin, so writing it requires SuperAdmin. With limits on both
```
denmaloyreb, line 65:
```suggestion
                    including the ones that hold `SuperAdmin`, so writing it requires `SuperAdmin`. With limits on both
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -50,6 +50,43 @@ spec:
...
+                    emails:
+                      type: array
+                      description: 'Exact email addresses this provider may assert. Compared case-insensitively.'
+                      items:
+                        type: string
+                        minLength: 1
+                    emailDomains:
+                      type: array
+                      description: |
+                        Email domains this provider may assert, without the `@`. Matched exactly against the part of
```
denmaloyreb, line 77:
```suggestion
                        Email domains this provider may assert, without the `@`.

                        Matched exactly against the part of
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -50,6 +50,43 @@ spec:
...
+                        the email after the last `@`; subdomains must be listed explicitly.
+                      items:
+                        type: string
+                        minLength: 1
+                    groups:
+                      type: array
+                      description: |
+                        Group names this provider may assert. Applied on top of the provider-specific group filters
```
denmaloyreb, line 85:
```suggestion
                        Group names this provider may assert.

                        Applied on top of the provider-specific group filters
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -719,6 +756,43 @@ spec:
                   type: string
                   description: |
                     The provider name to show on the authentication provider selection page. The selection page will not be displayed if there is only one provider configured.
+                allowedIdentities:
+                  type: object
+                  description: |
+                    Limits the identities this provider is allowed to assert, regardless of the provider type.
+                    Dex refuses to sign in a user whose email is outside `emails` and `emailDomains` when either is set,
```
denmaloyreb, line 763:
```suggestion
                    Limits the identities this provider is allowed to assert, regardless of the provider type.

                    Dex refuses to sign in a user whose email is outside `emails` and `emailDomains` when either is set,
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -719,6 +756,43 @@ spec:
...
+                    and narrows the `groups` claim to the intersection with `groups` when it is set (an empty
+                    intersection refuses the sign-in). Each list is optional; an omitted or empty list does not limit
+                    that attribute.
+
+                    The limits also decide who may create or change the provider. A provider that can assert an
+                    email or a group that already carries roles in the cluster may be written only by a requester
+                    who is allowed to assign those roles. Without an email limit the provider can assert any email,
+                    including the ones that hold SuperAdmin, so writing it requires SuperAdmin. With limits on both
```
denmaloyreb, line 771:
```suggestion
                    including the ones that hold `SuperAdmin`, so writing it requires `SuperAdmin`. With limits on both
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -719,6 +756,43 @@ spec:
...
+                    emails:
+                      type: array
+                      description: 'Exact email addresses this provider may assert. Compared case-insensitively.'
+                      items:
+                        type: string
+                        minLength: 1
+                    emailDomains:
+                      type: array
+                      description: |
+                        Email domains this provider may assert, without the `@`. Matched exactly against the part of
```
denmaloyreb, line 783:
```suggestion
                        Email domains this provider may assert, without the `@`.

                        Matched exactly against the part of
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -719,6 +756,43 @@ spec:
...
+                        the email after the last `@`; subdomains must be listed explicitly.
+                      items:
+                        type: string
+                        minLength: 1
+                    groups:
+                      type: array
+                      description: |
+                        Group names this provider may assert. Applied on top of the provider-specific group filters
```
denmaloyreb, line 791:
```suggestion
                        Group names this provider may assert.

                        Applied on top of the provider-specific group filters
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -719,6 +756,43 @@ spec:
...
+
+                    The limits also decide who may create or change the provider. A provider that can assert an
+                    email or a group that already carries roles in the cluster may be written only by a requester
+                    who is allowed to assign those roles. Without an email limit the provider can assert any email,
+                    including the ones that hold SuperAdmin, so writing it requires SuperAdmin. With limits on both
+                    attributes only the roles granted to the listed identities count.
+                  properties:
+                    emails:
+                      type: array
+                      description: 'Exact email addresses this provider may assert. Compared case-insensitively.'
```
denmaloyreb, line 776:
```suggestion
                      description: |
                        Exact email addresses this provider may assert.

                        Compared case-insensitively.
```

modules/150-user-authn/crds/dex-provider.yaml
```diff
@@ -50,6 +50,43 @@ spec:
...
```
denmaloyreb, line 70:
```suggestion
                      description: |
                        Exact email addresses this provider may assert.

                        Compared case-insensitively.
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -23,6 +23,30 @@ spec:
...
+                    ничего не ограничивает.
+
+                    Ограничения также определяют, кто может создать или изменить провайдер. Провайдер, способный
+                    утвердить email или группу, на которых в кластере уже есть роли, может записать только тот, кому
+                    разрешено назначать эти роли. Без ограничения по email провайдер может утвердить любой адрес, в том
+                    числе с SuperAdmin, поэтому его запись требует SuperAdmin. С ограничениями по обоим признакам
+                    учитываются только роли, выданные перечисленным идентичностям.
+                  properties:
+                    emails:
+                      description: 'Точные email-адреса, которые провайдер может утвердить. Сравнение без учёта регистра.'
```
denmaloyreb, line 41:
```suggestion
                      description: |
                        Точные email-адреса, которые провайдер может утвердить.

                        Сравнение без учёта регистра.
```

modules/150-user-authn/crds/doc-ru-dex-provider.yaml
```diff
@@ -418,6 +442,30 @@ spec:
...
```
denmaloyreb, line 460:
```suggestion
                      description: |
                        Точные email-адреса, которые провайдер может утвердить.

                        Сравнение без учёта регистра.
```

denmaloyreb (approved):

z9r5 (approved):

PR #22722, opened 2026-09-01
[admission-policy-engine] Add audit rules benchmark scripts

Suselz:
## Description

Adds diagnostic/benchmarking tooling for `admission-policy-engine` (Gatekeeper): scripts to measure resource consumption both statically (per-rule) and on a live cluster.

- `charts/constraint-templates/tests/tools/bench_rules.py` — per-`ConstraintTemplate` Rego eval time (`opa eval --metrics`), parallelized.
- `charts/constraint-templates/tests/tools/rulebench/` (+ `rulebench.sh` wrapper) — same, plus real allocation cost (`B/op`, `allocs/op`) via OPA's Go SDK + `testing.Benchmark`. Isolated Go module (own `go.mod`) so it doesn't add a dependency to the repo's root module.
- `tools/live_resource_check.sh` — live-cluster sweep: pods/restarts, resource usage over time, Gatekeeper's own Prometheus metrics, audit-cycle timing, cluster-wide scale/events.
- `tools/audit_cycle_cost.sh` — measures the exact CPU-seconds and memory allocations of one audit cycle by diffing Go-runtime counters across the cycle's logged start/end — no `--enable-pprof`, no Deployment change needed.
- `tools/AGENTS.md` and `tests/AGENTS.md` updates — consolidate the findings from using these tools (e.g. audit CPU is dominated by API discovery/LIST, not Rego eval; which rules are relatively most expensive and why) so future investigations don't start from zero.

No production manifests/behavior are touched — dev/diagnostic tooling only.

There was no way to answer "why is admission-policy-engine using this much CPU/memory" other than reading raw `kubectl top`/logs by hand, and no way to attribute cost to a specific rule or audit cycle. These scripts were used to investigate and fix real idle-resource-consumption issues in the module (tracked separately) and are kept as reusable tooling for future investigations.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: chore
summary: Add scripts to benchmark per-rule cost and diagnose live resource consumption for Gatekeeper.
impact_level: low
```


modules/015-admission-policy-engine/tools/audit_cycle_cost.sh
```diff
@@ -0,0 +1,212 @@
...
+echo ""
+
+echo "waiting for the start of a NEW audit cycle (up to ${TIMEOUT}s)..."
+deadline=$(( $(date +%s) + TIMEOUT ))
+last_id=""
+start_id=""
+while :; do
+  line="$(kubectl -n "$NAMESPACE" logs "$AUDIT_POD" -c manager --tail=50 2>/dev/null | grep 'audit_started' | tail -1)"
+  id="$(grep -oE '"audit_id":"[^"]+"' <<<"$line" | head -1)"
+  if [ -n "$id" ] && [ "$id" != "$last_id" ]; then
```
AlwxSin: 🔴 **"Wait for a NEW audit cycle" grabs the previous / in-progress cycle on the first poll**

`last_id` starts empty, and the very first iteration evaluates `[ -n "$id" ] && [ "$id" != "$last_id" ]` before the current id is ever recorded (the `last_id="$id"` assignment below is only reached when the `if` fails). So if the last 50 log lines contain *any* `audit_started` — a cycle currently in progress, or a recently finished one whose post-cycle burst was short — the loop breaks immediately with a stale `start_id`.

Consequences: the `before` snapshot is taken mid-cycle (or after the cycle already ended), `audit_finished` for that id is found on the first poll, and the script prints a partial or near-zero cost under the header "measured, not estimated".

It only works today because with ~64 constraints the per-constraint status burst pushes `audit_started` out of `--tail=50` between cycles. Fewer constraints, a more verbose log, or simply starting the script while a cycle is running (16s of every 300s; 27% of the time with the current 60s interval) trips it.

Fix: seed `last_id` from the current log **before** entering the loop, then wait until `id != last_id`.

modules/015-admission-policy-engine/tools/live_resource_check.sh
```diff
@@ -0,0 +1,356 @@
...
+# ---------------------------------------------------------------------------
+AUDIT_POD="$(kubectl -n "$NAMESPACE" get pods -l control-plane=audit-controller -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)"
+if [ -n "$AUDIT_POD" ]; then
+  LOG_FILE="$OUTDIR/audit_log_tail.txt"
+  kubectl -n "$NAMESPACE" logs "$AUDIT_POD" -c manager --tail=3000 2>/dev/null \
+    | grep -E 'audit_started|audit_finished' > "$LOG_FILE" || true
+  n_cycles=$(grep -c audit_finished "$LOG_FILE" 2>/dev/null || echo 0)
+  log "audit pod: $AUDIT_POD, cycles found in last 3000 log lines: $n_cycles"
+  if [ "${n_cycles:-0}" -gt 0 ]; then
+    grep -oE '"duration":"[0-9.]+s"' "$LOG_FILE" | grep -oE '[0-9.]+' | awk '
```
AlwxSin: 🟡 **Duration regex silently drops every cycle of 60s or longer**

`"duration":"[0-9.]+s"` matches Go's `Duration.String()` only below one minute. At 65.5s Go prints `1m5.548640275s`, which this regex does not match, so the cycle vanishes from `cycles/avg/max` entirely (verified with a one-line Go program).

Section 5 exists to flag 3x outliers, and those are exactly the ones that cross 60s. The README in this PR projects 64s/cycle at 1000 pods and 160s at 2500, so on any mid-size cluster this section reports zero cycles or a misleadingly low average.

Parse the `h`/`m`/`s` components (e.g. `([0-9]+h)?([0-9]+m)?([0-9.]+)s` and sum them in awk) instead of assuming seconds-only.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/rulebench/main.go
```diff
@@ -0,0 +1,310 @@
...
+	// contention - only the timing numbers get noisier under concurrency,
+	// and only by about as much as normal run-to-run jitter already is).
+	if v := os.Getenv("RULEBENCH_JOBS"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			jobs = n
+		}
+	}
+
+	results := make([][]string, len(entries))
+	sem := make(chan struct{}, jobs)
```
AlwxSin: 🟡 **Concurrency is a no-op for the timed part; the docs describe a trade-off that cannot happen**

`testing.B.runN` takes a package-level `benchmarkLock` (`$GOROOT/src/testing/benchmark.go:69,198`), so concurrent `testing.Benchmark` calls serialize their timed runs. Measured: 4 goroutines each running a 2.15s benchmark take 7.84s total (3.6x, i.e. sequential).

So the semaphore, `RULEBENCH_JOBS`, and the README paragraphs (`tests/README.md:278-289`) about "noisier ns/op under contention" and "`RULEBENCH_JOBS=1` for precision" describe contention that never occurs; the observed ~13% speedup is just `PrepareForEval` + YAML loading overlapping with someone else's timed run.

The one real side effect goes the *other* way: another goroutine's compile and sanity `Eval` execute inside a running benchmark's timed window and land in the process-wide `MemStats.Mallocs` that `ReportAllocs` reads, so `allocs/op`/`B/op` get a small (bounded, one compile per constraint) bump — while the README claims they are "exact ... unaffected by contention".

Simplest fix: drop the pool and the env var, run sequentially, delete those paragraphs. If you keep the pool, rewrite the docs to say what it actually parallelizes (prepare phase only) and that alloc numbers can be slightly inflated by it.

modules/015-admission-policy-engine/tools/audit_cycle_cost.sh
```diff
@@ -0,0 +1,212 @@
...
+    exit 1
+  fi
+  if [ "$(date +%s)" -gt "$deadline" ]; then
+    echo "ERROR: timed out waiting for the cycle to finish" >&2
+    exit 1
+  fi
+  sleep 1
+done
+after_wall=$(date +%s.%N)
+after="$(curl -s -m 2 "http://127.0.0.1:${PORT}/metrics")"
```
AlwxSin: 🟡 **The `after` snapshot is never validated; a dead port-forward yields negative CPU**

The wait for `audit_finished` can last minutes (`TIMEOUT * 3`), and a backgrounded `kubectl port-forward` is known to drop on idle. If this `curl` fails, `after` is empty, every `val` returns nothing, and the awk block below happily prints e.g. `CPU-seconds consumed -123.456s` / `heap alloc, net change -200 MiB` under the header "measured, not estimated".

Check that both snapshots are non-empty and contain `process_cpu_seconds_total`; on failure retry once (re-establishing the port-forward), then exit non-zero rather than printing garbage.

modules/015-admission-policy-engine/tools/AGENTS.md
```diff
@@ -0,0 +1,172 @@
...
+   cross-check: `allowed-proc-mount`, `read-only-root-filesystem`,
+   `allow-privilege-escalation`, `allow-privileged`, `allowed-host-paths`,
+   `allowed-cluster-roles`, `allowed-users`. Common cause: per-container SPE
+   (SecurityPolicyException) resolution builds and holds intermediate
+   objects even when no exception applies. Memory spread across rules
+   (~8x) is much narrower than time spread (~100x) - most rules share a
+   common per-eval allocation floor from OPA itself (~12,000 B/op, ~230
+   allocs/op) before any rule-specific logic runs.
+
+5. **`--audit-interval` was 60s** (upstream Gatekeeper's own default is
```
AlwxSin: 🟡 **Docs present unmerged changes and a non-existent report as established facts**

This item says `--audit-interval` "was 60s" with a fix visible "in the Deployment manifests"; `tools/README.md:180` says it was "reduced to 300s, explicit `limits` added"; `tools/README.md:307` talks about "this module's own earlier patch" setting a 1000m limit.

`origin/main` has `--audit-interval=60` at `templates/audit-deployment.yaml:69` and no static limits, and none of your open PRs contain that change. "The main resource-audit report" is referenced five times across `tools/README.md` and `live_resource_check.sh` and does not exist in the repo.

AGENTS.md is explicitly written for agents that will follow these pointers literally — right now they find the opposite of what is written. Either land that PR first and link it, or reword these as "proposed" with a link to the tracker card / the actual report.

modules/015-admission-policy-engine/charts/constraint-templates/tests/README.md
```diff
@@ -183,3 +190,113 @@ go run $constraint_testgen coverage -tests-root ./ -format table
...
+(`*allowed*` / `*disallowed*`, falling back to the first two files) — treat
+absolute numbers as approximate, the relative ranking across constraints is
+the useful part. See each tool's docstring/doc-comment for details.
+
+**What these numbers are (and aren't) representative of.** Both tools
+measure exactly what the **webhook** does for one admission review: one
+object, one constraint, no cluster listing. They are directly representative
+of webhook/admission-request latency. They are **not** representative of the
+**audit** loop's total cost: measured on a live cluster with
+[`../../../../tools/audit_cycle_cost.sh`](../../../../tools/audit_cycle_cost.sh)
```
AlwxSin: 🔵 **Broken relative paths in the new docs**

Resolved every added relative link against the PR tree:

- here and at lines 220, 222: `../../../../tools/…` from `tests/` resolves to `modules/tools/` — one level too many, should be `../../../tools/…` (the same-directory `AGENTS.md` gets it right).
- `tools/README.md:309`: `../charts/constraint-templates/templates/audit-deployment.yaml` does not exist; the real file is `../templates/audit-deployment.yaml`.
- `tools/AGENTS.md:114`: `../charts/constraint-templates/templates/policies/security-policy/constraint.yaml` does not exist; the real file is `../templates/policies/security-policy/constraint.yaml`.

modules/015-admission-policy-engine/tools/audit_cycle_cost.sh
```diff
@@ -0,0 +1,212 @@
...
+set -uo pipefail
+
+NAMESPACE="d8-admission-policy-engine"
+TIMEOUT=180
+
+while getopts "n:t:h" opt; do
+  case "$opt" in
+    n) NAMESPACE="$OPTARG" ;;
+    t) TIMEOUT="$OPTARG" ;;
+    h) sed -n '2,30p' "$0" | sed 's/^# \{0,1\}//'; exit 0 ;;
```
AlwxSin: 🔵 **`-h` prints the Apache license instead of the usage**

`sed -n '2,30p'` predates the "Fix CI: copyright headers" commit, which inserted 14 license lines above the description. Now `-h` shows the license plus the first half of the description and cuts off before `Usage:`. Same in `live_resource_check.sh:63`.

Anchor on a marker instead of line numbers — e.g. print from the line matching the script's own name up to the end of that comment block.

modules/015-admission-policy-engine/tools/live_resource_check.sh
```diff
@@ -0,0 +1,356 @@
...
+
+# ---------------------------------------------------------------------------
+head1 "5. Audit-cycle timing trend (from logs)"
+# ---------------------------------------------------------------------------
+AUDIT_POD="$(kubectl -n "$NAMESPACE" get pods -l control-plane=audit-controller -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)"
+if [ -n "$AUDIT_POD" ]; then
+  LOG_FILE="$OUTDIR/audit_log_tail.txt"
+  kubectl -n "$NAMESPACE" logs "$AUDIT_POD" -c manager --tail=3000 2>/dev/null \
+    | grep -E 'audit_started|audit_finished' > "$LOG_FILE" || true
+  n_cycles=$(grep -c audit_finished "$LOG_FILE" 2>/dev/null || echo 0)
```
AlwxSin: 🔵 **Assorted tool inconsistencies and rough edges**

- Here: `grep -c` already prints `0` on no match and exits 1, so `|| echo 0` makes `n_cycles` two lines (`0\n0`) and the following `-gt` test errors out with "integer expression expected".
- Line 70: the default `./ape-diag-<ts>` output dir lands inside the repo when run as documented from `tools/`, and is not gitignored.
- `--only` is a substring match in `bench_rules.py:271` and an exact match in `rulebench/main.go:256`; Python scans every group dir, Go hardcodes `operation`/`security` (`main.go:246`). The README calls them "same arguments".
- `rulebench/main.go:238`: `os.Args[1]` panics without arguments (the wrapper guards, a direct `go run .` does not).
- `bench_rules.py:287`: `future.result()` re-raises, so one `opa` timeout (120s) or a missing `opa` binary kills the whole 39-constraint run and discards everything already collected. Catch per-constraint and record it as an `error` row instead.

tools/docs/spelling/wordlist
```diff
@@ -2853,3 +2853,24 @@ ZvirtInstanceClass
 юзера
 юнит
 юнита
+ConstraintTemplate
+ConstraintTemplates
+GC
+GVKs
+SDK
+SecurityPolicyException
+adyakonov
```
AlwxSin: 🔵 **Shared spelling dictionary learns a personal dev-cluster hostname**

`adyakonov` is added (here and in `dev_OPS.dic`) so that `adyakonov-cloud` passes spell-check in ~10 places across the new docs, alongside fragments like `didn`, `args`, `diff`.

Describing the cluster as "a 2-node dev cluster (143 pods)" in the docs avoids teaching the repo-wide dictionary a hostname.

Suselz:
Thanks for the thorough review — all 9 points were valid, fixed in f8b16d6fcc:

- **`audit_cycle_cost.sh`**: fixed the `last_id` seeding bug (real, confirmed by reading the code), added snapshot validation + retry, fixed `-h`.
- **`live_resource_check.sh`**: fixed the duration regex (confirmed with a synthetic test — `1m5.5s` etc. now parse correctly), fixed the `grep -c || echo 0` double-line bug, moved the default output dir outside the repo, fixed `-h`.
- **`rulebench`**: confirmed `benchmarkLock` in `$GOROOT/src/testing/benchmark.go` — you're right, the concurrency was a no-op for the timed part and risked polluting `allocs/op`. Reverted to sequential, removed `RULEBENCH_JOBS`. Also fixed the `os.Args[1]` panic and aligned `--only` to substring matching.
- **`bench_rules.py`**: `future.result()` is now caught per-future instead of killing the whole run (tested by removing `opa` from `PATH`).
- **Docs**: removed the false "already fixed to 300s / limits added" claims and the "main resource-audit report" references (that report was never a repo artifact), fixed the broken relative paths, and replaced `adyakonov-cloud` with a generic description (also dropped from the shared spelling wordlist).

All fixes verified locally (bash -n, go build/vet/golangci-lint, python parse, and a few targeted repro tests) before pushing.

modules/015-admission-policy-engine/tools/audit_cycle_cost.sh
Suselz: Confirmed and fixed in f8b16d6fcc: seeded `last_id` from the log before entering the wait loop, exactly as suggested — the loop now only breaks on an `audit_id` genuinely different from whatever was already there when it started polling.

modules/015-admission-policy-engine/tools/live_resource_check.sh
Suselz: Confirmed — verified with a synthetic test (7.5s / 65.5s / 58.4s / 3723.5s inputs, all parsed correctly). Fixed in f8b16d6fcc: now parses the optional `h`/`m` components before the trailing seconds instead of assuming seconds-only.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/rulebench/main.go
Suselz: Verified `benchmarkLock` directly in `$GOROOT/src/testing/benchmark.go:69` — you're right, it serializes the timed run process-wide. Reverted to sequential execution in f8b16d6fcc, dropped the worker pool and `RULEBENCH_JOBS`, and corrected the README section that described the (nonexistent) contention trade-off.

modules/015-admission-policy-engine/tools/audit_cycle_cost.sh
Suselz: Fixed in f8b16d6fcc: both snapshots are now validated to actually contain `process_cpu_seconds_total` before computing any deltas, with one retry (re-establishing the port-forward) before failing loudly instead of printing numbers computed from empty data.

modules/015-admission-policy-engine/tools/AGENTS.md
Suselz: Fixed in f8b16d6fcc: reworded finding 5 to state `--audit-interval` is currently `60` in the manifest and that lowering it is a proposed, not-yet-applied change (with a pointer to the actual file to check), and removed all references to "the main resource-audit report" — it was never a repo artifact, just an earlier ephemeral analysis.

modules/015-admission-policy-engine/charts/constraint-templates/tests/README.md
Suselz: Fixed in f8b16d6fcc: corrected all four `../../../../tools/...` occurrences to `../../../tools/...`, and the two `charts/constraint-templates/templates/...` paths that should point at the module's own `templates/`. Also found and fixed one more broken instance you didn't flag, in `bench_rules.py`'s docstring.

modules/015-admission-policy-engine/tools/audit_cycle_cost.sh
Suselz: Fixed in f8b16d6fcc in both scripts: anchored the extraction on the license's last line instead of a fixed line range, so it won't rot again the next time something gets inserted above it.

modules/015-admission-policy-engine/tools/live_resource_check.sh
Suselz: Fixed all five in f8b16d6fcc:
- `grep -c ... || echo 0` double-line bug: removed the redundant `|| echo 0` (`grep -c` already prints `0` on no match).
- Default output dir: now `$TMPDIR/ape-diag-<ts>`, outside the repo.
- `--only` substring vs exact match: aligned `rulebench` to substring matching like `bench_rules.py`'s, so "same arguments" is now actually true.
- `os.Args[1]` panic: now guarded with a usage message and `exit 2`.
- `future.result()` killing the whole run: now caught per-future in `bench_rules.py` and recorded as an error row (tested by removing `opa` from `PATH`).

tools/docs/spelling/wordlist
Suselz: Fixed in f8b16d6fcc: replaced `adyakonov-cloud` throughout with a generic "the reference cluster" description and removed it from the shared `wordlist`/`dev_OPS.dic`.

modules/015-admission-policy-engine/tools/audit_cycle_cost.sh
```diff
@@ -0,0 +1,272 @@
...
+  kill "${PF_PID:-}" >/dev/null 2>&1
+  wait "${PF_PID:-}" 2>/dev/null
+  start_port_forward
+  if snap="$(fetch_snapshot)"; then
+    printf '%s' "$snap"
+    return 0
+  fi
+  echo "ERROR: could not fetch a valid $label metrics snapshot after retrying the port-forward." >&2
+  echo "       Refusing to print numbers computed from empty/partial data." >&2
+  exit 1
```
AlwxSin, line 149: 🟡 **Finding 4 follow-up: `exit 1` here never stops the script, and the retried port-forward leaks**

The validation and retry are right, but both call sites wrap this in a command substitution (`before="$(take_snapshot_or_die "start")"` at line 181, `after=…` at line 210), so the function runs in a subshell. `exit 1` at line 141 terminates only that subshell; the script has `set -uo pipefail` without `-e`, so the assignment of an empty string succeeds and execution continues with an empty snapshot — printing exactly the negative-CPU garbage this was meant to prevent. Reproduced:

```
$ bash -c 'set -uo pipefail; PF_PID=old; f(){ PF_PID=new; exit 1; }; before="$(f)"; echo "continued: before=[$before] PF_PID=$PF_PID"'
continued: before=[] PF_PID=old
```

Second consequence of the same subshell: `start_port_forward` at line 134 assigns `PF_PID` inside the subshell. The parent keeps the PID of the process you just killed, so `trap cleanup EXIT` never reaps the re-established port-forward — one leaked `kubectl port-forward` per retry.

Fix: return the snapshot through a global instead of stdout and call the function directly (`take_snapshot_or_die start; before="$SNAP"`), so both `exit 1` and the `PF_PID` update happen in the parent shell. Minimum viable alternative: `before="$(…)" || exit 1` at both call sites *and* move the port-forward restart out of the function.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/rulebench/main.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	root := os.Args[1]
+	only := ""
+	if len(os.Args) > 2 {
+		only = os.Args[2]
+	}
+
+	type entry struct{ group, name, dir string }
+	var entries []entry
+	for _, group := range []string{"operation", "security"} {
```
AlwxSin: 🔵 **Finding 8 leftover: group list is still hardcoded**

Four of the five bullets are done, thanks. This one remains: `rulebench` iterates a fixed `[]string{"operation", "security"}` while `bench_rules.py` scans every directory under the constraints root (skipping `_`/`.`-prefixed). A new group would be silently skipped by one tool and picked up by the other. `os.ReadDir(root)` with the same skip rule keeps the two tools in step.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/bench_rules.py
```diff
@@ -0,0 +1,308 @@
...
+Two samples are measured per constraint when available - "allowed" (no
+violation) and "disallowed" (violation) - picked from the constraint's own
+rendered/test_samples/ by filename convention. Both code paths are reported
+because some rules do materially more work on one branch than the other.
+
+IMPORTANT - what this number is (and isn't) representative of: one
+evaluation here is exactly what the WEBHOOK does for one admission review
+(one object, one constraint, no cluster listing). It is NOT representative
+of the AUDIT path's total cost: on a live cluster (measured with
+../../../../tools/audit_cycle_cost.sh), pure Rego-eval time was under 2%
```
AlwxSin: 🔵 **This path change broke it relative to the docstring's documented cwd**

Every other relative path in this docstring is relative to the working directory the examples use, `test_cases/constraints` — `../../docs/TESTING_GUIDE.md`, `../../tools/constraint_testgen`, `../../README.md` only resolve from there, not from this file's location. From that base the original five-`..` path was correct, and the new four-`..` one resolves to `charts/tools/audit_cycle_cost.sh`, which does not exist. Resolved from both bases against this head:

```
../../docs/TESTING_GUIDE.md            file-relative: no   cwd-relative: OK
../../tools/constraint_testgen         file-relative: no   cwd-relative: OK
../../../../tools/audit_cycle_cost.sh  file-relative: OK   cwd-relative: no   <- this line
../../../README.md                     file-relative: no   cwd-relative: no   <- line 63
```

Line 63's `../../../README.md` is dead from either base (`charts/constraint-templates/README.md` does not exist; `tests/README.md` is presumably meant, i.e. `../../README.md`). Pick one base for the docstring and say which — simplest is to revert this line to five `..` (cwd-relative, matching its neighbours) and fix line 63.

modules/015-admission-policy-engine/tools/audit_cycle_cost.sh
Suselz: Confirmed and fixed in b3009a9c09, exactly as suggested: `take_snapshot_or_die` now writes into a global `$SNAP` and is called directly (`take_snapshot_or_die start; before="$SNAP"`) instead of via `$(...)`, so both `exit 1` and the retried `PF_PID` land in the real running shell. Verified with two behavioral tests mirroring your repro — confirmed `exit 1` now actually stops the script, and `PF_PID` correctly reflects the retry after a successful second attempt.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/rulebench/main.go
Suselz: Fixed in b3009a9c09: `rulebench` now scans `root` with `os.ReadDir` and the same `"_"`/`"."`-prefix skip rule as `bench_rules.py`, instead of the hardcoded `{"operation","security"}` list. Verified the full run still finds the same 39 constraints (80 output lines) and `_test-samples` is still correctly skipped.

modules/015-admission-policy-engine/charts/constraint-templates/tests/tools/bench_rules.py
Suselz: You're right, and thanks for checking both bases explicitly — reverted `audit_cycle_cost.sh` to the correct 5-`..` form and fixed the actually-broken neighbor (old line 63's `../../../README.md`, dead from either base) to `../../README.md`. Also added one sentence stating explicitly that the whole docstring is relative to `test_cases/constraints`, per your ask. Fixed in b3009a9c09; re-verified every path in the docstring resolves from that cwd.

AlwxSin (approved):

z9r5 (approved):

Nikolay1224 (approved):

PR #22731, opened 2026-09-02
[user-authz] Fix permission-browser BulkSAR timeout for Editor

Jabejixo:
## Description

Console `cani` is proxied to permission-browser `BulkSubjectAccessReview` (~258 `resourceAttributes`). SuperAdmin-like users finish in a few hundred milliseconds. Editor / ClusterAdmin with a real `namespaceSelector` or `limitNamespaces` hit the handler timeout (~30s).

The timeout is live discovery on the multi-tenancy Authorize hot path, not CRB count. `hasAnyFilters` makes every cluster-scoped item call `ServerGroups` + `ServerResourcesForGroupVersion` with `context.TODO()` and a 32s client-go timeout. Negative lookups are not cached. `ResourceScopeCache` already existed for AccessibleNamespaces / IdentityRead and was not passed into `NewEngine`.

This PR:

1. Locks the Editor vs SuperAdmin BulkSAR / ANS contract in unit tests.
2. Wires `ResourceScopeCache` into the MT engine and treats unknown GVRs as namespaced (fail-closed, same as the webhook). Live discovery is gone from Authorize.
3. Snapshots the subject's ClusterRoleBinding rules once per BulkSAR and reuses them for `Authorize` / `AllowsIndependently`. Label selectors are compiled in `renewDirectories`.

CAR / AR filter decisions are not cached. Only "is this GVR namespaced?" is. `AllowsIndependently` still skips `user-authz:` + `heritage=deckhouse` + `module=user-authz` CRBs.

A new cluster-scoped CRD can show a false Deny in the console for up to the 5-minute scope-cache refresh. That is a reporting miss, not an access hole. An API group whose discovery fails (an unavailable APIService, a CRD group mid-restart) keeps its entries from the previous refresh, so an outage does not turn its cluster-scoped resources into Deny; only a group that discovery no longer lists at all is evicted. The cache reads `ServerGroupsAndResources`, so subresources are present and the RBAC wildcard matching of `HasNamespacedResourceMatching` sees them.

Editor console cani timed out. SuperAdmin on the same payload did not. The permission browser reported Allow for discovery misses (fail-open through a CAR CRB) while the webhook Denied them.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: fix
summary: Speed up permission-browser BulkSAR for users with namespace filters
impact_level: default
```

## Manual Kubernetes measurements

Measured against the same 258-resource console payload from the incident, with
`time.perf_counter()` around `kubectl create --raw` and a 90s request timeout.
The incident was reproduced under the same relevant conditions, then repeated
with the fixed PR image using the same payload and measurement procedure.

| Subject | Pre-fix (10 runs) | Fixed PR (30 runs) |
| --- | ---: | ---: |
| `pbtest-20260204-car@example.com` | p50 34.454s, p95 34.789s, p99 34.886s, 0/10 completed | p50 263ms, p95 321ms, p99 923ms, 30/30 completed |
| `watchtest1`, groups=`watchtest1` | p50 34.521s, p95 34.591s, p99 34.598s, 0/10 completed | p50 261ms, p95 295ms, p99 302ms, 30/30 completed |

The pre-fix requests consistently hit the handler timeout. The fixed PR
returned all 258 results; the one 923ms p99 sample was a cold/first-request
outlier, with the remaining requests around 250-340ms. A size sweep on the
fixed PR grew from 199ms at one item to 309ms at 258 items.

### DIFF
--- a/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/apiserver/apiserver.go
+++ b/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/apiserver/apiserver.go
@@ -154,7 +154,7 @@ func initAuthorizers(init *initResult, configPath string, scopeCache *resolver.R
 			configPath,
 			init.informerFactory.Core().V1().Namespaces().Lister(),
 			init.informerFactory.Core().V1().Namespaces().Informer().HasSynced,
-			init.clientset.Discovery(),
+			scopeCache,
 		)
 		if err != nil {
 			klog.Warningf("Failed to initialize multi-tenancy engine: %v. Multi-tenancy restrictions will not be applied.", err)
--- a/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/apiserver/apiserver.go
+++ b/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/apiserver/apiserver.go
@@ -146,6 +146,13 @@ func initAuthorizers(init *initResult, configPath string, scopeCache *resolver.R
 		configPath = "/etc/user-authz-webhook/config.json"
 	}
 
+	// Same guard for the engine: a nil *ResourceScopeCache must arrive as a nil
+	// interface, not as a non-nil interface holding a nil pointer.
+	var resourceScope multitenancy.ResourceScope
+	if scopeCache != nil {
+		resourceScope = scopeCache
+	}
+
 	// Create multi-tenancy engine
 	var mtEngine *multitenancy.Engine
 	if init.clientset != nil {
@@ -154,7 +161,7 @@ func initAuthorizers(init *initResult, configPath string, scopeCache *resolver.R
 			configPath,
 			init.informerFactory.Core().V1().Namespaces().Lister(),
 			init.informerFactory.Core().V1().Namespaces().Informer().HasSynced,
-			scopeCache,
+			resourceScope,
 		)
 		if err != nil {
 			klog.Warningf("Failed to initialize multi-tenancy engine: %v. Multi-tenancy restrictions will not be applied.", err)
--- a/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/scope_cache.go
+++ b/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/scope_cache.go
@@ -12,7 +12,11 @@ import (
 
 	"k8s.io/client-go/discovery"
 	"k8s.io/klog/v2"
+
+	"permission-browser-apiserver/pkg/authorizer/multitenancy"
 )
+
+var _ multitenancy.ResourceScope = (*ResourceScopeCache)(nil)
 
 const (
 	// defaultRefreshInterval is how often the scope cache refreshes from discovery.
@@ -74,6 +78,19 @@ func (c *ResourceScopeCache) IsNamespaced(group, resource string) bool {
 		return false
 	}
 	return namespaced
+}
+
+// Scope reports whether the resource is namespaced and whether the snapshot
+// contains it. Unlike IsNamespaced, a miss is not coerced to cluster-scoped:
+// multi-tenancy uses !known as "treat like namespaced" so a discovery hole
+// cannot fail-open. IsNamespaced stays fail-closed-as-cluster-scoped for
+// AccessibleNamespaces (a false namespaced=true would list every namespace).
+func (c *ResourceScopeCache) Scope(group, resource string) (namespaced, known bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	namespaced, known = c.scopeMap[group+"/"+resource]
+	return namespaced, known
 }
 
 // HasResource reports whether the discovery snapshot serves the resource at
--- a/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/scope_cache.go
+++ b/ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/scope_cache.go
@@ -85,11 +85,11 @@ func (c *ResourceScopeCache) IsNamespaced(group, resource string) bool {
 // multi-tenancy uses !known as "treat like namespaced" so a discovery hole
 // cannot fail-open. IsNamespaced stays fail-closed-as-cluster-scoped for
 // AccessibleNamespaces (a false namespaced=true would list every namespace).
-func (c *ResourceScopeCache) Scope(group, resource string) (namespaced, known bool) {
+func (c *ResourceScopeCache) Scope(group, resource string) (bool, bool) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 
-	namespaced, known = c.scopeMap[group+"/"+resource]
+	namespaced, known := c.scopeMap[group+"/"+resource]
 	return namespaced, known
 }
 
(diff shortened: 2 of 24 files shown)


ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/scope_cache.go
```diff
@@ -76,6 +80,19 @@ func (c *ResourceScopeCache) IsNamespaced(group, resource string) bool {
 	return namespaced
 }
 
+// Scope reports whether the resource is namespaced and whether the snapshot
+// contains it. Unlike IsNamespaced, a miss is not coerced to cluster-scoped:
+// multi-tenancy uses !known as "treat like namespaced" so a discovery hole
+// cannot fail-open. IsNamespaced stays fail-closed-as-cluster-scoped for
+// AccessibleNamespaces (a false namespaced=true would list every namespace).
+func (c *ResourceScopeCache) Scope(group, resource string) (bool, bool) {
```
AlwxSin, line 91: 🟡 **Finding 1 — a partial discovery failure evicts a group from the scope map, and that now decides Deny**

Real location: `refresh` at `scope_cache.go:207-268`, which this PR does not touch — commenting here because `Scope` is what promotes that map from a reporting aid to an authorize input.

`refresh` rebuilds `scopeMap` from `ServerPreferredResources()` and swaps it in wholesale. On `ErrGroupDiscoveryFailed` client-go returns partial results — the failed GroupVersions are simply absent (`discovery_client.go`: `apiResourceList, ok := groupVersionResources[groupVersion]; if !ok { continue }`) — and the only preservation guard is `len(newMap) == 0`, i.e. total failure. Granularity is per GroupVersion, not per group: `fetchGroupVersionResources` issues one `ServerResourcesForGroupVersion` per GV and records each failure separately, so a group serving both `v1` and `v1beta1` survives one failing version — but an unreachable APIService fails every version it serves, and the usual aggregated API serves one.

So entries that were present a minute ago disappear, while the comment above the function promises "On error, the existing cache is preserved (stale data is better than no data)". That was true while the map only fed `IsNamespaced`, where a miss is the safe direction. It is no longer true now that a miss feeds `Scope` on the authorize path: while any one APIService is unavailable — a down metrics-server, a broken aggregated API, a CRD group mid-restart — the console reports Deny for that group's genuinely cluster-scoped resources, for every user with `limitNamespaces` or `namespaceSelector`, until the group comes back.

The direction is the accepted trade (before this PR the same situation produced a false Allow), but this is a different case from the one the description documents: "a new cluster-scoped CRD ... up to the 5-minute scope-cache refresh" is bounded by the refresh interval, and this is not. Note the webhook this component mirrors goes the other way explicitly (`webhook/src/internal/cache/cache.go:439-446`): on a renewal error, "if there is an error, we could just use stale cache".

Fix is local to `refresh`: merge into the existing map, or overwrite only the group/versions actually present in `resourceLists`, instead of replacing it. Eviction then requires a successful discovery of that GV — which is also how a genuinely deleted CRD disappears, since a removed API group drops out of `ServerGroups()` rather than failing per-version.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/apiserver/apiserver.go
```diff
@@ -154,7 +154,7 @@ func initAuthorizers(init *initResult, configPath string, scopeCache *resolver.R
 			configPath,
 			init.informerFactory.Core().V1().Namespaces().Lister(),
 			init.informerFactory.Core().V1().Namespaces().Informer().HasSynced,
-			init.clientset.Discovery(),
+			scopeCache,
```
AlwxSin: 🔵 **Finding 2 — `NewEngine`'s nil-safety promise does not survive a typed nil**

`NewEngine`'s doc says "resourceScope may be nil: every lookup then reports !known". This call passes `scopeCache` — a `*resolver.ResourceScopeCache` — straight into the `ResourceScope` interface parameter, so a nil pointer would arrive as a non-nil interface holding nil, `e.resourceScope != nil` would be true, and `Scope` / `HasData` would panic on `c.mu.RLock()` rather than report `!known`.

Unreachable today: `scopeCache` is nil only when `initRes.clientset == nil`, and `mtEngine` is built under the same condition, so `NewEngine` is never reached on that path. Raising it because the guard two lines above does handle exactly this for `registry` (`if scopeCache != nil { registry = scopeCache }`), and `TestEngine_Authorize_NilScopeIsDenied` covers the literal nil rather than the typed one — so nothing would catch it if the two conditions ever drift apart. Either reuse the same guard here or make `Scope` / `HasData` nil-receiver-safe.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/scope_cache.go
```diff
@@ -76,6 +80,19 @@ func (c *ResourceScopeCache) IsNamespaced(group, resource string) bool {
...
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	namespaced, known := c.scopeMap[group+"/"+resource]
+	return namespaced, known
```
AlwxSin, line 100: 🔵 **Finding 3 — out of scope, but this PR makes it matter: `HasNamespacedResourceMatching` has no subresources to match**

Real location: `HasNamespacedResourceMatching` / `matchesResource` at `scope_cache.go:118-160`, pre-existing and untouched here.

`ServerPreferredResources` drops every resource whose name contains "/" (`discovery_client.go`: `if strings.Contains(apiResource.Name, "/") { continue }`), so `scopeMap` never holds a subresource key. `matchesResource`'s `"*/" + subresource` branch can therefore never fire, and the doc comment "Both top-level resources and subresources participate so wildcard rules are evaluated with the same semantics as Kubernetes RBAC" does not describe what the code does.

It does not affect `Scope`: `attrs.GetResource()` is always the top-level resource, with the subresource in its own field. Mentioning it only because this PR promotes `scopeMap` from a reporting aid to an authorize input, so its gaps are worth knowing — either drop the subresource branch and correct the comment, or populate subresources from `ServerGroupsAndResources` if the wildcard semantics were actually wanted.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/scope_cache.go
Jabejixo: Confirmed and fixed in the last commit. `refresh` now reads `ServerGroupsAndResources`, takes the failed GroupVersions from the `ErrGroupDiscoveryFailed` it returns and keeps the previous entries of those groups, replaces the groups discovery did return, and drops only the groups that are neither returned nor failed (a removed API group). Any other error, or an empty result, preserves the whole snapshot, so the comment above the function is true again. Covered by `TestRefresh_PartialDiscoveryError_KeepsTheFailedGroups` and `TestRefresh_NonDiscoveryError_PreservesCache`.

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/apiserver/apiserver.go
Jabejixo: Fixed: `initAuthorizers` passes a nil `multitenancy.ResourceScope` when the cache is nil, the same guard the registry already had, and `Scope`/`HasData`/`HasResource` answer like an empty cache on a nil receiver as well (`TestNilCache_AnswersLikeEmpty`).

ee/be/modules/140-user-authz/images/permission-browser-apiserver/src/pkg/resolver/scope_cache.go
Jabejixo: You are right, `ServerPreferredResources` drops every name with a `/`, so the `*/status` branch never fired and the existing subresource test only passed because of the mock. Since the wildcard semantics were the intent, the cache now reads `ServerGroupsAndResources`, which lists subresources (and every version), so `HasNamespacedResourceMatching` sees what RBAC sees and the comment is accurate. `Scope` is unaffected: it is called with the top-level resource. The branch is also rebased onto main so the fuzz-images job can find the images it looks for.

AlwxSin (approved):

PR #22757, opened 2026-09-02
[admission-policy-engine] Add operationPolicy for GPU resorces

Suselz:
## Description

Adds a new `gpuResourceRestriction` policy to the `OperationPolicy` resource.

The policy denies Pods that request GPU resources unless the target namespace carries a configurable label. Both `resources.requests` and `resources.limits` of every container, init container, and ephemeral container are inspected; the set of resource names treated as GPU resources is a configurable list of regex patterns (default `^nvidia\.com/.*$`).

Implementation:

- `OperationPolicy` CRD (EN + RU) — new `policies.gpuResourceRestriction` section with `namespaceLabel` (`key`/`value`, defaults `gpu.deckhouse.io/enabled: "true"`) and `gpuResourcePatterns`.
- New Gatekeeper `ConstraintTemplate` `D8GpuResourceRestriction`; the namespace label is read from `data.inventory`, so `Namespace` was added to the Gatekeeper `Config` `syncOnly` list.
- Helm rendering of the `D8GpuResourceRestriction` constraint from the CR, Go API struct for the hook.
- Tests: gator constraint suite (16 cases, 2 blocks, 100% field coverage), hook tests, helm render expectations.
- Docs: module FAQ (EN + RU).

The policy fails closed: while a namespace is missing from the Gatekeeper cache (e.g. right after the Gatekeeper pods restart), Pods requesting GPU resources in that namespace are denied. This is documented, and the violation message distinguishes it from a missing label.

GPU nodes are a scarce and expensive resource. Without an admission-level guard, any team can schedule GPU workloads in any namespace they have access to. This policy lets a cluster administrator explicitly allow-list namespaces for GPU usage with a single label, instead of relying on quotas per namespace or on convention.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: feature
summary: Added the `gpuResourceRestriction` policy to `OperationPolicy`, which denies Pods requesting GPU resources in namespaces without the configured label.
```


modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/gpu-resource-restriction.yaml
```diff
@@ -0,0 +1,131 @@
...
+
+              violation contains {"msg": msg} if {
+                not input.review.operation == "DELETE"
+                namespace_in_cache
+                not namespace_allows_gpu
+                container := input_containers[_]
+                resources := gpu_resources(container)
+                count(resources) > 0
+                msg := sprintf(
+                  "container <%v> requests GPU resources <%v>, but namespace <%v> does not have label <%v: %v>",
```
AlwxSin: 🟡 **Denial message lies when the label is present with a different value**

`namespace_allows_gpu` requires `key == value`, but the message says `namespace <X> does not have label <key: value>`. For a namespace carrying `gpu.deckhouse.io/enabled: "false"` the operator is told the label is missing and goes looking in the wrong place.

The test matrix bakes this in: `tests/test_cases/constraints/operation/gpu-resource-restriction/test-matrix.yaml:114` asserts "does not have label" for `namespaceWithWrongGpuLabelValue`.

Suggestion: `namespace <%v> does not have label <%v> set to <%v>`, or report the value actually found.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/gpu-resource-restriction.yaml
```diff
@@ -0,0 +1,131 @@
...
+                object.get(data.inventory, ["cluster", "v1", "Namespace", namespace_name], null) != null
+              }
+
+              namespace_allows_gpu if {
+                namespace_labels[input.parameters.namespaceLabel.key] == input.parameters.namespaceLabel.value
+              }
+
+              # Resource names requested by a container (both requests and limits) that match
+              # any of the configured GPU patterns.
+              gpu_resources(container) := names if {
```
AlwxSin, line 94: 🟡 **Zero-quantity GPU resources are denied**

`gpu_resources()` only looks at resource *names*, never quantities. `resources.limits: {nvidia.com/gpu: 0}` is valid Kubernetes (extended resources must be non-negative integers) and means "no GPU" — the device plugin allocates nothing — but the policy denies the Pod. This is a common Helm-chart shape: `nvidia.com/gpu: {{ .Values.gpuCount | default 0 }}`.

Trap for the fix: in a real `AdmissionReview` quantities are serialized as **strings** (`"0"`), while the gator matrix writes them as YAML **ints** (`nvidia.com/gpu: 1`). A guard written as `v != 0` passes the tests and silently does nothing in production; `v != "0"` does the reverse. Normalize both and add a gator case with a string quantity.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/gpu-resource-restriction.yaml
```diff
@@ -0,0 +1,131 @@
...
+              # admission, but fall back to review.namespace for requests that omit it.
+              namespace_name := name if {
+                name := object.get(input.review, ["object", "metadata", "namespace"], "")
+                name != ""
+              } else := object.get(input.review, "namespace", "")
+
+              # Namespace labels taken from the audit cache (requires Namespace in the Gatekeeper
+              # Config syncOnly list). Missing namespace yields an empty label set, so the policy
+              # fails closed.
+              namespace_labels := object.get(
```
AlwxSin, line 70: 🟡 **Reading the namespace label from `data.inventory` reintroduces a cache-lag failure that Gatekeeper's own matcher guards against**

The fail-closed branch (lines 116-131) fires whenever the namespace has not reached the OPA store yet. Checked against the Gatekeeper this module ships (`oss.yaml` → 3.22.2):

- `data.inventory` **is** populated in the webhook pods — the cache manager and config controller are not gated by `--operation` (`main.go:533`). The mechanism works.
- But the OPA store is watch-driven with no fallback. Gatekeeper's own `match.namespaceSelector` path resolves the namespace via the cached client and, on a cache miss, **bypasses the cache and reads the API server directly** (`pkg/webhook/policy.go:687-699`). The inventory path has no such escape hatch.

Practical window: a namespace created and a GPU Pod applied in the same `kubectl apply -f ./manifests` / GitOps sync, before the sync controller observes the namespace. A Gatekeeper pod restart is *not* a real window — the readiness tracker holds a per-GVK `ForData` expectation (`pkg/readiness/ready_tracker.go:195,910`), so `/readyz` gates it. Blast radius is also bounded: the main validating webhook excludes `heritage: deckhouse` namespaces (`templates/validatingwebhookconfiguration.yaml:69-74`), so system namespaces cannot be hit.

Alternative worth weighing: drop the inventory lookup and render the negation into the constraint's own selector — `matchExpressions: [{key: <namespaceLabel.key>, operator: NotIn, values: ["<value>"]}]` merged into `constraint_selector`. Costs ~10 lines of selector-merging Helm and a less specific deny message; removes the failure mode, the `templates/config.yaml` `syncOnly` change, and half the Rego.

Not a blocker either way — but if the inventory approach stays, the fail-closed choice deserves to be a conscious, documented decision rather than a side effect.

modules/015-admission-policy-engine/crds/operation-policy.yaml
```diff
@@ -215,6 +215,49 @@ spec:
...
+
+                        Pods that request GPU resources are denied unless the namespace has the specified label.
+                        Both `resources.requests` and `resources.limits` of every container, init container,
+                        and ephemeral container are inspected.
+
+                        The namespace label is read from the Gatekeeper cache. While the namespace is missing
+                        from that cache, for example right after the Gatekeeper pods restart, the policy denies
+                        the Pods that request GPU resources in that namespace.
+                      properties:
+                        namespaceLabel:
```
AlwxSin, line 232: 🔵 **Default values live in two places and can drift**

`namespaceLabel.key`/`value` carry CRD `default`s, but `namespaceLabel` itself has no `default: {}` — so those defaults only apply if the user explicitly writes `namespaceLabel: {}`. In every other case the effective default comes from `dig` in `templates/policies/operation-policy/constraint.yaml:406-408`. Two sources of truth for the same constant.

Either add `default: {}` on `namespaceLabel` (the CRD becomes self-describing and `dig` becomes belt-and-braces), or drop the sub-defaults and keep only the Helm ones.

modules/015-admission-policy-engine/templates/config.yaml
```diff
@@ -13,3 +13,7 @@ spec:
       - group: ""
         version: "v1"
         kind: "Pod"
+{{/*  Namespaces are looked up via data.inventory by the D8GpuResourceRestriction constraint template */}}
```
AlwxSin: 🔵 **Stray blank line from the Helm comment**

`{{/* ... */}}` without dashes leaves an empty line inside the `syncOnly` list in the rendered manifest. Valid YAML, just untidy — `{{- /* ... */}}` or a plain `#` comment.

Also worth noting: no template test asserts the `syncOnly` contents, so nothing catches a regression there.

modules/015-admission-policy-engine/charts/constraint-templates/tests/test_cases/constraints/operation/gpu-resource-restriction/test-matrix.yaml
```diff
@@ -0,0 +1,482 @@
...
+            merge:
+              metadata:
+                name: allowed-no-resources
+                namespace: testns
+              spec:
+                containers:
+                  - name: cpu-app
+                    image: my.repo/app:latest
+        # Allowed: resource name that does not match the configured patterns
+        - name: example-allowed-non-matching-resource-name
```
AlwxSin, line 235: 🔵 **The documented unanchored-regex footgun has no test**

Both CRDs and the FAQ warn that `nvidia.com/gpu` also matches `not-nvidia.com/gpu`. This suite only tests the *anchored* pattern doing the right thing. A case with an unanchored pattern actually matching `not-nvidia.com/gpu` would pin the documented behaviour.

Same for `namespaceLabel.key: ""` — neither the CRD nor the ConstraintTemplate schema sets `minLength`, and an empty key renders a constraint that denies every GPU Pod unconditionally.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/gpu-resource-restriction.yaml
Suselz: Fixed. The message now reads `... but namespace <X> does not have label <key> set to <value>`, which is accurate both when the label is absent and when it carries a different value.

Test matrix assertions updated accordingly, including `example-disallowed-gpu-wrong-label-value` for `namespaceWithWrongGpuLabelValue`.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/gpu-resource-restriction.yaml
Suselz: Fixed. `gpu_resources()` now skips resources whose quantity normalizes to zero:

```rego
zero_quantity(quantity) if {
  to_number(quantity) == 0
}
```

`to_number` accepts both `0` and `"0"`, so the numeric form in the gator fixtures and the string form in a real `AdmissionReview` behave identically — which is exactly the trap you flagged. A quantity that cannot be parsed leaves the rule undefined, so `not zero_quantity(...)` holds and the resource still counts as a GPU request; the policy stays fail-closed on malformed input.

Three gator cases added to pin both directions:

- `example-allowed-zero-quantity-number` — `nvidia.com/gpu: 0`, allowed.
- `example-allowed-zero-quantity-string` — `nvidia.com/gpu: "0"`, allowed.
- `example-disallowed-quantity-string` — `nvidia.com/gpu: "1"`, denied.

modules/015-admission-policy-engine/charts/constraint-templates/templates/operation/gpu-resource-restriction.yaml
Suselz: Thanks for digging into the Gatekeeper internals — the correction about the restart window is accepted, and the readiness-tracker detail is something we had wrong.

Decision: the `data.inventory` lookup stays, as a conscious choice rather than a side effect.

Reasoning:

- Fail-closed is the intended behaviour for a deny policy on a scarce resource. The selector approach fails open: on a cache miss the constraint simply does not match and the Pod is admitted, which is the opposite of what this policy is for.
- The cache-miss state is diagnosable. The second violation rule emits a distinct message (`namespace <X> is not present in the Gatekeeper cache, so the label <key> cannot be checked`), so an operator hitting the window sees the cause instead of a misleading "label missing". The selector approach cannot report anything at all.
- The blast radius you measured — bounded by the `heritage: deckhouse` namespace exclusion — makes the residual risk acceptable for the window that actually exists.

What changed as a result of this comment: the documented window was wrong. The CRD descriptions (EN + RU) and the module FAQ said "right after the Gatekeeper pods restart"; they now describe the real case — a namespace and a Pod applied together before the namespace reaches the cache.

If the fail-open trade-off is preferred, the selector variant is a contained change and can be done in a follow-up; happy to revisit.

modules/015-admission-policy-engine/crds/operation-policy.yaml
Suselz: Fixed with the first option: `namespaceLabel` now carries `default: {}`, so the API server materializes `key` and `value` whenever `gpuResourceRestriction` is present, and the CRD is self-describing.

`dig` in `templates/policies/operation-policy/constraint.yaml` is kept as a guard for objects stored before this change, rather than as a second source of truth.

Also added `minLength: 1` to `namespaceLabel.key` in both the CRD and the ConstraintTemplate schema, per the related note in the test-matrix thread.

modules/015-admission-policy-engine/templates/config.yaml
Suselz: Fixed: the comment is now `{{- /* ... */}}`, so no blank line is left inside `syncOnly`.

On the missing coverage: added a spec to `template_tests/operation_policies_test.go` asserting the rendered `Config` — three entries in `spec.sync.syncOnly`, with `Namespace` and `Pod` at `v1` and `SecurityPolicyException` in `deckhouse.io`.

One caveat: that `Describe` still carries the pre-existing `Skip("legacy helm-render specs are isolated after constraint test runner migration")`, so the spec does not run in CI today. Verified locally with the skip lifted — 16/16 specs pass, then the skip was restored untouched.

modules/015-admission-policy-engine/charts/constraint-templates/tests/test_cases/constraints/operation/gpu-resource-restriction/test-matrix.yaml
Suselz: Both covered.

Unanchored regex: added a third gator block `operation-policy-unanchored-pattern` with `constraints/policy_3.yaml`, which uses the unanchored `nvidia\.com/gpu` pattern.

- `example-disallowed-unanchored-pattern-lookalike` asserts it denies a Pod requesting `not-nvidia.com/gpu`, pinning exactly the behaviour the docs warn about.
- `example-allowed-unanchored-pattern-with-label` shows the namespace label still governs for that look-alike name.

Empty label key: `minLength: 1` added to `namespaceLabel.key` in the CRD and in the ConstraintTemplate `openAPIV3Schema`, so a constraint that denies every GPU Pod unconditionally can no longer be created.

Suite after the changes: 21 cases across 3 blocks, and `run_all_tests.sh` reports 40/40 constraints, 117/117 OPA tests, coverage 100%.

AlwxSin (approved):

modules/015-admission-policy-engine/docs/FAQ_RU.md
```diff
@@ -640,3 +640,68 @@ spec:
 - Используйте `input.review.operation == "CONNECT"` для проверки операций `CONNECT`.
 - Информация о пользователе доступна в `input.review.userInfo.username` и `input.review.userInfo.groups`.
 - Неймспейс доступен в `input.review.namespace`.
+
+## Как ограничить использование GPU-ресурсов в неймспейсах?
+
+Политика `gpuResourceRestriction` в ресурсе [OperationPolicy](/modules/admission-policy-engine/cr.html#operationpolicy)
```
z9r5:
```suggestion
Политика `gpuResourceRestriction` в [OperationPolicy](cr.html#operationpolicy)
```

modules/015-admission-policy-engine/docs/FAQ.md
```diff
@@ -640,3 +640,67 @@ Key data and checks available when validating `CONNECT` operations:
 - Use `input.review.operation == "CONNECT"` to check for `CONNECT` operations.
 - User information is available in `input.review.userInfo.username` and `input.review.userInfo.groups`.
 - The namespace is available in `input.review.namespace`.
+
+## How do I restrict GPU resource usage in namespaces?
+
+The `gpuResourceRestriction` policy in [OperationPolicy](/modules/admission-policy-engine/cr.html#operationpolicy)
```
z9r5:
```suggestion
The `gpuResourceRestriction` policy in [OperationPolicy](cr.html#operationpolicy)
```

modules/015-admission-policy-engine/crds/doc-ru-operation-policy.yaml
```diff
@@ -93,6 +93,34 @@ spec:
...
+                      description: |
+                        Ограничивает использование GPU-ресурсов в подах.
+
+                        Поды, запрашивающие GPU-ресурсы, отклоняются, если на неймспейсе не установлен
+                        указанный лейбл. Проверяются `resources.requests` и `resources.limits` всех
+                        контейнеров, init-контейнеров и ephemeral-контейнеров пода.
+
+                        Лейбл неймспейса читается из кэша Gatekeeper. Пока неймспейс отсутствует в этом
+                        кэше, например когда неймспейс и под применяются одновременно и неймспейс ещё
+                        не попал в кэш, политика запрещает поды, запрашивающие GPU-ресурсы в этом неймспейсе.
```
z9r5:
```suggestion
                        Лейбл неймспейса читается из кеша Gatekeeper. Пока неймспейс отсутствует в этом
                        кеше, например когда неймспейс и под применяются одновременно и неймспейс ещё
                        не попал в кеш, политика запрещает поды, запрашивающие GPU-ресурсы в этом неймспейсе.
```

z9r5:
```suggestion
                        Лейбл неймспейса читается из кеша Gatekeeper. Пока неймспейс отсутствует в этом
                        кеше, например когда неймспейс и под применяются одновременно и неймспейс ещё
                        не попал в кеш, политика запрещает поды, запрашивающие GPU-ресурсы в этом неймспейсе.
```

modules/015-admission-policy-engine/docs/FAQ_RU.md
```diff
@@ -640,3 +640,68 @@ spec:
...
+После этого поды, запрашивающие GPU-ресурсы, будут разрешены только в неймспейсах с лейблом
+`gpu.deckhouse.io/enabled: "true"`.
+
+{% alert level="warning" %}
+Политика применяется к подам. Deployment или другой контроллер, запрашивающий GPU-ресурсы, создаётся
+успешно, а отказ отображается в событиях ReplicaSet.
+
+Лейбл неймспейса читается из кэша Gatekeeper. Пока неймспейс отсутствует в этом кэше, например когда
+неймспейс и под применяются одновременно и неймспейс ещё не попал в кэш, политика запрещает поды,
+запрашивающие GPU-ресурсы в этом неймспейсе.
```
z9r5:
```suggestion
Лейбл неймспейса читается из кеша Gatekeeper. Пока неймспейс отсутствует в этом кеше, например когда
неймспейс и под применяются одновременно и неймспейс ещё не попал в кеш, политика запрещает поды,
запрашивающие GPU-ресурсы в этом неймспейсе.
```

modules/015-admission-policy-engine/docs/FAQ.md
```diff
@@ -640,3 +640,67 @@ Key data and checks available when validating `CONNECT` operations:
...
+denies Pod creation when a Pod requests GPU resources and the namespace has no label allowing GPU usage.
+Both `resources.requests` and `resources.limits` of every container, init container,
+and ephemeral container are inspected. A resource with the quantity of `0` is not treated as a GPU request.
+
+To set up GPU resource restriction:
+
+1. Create an `OperationPolicy` resource with the `gpuResourceRestriction` setting.
```
z9r5:
```suggestion
1. Create an OperationPolicy with the `gpuResourceRestriction` setting.
```

modules/015-admission-policy-engine/docs/FAQ_RU.md
```diff
@@ -640,3 +640,68 @@ spec:
...
+## Как ограничить использование GPU-ресурсов в неймспейсах?
+
+Политика `gpuResourceRestriction` в ресурсе [OperationPolicy](/modules/admission-policy-engine/cr.html#operationpolicy)
+запрещает создание пода, если под запрашивает GPU-ресурсы, а на неймспейсе не установлен лейбл,
+разрешающий использование GPU. Проверяются `resources.requests` и `resources.limits` всех контейнеров,
+init-контейнеров и ephemeral-контейнеров пода. Ресурс с количеством `0` не считается запросом GPU.
+
+Чтобы настроить ограничение использования GPU-ресурсов, выполните следующие шаги:
+
+1. Создайте ресурс `OperationPolicy` с параметром `gpuResourceRestriction`.
```
z9r5:
```suggestion
1. Создайте OperationPolicy с параметром `gpuResourceRestriction`.
```

z9r5: Ну или можно "создайте объект". Up to you. Ресурс имхо не очень корректно и можно опустить, по аналогии с создайте Deployment, создайте ConfigMap

modules/015-admission-policy-engine/docs/FAQ_RU.md
```diff
@@ -640,3 +640,68 @@ spec:
 - Используйте `input.review.operation == "CONNECT"` для проверки операций `CONNECT`.
 - Информация о пользователе доступна в `input.review.userInfo.username` и `input.review.userInfo.groups`.
 - Неймспейс доступен в `input.review.namespace`.
+
+## Как ограничить использование GPU-ресурсов в неймспейсах?
+
+Политика `gpuResourceRestriction` в ресурсе [OperationPolicy](/modules/admission-policy-engine/cr.html#operationpolicy)
+запрещает создание пода, если под запрашивает GPU-ресурсы, а на неймспейсе не установлен лейбл,
```
z9r5: Тут (и в EN) судя по gpu-resource-restriction.yaml:109 должно быть более жесткое "запрещает поды", но тут тебе виднее. Просто подсветил

modules/015-admission-policy-engine/docs/FAQ.md
```diff
@@ -640,3 +640,67 @@ Key data and checks available when validating `CONNECT` operations:
 - Use `input.review.operation == "CONNECT"` to check for `CONNECT` operations.
 - User information is available in `input.review.userInfo.username` and `input.review.userInfo.groups`.
 - The namespace is available in `input.review.namespace`.
+
+## How do I restrict GPU resource usage in namespaces?
+
+The `gpuResourceRestriction` policy in [OperationPolicy](/modules/admission-policy-engine/cr.html#operationpolicy)
+denies Pod creation when a Pod requests GPU resources and the namespace has no label allowing GPU usage.
+Both `resources.requests` and `resources.limits` of every container, init container,
+and ephemeral container are inspected. A resource with the quantity of `0` is not treated as a GPU request.
```
z9r5:
```suggestion
and ephemeral container are inspected. A resource with a quantity of `0` is not treated as a GPU request.
```

modules/015-admission-policy-engine/docs/FAQ.md
```diff
@@ -640,3 +640,67 @@ Key data and checks available when validating `CONNECT` operations:
 - Use `input.review.operation == "CONNECT"` to check for `CONNECT` operations.
 - User information is available in `input.review.userInfo.username` and `input.review.userInfo.groups`.
 - The namespace is available in `input.review.namespace`.
+
+## How do I restrict GPU resource usage in namespaces?
+
+The `gpuResourceRestriction` policy in [OperationPolicy](/modules/admission-policy-engine/cr.html#operationpolicy)
+denies Pod creation when a Pod requests GPU resources and the namespace has no label allowing GPU usage.
```
z9r5:
```suggestion
denies pod creation when a pod requests GPU resources and the namespace has no label allowing GPU usage.
```

modules/015-admission-policy-engine/docs/FAQ.md
```diff
@@ -640,3 +640,67 @@ Key data and checks available when validating `CONNECT` operations:
...
+After that, the Pods that request GPU resources are allowed only in the namespaces
+with the `gpu.deckhouse.io/enabled: "true"` label.
+
+{% alert level="warning" %}
+The policy is applied to Pods. A Deployment or another controller that requests GPU resources
+is created successfully, and the denial is reported in the ReplicaSet events.
+
+The namespace label is read from the Gatekeeper cache. While a namespace is missing from that cache,
+for example when the namespace and the Pod are applied together and the namespace has not been cached yet,
+the policy denies the Pods that request GPU resources in that namespace.
```
z9r5:
```suggestion
the policy denies the pods that request GPU resources in that namespace.
```

modules/015-admission-policy-engine/crds/operation-policy.yaml
```diff
@@ -215,6 +215,52 @@ spec:
                         maxReplicas:
                           description: "The maximum number of replicas allowed, inclusive."
                           type: integer
+                    gpuResourceRestriction:
+                      type: object
+                      description: |
+                        Restricts GPU resource usage in Pods.
```
z9r5:
```suggestion
                        Restricts GPU resource usage in pods.
```

modules/015-admission-policy-engine/crds/operation-policy.yaml
```diff
@@ -215,6 +215,52 @@ spec:
...
+
+                        Pods that request GPU resources are denied unless the namespace has the specified label.
+                        Both `resources.requests` and `resources.limits` of every container, init container,
+                        and ephemeral container are inspected.
+
+                        The namespace label is read from the Gatekeeper cache. While the namespace is missing
+                        from that cache, for example when a namespace and a Pod are applied together and the
+                        namespace has not been cached yet, the policy denies the Pods that request GPU resources
```
z9r5:
```suggestion
                        Pods that request GPU resources are denied unless the namespace has the specified label.
                        Both `resources.requests` and `resources.limits` of every container, init container,
                        and ephemeral container are inspected.

                        The namespace label is read from the Gatekeeper cache. While the namespace is missing
                        from that cache, for example when a namespace and a pod are applied together and the
                        namespace has not been cached yet, the policy denies the pods that request GPU resources
```

modules/015-admission-policy-engine/crds/operation-policy.yaml
```diff
@@ -215,6 +215,52 @@ spec:
...
+                              minLength: 1
+                              description: "Label key that must be present on the namespace."
+                            value:
+                              type: string
+                              default: "true"
+                              description: "Expected label value."
+                        gpuResourcePatterns:
+                          type: array
+                          description: |
+                            List of regular expressions matched against GPU resource names in Pod specs.
```
z9r5:
```suggestion
                            List of regular expressions matched against GPU resource names in pod specs.
```

modules/015-admission-policy-engine/docs/FAQ.md
```diff
@@ -640,3 +640,67 @@ Key data and checks available when validating `CONNECT` operations:
...
+   d8 k label ns my-namespace operation-policy.deckhouse.io/enabled=true
+   ```
+
+1. Add the GPU label to the namespaces where GPU usage is allowed.
+
+   ```shell
+   d8 k label ns my-gpu-namespace gpu.deckhouse.io/enabled=true
+   ```
+
+After that, the Pods that request GPU resources are allowed only in the namespaces
```
z9r5:
```suggestion
After that, the pods that request GPU resources are allowed only in the namespaces
```

modules/015-admission-policy-engine/docs/FAQ.md
```diff
@@ -640,3 +640,67 @@ Key data and checks available when validating `CONNECT` operations:
...
+with the `gpu.deckhouse.io/enabled: "true"` label.
+
+{% alert level="warning" %}
+The policy is applied to Pods. A Deployment or another controller that requests GPU resources
+is created successfully, and the denial is reported in the ReplicaSet events.
+
+The namespace label is read from the Gatekeeper cache. While a namespace is missing from that cache,
+for example when the namespace and the Pod are applied together and the namespace has not been cached yet,
```
z9r5:
```suggestion
The policy is applied to pods. A Deployment or another controller that requests GPU resources
is created successfully, and the denial is reported in the ReplicaSet events.

The namespace label is read from the Gatekeeper cache. While a namespace is missing from that cache,
for example when the namespace and the pod are applied together and the namespace has not been cached yet,
```

z9r5 (commented):
Comments

modules/015-admission-policy-engine/docs/FAQ_RU.md
Suselz: Замечание верное, поправил в обеих версиях.

Проверил по коду: в правиле `violation` исключается только `DELETE`, а вебхук для отслеживаемых ресурсов перехватывает `CREATE`, `UPDATE` и `DELETE` (`validatingwebhookconfiguration.yaml:21-28`), плюс отдельным правилом `UPDATE` на `pods/ephemeralcontainers` (`validatingwebhookconfiguration.yaml:89-92`). То есть политика срабатывает и при изменении пода, в том числе при подключении ephemeral-контейнера. Сверх того, audit сообщает о нарушениях для уже запущенных подов — при проверке на стенде снятие лейбла с неймспейса дало `totalViolations=1` по существующему поду.

Так что «запрещает создание пода» описывало только часть поведения. Теперь:

- RU: «запрещает поды, запрашивающие GPU-ресурсы, если на неймспейсе не установлен лейбл... Проверка выполняется при создании и изменении пода, в том числе при добавлении ephemeral-контейнера».
- EN: «denies the Pods that request GPU resources when the namespace has no label allowing GPU usage. The check runs on Pod creation and update, including when an ephemeral container is added».

Заодно ушло расхождение с описаниями в CRD — там уже была формулировка «Поды, запрашивающие GPU-ресурсы, отклоняются...» / «Pods that request GPU resources are denied...».

z9r5 (approved):

PR #22758, opened 2026-09-02
[admission-policy-engine] Validate Gatekeeper content restrictions on mutator resources

Suselz:
## Description

The `Assign`, `AssignMetadata`, `ModifySet` and `AssignImage` CRDs in `admission-policy-engine` are imported from upstream Gatekeeper as-is — a build-time check (`images/gatekeeper/werf.inc.yaml`) diffs `crds/gatekeeper/*.yaml` against a fresh upstream checkout, so their OpenAPI schema cannot be edited directly. Gatekeeper's own mutation webhook still enforces several content restrictions that aren't expressed in that schema, so invalid resources currently pass `kubectl apply` and only fail later at the Gatekeeper webhook, sometimes with a confusing parser error, and sometimes (missing `match.kinds`) with **no error at all** — the mutation is just silently never invoked.

This PR adds a new `templates/admission-mutators-validation.yaml` with `ValidatingAdmissionPolicy`/`ValidatingAdmissionPolicyBinding` pairs (matching the existing pattern in `templates/admission.yaml`) that reject such resources immediately, with a clear message:

- `deny-invalid-assign-location.deckhouse.io` — `Assign.spec.location` cannot target the `metadata` field at all, quoted or not; use `AssignMetadata` for labels/annotations.
- `deny-invalid-assignmetadata-location.deckhouse.io` — `AssignMetadata.spec.location` must match Gatekeeper's own location grammar for `metadata.labels.<KEY>`/`metadata.annotations.<KEY>` (whitespace and quoting allowed the way Gatekeeper's scanner allows them; a bare key can't start with a digit).
- `deny-invalid-assignmetadata-value.deckhouse.io` — `AssignMetadata.spec.parameters.assign.value` must be a plain string.
- `deny-invalid-mutator-frommetadata-field.deckhouse.io` — `spec.parameters.assign.fromMetadata.field` is mandatory once `fromMetadata` is set, and only accepts `namespace` or `name`.
- `deny-invalid-assignmetadata-externaldata-datasource.deckhouse.io` — `AssignMetadata.spec.parameters.assign.externalData.dataSource` only supports `Username`; the message says explicitly that the field defaults to `ValueAtLocation` when omitted and that default is rejected too, since CRD defaulting runs before this policy sees the object.
- `deny-mutator-without-match-kinds.deckhouse.io` — `spec.match.kinds` is required, non-empty, and every entry must set non-empty, non-wildcard `apiGroups`/`kinds` on all four mutator kinds. This one is Deckhouse-specific: `images/constraint-exporter` builds the `d8-admission-policy-engine-config` MutatingWebhookConfiguration rules from `match.kinds` by resolving `apiGroups`×`kinds` through the REST mapper and silently dropping anything that doesn't resolve — so a schema-valid entry that doesn't resolve to a concrete resource type never fires, even though Gatekeeper itself accepts it. Traced from a real user report where an `Assign` without `match.kinds` produced zero errors, zero events, and just never mutated anything.

Policy names carry a `.deckhouse.io` suffix (cluster-scoped registry, so an unqualified name could collide with a user's own VAP and fail the Helm release), and every expression is guarded against a missing `spec` (not `required` in the vendored CRDs, so an object without one is schema-valid and would otherwise crash the CEL with `no such key: spec` instead of denying cleanly).

**Testing**: every expression is exercised by a table test in the new `template_tests/admission_mutators_validation_cel_test.go`, which reads `spec.validations.0.expression` back out of the *rendered* template and compiles/evaluates it with `cel-go` (already a direct dependency) against valid and invalid objects — including every gap called out in review (whitespace/quoting in `location`, a leading-digit key, `fromMetadata: {}`, wildcard `kinds`, a request with no `spec` at all). `admission_vap_test.go` keeps the structural checks (names, `resourceRules`, `failurePolicy`) so a broken CEL string and a broken policy shape fail in the test that actually explains which one broke.

Documented in `docs/FAQ.md` / `docs/FAQ_RU.md`, cross-linked to `docs/GATEKEEPER-CR.md`'s Mutation Custom Resources section (`GATEKEEPER-CR.md` itself is not an allowed `docs/` file name for modules per `tools/validation`, so it can't carry new content — see the linked FAQ entry for why).

Users hit these Gatekeeper/Deckhouse limitations blind. `match.kinds` in particular fails completely silently: `kubectl describe` shows nothing wrong, there are no events, and the mutation just never applies — see the real-world report this was built from: someone spent time debugging why their `Assign` did nothing before being told on Compass that `match.kinds` is required. The other restrictions produce Gatekeeper parser errors (`invalid location format ...`) that are hard to map back to "wrap the key in quotes".

## Know before update

`Assign`/`AssignMetadata`/`ModifySet`/`AssignImage` resources that violate one of the six restrictions above are rejected on `kubectl apply` / `helm upgrade` after this change, where they previously succeeded and either failed later at the Gatekeeper webhook or (for a missing `match.kinds`) silently never took effect. No Deckhouse component ships resources of these kinds, so this only affects user-authored mutators; a GitOps controller re-applying one that happens to violate a restriction starts failing that sync until the resource is fixed.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Reject Assign/AssignMetadata/ModifySet/AssignImage resources that violate Gatekeeper's own content restrictions (invalid `location`, missing `match.kinds`, etc.) at `kubectl apply` time instead of silently ignoring or failing them later.
impact: A pre-existing Assign/AssignMetadata/ModifySet/AssignImage resource that violates one of these restrictions (see the module FAQ) starts failing `kubectl apply`/`helm upgrade` after the upgrade. Fix the resource before upgrading, or drop it if it was already a silent no-op, most commonly a mutator missing `spec.match.kinds`, which never took effect anyway.
impact_level: high
```


modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
```diff
@@ -0,0 +1,197 @@
...
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["mutations.gatekeeper.sh"]
+        apiVersions: ["v1", "v1alpha1", "v1beta1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["assign", "assignmetadata", "modifyset", "assignimage"]
+  validations:
+    - expression: >-
+        has(object.spec.match) && has(object.spec.match.kinds) && size(object.spec.match.kinds) > 0
```
AlwxSin: 🟡 **`deny-mutator-without-match-kinds` catches only the emptiest case of the silent failure it targets**

`size(match.kinds) > 0` is the weakest of the three conditions `constraint-exporter` actually needs. It builds the webhook rules from the cross product of `apiGroups × kinds`, resolving every kind through the RESTMapper, and bails out with `continue` when the result is empty (`images/constraint-exporter/src/pkg/kinds/kinds.go:279` and `:343`). Neither `apiGroups` nor `kinds` is `required` inside a `match.kinds[]` entry in the vendored CRD, so all of these are schema-valid, pass this policy, and still silently never fire — the exact failure mode this PR sets out to eliminate:

| `spec.match.kinds` | this policy | fires? |
|---|---|---|
| `[{}]` | allow | no |
| `[{apiGroups: ["*"]}]` | allow | no |
| `[{kinds: ["Pod"]}]` (no `apiGroups`) | allow | no |
| `[{apiGroups: ["*"], kinds: ["*"]}]` | allow | no — `findGVKsForWildcard` matches on an exact `Kind` name |
| `[{apiGroups: [""], kinds: ["*"]}]` | allow | no — `RESTMapping` errors, `slog.Warn` only |
| `[{apiGroups: [""], kinds: ["Pod"]}, {kinds: ["Pod"]}]` | allow | second entry no |

The `kinds: ["*"]` case is already documented as broken in this module's own `docs/GATEKEEPER-CR.md` ("Using `*` in `spec.match.kinds` is not allowed. If `*` is specified, the mutation will not be applied"), so a known silent no-op stays uncaught.

One `all()` closes all six — compiled and evaluated with cel-go, allows both valid shapes above, denies all six:

```
has(object.spec.match) && has(object.spec.match.kinds) && size(object.spec.match.kinds) > 0 &&
object.spec.match.kinds.all(k,
  has(k.apiGroups) && size(k.apiGroups) > 0 &&
  has(k.kinds) && size(k.kinds) > 0 &&
  !('*' in k.kinds))
```

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
```diff
@@ -0,0 +1,197 @@
...
+    resourceRules:
+      - apiGroups: ["mutations.gatekeeper.sh"]
+        apiVersions: ["v1", "v1alpha1", "v1beta1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["assignmetadata"]
+  validations:
+    - expression: >-
+        !has(object.spec.parameters) || !has(object.spec.parameters.assign) ||
+        !has(object.spec.parameters.assign.externalData) ||
+        !has(object.spec.parameters.assign.externalData.dataSource) ||
```
AlwxSin, line 158: 🟡 **`externalData.dataSource` is defaulted by the CRD, so users get denied for a field they never wrote**

`crds/gatekeeper/assignmetadata-customresourcedefinition.yaml` gives this field `default: ValueAtLocation`. Structural-schema defaulting runs during decoding, before validating admission, so the `!has(...dataSource)` short-circuit on this line never fires for an object that omits it. This:

```yaml
parameters:
  assign:
    externalData:
      provider: my-provider
```

reaches the policy with `dataSource: ValueAtLocation` already filled in and is denied with a message naming a field the user never set. Gatekeeper rejects it too (`pkg/mutation/mutators/assignmeta/assignmeta_mutator.go:146` in v3.22.2), so the verdict is right — but the message defeats the stated goal of a clear, actionable error at apply time.

Worth saying explicitly in the message (lines 155-157) and in both FAQ entries (`docs/FAQ.md:474`, `docs/FAQ_RU.md:474`) that `dataSource` defaults to `ValueAtLocation` and has to be set to `Username` explicitly.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
```diff
@@ -0,0 +1,197 @@
...
+metadata:
+  name: "deny-mutator-without-match-kinds"
+  {{- include "helm_lib_module_labels" (list . (dict "app" "gatekeeper")) | nindent 2 }}
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["mutations.gatekeeper.sh"]
+        apiVersions: ["v1", "v1alpha1", "v1beta1"]
+        operations: ["CREATE", "UPDATE"]
```
AlwxSin, line 186: 🟡 **The behavioral change for existing clusters is under-declared in the changelog**

`CREATE` + `UPDATE` with `failurePolicy: Fail` (line 174) and `validationActions: [Deny]` (line 197), and — unlike `deny-gatekeeper-webhook-operation-label*` in `templates/admission.yaml` — no bypass for `system:sudouser` or `system:serviceaccount:d8-*`. After the upgrade any pre-existing mutator that trips one of these rules becomes un-appliable, so a GitOps controller re-applying a manifest set that contains one starts failing its sync, with no opt-out short of deleting a Helm-managed VAP.

The blast radius is bounded, and I checked it: no Deckhouse component ships mutator CRs (`mutations.gatekeeper.sh` across `deckhouse`, the standalone modules, `deckhouse-cse`, `dex`, `deckhouse-cli` only turns up CRDs, RBAC rules, docs and console API metadata), and Gatekeeper's mutator reconciler only writes `MutatorPodStatus`, never the mutator object (`pkg/controller/mutators/core/reconciler.go:329`) — so no controller gets locked out. Only user-authored resources are affected.

For those users this is still a breaking change rather than a `low`-impact fix. Either bump the changelog to `impact_level: default` with an `impact` note describing what to fix before upgrading, or add a `system:sudouser` escape hatch consistent with the rest of `admission.yaml`.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
```diff
@@ -0,0 +1,197 @@
...
+        resources: ["assignmetadata"]
+  validations:
+    - expression: >-
+        !has(object.spec.location) ||
+        object.spec.location.matches(
+          "^metadata\\.(labels|annotations)\\.([A-Za-z0-9_-]+|\"(\\\\.|[^\"\\\\])*\"|'(\\\\.|[^'\\\\])*')$"
```
AlwxSin: 🔵 **This regex is stricter than Gatekeeper's own path parser**

Two `location` values that Gatekeeper v3.22.2 accepts are denied here:

- `metadata . labels . foo` — the scanner calls `skipWhitespace()` before every token (`pkg/mutation/path/token/scanner.go:50`), so whitespace around separators is legal upstream.
- `metadata."labels".foo` and `"metadata".labels.foo` — `readString` strips the quotes and stores the bare reference in `Object.Reference` (`pkg/mutation/path/parser/node.go:69`), and `isValidMetadataPath` compares via `reflect.DeepEqual` against `[]parser.Node{&Object{"metadata"}, &Object{"labels"}}`, which follows the pointers. So a quoted `metadata` or `labels` segment is valid upstream but fails the anchored `^metadata\.(labels|annotations)\.` prefix.

Both are exotic, but a false positive is the expensive direction here: `failurePolicy: Fail`, `[Deny]`, no bypass, and a legal resource becomes un-appliable. Allowing optional quoting and whitespace on the first two segments would close it.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
```diff
@@ -0,0 +1,197 @@
...
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["mutations.gatekeeper.sh"]
+        apiVersions: ["v1", "v1alpha1", "v1beta1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["assign"]
+  validations:
+    - expression: >-
+        !has(object.spec.location) ||
+        (object.spec.location != "metadata" && !object.spec.location.startsWith("metadata."))
```
AlwxSin: 🔵 **Three invalid shapes slip through**

Gatekeeper rejects these; the policies allow them, so the confusing-parser-error experience this PR removes elsewhere stays for:

- `Assign` with `location: "metadata".labels.x` (this line) — `hasMetadataRoot` compares the reference *after* quote stripping (`pkg/mutation/mutators/assign/assign_mutator.go:188`), while the CEL check is a literal `startsWith("metadata.")`.
- `AssignMetadata` with `location: metadata.labels.123` (line 56) — the scanner emits `INT` for a leading digit (`pkg/mutation/path/token/scanner.go:54`) and `parser.Parse` only loops while `curToken.Type == token.IDENT`, so it errors out; `[A-Za-z0-9_-]+` accepts it.
- `fromMetadata: {}` or `field: ""` (line 117) — `FromMetadata.Validate()` rejects any unrecognised field including the empty string (`apis/mutations/unversioned/value.go:112`), but the CEL short-circuits on `!has(...field)`. The FAQ wording ("accepts only `namespace` or `name`") could also say the field is mandatory once `fromMetadata` is set.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
```diff
@@ -0,0 +1,197 @@
...
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["mutations.gatekeeper.sh"]
+        apiVersions: ["v1", "v1alpha1", "v1beta1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["assign"]
+  validations:
+    - expression: >-
+        !has(object.spec.location) ||
```
AlwxSin: 🔵 **A missing `spec` makes all six policies fail with an internal CEL error**

`spec` is not `required` in the vendored CRDs, so `kind: AssignMetadata` with no `spec` at all is schema-valid and reaches admission. Every expression here then dereferences `object.spec.*` and errors with `no such key: spec` (reproduced with cel-go). With `failurePolicy: Fail` the request is still denied — right outcome — but the user gets an evaluation error from up to six policies instead of the intended message.

`templates/admission.yaml` already uses the optional idiom (`object.metadata.?finalizers.orValue([])`); the same shape here (`object.spec.?location`, or a leading `!has(object.spec) ||`) would keep the messages meaningful. Affects lines 24, 54, 86, 117, 149 and 183.

(`kinds: null` also errors, on `size()`, but CRD schema validation runs before validating admission, so that one is unreachable.)

modules/015-admission-policy-engine/docs/FAQ.md
```diff
@@ -461,6 +461,19 @@ Then, in order to fulfill the requirements of the above security policies, the f
       readOnlyRootFilesystem: true
 ```
 
+## Why is an Assign or AssignMetadata mutation rejected or silently ignored?
```
AlwxSin, line 464: 🔵 **Restriction docs now live in two places without a cross-link**

`docs/GATEKEEPER-CR.md` / `_RU.md` is already the reference page for these mutation CRs and already documents mutator restrictions, including the `*`-in-`match.kinds` gotcha. The reasoning for putting this in `FAQ.md` instead holds — `GATEKEEPER-CR` is not in `allowedDocFileNames` (`tools/validation/doc_changes.go:79`), so touching it trips the validator and the existing file is only grandfathered in. But the result is two overlapping sections that will drift. At minimum, link this section to the Gatekeeper CR page.

modules/015-admission-policy-engine/template_tests/admission_vap_test.go
```diff
@@ -69,4 +69,54 @@ admissionPolicyEngine:
...
+
+		assignLocationVAP := f.KubernetesGlobalResource("ValidatingAdmissionPolicy", "deny-invalid-assign-location")
+		Expect(assignLocationVAP.Field("spec.matchConstraints.resourceRules.0.resources").String()).To(
+			MatchJSON(`["assign"]`),
+		)
+
+		assignMetadataLocationVAP := f.KubernetesGlobalResource("ValidatingAdmissionPolicy", "deny-invalid-assignmetadata-location")
+		Expect(assignMetadataLocationVAP.Field("spec.validations.0.expression").String()).To(
+			ContainSubstring(
+				`"^metadata\\.(labels|annotations)\\.([A-Za-z0-9_-]+|\"(\\\\.|[^\"\\\\])*\"|'(\\\\.|[^'\\\\])*')$"`,
```
AlwxSin: 🔵 **This assertion restates the template's own text rather than testing behaviour**

The `ContainSubstring` argument is the regex verbatim, so the assertion and the template are the same string: it fails on a typo, never on wrong semantics. The description says the expressions were verified against a battery of cases with `cel-python`, but none of that is committed, so the six expressions have no regression guard.

A small table test compiling each expression with `cel-go` (already a dependency, `go.mod:244`) and evaluating a handful of valid/invalid objects would cover it — that is also what catches the gaps in the other comments on this PR.

Unrelated, but worth knowing before anyone reindents the template: the `>-` folded scalar keeps a literal newline around the more-indented regex line, because that line is indented deeper than the block. Harmless for CEL.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
```diff
@@ -0,0 +1,197 @@
...
+# images/gatekeeper/werf.inc.yaml), so none of the restrictions below can be expressed in the CRD
+# schema itself. Gatekeeper's own mutation webhook enforces them anyway, but only *after* the object
+# is persisted, with confusing parser errors and, for a missing `match.kinds`, with no error at all
+# (see the `deny-mutator-without-match-kinds` policy below). These ValidatingAdmissionPolicies give
+# the same feedback immediately, at `kubectl apply` time.
+---
+apiVersion: {{ include "helm_lib_get_api_version_by_kind" (list . "ValidatingAdmissionPolicy") }}
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "deny-invalid-assign-location"
```
AlwxSin: 🔵 **Six unqualified names in a cluster-scoped registry**

`deny-invalid-assign-location` and its five siblings (lines 42, 75, 106, 138, 171) are generic names for cluster-scoped objects. If a user already has a VAP under one of them, the Helm release fails with an ownership-metadata error. The newest policy in `templates/admission.yaml` uses a qualified name (`deny-deckhouse-finalizers.deckhouse.io`); the same suffix here removes the collision surface.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
Suselz: Fixed. Replaced the `size(match.kinds) > 0` check with the `all()` form you gave, requiring every entry to set non-empty, non-wildcard `apiGroups`/`kinds`. Verified against your six-row table (and the missing-spec case) with cel-python before committing, and it's now a permanent regression test in `admission_mutators_validation_cel_test.go` that reads the expression back out of the rendered template.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
Suselz: Fixed the message to say explicitly that `dataSource` defaults to `ValueAtLocation` and has to be set to `Username` explicitly, and added the same note to both FAQ entries. Left the CEL logic as-is since you confirmed the verdict is correct — this was purely a message-clarity gap.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
Suselz: Bumped `impact_level` to `high` with an explicit `impact:` note in the PR body/changelog describing what breaks and what to do about it, rather than adding a bypass — these are content-validation policies (is this resource well-formed), not the system-protection kind `deny-gatekeeper-webhook-operation-label*` is, so an sudouser/d8-serviceaccount bypass would mean 'the platform is exempt from writing valid mutators,' which doesn't fit here, especially since you'd already confirmed no Deckhouse component ships one. Thanks for doing that blast-radius check, it's what let me make that call with confidence.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
Suselz: Fixed both directions on this regex: added optional whitespace around every separator and optional quoting on the first two segments (`metadata`/`labels`/`annotations`), and tightened the bare-key alternative to require the first character not be a digit (see the reply on the sibling comment about `metadata.labels.123`). Verified against a matrix including your two cases plus the existing ones with cel-python, and it's now covered by a table test in `admission_mutators_validation_cel_test.go`.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
Suselz: All three fixed:
- `Assign` location: replaced the `startsWith` check with a regex that also matches a quoted `"metadata"`/`'metadata'` root.
- `AssignMetadata` location: the bare-key alternative now requires the first character to be a letter, `_`, or `-` (not a digit), matching the scanner's INT-vs-IDENT split.
- `fromMetadata`: reworked so `field` is required once `fromMetadata` is set (`fromMetadata: {}` and `field: ""` are both denied now), and reworded the FAQ entry to say the field is mandatory, not just enum-constrained.

All verified with cel-python against your specific cases (plus round-tripped through cel-go in the new test) before committing.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
Suselz: Fixed exactly as suggested — every expression is now guarded with a leading `!has(object.spec) ||` (or, for `deny-mutator-without-match-kinds`, folded into the `has(object.spec) && ...` chain, since that one must still deny rather than allow when `spec` is missing). Added a `spec: {}`-and-`spec` absent case to every table in the new CEL test so this can't regress.

modules/015-admission-policy-engine/docs/FAQ.md
Suselz: Added a link from the new FAQ entry to `docs/GATEKEEPER-CR.md`'s Mutation Custom Resources section, and a note there about why the restrictions live in FAQ.md instead of that page (the `allowedDocFileNames` constraint you found). Agreed a full merge would be better long-term, but that's a bigger doc restructuring than this PR should carry — the cross-link at least stops a reader who starts on the CR page from missing this.

modules/015-admission-policy-engine/template_tests/admission_vap_test.go
Suselz: Replaced it. `admission_mutators_validation_cel_test.go` now reads `spec.validations.0.expression` back out of the *rendered* template for each of the six policies and compiles/evaluates it with `cel-go` (per your pointer, already a direct dependency) against a table of valid/invalid objects — including every gap raised elsewhere in this review, so those are permanent regression tests now, not just a one-off cel-python check before commit. `admission_vap_test.go` keeps the structural assertions (names, `resourceRules`, `failurePolicy`) so the two failure modes stay distinguishable. Thanks for the folded-scalar note too — switched the `expression:` fields to `|-` literal blocks so the line-break behavior isn't left to a YAML subtlety.

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
Suselz: Fixed — all six policies and bindings now carry a `.deckhouse.io` suffix (`deny-invalid-assign-location.deckhouse.io`, etc.), matching `deny-deckhouse-finalizers.deckhouse.io`. Updated the FAQ entries, the VAP test, and the new CEL test to the new names.

z9r5 (approved):

modules/015-admission-policy-engine/templates/admission-mutators-validation.yaml
```diff
@@ -0,0 +1,210 @@
...
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["mutations.gatekeeper.sh"]
+        apiVersions: ["v1", "v1alpha1", "v1beta1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["assign"]
+  validations:
+    - expression: |-
+        !has(object.spec) || !has(object.spec.location) ||
+        !object.spec.location.matches("^\\s*(metadata|\"metadata\"|'metadata')\\s*(\\.|$)")
```
AlwxSin: 🔵 **A `metadata` root followed by a list spec still slips through**

The rework covers the quoted and whitespaced spellings, but the pattern requires `.` or end-of-string right after the root. A list spec is the third thing that can follow, and Gatekeeper still rejects it.

`parser.Parse` appends the `Object` node *before* it looks for `[` (`pkg/mutation/path/parser/parser.go:75-84`), so `location: metadata[name: x].foo` parses to `[Object{metadata}, List{...}, Object{foo}]`. `hasMetadataRoot` only inspects `path.Nodes[0]` (`pkg/mutation/mutators/assign/assign_mutator.go:188`), and `Object` has a single `Reference` field (`pkg/mutation/path/parser/node.go:69`), so the `reflect.DeepEqual` hits and the mutator is rejected.

Checked with cel-go against the rendered expression: `metadata[name: x].foo` and `"metadata"[name: x].foo` are the only two of twelve `location` cases that come back `allow` where Gatekeeper says no.

This is not something the fix introduced — it was there in the first revision too and I missed it; the shape I did flag (`"metadata".labels.x`) is caught now. Adding `\[` to the trailing alternation closes it:

```
"^\\s*(metadata|\"metadata\"|'metadata')\\s*(\\.|\\[|$)"
```

Worth a case in the new table test either way, next to the quoted-root ones at `template_tests/admission_mutators_validation_cel_test.go:141-143`.

Suselz: Fixed — added `\[` to the trailing alternation, so `metadata[name: x].foo` and `"metadata"[name: x].foo` are denied too. Added both to the CEL table test right next to the quoted-root cases you pointed at.

AlwxSin (approved):

AlwxSin (approved):

PR #22785, opened 2026-09-03
[docs] Code: pull mirroring branch filter and its effect on tag synchronization

Dreamseeker92:
## Description

Documents the "Branch filter" block in the pull mirroring settings of Deckhouse Code (EN + RU):

- describes the three options: "Mirror all branches", "Mirror only protected branches" (unavailable until the project has at least one protected branch, parent group's protected branches included) and "Mirror matching branches" (RE2 regular expression matched against the whole branch name);
- adds a new subsection about how the branch filter affects tag synchronization: with a filter enabled, a remote tag is synced only if its commit is contained in one of the branches of the mirror repository; skipped tags are re-checked on every subsequent sync; the check runs against all branches of the mirror, not only the ones matching the filter; already synced tags are kept if the filter changes later;
- updates the "Importing a repository by URL as a pull mirror" section to point to the new "Branch filter" section instead of mentioning only the protected-branches option.

Docs-only change, no code is touched.

The branch filter silently narrows tag synchronization as well, which is not obvious from the UI: with "Mirror only protected branches" or "Mirror matching branches" enabled, part of the source repository tags are not created in the mirror. Users perceive this as lost tags. The documentation now explains the behavior explicitly and states that skipped tags appear as soon as their commit reaches a mirrored branch.

Not necessarily.

## Checklist

## Changelog entries

```changes
section: docs
type: fix
summary: Document the pull mirroring branch filter in Deckhouse Code and its effect on tag synchronization.
impact_level: low
```


maxinsky (approved):

z9r5:
@dacetascien FYI

PR #22793, opened 2026-09-03
[multitenancy-manager] Keep grant webhook CA in sync after cert rotation

Jabejixo:
## Description

Grant defaulting (`cluster-objects-grants-defaulting`) and validation (`cluster-objects-grants-validator`) webhooks wrote `caBundle` only when the configuration was first created.

The serving certificate in `admission-webhook-certs` is reissued when SANs change (typical on a fresh cluster once `clusterDomain` is discovered). Helm updates the Secret and the Helm-managed project webhooks; the pod's `certwatcher` loads the new leaf. The grant webhooks kept the old CA.

The apiserver then rejects `/defaults` with `x509: certificate signed by unknown authority` / `ECDSA verification failure`. Project creation still works (Helm webhook). RoleBindings in a project namespace fail.

This change:
- always writes the desired webhook as a single-element `webhooks` list (`newGrant*Webhook` + current rules), so Name / Service / SideEffects and extra entries cannot drift;
- sets `caBundle` from the current CA (Secret preferred, values as fallback) when a CA is available;
- on an existing configuration with no CA yet, keeps the cluster CA and still reconciles the rest (does not stall `/modules/160-multitenancy-manager`);
- fails instead of publishing `failurePolicy: Fail` without a CA;
- watches `admission-webhook-certs` so a rotation retriggers the hooks without a Grantable CR change.

Broken clusters self-heal on upgrade: the new Secret watch fires a synchronization run, and the hooks rewrite `caBundle`. No manual `kubectl` / object delete is required.

No documentation changes.

On CSE/DKP 1.77 a RoleBinding in a project namespace can fail admission even though `multitenancy-manager` is Ready:

```
failed calling webhook "cluster-objects-grants-defaulting.multitenancy.deckhouse.io":
tls: failed to verify certificate: x509: certificate signed by unknown authority
(possibly because of "x509: ECDSA verification failure" while trying to verify
candidate authority certificate "multitenancy-manager.d8-multitenancy-manager.svc")
```

Confirmed on a test cluster: Secret CA fingerprint and MutatingWebhookConfiguration `caBundle` fingerprint differ.

Yes. The grant webhooks shipped in v1.77.0. The bug is present on the whole 1.77 line and on current `main`. Cherry-pick onto `release-1.77` applies as a single commit with no conflicts; hook tests pass there.

After the patch is applied, already-broken clusters heal themselves on upgrade. The new watch on `admission-webhook-certs` produces a synchronization run; the hooks overwrite the stale `caBundle`. Operators do not need to delete or recreate webhook configurations.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: fix
summary: Refresh grant webhook CA bundles after the admission certificate is rotated.
impact_level: default
```


modules/160-multitenancy-manager/hooks/configure_grant_defaulting_webhook.go
```diff
@@ -69,45 +91,38 @@ func configureDefaultingWebhook(ctx context.Context, input *go_hook.HookInput, d
...
+	// makes the apiserver reject the webhook with x509 ECDSA verification failure.
+	// Skip the CA write when it is not issued yet: the rest must still reconcile,
+	// or the hook fails the queue the way #20700 was written to prevent.
+	if caBundle != "" {
+		whConfig.Webhooks[0].ClientConfig.CABundle = []byte(caBundle)
+	}
 	whConfig.Webhooks[0].NamespaceSelector = projectNamespaceSelector
 	whConfig.Webhooks[0].MatchConditions = systemWriterMatchConditions
+	whConfig.Webhooks[0].FailurePolicy = ptr.To(admissionregistrationv1.Fail)
 	whConfig.Webhooks[0].TimeoutSeconds = ptr.To(int32(10))
```
AlwxSin: 🔵 **Three build paths and a field-by-field reconcile where a wholesale replace would do**

The hook now assembles the webhook in three places: the NotFound branch, the empty-list branch, and the in-place reconcile that copies CA / selector / matchConditions / failurePolicy / timeout one field at a time. Fields not in that list (`ClientConfig.Service`, `Name`, `SideEffects`, `AdmissionReviewVersions`) stay whatever they are in the cluster, and extra entries in `webhooks[1:]` survive.

Since `newGrantDefaultingWebhook()` already exists, replacing `webhooks` outright collapses all three paths and heals every field:

```go
wh := newGrantDefaultingWebhook()
wh.Rules = grantableWebhookRules(input)
switch {
case caBundle != "":
	wh.ClientConfig.CABundle = []byte(caBundle)
case whConfigExists && len(whConfig.Webhooks) > 0 && len(whConfig.Webhooks[0].ClientConfig.CABundle) > 0:
	wh.ClientConfig.CABundle = whConfig.Webhooks[0].ClientConfig.CABundle
default:
	return errWebhookCertNotIssued
}
whConfig.Webhooks = []admissionregistrationv1.MutatingWebhook{wh}
```

Same shape for the validating hook (`configure_grant_validation_webhook.go:153-171`). The existing tests pass unchanged against this; the "empty webhooks list" contexts become redundant.

Not a blocker, the fix itself is correct.

Jabejixo: Refactored to build the webhook once and replace webhooks with a single-element slice, so all paths share one reconcile and fields like Name/Service can’t drift. Dropped the redundant empty-list tests.

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [22822](https://github.com/deckhouse/deckhouse/pull/22822) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #22794, opened 2026-09-03
[descheduler] add feature flags support with evictionsInBackground

ergoz:
## Description

Makes the module always run the descheduler with the upstream `EvictionsInBackground` feature gate enabled, and adds an e2e scenario covering it.

The `descheduler` container now always receives `--feature-gates EvictionsInBackground=true`. There is no ModuleConfig setting for it — the behavior is part of the module and requires no configuration. The template keeps the enabled gates in a single list, so adding another one later is a one-line change:

```gotemplate
{{- $featureGates := list "EvictionsInBackground=true" }}
- "--feature-gates"
- {{ $featureGates | join "," | quote }}
```

**What the gate does.** By default, the descheduler expects every eviction to complete immediately: the eviction API request either succeeds or is rejected. With the gate enabled, an eviction of a pod annotated with `descheduler.alpha.kubernetes.io/request-evict-only` is treated as a *request* to start the eviction: the controller that owns the pod rejects the eviction API call with `429 Too Many Requests` and an error containing `Eviction triggered evacuation`, then marks the pod with `descheduler.alpha.kubernetes.io/eviction-in-progress` while it performs the actual eviction. The descheduler tracks such pods and counts them against its eviction limits until the pod is gone or the annotation is removed, so it does not start more evictions than allowed while the previous ones are still in progress.

**Why enabling it unconditionally is safe.** The gate changes nothing on its own. Pods without the `descheduler.alpha.kubernetes.io/request-evict-only` annotation take exactly the same code path as before, and even annotated pods are evicted normally unless some controller answers the eviction request as described above. No new RBAC is required either: the gate only adds an event handler on the pod informer that already exists, and `pods: get/watch/list` is already granted to the descheduler ServiceAccount.

**Impact on cluster components.** On upgrade, the `descheduler` Deployment in `d8-descheduler` gets a new container argument and its single replica restarts once. No other component is affected, and descheduling behavior is unchanged for every workload that does not implement its own eviction policy.

The eviction API contract — succeed or be rejected right away — does not fit workloads whose eviction is a long-running operation. The canonical case is a pod backing a virtual machine: it has to be live-migrated first, which may take minutes and may need retries. Without the gate, the descheduler either fails such an eviction or keeps retrying it while its eviction limits are already exhausted by evictions that are silently still in flight.

Upstream solves this with the `EvictionsInBackground` feature gate ([KEP-1397](https://github.com/kubernetes-sigs/descheduler/issues/1397)), but it is alpha and can only be turned on with the `--feature-gates` CLI argument, which the module did not pass.

Not necessarily.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: descheduler
type: feature
summary: Always run the descheduler with the upstream `EvictionsInBackground` feature gate enabled.
impact: The `descheduler` Deployment in `d8-descheduler` restarts once on upgrade to pick up the new argument. Descheduling behavior is unchanged for workloads that do not implement their own eviction policy.
impact_level: low
```


yalosev (approved):

z9r5 (approved):

deckhouse-BOaTswain:
Cherry pick PR [22801](https://github.com/deckhouse/deckhouse/pull/22801) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 56ac7a35437a76e554a1c6cc5702a9826cc1acf6 to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/33796775509) for details.

ergoz:
backported manually in #22803

PR #22824, opened 2026-09-04
[user-authz] Aggregate custom ClusterRoles into one role per access level

Jabejixo:
## Description

A ClusterRole annotated with `user-authz.deckhouse.io/access-level: <Level>` extends the `<Level>` high-level role and every role above it. Until now the chart rendered one `ClusterRoleBinding`/`RoleBinding` per such ClusterRole for every `ClusterAuthorizationRule`/`AuthorizationRule`, so the size of the `user-authz` Helm release grew as *(number of rules) × (number of annotated ClusterRoles of the level and below)*. Most annotated ClusterRoles are shipped by other Deckhouse modules (`d8:user-authz:<module>:<level>`), so enabling a module added one binding to every rule.

This PR changes the model:

- The chart renders six aggregated ClusterRoles `user-authz:<level>:custom` (`user`, `privileged-user`, `editor`, `admin`, `cluster-editor`, `cluster-admin`) with an `aggregationRule` that selects annotated ClusterRoles by the `user-authz.deckhouse.io/access-level` label for the level and all levels below it (the same cascade the bindings implemented). `rules` is left to the Kubernetes aggregation controller.
- Every rule with an `accessLevel` (except `SuperAdmin`) gets **one** binding to the aggregated role of its level, labeled `user-authz.deckhouse.io/binding-kind: aggregated-custom`. Bindings for `additionalRoles`, `portForwarding` and `allowScale` are unchanged.
- `hooks/handle_custom_cluster_roles.go` now keeps the `user-authz.deckhouse.io/access-level` **label** of every annotated ClusterRole equal to its annotation and removes the label when the annotation is gone, so the aggregation follows the existing annotation contract; nothing changes for modules or users who create such ClusterRoles. The hook no longer publishes the roles into `userAuthz.internal.customClusterRoles` (nothing renders from it any more), so a custom ClusterRole that appears, changes or disappears does not trigger a module release: the aggregation controller applies it to the aggregated roles on its own.
- The per-role bindings are not rendered any more. Two hooks retire them without a gap in permissions: right before the release `hooks/keep_legacy_custom_role_bindings.go` stamps `helm.sh/resource-policy: keep` on them, so the release engine plans nothing for them (nelm registers two operations per object that leaves the manifest, delete and track absence, and plan-building cost grows with the square of the number of operations, while a kept object costs one GET) and they keep granting access while the release creates the aggregated bindings; after the release has succeeded `hooks/delete_legacy_custom_role_bindings.go` deletes them in parallel, when the replacement is already in place. No transitional flags, one release.
- Two alerts on the Helm operation timings Deckhouse already exports for every module (`deckhouse_helm_operation_seconds`): `D8UserAuthzReleaseSlow` when applying the `user-authz` release takes more than 5 minutes on average over 30 minutes (the release timeout is 20), and `D8UserAuthzRenderSlow` when rendering the templates takes more than 30 seconds (normally about one), the earliest sign of an oversized release. Only `_sum`/`_count` of these histograms reach Prometheus, hence the averages.
- The template test suite derives its paths from the test file location instead of the hardcoded `/deckhouse` checkout path. It still needs the CI layout, because `charts/helm_lib` is a symlink into `/deckhouse`.

Effective permissions do not change: RBAC is a union of rules, and the aggregated role holds the union of the same ClusterRoles the per-role bindings referenced.

No cluster components are restarted by this change.

In clusters with thousands of `ClusterAuthorizationRule` objects and a few dozen annotated ClusterRoles the release contains tens of thousands of bindings. Rendering, planning and applying such a release takes longer than the module release timeout, the module fails to converge and platform updates stall. After this change the number of bindings per rule is constant (2–4), the release becomes an order of magnitude smaller and its size no longer depends on how many modules are enabled.

### Tested variants

All three revisions were tried on the same test cluster in the failing state: 600 `ClusterAuthorizationRule` + 50 `AuthorizationRule` objects, 55 annotated ClusterRoles, 8516 ClusterRoleBindings in the last successful release (the full manifest of about 19.4k bindings never got released at all), nelm as the release engine with the default 20-minute module release timeout, one master node.

| Revision | How the old per-role bindings go away | Objects the engine has to plan | Outcome |
|---|---|---|---|
| 1 — two releases | kept in the manifest until the hook sees the aggregated binding of the rule, dropped by the next release | ≈ 9.2k installs (8177 old + 656 aggregated + the rest) | 4 consecutive attempts, each `action timed out after 20m0s`, zero objects created |
| 2 — one release, engine deletes | left out of the manifest, deleted by nelm at `pre-pre-uninstall` | ≈ 950 installs + 8177 deletes (2 operations each) | `action timed out after 20m0s`, zero progress |
| 3 — one release, hook deletes before it | deleted in parallel by a `beforeHelm` hook right before the release | ≈ 950 installs, 0 operations for the deleted objects (a GET each) | release passed in 7 min 13 s; permissions of annotated ClusterRoles unavailable for 8 min 21 s |
| 4 — one release, keep then delete (this PR) | `beforeHelm` hook stamps `helm.sh/resource-policy: keep` on them, `afterHelm` hook deletes them once the release has created the aggregated bindings | ≈ 950 installs, 0 operations for the kept objects (a GET each) | release passed in 8 min 35 s; no probe that was allowed before the update was denied at any point |

Measurements of revision 4 (the incident was reproduced again on the same stand: the `main` release timed out after 20 minutes with 2374 of ~19k bindings created, so the test users had partial permissions when the update started):

| Step | Value |
|---|---|
| Hook: keep on the old bindings | 2578 objects (2282 ClusterRoleBindings + 296 RoleBindings) in 77 s |
| nelm upgrade | 8 min 35 s (GET pass over the 2.4k kept objects, plan of ~1.3k objects, creation of 6 ClusterRoles and ~1000 bindings) |
| Hook: deletion of the old bindings after the release | 2578 objects in 25 s |
| Permissions during the update | every probe that was allowed before the update stayed allowed throughout (SubjectAccessReview every second, 8 probes); the probes that had no bindings because of the incident became allowed as the release created them |
| Effective permissions after the update vs the fully converged baseline | no difference |

Measurements of revision 3, for comparison:

| Step | Value |
|---|---|
| Hook: deletion of the old bindings | 9238 objects (8177 ClusterRoleBindings + 1061 RoleBindings) in 130 s, 16 workers |
| nelm upgrade | 7 min 13 s, of which about 7 min is the GET pass over the objects of the previous release that the hook had already deleted (~22 objects/s), then creation of 6 ClusterRoles and 656 bindings |
| Permissions through annotated ClusterRoles unavailable | 8 min 21 s (from the first deletion by the hook to the creation of the aggregated bindings); the access-level bindings kept working throughout, no API errors |
| Effective permissions before/after (SubjectAccessReview snapshot of every level, 54 probes) | no difference |
| Applying an `accessLevel` change of one rule | 1 min 49 s (before the change the release never completed) |
| Custom ClusterRole created / rules changed / moved to another level / annotation dropped / deleted | effect visible in SubjectAccessReview after 0–1 s for all six levels, cascade preserved, no module release |
| Release secret | 308 KB → 199 KB |

Why revisions 1 and 2 fail: nelm's plan builder (`pkg/plan/plan.go`, `planChainBuilder.Do`) walks the whole operation graph for every operation it adds, so building the plan costs the square of the number of operations, and a deleted object costs two operations (delete and track absence). A release that removes thousands of objects is therefore as expensive as the one that created them. Taking the old bindings out of the engine's plan (deleting them beforehand in revision 3, keeping them in revision 4) is what makes the first release small enough to pass; the remaining cost is the linear GET pass over the objects of the previous release.

Side effects worth knowing:

- No permission is unavailable during the update: the old per-role bindings keep granting access until the aggregated bindings exist, and are deleted right after the first release.
- In audit logs and `SubjectAccessReview` reasons, access granted through an annotated ClusterRole is now attributed to `user-authz:<level>:custom` instead of the individual ClusterRole.
- The `user-authz.deckhouse.io/access-level` label appears on annotated ClusterRoles, including user-created ones; the module keeps it equal to the annotation and removes a stray one, and the docs say it is internal.

The problem blocks platform updates on large clusters, and the fix is limited to the module chart and two hooks, so it is safe to backport.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: fix
summary: Grant custom ClusterRoles (annotated with `user-authz.deckhouse.io/access-level`) through one aggregated ClusterRole per access level, so the number of bindings per ClusterAuthorizationRule/AuthorizationRule no longer depends on the number of such roles.
impact: |
  During the first `user-authz` release after the update the per-role bindings are protected from the release engine, the aggregated ones are created, and the per-role bindings are deleted right after the release. Permissions granted through annotated ClusterRoles stay in place throughout.
  The `user-authz.deckhouse.io/access-level` label is now set automatically on annotated ClusterRoles.
  In audit logs, access granted through annotated ClusterRoles is attributed to `user-authz:<level>:custom` instead of the individual ClusterRole.
impact_level: high
```


deckhouse-BOaTswain:
:red_circle: `e2e: DVP` for `deckhouse:user-authz/aggregate-custom-cluster-roles` [failed](https://github.com/deckhouse/deckhouse/actions/runs/33971619032) in 1h5m24s.




  `/e2e/abort dvp;WithoutNAT;containerd;1.34 33971619032 33971619032-2-con-1-34 /sys/deckhouse-oss/install:pr22824 ` - for abort failed cluster





:red_circle: `e2e: DVP` for `deckhouse:user-authz/aggregate-custom-cluster-roles` [failed](https://github.com/deckhouse/deckhouse/actions/runs/33971619032) in 33m12s.




  `/e2e/abort dvp;WithoutNAT;containerd;1.34 33971619032 33971619032-1-con-1-34 /sys/deckhouse-oss/install:pr22824 ` - for abort failed cluster




:red_circle: `e2e: DVP, Containerd, Kubernetes 1.34` failed in 31m45s.

</details>
  :fast_forward: `e2e: DVP, Containerd, Kubernetes 1.34` for `deckhouse:user-authz/aggregate-custom-cluster-roles` [started](https://github.com/deckhouse/deckhouse/actions/runs/33971619032).


:red_circle: `e2e: DVP, Containerd, Kubernetes 1.34` failed in 31m0s.

</details>

Jabejixo:
/e2e/abort dvp;WithoutNAT;containerd;1.34 33971619032 33971619032-1-con-1-34 /sys/deckhouse-oss/install:pr22824

deckhouse-BOaTswain:
:green_circle: `destroy cluster: DVP` for `refs/heads/main` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/33973428625) in 2m20s.


:green_circle: `destroy cluster: DVP, Containerd, Kubernetes 1.34` succeeded in 1m56s.

</details>

deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:user-authz/aggregate-custom-cluster-roles` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/33973648643) in 42m48s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@51.250.11.218`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 41m22s.

</details>

z9r5 (approved):

modules/140-user-authz/hooks/delete_legacy_custom_role_bindings.go
```diff
@@ -0,0 +1,105 @@
...
+const (
+	// legacyCustomRoleBindingMarker is the part of the name that only the per-custom-role bindings had.
+	legacyCustomRoleBindingMarker = ":custom-cluster-role:"
+
+	deleteLegacyBindingsWorkers = 16
+)
+
+var _ = sdk.RegisterFunc(&go_hook.HookConfig{
+	Queue:        internal.Queue("delete-legacy-custom-role-bindings"),
+	OnBeforeHelm: &go_hook.OrderedConfig{Order: 10},
```
AlwxSin: 🟡 **Finding 1 — the access gap is unavoidable only in the backport; the stack itself already has the fix**

The trade-off is documented honestly and the three measured revisions do rule out leaving the deletion to nelm. But there is a fourth option the table does not cover, and #22828 in this same stack already implements it for the other bindings: stamp `helm.sh/resource-policy: keep` on the legacy bindings *before* the release (so the engine plans nothing for them — the same reason `set_keep_policy_on_authorization_bindings.go` exists), let the release create the aggregated bindings, then delete the legacy ones from an `OnAfterHelm` hook. The release cost stays what revision 3 measured (~950 installs plus the linear GET pass) and the 8m21s window disappears, because the replacement is in place before anything is removed.

This matters specifically for the backport. In `main` the legacy bindings end up being removed by the controller with creation before deletion (`reconciler.go:390-397`), so the released `main` never has the gap — only 1.77/1.76 carry it, and that is exactly where #22828 does not follow. Worth measuring on the same stand before deciding, but the tool is already in the stack.

modules/140-user-authz/hooks/handle_custom_cluster_roles.go
```diff
@@ -75,68 +89,52 @@ var _ = sdk.RegisterFunc(&go_hook.HookConfig{
...
 
 func customClusterRolesHandler(_ context.Context, input *go_hook.HookInput) error {
-	customClusterRoles, err := snapshotsToInternalValuesCustomClusterRoles(input.Snapshots.Get(customClusterRoleSnapshots))
+	roles, err := snapshotsToCustomClusterRoles(input.Snapshots.Get(customClusterRoleSnapshots))
 	if err != nil {
 		return fmt.Errorf("failed to convert custom cluster roles snapshots: %w", err)
 	}
 
-	input.Values.Set("userAuthz.internal.customClusterRoles", customClusterRoles)
+	syncAccessLevelLabels(input, roles)
```
AlwxSin, line 97: 🔵 **Finding 2 — granting now keys off the label, and the label is not the documented contract**

After this change the aggregated ClusterRoles select on the `user-authz.deckhouse.io/access-level` **label**, which the aggregation controller acts on within about a second and without a module release. `syncAccessLevelLabels` only reconciles that label to the annotation afterwards.

The documented user-facing contract is still the annotation. So the effective grant path is now one an operator can trigger directly by labelling a ClusterRole, bypassing the annotation the docs describe, and a hand-edited label survives until the next hook run. Worth one line in the docs saying the label is internal and must not be set by hand — or, if you would rather not rely on that, having `syncAccessLevelLabels` treat a label without a matching annotation as drift to remove (which it already does for the removal case).

modules/140-user-authz/template_tests/module_test.go
```diff
@@ -118,17 +113,23 @@ var testCRDsWithCRDsKeyJSON, _ = ConvertYAMLToJSON([]byte(testCLusterRoleCRDsWit
 var _ = Describe("Module :: user-authz :: helm template ::", func() {
 	f := SetupHelmConfig(``)
 
+	// The BE-edition templates live under ee/be; link them into the module so the render covers
+	// them. Paths are derived from this file, so the suite runs both in CI (/deckhouse) and locally.
```
AlwxSin: 🔵 **Finding 3 — the suite still does not run outside the CI container**

The comment (and the PR description) claim the suite now runs both in CI and locally. It does not: `modules/140-user-authz/charts/helm_lib` is a symlink to `/deckhouse/helm_lib/charts/deckhouse_lib_helm`, an absolute path that only exists inside the CI image. `loader.Load` fails on it before any test body runs — 51 of 51 specs fail locally. Repointing that one symlink at the checkout makes all 51 pass, so the path derivation added here is correct and is not what blocks it.

The symlink is a repository-wide convention (33 modules do the same), so fixing it is out of this PR's scope. The ask is just to not claim local execution in the description until that symlink is resolved relative to the checkout.

modules/140-user-authz/hooks/delete_legacy_custom_role_bindings.go
Jabejixo: Done in 4354ffe426: `keep_legacy_custom_role_bindings.go` stamps `helm.sh/resource-policy: keep` on the per-custom-role bindings before the release and refuses to let the release run while any of them is unprotected, `delete_legacy_custom_role_bindings.go` deletes them from `afterHelm`. Measured on the same stand from a reproduced incident (release timed out, 2374 of ~19k bindings created): keep 2578 objects in 77 s, nelm upgrade 8 min 35 s, deletion 25 s, and none of the probes that were allowed before the update was denied at any point. Numbers are in the description; #22828 drops both hooks since the controller replaces bindings creation-first.

modules/140-user-authz/hooks/handle_custom_cluster_roles.go
Jabejixo: The removal case was already covered: `applyCustomRoleFilter` keeps a role that carries the label without a valid annotation in the snapshot, `syncAccessLevelLabels` removes the label, and the hook is driven by ClusterRole events, so a hand-set label lives seconds, not until the next release. Added the sentence you asked for to USAGE/USAGE_RU in c5d3b4446f: the label is internal, the module keeps it equal to the annotation and removes a stray one.

modules/140-user-authz/template_tests/module_test.go
Jabejixo: Agreed, the symlink makes the claim false. c5d3b4446f rewords the comment in `module_test.go` and the description: the paths are derived from the test file instead of the hardcoded checkout path, but the suite still needs the CI layout because `charts/helm_lib` points into `/deckhouse`.

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [22967](https://github.com/deckhouse/deckhouse/pull/22967) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [22968](https://github.com/deckhouse/deckhouse/pull/22968) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #22828, opened 2026-09-04
[user-authz] Reconcile rule bindings with user-authz-controller instead of the Helm chart

Jabejixo:
## Description

Stacked on #22824 (aggregated custom ClusterRoles); review the commits of this branch on top of it, the "Files changed" tab also shows the base PR until it is merged.

The ClusterRoleBindings and RoleBindings that grant users the access level of a `ClusterAuthorizationRule` or `AuthorizationRule` were rendered by the module Helm chart from values produced by hooks: every rule change re-rendered the whole release, and on clusters with thousands of rules the release grew large enough to exceed the release engine timeout. This PR moves them out of the chart into a new controller.

**`user-authz-controller`** (new CE image, `modules/140-user-authz/images/user-authz-controller`, controller-runtime, modeled on `user-authn-controller`):

- One reconciler per rule kind (`ClusterAuthorizationRule` → ClusterRoleBinding, `AuthorizationRule` → RoleBinding). The desired bindings — access level, aggregated custom role (`user-authz:<level>:custom`), `port-forward`, `scale`, `additionalRoles` — carry exactly the names, roleRefs and labels the chart used to render, so existing objects are taken over name-for-name.
- Reconcile creates missing bindings, updates drifted ones in place (roleRef, subjects, labels, ownerReferences, stray annotations) and deletes the bindings of the rule that are no longer desired. Creation always precedes deletion, so the per-custom-role bindings of versions before #22824 are removed only after the aggregated binding exists.
- Only objects with the module labels (`heritage=deckhouse`, `module=user-authz`) and named `user-authz:<rule>:*`, or carrying the controller's own ownerReference to the rule (so a binding whose labels were stripped is repaired rather than refused), are considered; foreign bindings and bindings of other rules are never touched. Subjects are compared with the `apiGroup` the API server defaults, so a converged binding is never rewritten. Bindings inherited from the chart lose the Helm ownership labels/annotations and get an `ownerReference` to their rule, so garbage collection removes them together with the rule.
- Bindings of a deleted rule that have no ownerReference (inherited ones) are cleaned up on the rule's delete event and, via a watch on the bindings themselves, on controller start.
- The name of a binding fixes its `roleRef`, so a binding under a controller name with a different `roleRef` was planted or edited by hand: it is deleted and recreated rather than treated as an unrecoverable error, so nobody who can create a labelled binding in a namespace can block the reconcile of a rule and with it the revocation of access. Foreign labels and annotations on adopted bindings are preserved; only the Helm ownership label and annotations are removed.
- `ClusterAuthorizationRule` and `AuthorizationRule` get a **status subresource**: `Ready` condition (`BindingsApplied` / `InvalidSpec` / `ApplyError`), `bindings` count and `observedGeneration`, plus `kubectl get` columns (ACCESS LEVEL, READY, BINDINGS, AGE). A rule with an unsupported `accessLevel` yields `Ready=False`/`InvalidSpec` and a Warning event, and its existing bindings are left untouched; previously a single such rule failed the render of the whole module.
- Deployment (leader election always on, so a rolling update never has two writers; VPA when available), PDB, PodMonitor, ServiceAccount and RBAC. The SA gets `bind` on `clusterroles` without a name restriction: a rule may reference any ClusterRole (access-level roles, aggregated roles, `additionalRoles`).
- Bindings are looked up through a cache index by rule name; a reconcile never lists the whole cache. Write errors of one binding do not stop the others: all are reported together in the status, an error the API server marks as invalid (an immutable `roleRef` that changed) is terminal and not retried, and nothing is deleted while a desired binding could not be written. An object that already exists behind the cache is adopted through the API reader; a rule missing from the cache is re-checked against the API server before its bindings are cleaned up.
- Workers, client rate limits and the log level are explicit (`MAX_CONCURRENT_RECONCILES`, `KUBE_CLIENT_QPS`, `KUBE_CLIENT_BURST` as Deployment environment variables, defaults 8 / 100 / 200; `LOG_LEVEL`, default `info`); the informer cache holds only module-labeled bindings, without `managedFields`, and the resource requests are sized for it.

**Migration in one release** — `hooks/set_keep_policy_on_authorization_bindings.go` (`OnBeforeHelm`, same approach node-manager used when it moved CAPI objects into node-controller): lists the rule bindings still labeled `app.kubernetes.io/managed-by=Helm` (both release engines put the label on the objects they apply, the controller removes it on adoption), stamps `helm.sh/resource-policy: keep` on them with a dedicated client (100 QPS, 16 workers, paged lists), re-lists and **refuses to let the release run** if any binding is still unprotected. Both release engines honour the annotation on the live object, so the bindings survive the removal of the template, and the controller adopts them afterwards. Once adopted, the bindings no longer match the hook's selector, so the hook becomes a no-op.

**Chart and hooks**:

- `templates/cluster-role-bindings.yaml` is removed together with its template tests.
- `hooks/delete_authorization_bindings_on_disable.go` (`afterDeleteHelm`) removes the rule bindings when the module is disabled: they belong to no release anymore, and a binding of `additionalRoles` points at a ClusterRole outside the module, so without the hook it would keep granting access after the module is gone. Re-enabling the module recreates the bindings from the rules.
- The `ClusterAuthorizationRule`/`AuthorizationRule` hooks keep publishing the rules into values — the authorization webhook configuration is rendered from them — and are otherwise unchanged.
- The two hooks added in #22824 that retire the old per-custom-role bindings (keep before the release, delete after it) are removed: with the controller in charge, the keep-policy hook protects every existing binding from the release engine and the controller creates the replacement bindings of a rule before deleting the unwanted ones, so the permissions granted through custom ClusterRoles stay in place throughout the migration.
- The controller writes as `system:serviceaccount:d8-user-authz:controller`, an identity two admission checks did not know about, while the chart wrote as the exempt `system:serviceaccount:d8-system:deckhouse`. It is now exempt in the multitenancy-manager grant webhooks (the CEL match conditions, their handler-level mirror and the project webhooks that share the list): they intercept every RoleBinding written into a project namespace and deny one whose ClusterRole is not granted to projects, which `user-authz:<level>:custom`, `user-authz:port-forward` and `user-authz:scale` are not, so every AuthorizationRule in a project would have stopped converging. It is also exempt in the module's own `role-validating.deckhouse.io` webhook, which denies a ClusterRoleBinding to a `rbac.deckhouse.io/kind: use` ClusterRole and would have turned a rule with such a role in `additionalRoles` into a permanent `ApplyError`. Verified on a test cluster: an AuthorizationRule in a Project converges in 4 s, the same RoleBinding written by a cluster admin is denied by the grant webhook, and written under the controller's identity it passes.
- The custom-roles hook only maintains the `user-authz.deckhouse.io/access-level` label consumed by the aggregated ClusterRoles; the `customClusterRoles` values are gone.
- `testing/library/images_tags_generated.go` gets the new image key (the generator needs a werf-visible checkout; the entry is added by hand in the generated order).

No cluster components are restarted by this change; the module gets one new Deployment.

The release of the module must not scale with the number of access rules. After this PR the chart contains only constant objects; rules are reconciled individually, on their own events, with status on the object. Together with #22824 this removes the failure mode where a large set of rules makes the module release time out and blocks platform updates.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: feature
summary: Bindings of ClusterAuthorizationRule and AuthorizationRule are reconciled by the new `user-authz-controller` instead of the module Helm chart; rules get a status subresource with a `Ready` condition.
impact: |
  The module deploys a new `user-authz-controller` Deployment in `d8-user-authz`. Existing bindings are adopted in place; access is not interrupted.
  `kubectl get clusterauthorizationrules` and `kubectl get authorizationrules` show READY and BINDINGS columns; a rule with an unsupported `accessLevel` or an invalid `additionalRoles` entry gets `Ready=False` instead of breaking the module and no longer blocks the other rules.
  Disabling the module now removes the bindings of the rules, as it did when the chart rendered them.

  The release that performs the migration costs one pass in which the release engine fetches every
  object of the previous manifest to see that the bindings must be kept. It is proportional to the
  number of bindings and happens exactly once. Up to about 5000 bindings it just makes that one
  release slower; above that it can exceed the 20-minute module release timeout, so run the upgrade
  with `HELM_TIMEOUT=60m` or on the previous release engine (`USE_NELM=false`) and put the setting
  back afterwards. Count with
  `kubectl get clusterrolebindings -l heritage=deckhouse,module=user-authz --no-headers | wc -l`.
  If the marks cannot be set the release does not start and the bindings are left untouched, which
  is deliberate: a release that ran without them would delete the bindings.
impact_level: default
```


Jabejixo:
CI note: two failures in this run are unrelated to the change and flaky. `Tests`: `modules/150-user-authn/hooks` `discover dex ca :: Should use only ingress-tls-customcertificate secret` (expected `testca`, got empty); this PR touches no user-authn file, and the user-authz hooks and template tests pass. `Dhctl Tests`: `pkg/server/rpc/dhctl TestStartSenderReportsSendError` timed out after 1s; this PR touches no dhctl file. Both jobs need a re-run once the workflow is no longer waiting on the manual e2e step.

z9r5 (approved):

modules/140-user-authz/templates/controller/rbac-for-us.yaml
```diff
@@ -0,0 +1,102 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: controller
```
AlwxSin, line 5: 🔴 **Finding 4 — this ServiceAccount is not in the exemption lists the Helm writer was in (blocking)**

Until now the rule bindings were written by `system:serviceaccount:d8-system:deckhouse` applying the chart. Two existing validating webhooks exempt that identity **by name**, and neither knows about `system:serviceaccount:d8-user-authz:controller` — it appears nowhere else in the tree except the audit-policy targets (`modules/040-control-plane-manager/hooks/audit_policy_basic_targets_generated.go:136`).

**a) `cluster-objects-grants-validator`** — `modules/160-multitenancy-manager/hooks/configure_grant_validation_webhook.go:59-80`. This needs no misconfiguration to hit: every RoleBinding the controller writes for an AuthorizationRule in a project namespace goes through the grant allow-list. The exempt set is `system:apiserver`, `d8-system:deckhouse`, the multitenancy-manager SA, the `system:serviceaccounts:d8-system` / `kube-system` groups, `system:masters` and `system:nodes`. The controller lives in `d8-user-authz`, so it matches none of them.

The comment above that list describes this exact failure, from when it already happened:

> observed: user-authz emitting "RoleBinding/...:d8:user-authz:*:user" into every project namespace -> "webhook retry timed out after 2m0s"

That is why `exclude-deckhouse-controller` exists. Moving the same writes to a non-exempt SA puts them back on the path the exemption was added to remove, and with `failurePolicy: Fail` a denial *or* merely a slow/unreachable backend both stall convergence.

It also invalidates the reasoning of a merged decision: #22609 removed the `project-managed-by-namespace` exemption from `projectNamespaceSelector` precisely because "they are applied by the deckhouse controller, which `systemWriterMatchConditions` below excludes at the apiserver, before any call reaches this webhook". That stops being true for user-authz bindings the moment this controller owns them.

**b) `role-validating.deckhouse.io`** — `modules/140-user-authz/webhooks/validating/role_binding:41-45` exempts exactly `system:apiserver` and `d8-system:deckhouse`, and denies any ClusterRoleBinding CREATE whose `roleRef` names a ClusterRole labelled `rbac.deckhouse.io/kind: use` ("Use RoleBinding instead"). `additionalRoles` has no name validation in the CRD, and such roles are user-facing (`d8:use:role:admin`, `d8:use:capability:kubernetes:forward_port`), so a CAR carrying one renders today because the chart wrote it as the exempt SA. Weaker than (a) — the CRD calls `additionalRoles` "reserved for emergencies", its examples use the legacy `cluster-*-all` names, and no doc suggests putting a `d8:use:` role there — but it turns a discouraged-yet-working configuration into `Ready=False` / `ApplyError` retrying forever.

Both are one line each: add `system:serviceaccount:d8-user-authz:controller` to the two `matchConditions` lists. #22837's `D8UserAuthzControllerReconcileErrorsHigh` makes it observable, but it fires 10 minutes after the rules have already stopped converging.

Jabejixo: Confirmed and fixed in 8ce9d91a6f. The controller SA is now exempt in `systemWriterMatchConditions` of the grant webhooks, in the handler-level `systemBypassUsernames` mirror (the comment there requires the two to stay in sync), in the shared list of `templates/admission/validation.yaml`, and in `webhooks/validating/role_binding`. Verified on a test cluster: an AuthorizationRule in a Project got its three RoleBindings (`editor`, `editor:custom`, `port-forward`) in 4 s and `Ready=True`; the same RoleBinding to the non-delegatable `user-authz:port-forward` written by a cluster admin was denied by `cluster-objects-grants-validator` ("not available to project"), and written under `--as=system:serviceaccount:d8-user-authz:controller` it passed. One correction to the analysis: `additionalRoles` exists only on ClusterAuthorizationRule, so for the grant webhook the affected bindings are `user-authz:<level>:custom`, `user-authz:port-forward` and `user-authz:scale`; `additionalRoles` matters for the `role-validating.deckhouse.io` case.

AlwxSin (approved):

himax1991 (approved):

PR #22832, opened 2026-09-04
[user-authz] Move dict and manage bindings from hooks into user-authz-controller

Jabejixo:
## Description

Stacked on #22828 (user-authz-controller); review the commits of this branch on top of it, the "Files changed" tab also shows the base PRs until they are merged.

Two more object reconcilers leave the addon-operator hook queue for `user-authz-controller`:

- **dict-bindings** (`internal/controller/dictbindings`) replaces `hooks/handle_dict_bindings.go`. Every subject that holds a use role — via a user-created RoleBinding to `d8:use:role:*`, or via a module-created RoleBinding of a namespaced rule (`user-authz:user|privileged-user|editor|admin`) — gets one ClusterRoleBinding to `d8:use:dict`; bindings of subjects that no longer hold any use role are removed. The subject key (`kind[:namespace]:name`, ServiceAccount as `sa`, lower-cased) follows the hook, except that it is no longer cut at 55 characters: two subjects sharing a long prefix are two subjects. Bindings the hook created with generated names are recognised by their subject and kept; bindings without a subject, with a wrong `roleRef`, or duplicating another binding of the same subject are removed. New bindings get a name derived from the subject key (`d8:dict:<hash>`), which makes a subject impossible to grant twice by a race.
- **manage-bindings** (`internal/controller/managebindings`) replaces `hooks/handle_manage_bindings.go`. A ClusterRoleBinding to a manage role (`d8:manage:*`) is projected into a RoleBinding `d8:use:<use-role>:binding:<name>` in every namespace the manage role reaches through its aggregation rule, followed to any depth (`matchLabels` and `matchExpressions`, with cycle protection). Drifted projections are repaired in place and foreign labels/annotations on them are preserved; a projection with a stale `roleRef` is recreated; automated use RoleBindings that are no longer expected are removed. Object names, labels and annotations are those the hook produced.

Both reconcilers recompute the whole set on any relevant event (single request key), the same way the hooks did on each run, reading their inputs through cache indexes (contributing RoleBindings; ClusterRoleBindings by `roleRef`, one lookup per manage role — a manage role is any ClusterRole labelled `rbac.deckhouse.io/kind=manage`, whatever its name, as with the hook) so that a reconcile does not copy every binding of the cluster; creation and update come before deletion, and every object is attempted even when another one failed. The removed hooks' scenarios are ported as fake-client tests of the reconcilers, plus tests for drift repair, the subject key and legacy generated names.

The informer cache of the controller now holds every (Cluster)RoleBinding, because both reconcilers depend on user-created bindings; `managedFields` are stripped from cached objects and ClusterRoles are cached only for manage roles (`rbac.deckhouse.io/kind=manage`).

The two remaining metric hooks stay in the module on purpose: `discovery_legacy_custom_roles.go` also publishes a release requirement value that must live in the deckhouse-controller process (the DKP 1.78 release check reads it), and `alert_deprecated_spec.go` is a single metric consumed by the same alert rules.

No cluster components are restarted by this change.

These bindings are derived from live cluster objects, not from module values, so they belong to a controller with per-event reconciliation rather than to hooks that share the module queue and run on every module release. Together with #22824 and #22828 this leaves the user-authz release with constant objects only.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: chore
summary: The `d8:use:dict` grants and the projections of manage bindings into namespaced use RoleBindings are reconciled by `user-authz-controller` instead of module hooks.
impact: |
  Two behaviour changes of the experimental role model: subjects of use RoleBindings whose identity differs only after the 55th character used to share one `d8:use:dict` grant and now each get their own; the namespaces a manage role reaches are resolved through its aggregation rule to any depth (including `matchExpressions`), not only one level down, so a deeper role hierarchy projects into every module namespace it aggregates.
impact_level: default
```


Jabejixo:
CI note: `Main web build` failed in the `Import secrets` step (Vault ID token request timed out) before any repository code ran; this PR touches no documentation. The job cannot be rerun while the workflow is still running (e2e pending), so it needs a re-run once the workflow completes.

Jabejixo:
CI note: the `Tests` failure is the known flaky `modules/150-user-authn/hooks` `get dex client crds :: Should update entry in internal values` spec; this PR touches no user-authn file, and the user-authz hooks and template tests pass. The job needs a re-run once the workflow is no longer waiting on the manual e2e step.

z9r5 (approved):

AlwxSin (approved):
Re-reviewed at this head: no findings, and the controller suites pass. No blocking issues.

PR #22868, opened 2026-09-07
[user-authz] Exempt platform identities from the identity collision check

Jabejixo:
## Description

The `identity_collision` webhook (`d8-user-authz-*-authorization-rule-collision.deckhouse.io`) refuses to create a `User` or a `Group` whose identity is already a subject of a `ClusterAuthorizationRule` / `AuthorizationRule`, unless the collision is acknowledged with an annotation. It applied to every requester, including the platform itself.

This PR excludes the platform's own identities from the check, in two places:

- `matchConditions` on both webhooks, so the API server does not call the hook for them at all and they keep working while the hook is unavailable (`failurePolicy: Fail`);
- the same list inside the hook, since `matchConditions` are a pre-filter and not the authority.

The set is the platform's break-glass list, the same one the heritage ValidatingAdmissionPolicies exclude: `system:apiserver`, `system:sudouser`, `system:kube-controller-manager`, `system:kube-scheduler`, `system:volume-scheduler`, `dhctl`, `observability`, the `d8-system:deckhouse` ServiceAccount, Commander's `d8-commander:cluster-manager` ServiceAccount (the service account only; the rest of that namespace stays subject to the check), and the groups `system:masters`, `system:serviceaccounts:kube-system`, `system:serviceaccounts:d8-system`. Every other webhook of the module already has such a list; this one was the only exception.

Test fixtures used to run every case as `kubernetes-admin` in `system:masters`; they now run as an ordinary requester, and the exempt identities are covered explicitly, including a config contract test that keeps `matchConditions` in step with the in-hook list.

The installer applies the initial `ClusterAuthorizationRule` and the `User` it names from one manifest. On a first bootstrap the User usually lands before the rule reaches the webhook's snapshot, so it passes. On a retried bootstrap (for example after nodes were late to join) the rule from the previous attempt already exists, and the webhook refuses the installer's `User` for granting privileges to itself:

```
admission webhook "d8-user-authz-user-authorization-rule-collision.deckhouse.io" denied the request:
users.deckhouse.io ".spec.email" "admin@example.com" is already granted privileges by
clusterauthorizationrules.deckhouse.io "admin"; use a different value or set the
"user-authz.deckhouse.io/allow-authorization-rule-collision: true" annotation to confirm the collision
```

The installer runs as `dhctl` in `system:masters` and by design treats admission denials as transient, so the bootstrap ends in a ten-minute `Timeout while "Create deckhouse.io/v1, Kind=User resources"` instead of a clear error. The getting-started manifests and cluster-management tooling that drive the installer carry no annotation, so every retried installation on 1.76.11+ / 1.77.0+ hits this.

Excluding the installer does not weaken the check: `system:masters` is the API server's superuser and can create the same rule itself. The protection against an ordinary requester quietly attaching a local identity to someone else's rule stays for everyone else.

The check was introduced in 1.76.11 and 1.77.0 (#22218, #22359, #22352). Since then a retried bootstrap of a new cluster fails without manual intervention on every release carrying it.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: fix
summary: The identity collision check no longer applies to the platform's own identities, so a retried bootstrap creates the initial User again instead of timing out.
impact: |
  The installer, the deckhouse ServiceAccount, holders of the cluster
  administrator kubeconfig and the kube-system / d8-system ServiceAccounts
  can create a User or a Group named by an existing authorization rule
  without the acknowledgement annotation. Every other requester is checked
  as before.
impact_level: low
```

### DIFF
--- a/modules/140-user-authz/webhooks/validating/identity_collision.py
+++ b/modules/140-user-authz/webhooks/validating/identity_collision.py
@@ -37,8 +37,19 @@
 # would take the whole hook down in a cluster where user-authz is disabled. Here the dependency is
 # the other way round and is harmless: the validating rules below name resources owned by
 # user-authn, and if that module is disabled its CRDs are absent, the rules simply never match.
+#
+# The check is about a requester who may create a User or a Group but does not own the rules that
+# would pick it up. The identities that install and operate the platform are not that requester:
+# the installer applies the initial ClusterAuthorizationRule and the User it names from one manifest,
+# in whatever order the objects land, and on a retried bootstrap the rule already exists when the
+# User is created again, so the check would refuse the installer for granting privileges to itself.
+# Those identities are excluded twice: in matchConditions, so the API server does not call the hook
+# for them at all (and keeps them working while the hook is unavailable, which matters under
+# failurePolicy: Fail), and again inside the hook, because matchConditions are a pre-filter and not
+# the authority. The list is the platform's own break-glass set, the same one the heritage
+# ValidatingAdmissionPolicies exclude (modules/002-deckhouse/templates/validation.yaml).
 
-from typing import Callable, NamedTuple, Optional
+from typing import Any, Callable, NamedTuple, Optional
 
 from deckhouse import hook
 from dotmap import DotMap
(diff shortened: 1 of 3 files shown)


AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [22871](https://github.com/deckhouse/deckhouse/pull/22871) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [22873](https://github.com/deckhouse/deckhouse/pull/22873) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #22916, opened 2026-09-08
[service-with-healthchecks] EndpointSlice ownerReference and child Service improvements

apolovov:
## Description
* Owner reference added to child EndpointSlices and to the child Service.
  * existing objects are adopted in place, without being recreated
  * `blockOwnerDeletion` is deliberately not set, so no access to the
    `servicewithhealthchecks/finalizers` subresource is required
  * deleting a SWH now collects the child Service and the EndpointSlices of every node
* The child Service was never re-checked for its owner reference: one created before the
  reference was introduced kept running without it forever. The owner is now part of the
  desired state.
* Child Service adoption reworked, new condition `ChildServiceReady`:
  * a Service controlled by another resource, or a pre-existing one whose spec differs, is
    left untouched — not rewritten, not adopted, no owner reference set
  * the clash is reported as `ChildServiceConflict` and re-checked on a timer, so the
    resource recovers on its own once the Service is removed or aligned
  * while the conflict lasts the agents publish no EndpointSlices under that name and
    withdraw the ones they published earlier, otherwise kube-proxy would balance over the
    union of our endpoints and the endpoints of the foreign Service
  * addresses of a foreign Service are no longer reported in the status
* A selector left on the child Service is now cleared. With a selector,
  kube-controller-manager publishes its own EndpointSlices next to the ones the agents
  write, and kube-proxy balances over both sets.
* `SWH.spec.clusterIP` support:
  * this setting is being transferred to child Service on creation if set, including
    `None` (headless), which used to be ignored
  * actual clusterIP from child Service is being transferred to `SWH.status.clusterIP`
    and `SWH.status.clusterIPs`, and is never written back into the spec
  * an address that can no longer be applied is reported as
    `ChildServiceClusterIPImmutable` instead of being retried forever
  * new `CLUSTER-IP` printer column
* The child Service was rewritten on every reconciliation, because its spec was compared
  before the API server filled it in. Fixed for:
  * `SWH.spec.type` — `ClusterIP` default in CRD
  * `SWH.spec.internalTrafficPolicy` — `Cluster` default in CRD
  * `externalTrafficPolicy` — consider `Cluster` as default for external Services
    (NodePort and LoadBalancer), empty for the rest
  * `targetPort` is equal `port` if not set
  * `nodePort` — the allocated one is kept. It used to be reset to 0, so the API server
    handed out a new one and the node port of a load balancer moved on every sync.
  * ports are compared semantically: `appProtocol` is a pointer and was compared by address

**Breaking:** the status condition `ChildService` is renamed to `ChildServiceReady`. The old
one is removed from the status on the first reconciliation after the update.

It is potential problem with uncontrolled endpointslices left after SWH deletion.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
type: chore
summary: EndpointSlices reconciliation improvements.
impact_level: default
```

### DIFF
--- a/ee/modules/610-service-with-healthchecks/crds/doc-ru-servicewithhealthchecks.yaml
+++ b/ee/modules/610-service-with-healthchecks/crds/doc-ru-servicewithhealthchecks.yaml
@@ -348,6 +348,14 @@ spec:
             status:
               description: Определяет наблюдаемое состояние `ServiceWithHealthchecks`.
               properties:
+                clusterIP:
+                  description: |-
+                    Адрес дочернего сервиса: либо запрошенный в `spec.clusterIP`, либо выданный кластером.
+
+                    Адрес отражается здесь и никогда не записывается обратно в спецификацию.
+                clusterIPs:
+                  description: |-
+                    Все адреса дочернего сервиса, по одному на каждое семейство IP-адресов.
                 conditions:
                   description: Текущее состояние сервиса
                   items:
--- a/ee/modules/610-service-with-healthchecks/crds/servicewithhealthchecks.yaml
+++ b/ee/modules/610-service-with-healthchecks/crds/servicewithhealthchecks.yaml
@@ -318,6 +318,7 @@ spec:
                   The default value,
                   `Cluster`, uses the standard behavior of routing to all endpoints evenly
                   (possibly modified by topology and other features).
+                default: Cluster
                 type: string
               ipFamilies:
                 description: |-
@@ -570,6 +571,7 @@ spec:
                   ExternalName aliases this service to the specified `externalName`.
                   Several other fields do not apply to ExternalName services.
                   [More info.](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types)
+                default: ClusterIP
                 type: string
             required:
             - healthcheck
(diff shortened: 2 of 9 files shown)


z9r5 (approved):

ee/modules/610-service-with-healthchecks/crds/doc-ru-servicewithhealthchecks.yaml
```diff
@@ -348,6 +348,14 @@ spec:
             status:
               description: Определяет наблюдаемое состояние `ServiceWithHealthchecks`.
               properties:
+                clusterIP:
+                  description: |-
+                    Адрес дочернего сервиса: либо запрошенный в `spec.clusterIP`, либо выданный кластером.
+
+                    Адрес отражается здесь и никогда не записывается обратно в спецификацию.
```
apolovov:
```suggestion
```

ee/modules/610-service-with-healthchecks/crds/servicewithhealthchecks.yaml
```diff
@@ -570,13 +571,27 @@ spec:
...
             - healthcheck
             type: object
           status:
             description: Defines the observed state of ServiceWithHealthchecks
             properties:
+              clusterIP:
+                description: |-
+                  The address of the child Service: the one requested in `spec.clusterIP`, or the one the cluster allocated.
+
+                  The address is reported here and is never written back into the spec.
```
apolovov:
```suggestion
```

AndreyPavlovFlant (approved):

PR #22921, opened 2026-09-09
[user-authz] Create the controller's namespace regardless of MultiTenancy

duckhawk:
## Description

`user-authz-controller` (added in #22828 and #22832) is rendered unconditionally, but two objects it depends on were left behind the `userAuthz.enableMultiTenancy` gate:

- `Namespace/d8-user-authz` — the namespace the controller's `Deployment`, `PodDisruptionBudget`, `ServiceAccount` and `VerticalPodAutoscaler` are placed in;
- `Secret/deckhouse-registry` — the pull secret in the controller's `imagePullSecrets`.

Both become unconditional. The gate stays on what genuinely belongs to MultiTenancy: the kube-rbac-proxy CA and the `d8-user-authz-multitenancy-state` ConfigMap, whose absence the `multitenancy.py` validating webhook already reads as "disabled".

With `enableMultiTenancy: false` the `user-authz` release cannot converge. Server-side apply rejects the controller's objects, the module retries indefinitely, and the whole cluster never finishes its first convergence:

```
run helm install: install nelm release "user-authz": failed release "user-authz"
(namespace: "d8-system"): execute release install plan: wait for operations completion:
execute operation: create resource: server-side apply resource
"PodDisruptionBudget/user-authz-controller (namespace=d8-user-authz)":
retryable on webhook error: server-side apply: namespaces "d8-user-authz" not found
```

Observed on a freshly bootstrapped cluster built from `main` (Kubernetes 1.34.11, `enableMultiTenancy: false`), 51 minutes after install:

| | |
| --- | --- |
| `ModuleRun:user-authz` | failed **49** times, requeued every 32 s |
| `main` queue | stuck on `ConvergeModules:main:::Operator-Startup` |
| `deploy/deckhouse` | `0/1` available, container `deckhouse` not ready |
| readiness probe | `HTTP probe failed with statuscode: 500`, ×623 |
| `ns/d8-user-authz` | absent |

Every node was `Ready` and every other module converged — the cluster looks healthy but `deckhouse` never reports `Available`, so anything gated on cluster readiness (CI e2e runs against `main`, for one) fails on that check rather than on its own subject.

Creating the namespace by hand is not enough to recover: without the gated pull secret the controller lands in `ImagePullBackOff`.

Not necessarily — the affected code is only in `main`.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: fix
summary: Create the user-authz-controller namespace and registry secret regardless of MultiTenancy, so the module converges with MultiTenancy disabled.
```

### DIFF
--- a/modules/140-user-authz/templates/namespace.yaml
+++ b/modules/140-user-authz/templates/namespace.yaml
@@ -4,23 +4,28 @@ prometheus.deckhouse.io/rules-watcher-enabled: "true"
 security.deckhouse.io/pod-policy: "restricted"
 security.deckhouse.io/enable-security-policy-check: "true"
 {{- end }}
-{{- if .Values.userAuthz.enableMultiTenancy }}
 ---
+# Unconditional: the namespace holds user-authz-controller, which reconciles rule
+# bindings whether or not MultiTenancy is on. Gating it on enableMultiTenancy left the
+# release unable to converge — the controller's Deployment, PDB, ServiceAccount and VPA
+# are rendered unconditionally, and server-side apply refused them with
+# `namespaces "d8-user-authz" not found`.
 apiVersion: v1
 kind: Namespace
 metadata:
   name: d8-{{ .Chart.Name }}
   {{- include "helm_lib_module_labels" (list . (fromYaml (include "labels" .))) | nindent 2 }}
+{{- if .Values.userAuthz.enableMultiTenancy }}
 ---
 {{- include "helm_lib_kube_rbac_proxy_ca_certificate" (list . (printf "d8-%s"  .Chart.Name)) }}
 ---
 # Mirrors userAuthz.enableMultiTenancy (already merged with the edition's config-schema
 # default) into a real Kubernetes object: the multitenancy.py validating webhook cannot
-# read module Values directly, only Kubernetes objects. Rendering it in the same block as
-# the namespace above means it only exists — and only ever needs to say "true" — exactly
-# when MultiTenancy is effectively enabled; its absence is read by the webhook as
-# "disabled", and it's created/removed atomically with the rest of this namespace, in the
-# same Helm apply, with no separate hook or bootstrap ordering to get wrong.
+# read module Values directly, only Kubernetes objects. Rendering it under this gate
+# means it only exists — and only ever needs to say "true" — exactly when MultiTenancy
+# is effectively enabled; its absence is read by the webhook as "disabled", and it's
+# created/removed in the same Helm apply as the rest of the multitenancy objects, with
+# no separate hook or bootstrap ordering to get wrong.
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/modules/140-user-authz/template_tests/module_test.go
+++ b/modules/140-user-authz/template_tests/module_test.go
@@ -358,6 +358,31 @@ var _ = Describe("Module :: user-authz :: helm template ::", func() {
 		})
 	})
 
+	Context("Without enabledMultiTenancy", func() {
+		BeforeEach(func() {
+			f.ValuesSetFromYaml("global.enabledModules", `["operator-prometheus", "operator-prometheus-crd"]`)
+			f.HelmRender()
+		})
+
+		// user-authz-controller is rendered whether or not MultiTenancy is on, so the objects
+		// it needs — the namespace it lives in and the secret it pulls its image with — have
+		// to be unconditional as well. While they were gated, the release could not converge:
+		// server-side apply rejected the controller with `namespaces "d8-user-authz" not found`.
+		It("Should create the namespace and registry secret the controller needs", func() {
+			Expect(f.RenderError).ShouldNot(HaveOccurred())
+
+			Expect(f.KubernetesGlobalResource("Namespace", "d8-user-authz").Exists()).To(BeTrue())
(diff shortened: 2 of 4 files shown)


Jabejixo (approved):

modules/140-user-authz/templates/namespace.yaml
```diff
@@ -4,23 +4,28 @@ prometheus.deckhouse.io/rules-watcher-enabled: "true"
...
-# read module Values directly, only Kubernetes objects. Rendering it in the same block as
-# the namespace above means it only exists — and only ever needs to say "true" — exactly
-# when MultiTenancy is effectively enabled; its absence is read by the webhook as
-# "disabled", and it's created/removed atomically with the rest of this namespace, in the
-# same Helm apply, with no separate hook or bootstrap ordering to get wrong.
+# read module Values directly, only Kubernetes objects. Rendering it under this gate
+# means it only exists — and only ever needs to say "true" — exactly when MultiTenancy
+# is effectively enabled; its absence is read by the webhook as "disabled", and it's
+# created/removed in the same Helm apply as the rest of the multitenancy objects, with
+# no separate hook or bootstrap ordering to get wrong.
```
AlwxSin, line 28: With the namespace moved out of this block, the comment in `modules/140-user-authz/webhooks/validating/multitenancy.py` is the last place describing the old invariant, and it now reads false:

- `:30` — the ConfigMap is no longer rendered "in that same block, alongside the d8-user-authz namespace itself";
- `:33-34` — "NB: d8-user-authz (and everything rendered in that block, including the ConfigMap) only exists when enableMultiTenancy is true".

The hook's logic is unaffected — it reads the ConfigMap, which stays gated, so an absent snapshot still means "disabled". But that NB is exactly what the next person to touch this gate reads first, and it would tell them the namespace is proof of MultiTenancy. Worth fixing in this PR.

AlwxSin (commented):
Diagnosis and fix both check out against `origin/main`. Leaving the trace here since the two PRs this repairs are already merged.

**Root cause confirmed.** `templates/controller/{deployment,pdb,rbac-for-us}.yaml` place a Deployment, VPA, PDB and ServiceAccount into `d8-user-authz` unconditionally, while `namespace.yaml:7` and `registry-secret.yaml:1` are still gated on `enableMultiTenancy` — which defaults to `false` in `modules/140-user-authz/openapi/config-values.yaml:9`. `ee/cse/modules/140-user-authz/openapi/config-values.yaml:9` is the only other copy of that schema and the only one defaulting to `true`, so every edition except CSE cannot converge the module on a default install. That matches the `namespaces "d8-user-authz" not found` loop described above.

**Why CI did not catch it.** `template_tests/module_test.go:149` renders with `enableMultiTenancy: false`, and the test added in #22828 asserts the controller Deployment renders — it does. Helm rendering never validates that a manifest's namespace exists; only the apply does, so the failure was invisible to the layer gating the merge. The Context added here is a real regression test: it inherits MT=false and asserts `Namespace/d8-user-authz` exists, which fails on current `main`.

**The shape of the fix looks right, and I could not find a side effect.** The controller reconciles rule bindings regardless of MultiTenancy, so its namespace and pull secret have to be unconditional, and what stays behind the gate is what genuinely belongs to it. Webhook semantics survive: `multitenancy.py` reads the ConfigMap rather than the namespace, and `is_multitenancy_enabled()` returns `false` on an empty snapshot. Also checked:

- `modules/040-control-plane-manager/hooks/discover_modules.go:43` watches a ConfigMap carrying the `control-plane-configurator` label, not the existence of the namespace, so the authz webhook is not switched on in kube-apiserver as a side effect of creating it;
- the controller PDB is `minAvailable: {{ helm_lib_is_ha_to_value (list . 1 0) }}`, i.e. `0` on a single-master cluster, so an always-present PDB cannot block a drain;
- the PodMonitor lives in `d8-monitoring`, and every `ee/be` object in `d8-user-authz` (webhook DaemonSet, permission-browser-apiserver) is still gated on `enableMultiTenancy`.

One stale comment to update, inline.

AlwxSin (approved):

modules/140-user-authz/templates/namespace.yaml
duckhawk: Fixed in 654d9242a6 — comment only, the hook still reads the gated ConfigMap.

Both statements are gone: the note now says the ConfigMap alone carries the answer and that the namespace, being unconditional, proves nothing about MultiTenancy. I also checked the other places mentioning the state ConfigMap; `multitenancy_test_factories.py:277` still describes it as "rendered by templates/namespace.yaml", which stays true, so nothing else needed changing.

duckhawk:
Verified on a fresh bootstrap from these images (`pr22921`), a 5-node cluster with `enableMultiTenancy: false`:

```
$ kubectl -n d8-system get deploy deckhouse
NAME        READY   UP-TO-DATE   AVAILABLE   AGE
deckhouse   1/1     1            1           26m

$ kubectl get ns d8-user-authz
d8-user-authz   Active   19m

$ kubectl -n d8-user-authz get pods
user-authz-controller-68558b9846-qtl8l   1/1   Running   4 (17m ago)   19m
```

The release converges, `deckhouse` reports `Available`, and all five nodes are `Ready` — against the same install that looped on `namespaces "d8-user-authz" not found` before. The controller restarts are unrelated: it races kube-apiserver coming up during bootstrap (`create runtime manager: ... dial tcp 10.223.0.1:443: connect: connection refused`) and settles once the API answers.

PR #22688, opened 2026-08-31
[control-plane-manager] Move network (CIDR) cluster params to ModuleConfig

meatuses:
## Description
Cluster network parameters `podSubnetCIDR`, `serviceSubnetCIDR`, `podSubnetNodeCIDRPrefix` are migrated to `mc control-plane-manager` `spec.settings.network`. Old parameters in `cluster-configuration` are still supported but deprecated.

- Parameters are supported in both `mc` and older `cluster-configuration`, moduleconfig taking precedence over cluster-configuration. But setting them in both configs on bootstrap would fail the preflight check so new cluster would have correct configs.
- Parameters marked as deprecated in cluster-configuration in dhctl, warns when starting the bootstrap. Cluster bootstrap is still supported using new and old methods of setting cluster subnets.
- Alert `D8ObsoleteNetworkFieldsInClusterConfiguration` is fired after the DKP update, nudging user to migrate. Simple copy-and-paste instruction is provided in the alert description.
- Validation controls that parameters added to the ModuleConfig have the same values as in cluster-configuration for fail-safe migration.
- Parameters in MCs `spec.settings.network` are immutable by validation after the first setting. Special annotation `deckhouse.io/allow-unsafe=true` added to `mc control-plane-manager` allows editing them if specifically required.
- Special parsing for these parameters added to `node-controller`, since it reads resources directly.
- Hook `cluster_ip_ranges.go` is fixed and now checks the `d8-cluster-configuration` secret in the correct namespace `kube-system`. When the secret exists, this hook skips setting cluster network values  in `global.discovery` (these values otherwise set by `cluster_configuration.go`).

Migrations are kept at least until next CSE release after 1.77.

## Why do we need it, and what problem does it solve?
Another step in migrating parameters from cluster-configuration.

## Checklist
- [x] The code is covered by unit tests.
- [x] e2e tests passed.
- [x] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: control-plane-manager
type: feature
summary: Cluster network parameters podSubnetCIDR, serviceSubnetCIDR, podSubnetNodeCIDRPrefix are migrated to mc control-plane-manager spec.settings.network; these parameters in `cluster-configuration` are deprecated.
impact_level: default
```

PR #22971, opened 2026-09-10
[cert-manager] Add first-class Yandex Cloud DNS ACME DNS-01 support

IlshatKhisamov:
Wire ModuleConfig settings to deploy the Yandex DNS webhook and auto-create the yandex ClusterIssuer, matching the UX of Cloudflare/CloudDNS providers.

## Description

Adds first-class **Yandex Cloud DNS** support for ACME DNS-01 challenges in the `cert-manager` module.

When both module settings are set:

- `yandexFolderID`
- `yandexServiceAccountJSON` (Base64-encoded authorized key JSON)

Deckhouse Kubernetes Platform automatically:

1. Deploys the Yandex Cloud DNS ACME webhook (`Deployment`, `Service`, RBAC, `APIService`, TLS).
2. Creates Secret `yandex` and ClusterIssuer `yandex` in namespace `d8-cert-manager`.
3. Uses Let's Encrypt production ACME with DNS-01 via webhook (`groupName: acme.cloud.yandex.com`, `solverName: yandex-cloud-dns`).

Users then create a Certificate with `issuerRef.name: yandex` (including wildcard certificates).

**Cluster impact:** enabling the settings starts the webhook in `d8-cert-manager`. No restarts of ingress-controllers, control-plane, or Prometheus are expected. Existing ClusterIssuers and certificates are unchanged unless they start using issuer `yandex`.

Also includes review hardening of the webhook (CleanUp client init, operation `Wait` with timeouts, scoped Secret RBAC, OpenAPI coupling of the two settings) and documentation updates (USAGE / FAQ / CERTIFICATES EN+RU).

Previously Yandex Cloud DNS required a manual third-party webhook install and a hand-written ClusterIssuer (documented only in FAQ).

That is inconsistent with Cloudflare / Route53 / DigitalOcean / Google Cloud DNS, which are configured via ModuleConfig and get a ready ClusterIssuer automatically.

This change closes that gap: Yandex Cloud DNS becomes a built-in DNS-01 provider with the same UX, so operators can issue and renew certificates (including wildcards) without installing an external chart.

Related task: #54137309.

Not necessarily.

This is a feature for the release milestone (`v1.78.0`), not a patch backport.

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
section: cert-manager
type: feature
summary: Add first-class Yandex Cloud DNS ACME DNS-01 support via ModuleConfig settings.
impact_level: default
```


modules/101-cert-manager/templates/yandex-dns-webhook/apiservice.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if include "cert_manager.yandex_dns_webhook_ready" . }}
+---
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: v1alpha1.acme.cloud.yandex.com
```
AlwxSin, line 6: 🟡 **2. No migration path off a manually installed upstream webhook**

The docs this PR removes (`modules/101-cert-manager/docs/FAQ.md`, plus the four `CERTIFICATES.md` pages) instructed users to install the upstream `cert-manager-webhook-yandex` chart by hand. That chart registers this same cluster-scoped singleton `v1alpha1.acme.cloud.yandex.com`, so the exact population this feature targets may already own the object — and nothing warns them.

Two outcomes, neither good:

- Helm refuses to adopt a resource it does not own, and the **entire** cert-manager release fails to deploy, not just the Yandex part.
- Helm adopts it, repointing the APIService at `yandex-dns-webhook.d8-cert-manager`. Their existing `yc-clusterissuer` then breaks silently: the new webhook's Role grants only `get` on secret `yandex` in `d8-cert-manager`, so reading their own `cert-manager-secret` returns 403 and renewals stop.

I have not verified which branch the Helm client actually takes here. Either way the fix is cheap — a short migration note in `USAGE.md` / `USAGE_RU.md` telling users to uninstall the manual chart and delete the APIService before enabling these settings.

modules/101-cert-manager/images/cert-manager-webhook-yandex/src/main.go
```diff
@@ -0,0 +1,329 @@
...
+	saKey, err := credentials.ServiceAccountKey(key)
+	if err != nil {
+		return nil, "", err
+	}
+
+	if c.apiEndpoint != "" {
+		apiEndpoint = c.apiEndpoint
+	}
+
+	sdk, err := ycsdk.Build(ctx, options.WithCredentials(saKey), options.WithDiscoveryEndpoint(apiEndpoint))
```
AlwxSin: 🟡 **3. `ycsdk.SDK` is never shut down — gRPC connections leak per challenge**

`ycsdk.Build()` opens three things that outlive this function: a standalone discovery connector (`buildEndpointsResolver` -> `transport.NewSingleConnector`, plus a `ListApiEndpoints` round trip), an IAM authenticator connection, and a `ConnPool`. `SDK.Shutdown(ctx)` exists and closes the pool, but nothing calls it.

`dnssdk.NewDnsZoneClient(sdk)` retains `sdk` as its connector, so `buildDNSClient` needs to return the `*SDK` for `Present`/`CleanUp` to defer a shutdown. One caveat on the fix: on the timeout path the request context is already expired by the time the deferred shutdown runs, so use a fresh bounded context rather than `ctx`.

The leak itself is inherited — upstream's `setConfig` also rebuilt the SDK on every `Present` — but this PR roughly doubles the rate by having `CleanUp` build one too, and cert-manager calls `Present` repeatedly while the self-check loop runs. The 100Mi VPA cap keeps it bounded by OOMKill rather than unbounded growth, which is its own milder problem.

Worth saying explicitly: moving client construction into `CleanUp` fixes a genuine upstream bug, where `CleanUp` used `c.dnsZoneClient` without ever calling `setConfig` and would nil-panic on a replica that had not served a `Present`.

ee/cse/modules/101-cert-manager/openapi/doc-ru-config-values.yaml
```diff
@@ -1 +0,0 @@
-../../../../../modules/101-cert-manager/openapi/doc-ru-config-values.yaml
```
AlwxSin, line 1: 🟡 **5. CSE doc-ru symlink replaced with a byte-identical copy**

This was a symlink to `modules/101-cert-manager/openapi/doc-ru-config-values.yaml`; it is now a materialized 156-line file. Diffing it against the main module's version shows they are identical — there is no CSE-specific content here.

Every other CSE module keeps the symlink (`015-admission-policy-engine`, `040-control-plane-manager`, `140-user-authz`). Unlike `config-values.yaml`, which is a deliberate fork (CSE sets `disableLetsencrypt.default: true`), the RU descriptions are shared. The copy buys nothing and guarantees the two files drift the next time a setting description changes — restoring the symlink and editing only the main file should be enough.

modules/101-cert-manager/templates/cert-manager/clusterissuer-yandex.yaml
```diff
@@ -0,0 +1,27 @@
...
+---
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: yandex
+  namespace: d8-cert-manager
+  {{- include "helm_lib_module_labels" (list . (dict "app" "cert-manager")) | nindent 2 }}
+spec:
+  acme:
+    server: https://acme-v02.api.letsencrypt.org/directory
```
AlwxSin, line 11: 🔵 **6. Built-in ClusterIssuer is production-only, while the docs it replaces used staging**

The FAQ and CERTIFICATES examples removed in this PR pointed at `acme-staging-v02.api.letsencrypt.org`. The new `yandex` ClusterIssuer hardcodes the production directory with no staging counterpart, unlike `letsencrypt` / `letsencrypt-staging`.

The first failures users hit here are the iteration-heavy ones — wrong folder ID, service account missing `dns.editor`, zone living in a different folder — and every retry now consumes production Let's Encrypt rate limit.

This matches the existing cloudflare/clouddns/route53 issuers, so it is not a regression in the module, but it is a regression relative to the documentation being deleted. Also worth noting that CSE deliberately defaults `disableLetsencrypt: true` and this issuer is not gated by that flag — again, same as the existing ones. Either a `yandex-staging` issuer or a paragraph in `USAGE.md` on validating the config against staging first would close the gap.

modules/101-cert-manager/images/cert-manager-webhook-yandex/src/main.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	if _, err := op.Wait(ctx); err != nil {
+		return errors.Wrap(err, "waiting for UpsertRecordSets (cleanup)")
+	}
+
+	return nil
+}
+
+func getDNSZone(ctx context.Context, dnsClient dnssdk.DnsZoneClient, folder, resolvedZone string) (*dns.DnsZone, error) {
```
AlwxSin: 🔵 **7. The rewritten zone-resolution logic has no Go tests**

`getDNSZone` was substantially rewritten relative to the vendored upstream: `util.FindZoneByFqdn` dropped in favour of `ch.ResolvedZone`, trailing-dot normalization added, pagination added, and an exact `normalizeZone(dnsZone.Zone) == zoneName` match added — the last of which fixes a real upstream bug where any public zone in the response was returned regardless of name.

None of it is covered by a test. `dnssdk.DnsZoneClient` is an interface, so a fake client plus a table test for `normalizeZone` is a small amount of work for the amount of hand-written logic here.

Related: the PR checklist ticks `The code is covered by unit tests`. That is accurate for `generate_yandex_webhook_certs_test.go` and the three new `template_tests` contexts, which are good — but not for this file.

modules/101-cert-manager/images/cert-manager-webhook-yandex/src/main.go
```diff
@@ -0,0 +1,329 @@
...
+	zoneName := normalizeZone(resolvedZone)
+	if zoneName == "" {
+		return nil, errors.New("resolved zone is empty")
+	}
+
+	pageToken := ""
+	for {
+		req := dns.ListDnsZonesRequest{
+			FolderId:  folder,
+			Filter:    `zone = "` + zoneName + `"`,
```
AlwxSin: 🔵 **8. Zone name interpolated into the Yandex filter expression without escaping**

Inherited verbatim from upstream, but this function was rewritten here anyway, so it is a cheap moment to fix it.

`zoneName` derives from `ch.ResolvedZone`, which cert-manager fills from a real DNS SOA response, so exploitability is remote. Since `normalizeZone` already exists as a single chokepoint for this value, rejecting anything outside `^[a-zA-Z0-9._-]+$` there would be a one-line hardening.

modules/101-cert-manager/images/cert-manager-webhook-yandex/src/main.go
```diff
@@ -0,0 +1,329 @@
+/*
+Copyright 2026 Flant JSC
```
AlwxSin: 🔵 **9. Flant copyright header prepended to vendored Yandex code**

This file is largely upstream `yandex-cloud/cert-manager-webhook-yandex` at the pinned SHA, so a blanket `Copyright 2026 Flant JSC` over the original attribution reads oddly — a modification notice would be the cleaner form.

The rest of the vendoring is done properly and worth keeping as-is: LICENSE and AUTHORS are included, `oss.yaml` records the upstream and SHA, `werf.inc.yaml` carries the provenance comment, and dependabot is wired up for the new `go.mod`.

modules/101-cert-manager/templates/yandex-dns-webhook/apiservice.yaml
IlshatKhisamov: Fixed in 809c6d98b2: added a migration note in `USAGE.md` / `USAGE_RU.md` — uninstall the manual chart and delete `v1alpha1.acme.cloud.yandex.com` before enabling the module settings.

modules/101-cert-manager/images/cert-manager-webhook-yandex/src/main.go
IlshatKhisamov: Fixed in 809c6d98b2: `buildDNSClient` now returns `*ycsdk.SDK`; `Present`/`CleanUp` defer `shutdownSDK` with a fresh bounded context (not the request `ctx`).

ee/cse/modules/101-cert-manager/openapi/doc-ru-config-values.yaml
IlshatKhisamov: Fixed in 809c6d98b2: restored the CSE `doc-ru-config-values.yaml` symlink to the main module file.

modules/101-cert-manager/templates/cert-manager/clusterissuer-yandex.yaml
IlshatKhisamov: Addressed in 809c6d98b2 with a USAGE note (same prod-only pattern as Cloudflare/Route53/CloudDNS): warn about production LE rate limits and how to validate via a temporary staging ClusterIssuer. Did not add a built-in `yandex-staging` issuer to stay consistent with the other DNS providers.

modules/101-cert-manager/images/cert-manager-webhook-yandex/src/main.go
IlshatKhisamov: Fixed in 809c6d98b2: added `main_test.go` with table tests for `normalizeZone` and `getDNSZone` (match, pagination, private-only skip, invalid name).

modules/101-cert-manager/images/cert-manager-webhook-yandex/src/main.go
IlshatKhisamov: Fixed in 809c6d98b2: `normalizeZone` now rejects anything outside `^[a-zA-Z0-9._-]+$` before the value is used in the Yandex filter.

modules/101-cert-manager/images/cert-manager-webhook-yandex/src/main.go
IlshatKhisamov: Fixed in 809c6d98b2: header is now `Copyright 2022 YANDEX LLC` + `Modifications Copyright 2026 Flant JSC`.

ee/cse/modules/101-cert-manager/openapi/doc-ru-config-values.yaml
IlshatKhisamov: Update: had to rematerialize this file in 065a3051a8. Restoring the symlink made `doc changes` validation fail (`config-values.yaml` changed while CSE `doc-ru-config-values.yaml` did not appear as changed vs main). Content stays identical to the main module file; symlink alone is not enough for that check.

Zhbert (approved):

ee/cse/modules/101-cert-manager/openapi/doc-ru-config-values.yaml
AlwxSin: Thanks — I checked the diagnosis against the validator source and it holds exactly as you describe. `tools/validation/doc_changes.go:27` matches **any** path ending in `openapi/config-values.yaml`, `ee/cse/` included, and `checkRelatedFileExists` looks the partner file up by path in `diffInfo.Files`. A symlink whose target changed never appears as a changed path, so the check cannot pass while the symlink is in place. No argument there.

One thing worth weighing before settling on this, though: there is precedent, and it went the other way. The two most recent PRs that changed `ee/cse/modules/040-control-plane-manager/openapi/config-values.yaml` with its `doc-ru` symlink intact — #22688 and #22910 — both carry the `skip/documentation-validation` label. So the established workaround in this repo is the label, not materializing the file, and every other CSE module (`015-admission-policy-engine`, `040-control-plane-manager`, `140-user-authz`) still ships a symlink.

That makes cert-manager the only CSE module with a real file here, and the failure mode afterwards is a quiet one: once the two copies diverge, `doc changes` will not catch it, because both files then show up as changed independently — which is precisely the condition it is looking for. Stale RU descriptions in CSE would just sit there.

Not a blocker and your call — I would rather you pick between the two knowingly than have the validator pick for you.

modules/101-cert-manager/images/cert-manager-webhook-yandex/werf.inc.yaml
```diff
@@ -0,0 +1,60 @@
...
+  before: install
+secrets:
+- id: GOPROXY
+  value: {{ .GOPROXY }}
+mount:
+{{ include "mount points for golang builds" . }}
+shell:
+  install:
+  - cd /src
+  - GOPROXY=$(cat /run/secrets/GOPROXY) go mod tidy
```
AlwxSin: 🔵 **10. `go mod tidy` inside the build mutates dependency metadata**

The patch edits `go.mod` (moving `go-sdk/services/dns` and `google.golang.org/grpc` from indirect to direct, dropping `miekg/dns`, `golang.org/x/mod`, `golang.org/x/tools`) but leaves `go.sum` alone, which is presumably why `go mod tidy` is here.

This is the only werf build in the repository that runs it — `grep 'go mod tidy' --include='werf.inc.yaml' modules/ ee/` comes back empty.

The distinction matters: `go mod download` verifies against the pinned `go.sum`, while `tidy` rewrites it — it resolves the full module graph and fills in whatever checksums are missing. That costs some build determinism and pulls in more network than a download, which tends to show up first on restricted builds.

Including the `go.sum` hunk in the patch and dropping back to `go mod download` alone would be cheaper and keep the build verifying rather than rewriting.

modules/101-cert-manager/images/cert-manager-webhook-yandex/werf.inc.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - GOPROXY=$(cat /run/secrets/GOPROXY) go mod download
+  setup:
+  - cd /src
+  - GOPROXY=$(cat /run/secrets/GOPROXY) CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags '-s -w -extldflags "-static"' -o /webhook .
```
AlwxSin, line 56: 🔵 **11. The new unit tests are never executed anywhere**

The tests in the patch are good and hit exactly the right things — pagination, exact public-zone match, private-only skip, and an `evil" OR zone = "x` case that also asserts `List` is not called at all. Thank you for those.

The problem is that nothing runs them. This artifact only does `go build`, and the sources exist solely inside the build artifact — no CI job touches them, and reproducing a run locally means cloning upstream and applying the patch by hand.

Both conventions exist in this repo: `modules/000-common/images/task/werf.inc.yaml:31`, `modules/038-registry/images/mirrorer/werf.inc.yaml:34` and `039-registry-packages-proxy` run `go test ./...` in the build, while `modules/038-registry/images/registry-agent/werf.inc.yaml:53` carries an explicit comment on why it deliberately does not.

Here the tests cover code we wrote, in a patch we maintain. If it silently breaks on the next SHA bump there is nothing to tell us. One line in `setup` turns them from decoration into a guard.

z9r5 (approved):

modules/101-cert-manager/images/cert-manager-webhook-yandex/werf.inc.yaml
IlshatKhisamov: Fixed

modules/101-cert-manager/images/cert-manager-webhook-yandex/werf.inc.yaml
IlshatKhisamov: Fixed

AlwxSin (approved):

PR #19455, opened 2026-04-27
[service-with-healthchecks] Stabilize module, fix status storm, and improve pod readiness evaluation

tarabrind:
## Description

This PR significantly stabilizes the `service-with-healthchecks` module, resolving severe status update storms, multiple controller crash loops, and incorrect pod readiness calculations. It also migrates the module's logging mechanism to the Deckhouse standard `pkg/log`.

The update only affects the `service-with-healthchecks` agent and controller pods. It does not cause restarts of any critical cluster components (ingress-controllers, control-plane).

## Why do we need it, and what problem does it solve?

Fix several critical bugs discovered during testing:

- **Status Storm (API Server Overload):** The agent and controller were continuously patching the status of `ServiceWithHealthchecks` resources (often generating 70+ requests per second for just 10 pods). This caused excessive `etcd` writes and `HTTP/2 GOAWAY` errors. The issue was solved by:
  - Sorting `EndpointStatuses` and `Conditions` before deep equality checks to prevent false positives due to map iteration randomness.
  - Changing `UpdateStatusWithCondition` to only update timestamps if the actual condition state (`Status`, `Reason`, or `Message`) changes.
  - Adding a `verboseStatus` configuration flag (defaulting to `false`) that gates the continuous updates of `LastProbeTime` (which otherwise creates new timestamps with monotonic clock readings, causing endless patching loop).
- **LoadBalancer Status Cleanup:** The parent `ServiceWithHealthchecks` object retained obsolete IPs if the user changed `.spec.type` from `LoadBalancer` to `ClusterIP`. The controller now explicitly clears the `LoadBalancer` status object when the type is switched.
- **Validation Error for ClusterIP:** The controller blindly copied `ExternalTrafficPolicy: Local` to the child `Service`, causing Kubernetes API rejection when the type was `ClusterIP`. The controller now enforces `ExternalTrafficPolicy: Cluster` for non-LoadBalancer/NodePort services.
- **ResourceVersion 0 Error:** The agent was calling `Update()` and `Patch()` sequentially on the same object inside the new-status initialization block, causing version conflicts. The redundant `Update()` was removed.
[...]

PR #22973, opened 2026-09-10
[service-with-healthchecks] Fix EndpointSlices never restored after a target pod is replaced.

AndreyPavlovFlant:
## Description

The agent's pod watch handler stopped iterating over the stored `ServiceWithHealthchecks` at the
first one whose selector did not match the pod — `sync.Map.Range` treats a `false` return as "stop".
Range order is randomized, so a pod event reached the right object only when it happened to be
visited first. The handler now examines every entry.

Both reconcilers also requeue themselves periodically (1 min in the agent, 5 in the controller), so
no missed event can pin the data plane. The resync writes nothing while the state is settled, so it
does not reintroduce the status storm fixed in #19455. **The controller and agent are restarted on
update**; affected load balancers recover on their own once the new agent starts.

Replacing a target pod — a worker VM being recreated or migrated, say — took the load balancer down
for good: the child Service kept its external IP but lost every endpoint. It could not recover by
itself, because with no target left there is no probe result to wake the reconciliation up either.
Outages lasted hours and hit several objects at once.

The bug is as old as the module but was unreachable: the status storm reconciled everything many
times per second, so a new pod was always picked up regardless. Fixing the storm in #19455 removed
that accidental resync, and the module has none of its own.

The storm fix shipped in v1.76.13, so every installation on v1.76.13+ is exposed: any target pod
replacement can silently take an external load balancer down. This has already caused production
outages. The change is small, unit-tested, and limited to this module.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: Fixed a load balancer losing all its endpoints for good after a target pod was replaced.
impact: |
  The `service-with-healthchecks` controller and agent are restarted.
  A `ServiceWithHealthchecks` that was left in `NotAllEndpointsAreReady` with no EndpointSlice
  recovers automatically once the updated agent starts, no manual action is needed.
impact_level: default
```

### DIFF
--- a/ee/modules/610-service-with-healthchecks/images/artifact/internal/agent/agent.go
+++ b/ee/modules/610-service-with-healthchecks/images/artifact/internal/agent/agent.go
@@ -41,6 +41,14 @@ const (
 	endpointServiceNameLabelKey = "kubernetes.io/service-name"
 	endpointControllerLabelKey  = "endpointslice.kubernetes.io/managed-by"
 	controllerName              = "servicewithhealthchecks"
+
+	// resyncPeriod bounds how long a ServiceWithHealthchecks may stay out of sync with the
+	// pods on this node. New target pods are learned from watch events only, and once the
+	// last target is gone there is no probe result left to wake the reconciliation up
+	// either — so a single missed event would otherwise keep the EndpointSlice empty until
+	// the cache resync (10h by default) or an agent restart. The resync is read-only in the
+	// steady state: neither the status nor the EndpointSlice is written when nothing changed.
+	resyncPeriod = time.Minute
 )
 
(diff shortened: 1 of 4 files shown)


apolovov (approved):

deckhouse-BOaTswain:
Cherry pick PR [22979](https://github.com/deckhouse/deckhouse/pull/22979) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [22980](https://github.com/deckhouse/deckhouse/pull/22980) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #22941, opened 2026-09-09
[deckhouse] Put the package sync of the old module stack under its own feature flag

Glitchy-Sheep:
## Description

Phase 1 of splitting `feat/introduce-module-v2` into small PRs. Module v2 is not touched: no CRD change, its controller stays off.

New flag `DECKHOUSE_ENABLE_MODULE_PACKAGE_SYNC` (`app.ModulePackageSyncEnabled()`). It turns on:
- the startup sync `pkgsync`: a PackageRepository for every user ModuleSource, a ModulePackageVersion for every embedded module and every ModuleRelease;
- the ModulePackageVersion controller that fills the drafts;
- the manager cache for ModulePackage and ModulePackageVersion.

`DECKHOUSE_ENABLE_MODULE_PACKAGES` now covers only the Module v2 controller and its cache.

The deployment template ships the new flag off.

Unchanged:
- source to repository name mapping: `deckhouse` -> `deckhouse-modules`, other sources keep their name;
- the `deckhouse` and `flant` sources are skipped, their repositories come from the templates;
- login, password and scan interval of an existing repository are never touched.
- the repository is not owned by the source and stays when the source is deleted: the sync is a one-off migration, it never deletes.

## Why do we need it, and what problem does it solve?

On main the sync runs on every cluster with no flag. The controller that fills its drafts sits behind the same flag as the Module v2 controller. That flag cannot be turned on: the v2 controller watches `v1alpha2.Module`, the CRD serves only v1alpha1, and the manager fails on cache sync.

With separate flags we can turn on the package objects of the old stack (repositories and versions) and check them on their own, before Module v2 becomes primary.

## Checklist
- [x] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [ ] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Gate the startup sync of module sources and releases into package objects behind DECKHOUSE_ENABLE_MODULE_PACKAGE_SYNC
impact_level: low
```

PR #22981, opened 2026-09-11
[deckhouse] Fill Module v1alpha2 from the cluster under the package sync flag

Glitchy-Sheep:
## Description

Stacked on #22941, which puts the startup package sync behind `DECKHOUSE_ENABLE_MODULE_PACKAGE_SYNC`. This PR adds the Module v1alpha2 objects to that sync.

Everything below happens only when the flag is on. With the flag off nothing changes: the cluster keeps `v1alpha1` as the served and stored version, and the sync does not run at all.

The work goes in the order the pieces depend on each other.

**1. The Module CRD is generated from its Go types**

- `crds/module.yaml` was written by hand, and `copy-crds` never copied the generated file
- So the file had already drifted: the `spec` field of the v1alpha1 type never reached the schema
- `copy-crds` now copies it, and the file is regenerated
- The v1alpha2 version comes in turned off, `+kubebuilder:unservedversion` on the type keeps it that way

**2. The flag turns v1alpha2 on**

- `EnsureCRDs` applies the Module CRD with v1alpha2 served and stored when the flag is on
- With the flag off it applies the file as it lies on disk, so every start puts the cluster back on v1alpha1
- The switch happens before any controller starts
- The manifest is edited as a plain document: rebuilding it through the typed `CustomResourceDefinition` would drop the fields that type does not carry

**3. The loader stops wiping annotations it does not own**

- `ensureModule` replaced the whole annotation map, so it dropped every key the module files do not declare
- It runs right after the sync in the same start, so it would erase the marks the sync writes
- It now writes the two description annotations and leaves the rest alone

**4. A Module is written for every module the cluster carries the files of**

Three sources, read in the order the files win each other:

- the embedded copy, it is what the binary carries on disk right now
- a ready `ModulePullOverride`, the loader put its dev copy on disk over the release
- the newest deployed `ModuleRelease`

A module none of them supplies is left alone. The sync never deletes an object.

**5. Every field the sync can work out is filled**

- `spec.packageRepositoryName`: `embedded` for a module from the image, the repository of the module source for a released one
- `spec.packageVersion`: the Deckhouse version, the release version, or the image tag of the pull override
- `spec.releaseChannel`: the channel of the deckhouse module config, the channel of the update policy the release was fetched by, or empty for a dev copy
- `spec.enabled`, `spec.settings`, `spec.settingsVersion`, `spec.maintenance`: mirrored from the `ModuleConfig` of the same name, and cleared when there is none
- annotations `modules.deckhouse.io/embedded` and `modules.deckhouse.io/dev`: where the files came from

A pull override names no repository, so it comes from the module object, then the module config source, then the only module source offering the module.

Module v2 describes what a module should run: which package, from which repository, enabled or not, with which settings. Today all of that lives in `ModuleConfig`, `ModuleRelease`, `ModulePullOverride` and on disk, and nothing writes it into a `Module`.

This is the step that puts the data in one place. The controllers keep working with v1alpha1, and that is on purpose: the goal here is only to fill the v2 projection with what can already be worked out.

With the flag on the old stack loses `properties`, `status.phase` and `status.hooksState`: the v1alpha2 schema does not carry them, and the apiserver prunes an object by the schema of the stored version. Module lifecycle on such a cluster is not to be trusted, and that is the accepted price of a flag that is off everywhere but a dev cluster.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Fill Module v1alpha2 objects from the cluster resources under the package sync feature flag.
impact_level: low
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -707,6 +707,7 @@ copy-crds:
 	@cp bin/crd/bases/deckhouse.io_applicationpackages.yaml deckhouse-controller/crds/applicationpackage.yaml
 	@cp bin/crd/bases/deckhouse.io_modulepackageversions.yaml deckhouse-controller/crds/modulepackageversion.yaml
 	@cp bin/crd/bases/deckhouse.io_modulepackages.yaml deckhouse-controller/crds/modulepackage.yaml
+	@cp bin/crd/bases/deckhouse.io_modules.yaml deckhouse-controller/crds/module.yaml
 
 .PHONY: generate-crds
 generate-crds: controller-gen
--- a/deckhouse-controller/crds/doc-ru-module.yaml
+++ b/deckhouse-controller/crds/doc-ru-module.yaml
@@ -61,3 +61,92 @@ spec:
                   properties:
                     editions:
                       description: Настройки работы модуля в редакциях Deckhouse.
+    - name: v1alpha2
+      schema:
+        openAPIV3Schema:
+          description: |
+            Описывает экземпляр модуля, которым управляет система пакетов.
+          properties:
+            spec:
+              description: Желаемое состояние модуля.
+              properties:
+                packageRepositoryName:
+                  description: |
+                    Имя репозитория, из которого ставится пакет модуля.
+
+                    Если не задано, используется репозиторий по умолчанию.
+                packageVersion:
+                  description: Версия пакета модуля, которую нужно установить.
+                releaseChannel:
+                  description: Канал обновлений пакета модуля.
+                enabled:
+                  description: |
+                    Включает или выключает модуль.
+
+                    Если не задано, решение принимает платформа.
+                maintenance:
+                  description: |
+                    Режим обслуживания модуля.
+
+                    - `NoResourceReconciliation`: режим для разработки и отладки модуля.
+
+                      В этом режиме:
+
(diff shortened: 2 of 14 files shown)


deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:feat/pkgsync-module-v2` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/34578309898) in 26m5s.



GCP-WithoutNAT-Containerd-1.34 - Connection string: `ssh  user@34.179.244.125`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.34` succeeded in 24m54s.

</details>

deckhouse-BOaTswain:
:green_circle: `e2e: GCP` for `deckhouse:feat/pkgsync-module-v2` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/34603126381) in 21m22s.



GCP-WithoutNAT-Containerd-1.34 - Connection string: `ssh  user@34.185.146.202`


:green_circle: `e2e: GCP, Containerd, Kubernetes 1.34` succeeded in 20m15s.

</details>

z9r5 (approved):

ldmonster (approved):

PR #22943, opened 2026-09-09
[user-authz] 2/2 — Remove the config.json ConfigMap and finish moving the components onto the shared library

Jabejixo:
## Review this together with #22942

These are two stacked PRs — `user-authz/controller-observability` → #22942 → **this** — and they
are one change. This one also carries the defect fixes for the code #22942 introduces.

---

## Description

The `config.json` machinery goes away and the multi-tenancy decision moves into
`go_lib/user-authz`, so the controller, the authorization webhook and permission-browser work from
one library instead of three copies of the same logic. The defects that the copies had drifted
into are fixed with it, each verified on a stand before it was touched.

- **The rendered `config.json` is gone**: the hooks that snapshot every rule into values, the values,
  the ConfigMap and the mounts. Nothing has read the file since the previous PR; this is what removes
  the render cost that grows with the number of rules. The one thing the render did that nothing else
  did - failing when a rule asks for namespace limits while `enableMultiTenancy` is off - is now the
  alert `D8UserAuthzRuleNeedsMultiTenancy`, which reports the same situation without making an
  unrelated rule un-renderable.
- **The decision itself moves into the library.** The previous PR moved the facts and left the
  sequence that consults them written twice, which is the half that had actually drifted: the 404
  above was a carve-out present in one copy and not the other. `go_lib/user-authz/decision` now holds
  it once - find a subject's entries, apply the ordering guard, fold them, answer the namespaced or
  the cluster-scoped question - and each consumer supplies only what it alone knows. Neither side had
  to change a test to keep passing, which is the evidence that the behaviour is the one they had;
  between them they lose 285 lines and gain 128.
- **A nonexistent resource answers 404 again, not 403.** A user covered by any ClusterAuthorizationRule
  got Forbidden for any resource that does not exist, SuperAdmin included. The webhook asked discovery
  whether the resource was namespaced and denied on any error, so "the resource does not exist" and
  "discovery could not be reached" were the same outcome. They are now separate facts: an absence that
  discovery actually reported is left to RBAC, so the API server produces the 404 the caller is owed,
  while a lookup that did not happen still denies. permission-browser had the same asymmetry through
  the same library call and reported allowed for a nonexistent core resource the API server forbade.
- **Every discovery lookup on the authorization path is bounded.** The group and the resource of a
  request are chosen by its sender, so an unknown one used to be the cheapest way to make the webhook
  ask the API server: one round trip per request, with a retry loop of up to twenty seconds inside
[...]

PR #22992, opened 2026-09-11
[user-authz] Readiness that means "can work", permission-browser served through a rollout, one writer in user-authn

Jabejixo:
## Description

Four availability defects in the user-authz and user-authn components, each measured on a stand
before it was touched. Stacked on #22943 (base `user-authz/authz-chain-fixes`); review this PR's own
diff only. No documentation changes.

- **A controller that cannot reach the API no longer reports Ready.** With its RBAC removed,
  `user-authz-controller` created no bindings for a new ClusterAuthorizationRule, logged nothing, and
  answered 200 on both probes for the whole time. Readiness checked only that the informer caches had
  synced once, and a synced cache stays synced after access is lost. `/readyz` now also performs one
  uncached `List` of one object under the controller's own identity (`api-access`), so the answer is
  about now. Liveness is unchanged on purpose: a restart cures nothing here, and the controller
  recovered by itself once access came back. `user-authn-controller` has the same code and gets the
  same check. `/readyz` runs `cache-sync` and `api-access` in sequence, 2 s each, and the handler
  does not stop at the first failure, so both readiness probes get `timeoutSeconds: 5` (was 3) and the
  template tests pin that.
- **Rolling out permission-browser with one replica no longer takes its API down.** The Deployment
  hardcoded `maxSurge: 0 / maxUnavailable: 1`, so outside HA the only pod was removed before its
  replacement existed and the APIService `v1alpha1.authorization.deckhouse.io` had no endpoints for the
  whole restart — 16 seconds measured, `Available=False`, discovery of that group failing. It now uses
  the shared `helm_lib_deployment_strategy_and_replicas_for_ha` helper (two replicas rolled one at a
  time in HA, the Deployment default that surges first otherwise). The PodDisruptionBudget the chart
  already carries (`maxUnavailable: 1`) is unchanged. With the surge the same rollout produced zero
  unservable samples.
- **`user-authn-controller` always elects a leader.** Election followed `HA_MODE`, which the
  Deployment set only in HA. Outside HA the Deployment keeps the default rolling update, so two pods
  run during every version change, both reconciling and writing the same UserAccounts and
  DexProviderChecks. Election is now unconditional with `LeaderElectionReleaseOnCancel`, as
  `user-authz-controller` already does, and the `HA_MODE` variable is gone from the Deployment.

  The grant on leases moves from the ClusterRole to the namespaced Role the controller already
  binds to, where `user-authz-controller` keeps its own. Measured on a stand with the previous
  revision of this PR: with its ClusterRoleBinding removed, `user-authz-controller` stood NotReady and
  was Ready again the second the binding returned, while `user-authn-controller` went NotReady the
  same way and then exited 48 s later with `leader election lost` — losing cluster-wide access had
  also taken its lease. With the grant in the Role, both controllers wait instead of restarting.

  The price of a lease is that controller-runtime exits the process when a renewal misses
  `RenewDeadline` — with the defaults (15 s lease, 10 s renew) every short API-server absence would
  restart the single non-HA pod, which is the opposite of what the readiness change is for. Both
  controllers therefore run the lease at 60 s / 40 s / 8 s (lease, renew, retry): a minute of API
  absence is survived with the cache intact; a longer one restarts the pod, by which time readiness
  has been reporting it for a minute. This is the trade the readiness design accepts: a restart on a
  long outage is harmless, two writers are not.
- **The authorization webhook logs review bodies for denials only.** It wrote the full
  SubjectAccessReview — identity, groups, resource, answer — for every decision at the default
  verbosity (about 125 KB/min from a 5 rps probe), with no switch. Denials are what an operator looks
  for in this log; allowed and no-opinion answers are background. The binary understands a
  `LOG_DECISIONS=none|denied|all` variable for development and tests, but the chart does not expose
  it — the DaemonSet is Helm-managed and a hand edit is reverted — so on a cluster the new default is
  the behaviour. Nothing consumed the old lines: no alert, no log-shipper rule, no document.

The first defect is the serious one: a cluster where new rules grant no access at all looked healthy
to the kubelet, to the Deployment, and to anyone reading `kubectl get pods`. The second is a routine
operation — a module update — silently breaking the Permission Browser API on every non-HA cluster.
The third is a write race that exists on every non-HA rollout. The fourth is log volume and
identity data on the master's disk with no way to turn it off.

## Checklist
- [ ] e2e tests passed.
- [x] Documentation updated according to the changes — not needed: no user-facing behaviour is
  documented for probes, rollout strategy or webhook log verbosity.

## Changelog entries

```changes
section: user-authz
type: fix
summary: user-authz-controller reports not ready when it cannot read the API, instead of staying Ready with synced-but-dead informers.
impact: |
  Readiness now includes one uncached read per probe, and the probe timeout grows from 3 to 5 seconds.
  A controller whose access to the API server is lost becomes NotReady within a few seconds and Ready
  again once access returns. The leader lease is renewed at 60 s / 40 s instead of 15 s / 10 s, so a
  short API-server absence no longer restarts the pod.
impact_level: low
---
section: user-authz
type: fix
summary: A rollout of permission-browser-apiserver with one replica keeps the authorization.deckhouse.io API served.
impact: |
  Outside HA the Deployment uses the default rolling update, which starts the new pod before removing
  the old one; in HA nothing changes.
impact_level: low
---
section: user-authn
type: fix
summary: user-authn-controller always uses leader election, so two pods during a rolling update no longer write the same objects.
impact: |
  HA clusters already elected a leader and see no change. On clusters without HA a Lease named
  user-authn-controller appears in d8-user-authn, and the pod now restarts if it cannot renew that
  lease for 40 seconds — that is, after a prolonged loss of the API server; a shorter absence is
  survived. Before, the single pod stayed up through any outage but two pods wrote the same objects
  during every rolling update. The RBAC grant on leases moves from the ClusterRole
  d8:user-authn:controller to the Role controller in d8-user-authn; the set of permissions is unchanged.
impact_level: low
---
section: user-authn
type: fix
summary: user-authn-controller reports not ready when it cannot read the API, instead of staying Ready with synced-but-dead informers.
impact: |
  Readiness now includes one uncached read per probe, and the probe timeout grows from 3 to 5 seconds.
  A controller whose access to the API server is lost becomes NotReady within a few seconds and Ready
  again once access returns.
impact_level: low
---
section: user-authz
type: fix
summary: The authorization webhook logs SubjectAccessReview bodies only for denied requests.
impact: |
  The full review body — identity, groups, resource, answer — is written for denials only; allowed
  and no-opinion answers are no longer logged. There is no configuration for this in the module.
impact_level: low
```


modules/150-user-authn/images/user-authn-controller/src/cmd/main.go
```diff
@@ -119,11 +119,17 @@ func setupRuntimeManager() (ctrl.Manager, error) {
...
 	timeout := 10 * time.Second
+	// Leader election is always on. It used to follow HA_MODE, but two pods of this Deployment
+	// exist outside HA too: the non-HA rolling update surges the new pod before the old one is
+	// gone, and both reconcile and write UserAccounts and DexProviderChecks for that window.
 	opts := manager.Options{
-		LeaderElection:          false,
-		Scheme:                  scheme,
-		GracefulShutdownTimeout: &timeout,
-		HealthProbeBindAddress:  ":9090",
+		LeaderElection:                true,
```
AlwxSin, line 135: 🟡 **Always-on leader election makes this controller exit on API loss — the opposite of the probe design in the same commit**

The readiness rationale in this PR is explicit: liveness stays a plain ping because "a restart cures nothing here, and the controller recovered by itself once access came back". Enabling leader election unconditionally takes that away.

In controller-runtime v0.23.3, `OnStoppedLeading` pushes `errors.New("leader election lost")` onto the manager's error channel (`pkg/manager/internal.go:637`), `Start` returns it, and `main` hands it to `exitOnError` → `os.Exit(1)`. So after roughly `RenewDeadline` (10 s by default) of an unreachable API server, the single non-HA pod dies and restarts — exactly the behaviour the readiness change was written to avoid, with CrashLoopBackOff on top if the outage lasts.

The measured scenario still holds: the RBAC removal touched ClusterAuthorizationRules, not leases, so the lease kept renewing and nothing restarted. The general API-loss case is the one that changed.

Changelog entry 3 says "Nothing else changes for HA clusters" and is silent about non-HA, where the change is largest. Either state it in `impact`, or reconcile the two designs.

ee/be/modules/140-user-authz/images/webhook/src/internal/web/hook/handler.go
```diff
@@ -67,13 +68,72 @@ type Handler struct {
...
+	// logDenied writes the reviews this webhook denied. The default.
+	logDenied decisionLogMode = iota
+	// logAll writes every review, as before.
+	logAll
+	// logNone writes no review bodies at all.
+	logNone
+)
+
+// DecisionLogEnv names the environment variable that selects the mode: none, denied or all.
+const DecisionLogEnv = "LOG_DECISIONS"
```
AlwxSin, line 94: 🟡 **`LOG_DECISIONS` has no reachable knob: the changelog advertises an env var nobody can set**

`git grep LOG_DECISIONS` over the whole tree hits exactly one line — this constant. `templates/webhook/daemonset.yaml` renders no such env, `openapi/` has no value for it, and there is no FAQ entry. The changelog impact says "Set `LOG_DECISIONS=all` on the `user-authz-webhook` container to restore the previous behaviour", but that container is Helm-managed and Deckhouse reconciles it, so a hand edit is reverted within a minute. In practice the new default becomes the only mode.

The behaviour change itself is safe — nothing consumes the log line (grepped `response body` across the tree: no alert rule, no log-shipper config, no doc). The escape hatch is the problem: either plumb it through the chart (a `userAuthz.webhook.logDecisions` value, or at minimum a conditional env in the DaemonSet), or reword the changelog so it does not promise something an operator cannot do.

modules/150-user-authn/images/user-authn-controller/src/cmd/main.go
```diff
@@ -137,27 +143,47 @@ func newManagerOptions(scheme *runtime.Scheme) manager.Options {
...
+// under its own identity right now - the informers stay "synced" after access is lost, so with
+// cache-sync alone a controller that could do nothing kept both probes at 200.
 func addHealthChecks(mgr manager.Manager) error {
 	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
 		return fmt.Errorf("add healthz check: %w", err)
 	}
 	if err := mgr.AddReadyzCheck("cache-sync", cacheSyncCheck(mgr.GetCache())); err != nil {
 		return fmt.Errorf("add readyz check: %w", err)
 	}
+	if err := mgr.AddReadyzCheck("api-access", apiAccessCheck(mgr.GetAPIReader())); err != nil {
```
AlwxSin, line 179: 🟡 **The user-authn readiness change has no changelog entry**

The PR body says "`user-authn-controller` has the same code and gets the same check", and this line confirms it. But changelog entry 1 is `section: user-authz` and its summary names only `user-authz-controller`; entry 3 (`section: user-authn`) covers only the Lease.

A user-authn operator gets a new way for that pod to flip NotReady with nothing under their module's section.

modules/140-user-authz/template_tests/module_test.go
```diff
@@ -321,6 +348,24 @@ var _ = Describe("Module :: user-authz :: helm template ::", func() {
...
 
+		// A rollout of the single non-HA replica used to take the aggregated API down: the strategy
+		// was hardcoded to maxSurge 0 / maxUnavailable 1, so the only pod was removed before its
+		// replacement existed and the APIService had no endpoints for the whole restart (16 s
+		// measured). Without HA the Deployment now keeps the default strategy, which surges first.
+		It("Should roll permission-browser without taking its API down when there is one replica", func() {
+			deployment := f.KubernetesResource("Deployment", "d8-user-authz", "permission-browser-apiserver")
+			Expect(deployment.Exists()).To(BeTrue())
+			Expect(deployment.Field("spec.replicas").Int()).To(BeEquivalentTo(1))
+			Expect(deployment.Field("spec.strategy.rollingUpdate.maxSurge").Exists()).To(BeFalse(),
```
AlwxSin: 🔵 **This assertion also passes for `strategy: {type: Recreate}`**

`type: Recreate` has no `rollingUpdate.maxSurge` either, so this green test is compatible with a strategy that takes the aggregated API down *harder* than the bug being fixed — which is not what the comment above it claims to guard.

```go
Expect(deployment.Field("spec.strategy").Exists()).To(BeFalse())
```

is the assertion that matches the claim.

modules/140-user-authz/images/user-authz-controller/src/cmd/main.go
```diff
@@ -229,18 +229,48 @@ func envInt(logger logr.Logger, name string, def int) int {
 	return value
 }
 
-const cacheSyncCheckTimeout = 2 * time.Second
+const (
+	cacheSyncCheckTimeout = 2 * time.Second
+	// apiAccessCheckTimeout keeps the readiness check inside the probe's own budget (3 s).
+	apiAccessCheckTimeout = 2 * time.Second
```
AlwxSin, line 247: 🔵 **Two sequential readyz checks at 2 s each against a 3 s kubelet timeout**

`healthz.Handler.serveAggregated` runs the checks in a plain `for` loop over the map and does not short-circuit on failure (controller-runtime v0.23.3, `pkg/healthz/healthz.go:51`). Worst case is `cacheSyncCheckTimeout` (2 s) + `apiAccessCheckTimeout` (2 s) = 4 s against the Deployment's `readinessProbe.timeoutSeconds: 3`.

It only bites on cold start with the API unreachable — once synced, `WaitForCacheSync` returns immediately, so the steady-state access-loss path is ~2 s and fine. The pod is NotReady either way, so this is not a correctness bug; what is lost is the probe body naming which check failed, which is exactly the diagnosability this PR is buying. For user-authz, where the initial adoption of tens of thousands of bindings is an explicit design concern, that window is not hypothetical.

Same arithmetic in `modules/150-user-authn/images/user-authn-controller/src/cmd/main.go:150-152`. Related: the Go constants and the YAML `timeoutSeconds` are coupled by a comment only — the user-authz template tests already have the precedent for a guard (the webhook DaemonSet test asserting `timeoutSeconds >= 5`).

modules/150-user-authn/images/user-authn-controller/src/cmd/main_test.go
```diff
@@ -30,15 +30,22 @@ func TestNewManagerOptions(t *testing.T) {
...
+			wantLeaderElection:   true,
+			wantLeaderElectionID: controllerName,
+			wantLeaderElectionNS: leaderElectionNamespace,
 		},
 		{
-			name:               "ha mode false",
-			haMode:             "false",
-			wantLeaderElection: false,
+			name:                 "ha mode false",
+			haMode:               "false",
```
AlwxSin: 🔵 **`haModeEnv` is dead, and this table's `haMode` dimension now tests nothing**

Nothing reads `HA_MODE` any more; the constant at `modules/150-user-authn/images/user-authn-controller/src/cmd/main.go:49` survives only because line 61 below calls `t.Setenv(haModeEnv, ...)`. The three cases ("ha mode unset/false/true") now assert identical expectations, so one case would say the same thing.

Also: the user-authz template test asserts that `HA_MODE` is gone from the controller's env (`modules/140-user-authz/template_tests/module_test.go:175`), but the equivalent removal in `modules/150-user-authn/templates/controller/deployment.yaml` got no such assertion, even though `modules/150-user-authn/template_tests/` exists.

modules/150-user-authn/images/user-authn-controller/src/cmd/main.go
Jabejixo: Confirmed at `internal.go:637` — a missed renewal exits the process, and the RBAC test never touched the lease, so it could not show this. Kept election on (the non-HA rollout race is the thing being fixed and user-authz already pays this price) but both controllers now run the lease at 60 s / 40 s / 8 s, so a short API-server absence is survived with the cache intact and only a prolonged one restarts the pod; the changelog now says so for non-HA.

ee/be/modules/140-user-authz/images/webhook/src/internal/web/hook/handler.go
Jabejixo: Right — the chart exposes nothing, so on a cluster the default is the only mode. Plumbing a value would mean documentation this PR deliberately avoids, so I reworded the changelog and the PR body instead: the variable is described as a development/test switch, and nothing consumed the old lines.

modules/150-user-authn/images/user-authn-controller/src/cmd/main.go
Jabejixo: Added a `section: user-authn` entry for the readiness check (and the probe timeout it needed), so the user-authn operator sees it under their own module.

modules/140-user-authz/template_tests/module_test.go
Jabejixo: Taken as written: the assertion is now that `spec.strategy` is absent altogether, which is what the comment claims and what `Recreate` would fail.

modules/140-user-authz/images/user-authz-controller/src/cmd/main.go
Jabejixo: Verified in `healthz.go:51` — no short-circuit, so 2 s + 2 s against a 3 s probe on a cold start. Both readiness probes now have `timeoutSeconds: 5`, and the template tests of both modules pin it (`>= 5`), the same guard the webhook DaemonSet test already has.

modules/150-user-authn/images/user-authn-controller/src/cmd/main_test.go
Jabejixo: Removed `haModeEnv`, collapsed the table to one case that also pins the lease timings, and added `modules/150-user-authn/template_tests/controller_test.go` asserting that `HA_MODE` is gone from the Deployment and that the readiness timeout covers both checks.

Jabejixo:
CI note: `Build fuzz images` fails with `line 42: v2: unbound variable` in the workflow shell before any image is built. It fails the same way on the unrelated concurrent branch `registry-fuzz-images` (run 34602991203), so it is the workflow, not this change.

AlwxSin (approved):

deckhouse-BOaTswain:
:red_circle: `e2e: DVP` for `deckhouse:user-authz/ha-probes` [failed](https://github.com/deckhouse/deckhouse/actions/runs/34823155370) in 51m15s.




  `ssh -i ~/.ssh/e2e-id-rsa  debian@10.12.0.146
` - connect for debugging;

  `/e2e/abort dvp;WithoutNAT;containerd;1.34 34823155370 34823155370-1-con-1-34 /sys/deckhouse-oss/install:pr22992 debian@10.12.0.146
` - for abort failed cluster




DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.0.146`


:red_circle: `e2e: DVP, Containerd, Kubernetes 1.34` failed in 49m41s.

</details>

Jabejixo:
/e2e/abort dvp;WithoutNAT;containerd;1.34 34823155370 34823155370-1-con-1-34 /sys/deckhouse-oss/install:pr22992 debian@10.12.0.146

deckhouse-BOaTswain:
:red_circle: `destroy cluster: DVP` for `refs/heads/main` [failed](https://github.com/deckhouse/deckhouse/actions/runs/34829585197) in 1m5s.


:red_circle: `destroy cluster: DVP, Containerd, Kubernetes 1.34` failed in 33s.

</details>

PR #22993, opened 2026-09-11
[user-authz] Fire the slow-render and slow-release alerts on repetition, not on a single observation

Jabejixo:
## Description

`D8UserAuthzRenderSlow` and `D8UserAuthzReleaseSlow` fired on a freshly bootstrapped cluster with
zero ClusterAuthorizationRules, while `CPUStealHigh` was firing on every node. Both alerts are an
average of `deckhouse_helm_operation_seconds` over a window, and an average over a window that holds
one observation is that observation: the single render of the bootstrap, done while every module
starts at once on a node with stolen CPU, kept the alert firing for the whole 30-minute window even
though every later render took about a second. `for: 1m` added nothing.

Both alerts now require repetition before they fire:

- `D8UserAuthzRenderSlow`: the average over 30 minutes still has to exceed 30 seconds, but there
  must be **at least three renders** in the window, and the condition must hold for 15 minutes. One
  slow bootstrap render cannot raise it.
- `D8UserAuthzReleaseSlow`: the window grows to one hour, there must be **at least two releases** in
  it, and the condition must hold for 15 minutes. The one-time migration pass to
  `user-authz-controller` and a slow bootstrap release cannot raise it on their own; a release that
  is slow again does.

Both are also one step less severe: `D8UserAuthzReleaseSlow` moves from severity 5 to 6 and
`D8UserAuthzRenderSlow` from 6 to 7. Neither means the module is down — a slow release fails loudly
through `D8DeckhouseCouldNotRunModule`, and a slow render is a symptom of the node — so neither
deserves to page as if it were.

The descriptions are rewritten to say where the cause actually is. Rendering is CPU work inside the
Deckhouse Pod and does not depend on the number of rules, so the old advice to count
ClusterAuthorizationRules pointed the reader at the wrong thing; the new text names CPU shortage
and bootstrap first and gives the PromQL to compare the module with the others and to look at CPU
steal. The release description no longer blames the size of the release, which stopped growing with
the number of rules once the bindings moved to the controller.

The generated alert reference (`docs/documentation/_data/deckhouse-alerts.yml`) is updated to
match; no other documentation changes.

An alert with a wrong root-cause hint, firing for half an hour on every fresh
cluster whose node is under CPU pressure, costs an engineer a debugging session that ends in
"nothing is wrong". The alerts exist to catch a release that genuinely stops fitting into its
timeout; that is a repeated condition, and the rules now say so.

The alerts shipped with the custom-role aggregation change, which is in `release-1.76` and
`release-1.77`; every cluster on those releases gets the same false alarm on bootstrap and on
every CPU-starved render. The change is two alert rules and their generated reference, with no
effect on the module's behaviour.

## Checklist
- [x] The code is covered by unit tests — the expressions are parsed with the Prometheus PromQL parser; there is no behaviour to unit-test beyond that.
- [ ] e2e tests passed.
- [x] Documentation updated according to the changes — the generated alert reference is in sync; nothing else describes these alerts.

## Changelog entries

```changes
section: user-authz
type: fix
summary: The D8UserAuthzRenderSlow and D8UserAuthzReleaseSlow alerts no longer fire on a single slow render or release, such as a cluster bootstrap under CPU pressure or the one-time migration pass.
impact: |
  D8UserAuthzRenderSlow requires at least three renders averaging above 30 seconds over 30 minutes,
  D8UserAuthzReleaseSlow at least two releases averaging above 5 minutes over one hour, and both must
  hold for 15 minutes. Severities are one step lower (6 and 7). Their descriptions now point at CPU
  shortage on the node rather than at the number of ClusterAuthorizationRules, which does not affect
  render time.
impact_level: low
```

### DIFF
--- a/docs/documentation/_data/deckhouse-alerts.yml
+++ b/docs/documentation/_data/deckhouse-alerts.yml
@@ -3539,26 +3539,19 @@ alerts:
       module: user-authz
       edition: ce
       description: |-
-        Applying the `user-authz` Helm release took more than 5 minutes on average over the last 30 minutes. A release that does not finish within the module release timeout (20 minutes) fails, changes to ClusterAuthorizationRule and AuthorizationRule objects stop being applied and platform updates stall (the `D8DeckhouseCouldNotRunModule` alert fires in that case).
+        Applying the `user-authz` Helm release took more than 5 minutes on average over the last hour, and it was applied at least twice in that hour, so this is not the single slow pass of a cluster bootstrap or of the one-time migration to `user-authz-controller`. A release that does not finish within the module release timeout (20 minutes) fails, changes to ClusterAuthorizationRule and AuthorizationRule objects stop being applied and platform updates stall (the `D8DeckhouseCouldNotRunModule` alert fires in that case).
 
-        The usual cause is the size of the release: thousands of ClusterAuthorizationRule/AuthorizationRule objects, or many ClusterRoles annotated with `user-authz.deckhouse.io/access-level` bound one by one. Check the number of rules and of the module bindings:
-
-        ```bash
-        d8 k get clusterauthorizationrules --no-headers | wc -l
-        d8 k get authorizationrules -A --no-headers | wc -l
-        d8 k get clusterrolebindings -l heritage=deckhouse,module=user-authz --no-headers | wc -l
-        d8 k get rolebindings -A -l heritage=deckhouse,module=user-authz --no-headers | wc -l
-        ```
-
-        Check the module queue and the Deckhouse log for the release:
+        The release itself no longer grows with the number of rules: bindings are managed by `user-authz-controller`, and custom ClusterRoles are aggregated. A slow release therefore points outside the module: an API server that answers slowly, an object stuck with a finalizer, a Deckhouse Pod short of CPU, or the release engine itself. Check what the release is waiting on:
 
         ```bash
         d8 s queue list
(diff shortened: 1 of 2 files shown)


Jabejixo:
CI note: `Build fuzz images` fails with `line 42: v2: unbound variable` in the workflow shell before any image is built. It fails the same way on the unrelated concurrent branch `registry-fuzz-images` (run 34602991203), so it is the workflow, not this change.

Jabejixo:
CI note: `Tests` failed on `Global hooks :: kubernetes_version :: Endpoint with multiple IP's in cluster` (`global-hooks/discovery/kubernetes_version_test.go:586`, an unexpected non-nil error). This PR touches two alert rules and the generated alert reference, nothing a global hook reads; the same test passes on the concurrent unrelated run 34602991203 and on #22992. Re-ran the failed jobs.

z9r5 (approved):

AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [23000](https://github.com/deckhouse/deckhouse/pull/23000) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [23002](https://github.com/deckhouse/deckhouse/pull/23002) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #23003, opened 2026-09-11
[admission-policy-engine] Scope policies in system namespaces by name

Suselz:
## Description

Namespaces holding the components of the platform were kept out of admission by the `heritage: deckhouse` namespace label. That label is set by the platform on its own namespaces, so it matched `d8-*` and never `kube-*`, and it carried no information about what should happen inside those namespaces — they were simply invisible to the module.

This PR scopes them by name and states in the module what happens there.

**Pod Security Standards.** Every namespace named `d8-*` or `kube-*` is checked against the `restricted` standard, regardless of its `security.deckhouse.io/pod-policy` label and of `settings.podSecurityStandards.defaultPolicy`, which govern non-system namespaces only. Violations reach the audit and Deckhouse Console.

**The action is decided in the module, not in the ModuleConfig.** It is the `system_namespaces_action` constant in the chart, and it is `warn`. Raising it to `deny` blocks the converge of every module whose workloads do not comply yet, which is a change to ship with a release rather than a setting to expose; lowering it would hide violations the platform wants to see. The companion constant `system_namespaces_excluded` names the system namespaces that are only ever warned about, whatever the action says and whatever label a module puts on them — it exists for a system namespace that legitimately hosts application workloads, and it is empty while the action stays `warn`. Both are platform internals and neither is documented.

**OperationPolicy and SecurityPolicy** follow the same constants, which matters most for a policy whose selector covers every namespace. A Gatekeeper constraint carries a single `enforcementAction` and cannot OR two namespace lists, so such a policy is rendered as two: the original name keeps the application namespaces with the action the policy asks for, and `d8-system-default-<name>` covers every system namespace with the action from the constant. A policy that cannot be split — one that names no system namespace, that excludes every system namespace it names, or whose list cannot be intersected with `d8-*` exactly, such as `*-system` — keeps one constraint, and that constraint is scoped to application namespaces: it still denies, so a system namespace has to stay out of it.

**No label of a namespace raises a user policy.** A module labels its own namespace `security.deckhouse.io/enable-security-policy-check` to declare it compliant with the platform's own standards; that says nothing about a rule a cluster operator wrote for application workloads. The label therefore raises the Pod Security Standards there and nothing else, and the system-scoped policy constraint carries no label selector at all. Before this, a policy with `enforcementAction: Deny` blocked workloads in every namespace a module had opted in, which is how a cluster-wide policy could stop a platform component from starting. Nothing is split that does not have to be — a policy that only warns or runs in dryrun, that names no system namespace, that already excludes every system namespace it names, or whose namespace list cannot be intersected with `d8-*` exactly, stays a single constraint.

**Webhook routing.** The main webhook now excludes system namespaces by a CEL match condition, and two new webhooks serve them: `system-namespaces` and `system-namespaces-enforce`. Both select the opt-in label **by value** (`In`/`NotIn ["true"]`), exactly as the constraints they serve do, so a namespace labeled with anything else cannot fall between routing and evaluation. Both use `failurePolicy: Ignore`: an unavailable Gatekeeper must never stop a system component from starting, which would otherwise deadlock cluster bootstrap and recovery. `deny-exec-heritage` keeps `failurePolicy: Fail` as the one deliberate exception — denying exec and attach cannot prevent a component from starting.

Every `request.namespace` reference is guarded with `has(request.namespace)`: a cluster-scoped admission request carries no namespace key, and an unguarded reference would error and, under `failurePolicy: Fail`, reject the Gatekeeper constraints the module applies to itself.

The `security-policy-exception` webhook is removed. Its only non-redundant role was routing system namespaces that opted in, which `system-namespaces-enforce` now covers; for every other namespace it duplicated the main webhook.

**Mutations** are scoped the same way and no longer run in `d8-*` or `kube-*`.

**The previous definition of a system namespace keeps working.** A namespace labeled `heritage: deckhouse` is not a user namespace whatever it is called, as it was when that label was the whole definition. It is excluded from Pod Security Standards in non-system namespaces, from every user-scoped policy constraint, from the main validating webhook and from mutations — all of it expressible as one more AND, so no extra constraint object and no hook. The `system-namespaces-enforce` webhook routes by the opt-in label alone, with no name condition, exactly as the `security-policy-exception` webhook it replaces did: a namespace a module opted in reaches Gatekeeper even if it is not named `d8-*`.

**Reserved name prefixes.** `d8-system-default-` and `d8-pod-security-` are rejected as OperationPolicy and SecurityPolicy names by a new validating webhook, so a policy cannot claim a name that a generated constraint already owns. A name longer than 235 characters is rejected for the same reason: with the longest prefix on top it would exceed the 253 characters Kubernetes allows an object name, and the failure would otherwise surface as a broken release at converge. No prefix may be a prefix of another, or two policies could derive the same constraint name; a test pins that.

**A SecurityPolicyException no longer requires the feature it permits.** `d8hostnetwork` and `d8hostprocesses` compared the pod against the exception alone, so `allowedValue: true` turned into a requirement: a pod whose `hostNetwork` was `false` differed from the `true` the exception permitted and was denied, with a message that contradicted itself — `hostNetwork: false | policy allows: false | SPE allows: true`. A module that enables the feature only for certain settings therefore could not ship a single exception. Not using a feature is never a violation of a rule that restricts it, so `false` is now permitted whatever the exception says; an exception in force still decides the `true` case and still narrows a constraint that would have allowed it. These were the only two rules comparing a boolean against an exception by hand — every other boolean field resolves through `lib.check_bool`, which tests the policy first.

**A SecurityPolicy that only forbids hostNetwork now denies again.** `allowedHostPorts` is optional, and omitting it places no restriction on host ports, while an empty list forbids all of them. The rule read `input.parameters.ranges` unconditionally, so for a policy that sets only `allowHostNetwork` the absent field left the whole violation rule undefined and the policy silently stopped denying anything, `hostNetwork` included. The check is total either way now, and an exception that names host ports governs them on its own.

The gap that let this through is in the test model, not in the rule: coverage is a set of (field, scenario) pairs, satisfied when each pair appears in some case, so the exception axis and the object axis are covered independently and never have to combine. That combination now has a name, `speUnused`, required of every boolean exception field.

The `semverCompare ">= 1.27"` guards around the `matchConditions` blocks are dropped, since that Kubernetes version is out of support.

Before this change the module had no visibility into the namespaces of the platform itself, and the mechanism that produced that blind spot was a label whose meaning was unrelated to the question being asked. The module now states the rule explicitly: system namespaces are observed, and whether a violation is blocked there is the operator's decision rather than a side effect of how a module labels itself.

**Behaviour changes worth calling out.**

The previous exclusion matched `d8-*` only, so `kube-*` namespaces — `kube-system` in particular — were subject to policies in full, including a `defaultPolicy: Restricted` that would deny there. They are now warn-only unless the operator says otherwise. This is a relaxation, deliberate and consistent with the rest of the change: the platform must not be able to block its own control plane through a cluster-wide policy setting.

Enforcement in system namespaces no longer depends on `defaultPolicy`. It used to: with `defaultPolicy: Restricted` the restricted constraint went missing, and with `defaultPolicy: Privileged` the baseline one, so a namespace that asked for enforcement was measured against the weaker standard while the stronger only warned.

Not necessarily.

## What a second review changed

The branch was re-reviewed from scratch after the stand run, and three defects came out of it:

- A denying policy whose namespace list cannot be intersected with `d8-*` — `*-system`, for instance — kept a single constraint that still denied, and `kube-system` matches that glob. Such a policy is now scoped to application namespaces like every other unsplit denying policy.
- A system namespace labeled `security.deckhouse.io/trivy-provider` reached no webhook at all: the exclusion was copied from a webhook that only ever served opted-in namespaces, and on the warn path it takes the namespace out of the one webhook that would route it.
- Every Gatekeeper constraint was validated twice, because a `namespaceSelector` cannot keep a cluster-scoped object out of the enforce webhook. It now asks for a namespace.

The exclusion machinery is gone with them. `system_namespaces_excluded` has been empty since it became a constant, which left an intersection helper, a namespace scope, a Pod Security Standards block, a policy variant and a reserved prefix that nothing could reach and no test could exercise.

## Checklist
- [ ] e2e tests passed.

Tested on a stand running this branch, with server-side dry-run probes for every namespace class. What the run confirms, in the order the documentation states it:

- A user policy with `enforcementAction: Deny` renders exactly two constraints — its own name and `d8-system-default-<name>` — and nothing else.
- That policy denies in an application namespace (`test`), and only warns in `d8-chrony`, in `kube-public`, and in `d8-monitoring` and `d8-user-authz`, which their modules have opted into enforcement. The last two are the case this PR changes.
- A namespace labeled `heritage: deckhouse` but named otherwise stays outside the policy altogether: no denial, no warning.
- Pod Security Standards reach system namespaces: a privileged pod warns in `d8-chrony` through `d8-pod-security-*-system`, and is denied in `d8-user-authz` through `d8-pod-security-*-deny-d8-default`.
- The reserved prefixes behave as documented: `d8-system-default-` and `d8-pod-security-` are rejected, the longest allowed name passes and one character more does not. (The stand run predates the removal of the third prefix: it was rejected there as well, and the limit was 234 while that prefix existed.)
- Each of the five single-constraint cases renders one object, including a policy scoped to system namespaces only, which keeps its name and warns.
- A mutation applies in `test` and in neither `d8-chrony` nor the `heritage`-labeled namespace.
- A cluster-scoped object still passes the main webhook, and the audit reports violations from the warn-only constraints (15 on `d8-pod-security-restricted-system`).

Two statements in the documentation did not survive the run and were corrected: a workload is not always let through in a system namespace — the standards deny it where a module has opted in — and a policy that names system namespaces only is a fifth case where nothing is split.

## Changelog entries

```changes
type: fix
summary: Scope policies in system namespaces by namespace name.
impact: |
  A namespace named `d8-*` or `kube-*`, and a namespace labeled `heritage: deckhouse` whatever it is called, is treated as the platform's own. Such a namespace is checked against the `restricted` Pod Security Standard: a violation is reported in the audit and in Deckhouse Console and the workload still starts, unless the module that owns the namespace has opted it into enforcement. None of this is configurable.

  User policies only warn there. An OperationPolicy or a SecurityPolicy with `enforcementAction: Deny` reports a violation in a system namespace and does not block the workload, whatever the namespace is labeled with, including a namespace a module has opted into enforcement.

  Enforcement in `kube-*` namespaces is relaxed: previously the exclusion matched `d8-*` only, so `kube-system` and the other `kube-*` namespaces were subject to user policies in full.

  Enforcement in system namespaces no longer depends on `defaultPolicy`. With `defaultPolicy: Restricted` the restricted constraint was previously not created, and with `defaultPolicy: Privileged` the baseline one, so a namespace that had opted into enforcement was checked against the weaker standard only.

  Gatekeeper mutations no longer run in `d8-*` and `kube-*` namespaces.

  The name prefixes `d8-system-default-` and `d8-pod-security-` are reserved: creating or updating an OperationPolicy or SecurityPolicy with such a name is rejected, as is a name longer than 235 characters, which would push a derived constraint name past the limit Kubernetes sets for an object name.

  Expect more audit data. Every pod in a system namespace is now measured against the full `restricted` set, and control-plane static pods violate most of it, so the constraints report violations that no constraint used to see. Each violation is one Prometheus series from `constraint-exporter` and one entry in the constraint status, up to the audit limit of 200 per constraint. None of it alerts: the audit rules fire on `deny` only.
impact_level: high
---
type: fix
summary: A SecurityPolicy that only forbids hostNetwork denies it again.
impact: |
  A SecurityPolicy that set `allowHostNetwork` without `allowedHostPorts` stopped denying anything at all, including the `hostNetwork` it forbade. Such a policy is enforced again. Review the policies that set only `allowHostNetwork`: workloads that ran under them may start being denied.
impact_level: high
---
type: fix
summary: A SecurityPolicyException no longer requires the feature it permits.
impact: |
  An exception permitting `hostNetwork`, `hostPID` or `hostIPC` denied every pod that did not use it, because the pod was compared against the exception alone. A module that enables such a feature only for certain settings could not ship a single exception: with the setting off, all of its pods were blocked. Not using a feature is no longer treated as a violation; an exception still decides whether the feature may be used.
impact_level: low
```


modules/015-admission-policy-engine/templates/_helpers.tpl
```diff
@@ -9,28 +9,281 @@
...
+    {{- $nsSelector := $match.namespaceSelector | default dict }}
+    {{- $namespaces := $nsSelector.matchNames | default list }}
+    {{- $excluded := $nsSelector.excludeNames | default list }}
+
+    {{- if eq $scope "user" }}
+      {{- $excluded = concat $excluded (include "system_namespaces" . | fromYamlArray) | uniq }}
+    {{- else if eq $scope "system-excluded" }}
+      {{/* The policy's own excludeNames stay in force: the operator's list widens what is spared
+           from enforcement, it does not drag back a namespace the policy author ruled out. */}}
+      {{- $namespaces = $match.d8ExcludedNamespaces }}
```
AlwxSin: 🟡 **`excludeNamespaces` is used verbatim as the constraint's namespace list, without intersecting it with the policy's own scope or with `d8-*`/`kube-*`**

`scoped_policy_variants` builds the `d8-system-excluded-<name>` variant with `d8ExcludedNamespaces = $systemExclude` (line 281), and this branch then replaces the policy's namespaces with it wholesale, dropping `$scope.names` entirely. `d8-pod-security-<std>-system-excluded` does the same at line 464 (`namespaces: {{ $systemExclude }}`). Two consequences:

* **A policy is evaluated outside its declared namespaces.** `syspolicy` (`matchNames: [d8-monitoring, kube-*]`, `Deny`) plus `systemNamespaces.excludeNamespaces: [d8-cert-manager]` renders `d8-system-excluded-syspolicy` with `namespaces: [d8-cert-manager]` — a namespace the policy author never named.
* **A non-system entry escapes the system scope.** `excludeNamespaces: [default]` or `[*-system]` renders a warn constraint that matches a *user* namespace, which the main webhook routes to Gatekeeper as usual. That contradicts the documented contract in `openapi/config-values.yaml:79` ("Names outside `d8-*` and `kube-*` have no effect") and `openapi/doc-ru-config-values.yaml:56`.

The test suite cannot catch this as written: the only spec that sets `excludeNamespaces` (`"Never blocks in the system namespaces the operator excluded"`) uses `exprpolicy`, which has no `matchNames` at all, so `namespaces: [d8-monitoring]` comes out coincidentally correct. `syspolicy` — the one fixture with an explicit `matchNames` — is never rendered with `excludeNamespaces` set.

Intersecting `$systemExclude` with `$scope.names` before setting `d8ExcludedNamespaces` (and with `system_namespaces` in the PSS block) fixes both. Everything involved is `warn`, so nothing is blocked wrongly — but the constraint scope is wrong and the documentation states the opposite.

modules/015-admission-policy-engine/webhooks/validating/reserved_policy_names.py
```diff
@@ -0,0 +1,79 @@
...
+# constraints, because a constraint carries a single enforcementAction. The extra constraints
+# are named by prefixing the policy name, so a policy that claims one of those prefixes would
+# render a constraint under a name another policy already owns and break the release.
+
+from typing import Optional
+
+from deckhouse import hook
+from dotmap import DotMap
+
+RESERVED_NAME_PREFIXES = (
```
AlwxSin, line 27: 🟡 **The `changes` block (and the PR description) describes a different implementation than the one on the branch**

Anchored here because this tuple is the authoritative list the changelog disagrees with. `CHANGELOG/CHANGELOG-v<X.Y>.0.yml` is autogenerated from the ```` ```changes ```` block, so the drift ships as release notes on an `impact_level: high` change:

* it reserves `d8-system-warn-` — a prefix the code does **not** use, and which `reserved_policy_names_test.py:26` explicitly asserts is *allowed*;
* it omits `d8-system-default-` and `d8-system-excluded-`, which are the prefixes actually reserved here;
* it never mentions `podSecurityStandards.systemNamespaces.enforcementAction` / `excludeNamespaces` — the entire new operator-facing API added in `1c544ca` / `35b6d7a`;
* it omits a tightening. Dropping the `defaultPolicy` gate on the TODO block in `_helpers.tpl` means `defaultPolicy: Restricted` plus a `d8-*` namespace labeled `enable-security-policy-check: "true"` now enforces `restricted` where previously no constraint was rendered at all (symmetrically for `Privileged` and baseline). That is ~14 platform namespaces (`d8-user-authn`, `d8-cloud-provider-*`, `d8-admission-policy-engine`, …). The practical risk is low — those workloads already pass both standards under the default `defaultPolicy: Baseline` — but the changelog currently describes only the `kube-*` relaxation.

The PR description prose carries the same `d8-system-warn-<name>` naming.

modules/015-admission-policy-engine/templates/_helpers.tpl
```diff
@@ -124,8 +381,116 @@ spec:
...
+  iteration of the default action, which is always present in
+  internal.podSecurityStandards.enforcementActions. That keeps it at one object per standard
+  instead of one per action.
+*/}}
+{{- if eq $policyAction ($context.Values.admissionPolicyEngine.podSecurityStandards.enforcementAction | default "deny" | lower) }}
+---
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: {{ $policyCRDName }}
+metadata:
+  name: d8-pod-security-{{$standard}}-system
```
AlwxSin, line 414: 🟡 **15 new always-on warn constraints over `d8-*` + `kube-*` multiply violation-metric cardinality and constraint status size**

This block renders for all 15 PSS constraints (10 baseline + 5 restricted) with no `pod-policy` opt-in, so every pod in `kube-system` and in every `d8-*` namespace is now audited against the full `restricted` set. Control-plane static pods violate nearly every restricted check (hostNetwork, hostPath, runAsRoot, missing seccomp, capabilities).

`constraint-exporter` emits **one Prometheus series per violation**, with `violation_msg` as a label (`images/constraint-exporter/src/pkg/gatekeeper/metrics.go:33-37`), and audit runs with `--constraint-violations-limit=200` (`templates/audit-deployment.yaml:71`). Worst case is roughly 15 × 200 = 3000 new high-cardinality series, plus the same violations stored in each constraint's `status.violations` in etcd, plus an admission warning header on every pod create in those namespaces.

This does not page — `monitoring/prometheus-rules/propagated-audit.yaml` filters on `violation_enforcement="deny"` — but the cost is real and is not mentioned anywhere in the PR. Worth either a lower per-constraint violation limit for the `-system` family, or a line in the changelog impact.

modules/015-admission-policy-engine/webhooks/validating/reserved_policy_names.py
```diff
@@ -0,0 +1,79 @@
...
+def validate(ctx: DotMap, output: hook.ValidationsCollector):
+    error = check_reserved_name(ctx.review.request.object.metadata.name)
+    if error is not None:
+        output.deny(error)
+        return
+
+    output.allow()
+
+
+def check_reserved_name(name: str) -> Optional[str]:
```
AlwxSin, line 71: 🔵 **The webhook guards the prefix but not the resulting name length**

`d8-system-excluded-` is 19 characters. A policy with a 240-character name passes this check, then renders a constraint whose name exceeds the 253-character RFC 1123 limit — and the failure surfaces as a broken Helm release at converge time rather than at policy creation. Since this webhook already owns the "a policy name must not break the generated constraints" invariant, `len(name) + len(longest prefix) <= 253` belongs here too.

modules/015-admission-policy-engine/templates/validatingwebhookconfiguration.yaml
```diff
@@ -147,24 +203,38 @@ webhooks:
...
     apiVersions: ["*"]
     resources: ["pods/ephemeralcontainers"]
     operations: ["UPDATE"]
-{{- if semverCompare ">= 1.27" .Values.global.discovery.kubernetesVersion }}
   matchConditions:
+    - name: 'system-namespaces'
+      expression: '{{ include "system_namespaces_cel" . }}'
     - name: 'exclude-virtualization'
       expression: '!(request.userInfo.groups.exists(e, (e == "system:serviceaccounts:d8-virtualization")))'
+  timeoutSeconds: 15
```
AlwxSin: 🔵 **`timeoutSeconds: 15` on the two new `failurePolicy: Ignore` webhooks**

Same for `system-namespaces` at line 167. `d8-*` and `kube-*` were previously excluded from admission entirely; now every pod create there goes through these two webhooks. The deadlock case is covered — a down Gatekeeper gives connection-refused and fails fast, and Gatekeeper's own webhook pods are exempt via `objectSelector` — but a Gatekeeper that is running and hung adds up to 15s per pod creation across the whole control plane. For warn-only paths a shorter timeout (3-5s) costs nothing.

modules/015-admission-policy-engine/templates/validatingwebhookconfiguration.yaml
```diff
@@ -104,41 +108,93 @@ webhooks:
...
-- name: deny-exec-heritage.admission-policy-engine.deckhouse.io
-  matchConditions:
-    - name: system-namespace-prefix
-      expression: 'request.namespace.startsWith("d8-") || request.namespace.startsWith("kube-")'
+{{/*
+  Webhook for system namespaces that have not opted into policy enforcement.
+
+  The webhook above skips every namespace named `d8-*` or `kube-*`, so without this one Gatekeeper
+  never sees their workloads. The constraints that match here run in warn mode only: Pod Security
+  Standards through `d8-pod-security-*-system`, and user policies through their `-system-warn`
```
AlwxSin: 🔵 **Comments reference constraint names that no longer exist**

The variant is `d8-system-default-<name>`, not `-system-warn` — `d8-system-warn-` is not a name the code generates, and `reserved_policy_names_test.py:26` asserts it is allowed.

Line 176 has a related one: `d8-pod-security-*-d8-default` only matches the default-action case; the non-default-action constraint is named `d8-pod-security-<std>-<action>-d8`.

modules/015-admission-policy-engine/templates/_helpers.tpl
Suselz: Half of this is no longer applicable, the other half was real and is fixed.

The operator-facing `systemNamespaces` settings are gone: the exclusion list is now the `system_namespaces_excluded` constant in the chart, so an operator can no longer put `default` or `*-system` into it, and the documentation that promised "names outside `d8-*` and `kube-*` have no effect" is gone with it.

The scoping bug stood on its own, though, and would have bitten the first maintainer to fill that constant. Both places now intersect instead of substituting, through a new `namespace_globs_intersect` helper: `scoped_policy_variants` intersects the constant with `$scope.names`, so a policy is never evaluated outside the namespaces its author named and the variant is skipped entirely when the intersection is empty; `pod_security_standard_base` intersects it with `system_namespaces`, so an entry that is not a system namespace cannot render a warn constraint over a user namespace.

The constant is empty as shipped, so the path stays dormant and untested by rendering; the two specs that used to set the values now pin the constants instead.

modules/015-admission-policy-engine/webhooks/validating/reserved_policy_names.py
Suselz: Outdated. The changelog block was rewritten twice since: `d8-system-warn-` is gone from it and from the prose, the reserved list matches this tuple exactly, and the `defaultPolicy` tightening you describe is called out as its own paragraph in the impact.

The `systemNamespaces` API the comment asks the changelog to document no longer exists — the action and the exclusions are chart constants now, deliberately not operator-facing, so there is nothing to announce there.

modules/015-admission-policy-engine/templates/_helpers.tpl
Suselz: Valid, and the cost is now in the changelog impact rather than nowhere: every pod in a system namespace measured against the full `restricted` set, one Prometheus series per violation from `constraint-exporter`, the same violations in each constraint status, up to the audit limit of 200 per constraint, and no alert because the audit rules fire on `deny` only.

Not lowering the limit for the `-system` family: `--constraint-violations-limit` is a flag on the audit process, not a per-constraint field, so it cannot be set for one family without setting it for all of them. If the volume turns out to hurt in practice, the lever to reach for is that flag, and it is a separate change.

modules/015-admission-policy-engine/webhooks/validating/reserved_policy_names.py
Suselz: Fixed. `MAX_NAME_LENGTH` is derived from the longest reserved prefix (253 − 19 = 234) and checked alongside the prefix, so the policy is rejected at creation rather than at converge. Three tests cover it, including one that pins the arithmetic by asserting that the longest allowed name plus every prefix still fits 253 characters.

modules/015-admission-policy-engine/templates/validatingwebhookconfiguration.yaml
Suselz: Fixed, both are `timeoutSeconds: 5` now, with a comment giving the reason you did: nothing on these two paths can deny a workload, so a hung Gatekeeper must give the pod back quickly instead of holding every create in the platform's own namespaces. A template test pins the bound.

`deny-exec-heritage` and the main webhook keep 15 seconds: those can deny, and a timeout there means a request goes through unevaluated.

modules/015-admission-policy-engine/templates/validatingwebhookconfiguration.yaml
Suselz: Fixed. The first comment now says `d8-pod-security-<standard>-system` and `d8-system-default-<name>`, and the second spells out both enforcing names — `d8-pod-security-<standard>-<action>-d8-default` for the module's own enforcement action and `d8-pod-security-<standard>-<action>-d8` for any other.

AlwxSin (approved):

docs/documentation/pages/admin/configuration/security/SECURITY-POLICIES.md
```diff
@@ -288,6 +288,44 @@ As with policy assignment, enforcement mode can be set:
...
+A policy with `enforcementAction: Deny` blocks workloads in application namespaces and only reports violations in a system namespace.
+No label of the namespace changes that: a module that hardens its own namespace raises the Pod Security Standards there, which says nothing about a policy you wrote for application workloads.
+
+A denying policy that reaches system namespaces is therefore rendered as two Gatekeeper constraints, both visible in the audit and in Deckhouse Console:
+
+- The policy's own name: For application namespaces, with the action the policy asks for.
+- `d8-system-default-<policy>`: For system namespaces, in `warn` mode.
+
+The `d8-system-default-` and `d8-pod-security-` prefixes are reserved: a policy whose name starts with one of them is rejected on creation.
+A policy name is limited to 234 characters for the same reason, so that the derived constraint names stay within the 253-character limit of a Kubernetes object name.
```
AlwxSin: 🟡 **The name limit is 235 now, not 234**

Dropping `d8-system-excluded-` from `RESERVED_NAME_PREFIXES` changed the longest prefix from 19 characters to 18, so `MAX_NAME_LENGTH = 253 - max(len(prefix))` evaluates to **235** (`reserved_policy_names.py:36`, with `d8-system-default-` at 18 and `d8-pod-security-` at 16).

234 is left in six places:

* this line, and `SECURITY-POLICIES_RU.md:318`
* `docs/site/pages/virtualization-platform/documentation/admin/platform-management/security/SECURITY-POLICIES.md:93` and its `_RU.md:94`
* the PR description twice — including *"a 234-character name passes and a 235-character one does not"*, which is now exactly inverted at the boundary
* the ```` ```changes ```` block, on an `impact_level: high` entry, so it ships as release notes

Worth noting why it slipped and why it will slip again: both new length tests build their input from the constant itself (`"a" * reserved.MAX_NAME_LENGTH`), so they pass at any value and cannot catch the prose. The fix is text-only, and nothing guards these numbers the next time the prefix list changes.

modules/015-admission-policy-engine/templates/validatingwebhookconfiguration.yaml
```diff
@@ -104,41 +119,103 @@ webhooks:
...
     matchExpressions:
     - key: security.deckhouse.io/enable-security-policy-check
-      operator: Exists
+      operator: In
+      values: ["true"]
   {{- if include "trivy.provider.enabled" . }}
     - key: security.deckhouse.io/trivy-provider
       operator: DoesNotExist
-  {{- end }} # If trivy-provides is disabled, namespaces with the label should also be checked
+  {{- end }} # If trivy-provider is disabled, namespaces with the label should also be checked
```
AlwxSin: 🔵 **The `trivy-provider` exclusion was removed from one of the two system webhooks, not both**

The comment added on `system-namespaces` this round (line 154) gets it right: that webhook is the fallback path, and a namespace taken out of it reaches no admission at all, because the main webhook already skips it by name. The same reasoning applies verbatim to `system-namespaces-enforce`, where the exclusion is still in place (lines 214-218).

A system namespace carrying both `security.deckhouse.io/enable-security-policy-check: "true"` and `security.deckhouse.io/trivy-provider` matches no webhook:

* main — skipped by the `system_namespaces_cel` match condition;
* `system-namespaces` — excluded by `enable-security-policy-check NotIn ["true"]`;
* `system-namespaces-enforce` — excluded by `trivy-provider DoesNotExist`.

The shape is inherited from the `security-policy-exception` webhook this one replaces, not introduced here, and reaching it is narrow: `trivy-provider` is applied by an operator (no module namespace template in the repo sets it), while `enable-security-policy-check` is written by the module that owns the namespace. But the asymmetry now sits in code rewritten in this round, next to a comment that argues against it.

docs/documentation/pages/admin/configuration/security/SECURITY-POLICIES.md
Suselz: Correct on the arithmetic and on the cause. `d8-system-default-` at 18 characters is the longest prefix left, so the webhook allows 235, and every place that quoted 234 was wrong.

Fixed in e129afde: the four pages say 235, and so do the pull request body and the `changes` block. The one remaining 234 is in the stand-run notes, where it is the number that was in force during the run — the third prefix still existed then.

The deeper point stands too: both length tests build their input from the constant, so no value of it can fail them. Added `test_the_documented_limit_is_the_one_in_force`, which asserts `MAX_NAME_LENGTH == 235` and says in a comment that the pages quote it. Changing the prefix list now fails a test that names the documentation instead of passing quietly.

modules/015-admission-policy-engine/templates/validatingwebhookconfiguration.yaml
Suselz: Agreed, and fixed in e129afde: the exclusion is off `system-namespaces-enforce` as well, with the same comment stating why.

The combination is narrow, as you say, but the asymmetry was the worse half of it — the reasoning written on one webhook argued against the code on the other.

Two notes on what the removal costs. A namespace labeled `security.deckhouse.io/trivy-provider` on the enforce path is now routed to Gatekeeper by this webhook; if the out-of-tree `operator-trivy` still routes the same namespace through a configuration of its own, such a request reaches Gatekeeper twice. That is a warning printed twice at worst, against a namespace that reached no webhook at all before. And `trivy.provider.enabled` had no other caller, so the helper is gone with its last use.

AlwxSin (approved):

docs/documentation/pages/admin/configuration/security/SECURITY-POLICIES_RU.md
```diff
@@ -272,7 +272,7 @@ status:
 Допустимые режимы применения политик:
 
 - `deny` — запрещает выполнений действий.
-- `dryrun` — не влияет на выполнение действий и используется для отладки. Информацию о событиях можно посмотреть в Grafana или в консоли с помощью команды `d8 k` или `kubectl`.
+- `dryrun` — не влияет на выполнение действий и используется для отладки. Информацию о событиях можно посмотреть в веб-интерфейсе Deckhouse или с помощью команды `d8 k`.
```
z9r5:
```suggestion
- `dryrun` — не влияет на выполнение действий и используется для отладки. Информацию о событиях можно посмотреть в веб-интерфейсе Deckhouse Platform или с помощью команды `d8 k`.
```

docs/documentation/pages/admin/configuration/security/SECURITY-POLICIES.md
```diff
@@ -288,6 +288,44 @@ As with policy assignment, enforcement mode can be set:
...
+Policies apply to them differently from application namespaces, and that difference is not configurable.
+The platform also labels the namespaces it creates with `heritage: deckhouse`. A namespace that carries the label but is named otherwise is left out of policies and mutations written for application namespaces; the Pod Security Standards below follow the names.
+
+Every namespace named `d8-*` or `kube-*` is checked against the `restricted` standard.
+The `security.deckhouse.io/pod-policy` label and the [`settings.podSecurityStandards.defaultPolicy`](/modules/admission-policy-engine/configuration.html#parameters-podsecuritystandards-defaultpolicy) parameter do not apply there.
+A violation is recorded in the audit and shown in Deckhouse Console, and the workload still starts.
+The exception is a namespace whose module has hardened it: there the standards are enforced and a violating workload is denied.
+
+These checks cannot be tuned from outside the platform.
+The labels that govern them, and the workloads they cover, belong to the module that owns the namespace, and Deckhouse returns both to their declared state the next time it applies the configuration.
```
z9r5:
```suggestion
The labels that govern them, and the workloads they cover, belong to the module that owns the namespace, and Deckhouse Platform returns both to their declared state the next time it applies the configuration.
```

docs/documentation/pages/admin/configuration/security/SECURITY-POLICIES_RU.md
```diff
@@ -290,6 +290,44 @@ status:
...
+Политики применяются к ним иначе, чем к прикладным неймспейсам, и это поведение не настраивается.
+Создаваемые платформой неймспейсы дополнительно помечены лейблом `heritage: deckhouse`. Неймспейс с этим лейблом, но с другим именем не попадает под политики и мутации, написанные для прикладных неймспейсов; Pod Security Standards ниже действуют по именам.
+
+Каждый неймспейс с именем `d8-*` или `kube-*` проверяется по стандарту `restricted`.
+Лейбл `security.deckhouse.io/pod-policy` и [параметр `settings.podSecurityStandards.defaultPolicy`](/modules/admission-policy-engine/configuration.html#parameters-podsecuritystandards-defaultpolicy) в них не действуют.
+Нарушение фиксируется в отчётах безопасности и отображается в веб-интерфейсе Deckhouse, нагрузка при этом запускается.
+Исключение — неймспейс, в котором владеющий им модуль включил принудительное применение: там стандарты применяются, и нарушающая их нагрузка блокируется.
+
+Настроить эти проверки извне платформы нельзя.
+Управляющие ими лейблы и сама нагрузка принадлежат владеющему неймспейсом модулю, и Deckhouse возвращает их к заявленному состоянию при следующем применении конфигурации.
```
z9r5:
```suggestion
Управляющие ими лейблы и сама нагрузка принадлежат владеющему неймспейсом модулю, и Deckhouse Platform возвращает их к заявленному состоянию при следующем применении конфигурации.
```

docs/site/pages/virtualization-platform/documentation/admin/platform-management/security/SECURITY-POLICIES.md
```diff
@@ -65,6 +65,44 @@ As with policy assignment, enforcement mode can be set:
...
+Policies apply to them differently from application namespaces, and that difference is not configurable.
+The platform also labels the namespaces it creates with `heritage: deckhouse`. A namespace that carries the label but is named otherwise is left out of policies and mutations written for application namespaces; the Pod Security Standards below follow the names.
+
+Every namespace named `d8-*` or `kube-*` is checked against the `restricted` standard.
+The `security.deckhouse.io/pod-policy` label and the [`settings.podSecurityStandards.defaultPolicy`](/modules/admission-policy-engine/configuration.html#parameters-podsecuritystandards-defaultpolicy) parameter do not apply there.
+A violation is recorded in the audit and shown in Deckhouse Console, and the workload still starts.
+The exception is a namespace whose module has hardened it: there the standards are enforced and a violating workload is denied.
+
+These checks cannot be tuned from outside the platform.
+The labels that govern them, and the workloads they cover, belong to the module that owns the namespace, and Deckhouse returns both to their declared state the next time it applies the configuration.
```
z9r5:
```suggestion
The labels that govern them, and the workloads they cover, belong to the module that owns the namespace, and Deckhouse Platform returns both to their declared state the next time it applies the configuration.
```

docs/site/pages/virtualization-platform/documentation/admin/platform-management/security/SECURITY-POLICIES_RU.md
```diff
@@ -66,6 +66,44 @@ Deckhouse Virtualization Platform (DVP) позволяет управлять б
...
+Политики применяются к ним иначе, чем к прикладным неймспейсам, и это поведение не настраивается.
+Создаваемые платформой неймспейсы дополнительно помечены лейблом `heritage: deckhouse`. Неймспейс с этим лейблом, но с другим именем не попадает под политики и мутации, написанные для прикладных неймспейсов; Pod Security Standards ниже действуют по именам.
+
+Каждый неймспейс с именем `d8-*` или `kube-*` проверяется по стандарту `restricted`.
+Лейбл `security.deckhouse.io/pod-policy` и [параметр `settings.podSecurityStandards.defaultPolicy`](/modules/admission-policy-engine/configuration.html#parameters-podsecuritystandards-defaultpolicy) в них не действуют.
+Нарушение фиксируется в отчётах безопасности и отображается в веб-интерфейсе Deckhouse, нагрузка при этом запускается.
+Исключение — неймспейс, в котором владеющий им модуль включил принудительное применение: там стандарты применяются, и нарушающая их нагрузка блокируется.
+
+Настроить эти проверки извне платформы нельзя.
+Управляющие ими лейблы и сама нагрузка принадлежат владеющему неймспейсом модулю, и Deckhouse возвращает их к заявленному состоянию при следующем применении конфигурации.
```
z9r5:
```suggestion
Управляющие ими лейблы и сама нагрузка принадлежат владеющему неймспейсом модулю, и Deckhouse Platform возвращает их к заявленному состоянию при следующем применении конфигурации.
```

Lada7878 (approved):

z9r5 (approved):

docs/documentation/_data/deckhouse-alerts.yml
```diff
@@ -5384,7 +5384,7 @@ alerts:
           )
           ```
 
-        - Alternatively, check the admission-policy-engine Grafana dashboard.
+        - Alternatively, check the admission-policy-engine reports in Deckhouse Console.
```
z9r5, line 5387: @Lada7878 Deckhouse Platform UI?

Lada7878 (approved):

z9r5 (approved):

deckhouse-BOaTswain:

  :fast_forward: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` for `deckhouse:ape-restricted-warn` [started](https://github.com/deckhouse/deckhouse/actions/runs/35971842934).


Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@158.160.52.3`

PR #23006, opened 2026-09-12
[admission-policy-engine] Fix a SecurityPolicy that only forbids hostNetwork

Suselz:
## Description

A `SecurityPolicy` that forbids host networking and says nothing about host ports silently did nothing:

```yaml
spec:
  enforcementAction: Deny
  policies:
    allowHostNetwork: false
```

`allowedHostPorts` is optional, so the chart renders no `parameters.ranges` for such a policy. The Rego read that parameter unconditionally through `port_ranges(false) := input.parameters.ranges`, and in Rego an absent key makes the expression undefined, which makes the whole enclosing `violation` rule undefined. No violation was produced — not for host ports, and not for `hostNetwork` either, even though that check had already concluded the pod was in breach.

The port check is now total. An absent `ranges` means what the CRD documents for an omitted field — no restriction on host ports — while an empty list keeps forbidding every one of them, and an exception that names host ports still governs them on its own.

Pod Security Standards were never affected: they always render `ranges`, empty when the standard forbids host ports.

The test suite gains a constraint with no `ranges` at all, covering a denied `hostNetwork: true`, an allowed pod without host networking, and an allowed host port.

This is a silent policy bypass. The policy reports `SYNCED: True`, the constraint exists in the cluster and Gatekeeper enforces it — it simply never finds a violation. Nothing in the status, the audit or the Grafana dashboard distinguishes it from a policy that is genuinely satisfied, so an operator has no way to notice that host networking is not actually being denied.

Found while testing #23003 on a cluster: a policy written to confirm that change did not deny what it was written to deny.

Worth backporting. A cluster relying on such a policy to forbid host networking is unprotected today, and the fix is confined to one Rego rule with no change to any other policy's behaviour.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix
summary: A SecurityPolicy that sets `allowHostNetwork` without `allowedHostPorts` no longer stops denying host networking.
impact: |
  Such a policy previously produced a constraint that never reported a violation, so host networking went undenied while the policy appeared to be in force. It now denies as written. Review clusters where a SecurityPolicy sets `allowHostNetwork: false` and omits `allowedHostPorts`: workloads that were running with `hostNetwork: true` in scope of that policy start being rejected.
impact_level: high
```

### DIFF
--- a/modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allow-host-network.yaml
+++ b/modules/015-admission-policy-engine/charts/constraint-templates/templates/security/allow-host-network.yaml
@@ -89,8 +89,7 @@ spec:
                 spe_ports_path := ["spec", "network", "hostPorts"]
                 spe_ports_applied := path_value_resolved(spe, spe_ports_path)
                 spe_ports := allowed_values_or_empty(spe, spe_ports_path)
-                effective_ranges := port_ranges(spe_ports_applied)
-                range_result := check_ports_with_protocol_in_ranges(ports, "hostPorts", effective_ranges, spe_ports)
+                range_result := host_port_result(ports, spe_ports_applied, spe_ports)
                 failed := failed_result(bool_result, range_result)
                 msg := violation_message("The hostNetwork or hostPort are not allowed", sprintf("%v: %v", [review_kind, obj.metadata.name]), failed.detail)
               }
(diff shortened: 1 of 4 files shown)


Suselz:
Moved into #23003 as 6e140468ce, so that the two fixes to `allow-host-network.yaml` land together and are tested as one. Closing this one; the branch is deleted.

PR #23011, opened 2026-09-14
[multitenancy-manager] Drop resourcesTemplate from ProjectTemplate v1alpha2

Jabejixo:
## Description

`ProjectTemplate` v1alpha2 is described by structured fields only: `spec.resourcesTemplate` is removed from the schema and the type, and with it the whole Helm-string render path.

- **Conversion.** v1alpha1 stays declared in the CRD (it is what a stored object may still carry) and unserved. A stored v1alpha1 template with a non-empty Helm string comes up from the conversion webhook without the string and with the `projects.deckhouse.io/legacy-helm-template: "true"` annotation; an empty or whitespace-only string is dropped without the mark. The `Project` conversion now converts `status.namespaces` in both directions (names in v1alpha2, `{name, kind}` objects in v1alpha3) and answers for objects without `spec` or `status`.
- **Guard.** The project manager refuses to render the projects of a marked template: they are parked in `Error` with the `ProjectTemplateUsable` condition set to `False`, and the Helm release is left exactly as it is. Rendering the empty structured shape over the release would delete every object the Helm string used to produce. Rewriting the template with structured fields and removing the annotation resumes the render.
- **The migration is self-contained.** The conversion carries the Helm string into the `projects.deckhouse.io/legacy-helm-template-body` annotation beside the mark, so the administrator asked to rewrite the template can read what it used to do: v1alpha1 is unserved, the copy in etcd is replaced by the first write of the object, and the pre-upgrade release values expire after three revisions. A body over 64 KiB is left out, because all annotations of an object together may not exceed 256 KiB and an object that cannot be written back cannot be fixed. Removing the mark on its own is refused by the template webhook while the template still renders nothing — that order persists an empty template with no mark, and the next reconcile prunes the release; rewriting and unmarking in one request, which is what `kubectl edit` and `kubectl apply` send, is accepted.
- **Built-in templates.** `simple` is structured now: namespace labels and annotations come from the `namespace` parameter, and the required-requests `OperationPolicy` is opt-in (`requiredRequests`, default `false`), so the template still renders the namespace and nothing else. The other three built-in templates were already structured.
- **Removed.** The Helm chart library (`helmlib`), the legacy render/analyze/validate entry points of the Helm client, the parameter injection check, the post-renderer branches that only a Helm string could exercise (ResourceQuota/AuthorizationRule filter, binding roleRef collection, `projects.deckhouse.io/unmanaged` and `skip-heritage-label`, the namespace-override warning), the `TemplateResourcesFiltered` and `TemplateRolesAllowed` conditions, and the v1alpha1 Go types. Helm remains the apply engine; the natively rendered manifests reach the post-renderer as before. The project webhook validates the parameters against the template schema and no longer rehearses a render. `rbac.deckhouse.io/disabled-for-direct-use-in-projects` keeps its remaining enforcement point, the ProjectRoleBinding webhook.

Documentation is not touched here; it is updated in the next PR of the stack together with the other documentation changes.

A non-empty `resourcesTemplate` silently won over the structured fields of the same template, which contradicted the type documentation and made a template with both shapes render something other than what it declared. A free-form Helm string is also applied by the module's ServiceAccount bound to cluster-admin, so authoring one was equivalent to holding cluster-admin. v1alpha2 exists only in this stack, so the field can go without a compatibility period; what does need care is the upgrade from v1alpha1, where every existing template is a Helm string — hence the conversion mark and the guard instead of a render.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: feature
summary: ProjectTemplate v1alpha2 is structured-only; the Helm resourcesTemplate string is removed from the API and a template that carried one is parked for rewriting instead of being rendered.
impact: |
  Breaking change for clusters with custom ProjectTemplates written as Helm strings.

  - `spec.resourcesTemplate` no longer exists in `ProjectTemplate` v1alpha2; a manifest that carries it is refused by the schema.
  - A template stored as v1alpha1 with a Helm string is converted without the string and marked with the `projects.deckhouse.io/legacy-helm-template: "true"` annotation. Projects of such a template switch to the `Error` state with the `ProjectTemplateUsable` condition and are not rendered; their existing objects are not touched. Rewrite the template with structured fields and remove the annotation to resume.
  - The built-in `simple` template renders the required-requests OperationPolicy only when the new `requiredRequests` parameter is `true`; existing projects on `simple` keep rendering the namespace only.
  - The `TemplateResourcesFiltered` and `TemplateRolesAllowed` project conditions and the `projects.deckhouse.io/unmanaged` and `projects.deckhouse.io/skip-heritage-label` labels are removed together with the Helm-string path.
impact_level: high
```


modules/160-multitenancy-manager/webhooks/conversion/projecttemplates
```diff
@@ -45,7 +52,13 @@ EOF
 
 function __on_conversion::v1alpha1_to_v1alpha2() {
   if converted=$(context::jq -rc '.review.request.objects//[] | map(
-     if .apiVersion == "deckhouse.io/v1alpha1" then .apiVersion = "deckhouse.io/v1alpha2" else . end
+     if .apiVersion == "deckhouse.io/v1alpha1" then
+       .apiVersion = "deckhouse.io/v1alpha2" |
+       (if (((.spec // {}).resourcesTemplate // "") | tostring | test("\\S")) then
+          .metadata.annotations["projects.deckhouse.io/legacy-helm-template"] = "true"
```
AlwxSin: 🔴 **The legacy Helm string is destroyed before the admin can read it**

`v1alpha1` is `served: false` (`crds/projecttemplate.yaml:26`), so the only way to read a stored template is through this conversion — which now strips `resourcesTemplate`. The controller then parks the projects with a message telling the administrator to "rewrite the template with structured fields", but there is no API path left to see what the template used to do.

The string survives in etcd until the first write of any kind (the apiserver persists at the storage version, v1alpha2), and it is readable only out-of-band: `helm get values <project-release> --revision <pre-upgrade>` still carries `projectTemplate.resourcesTemplate` from the old `buildValues`. With `HistoryMax: 3` that window closes after three reconciles that change the release, and nothing in USAGE (updated in #23022) mentions it.

Stashing the string next to the mark — e.g. `projects.deckhouse.io/legacy-helm-template-body` — would make the migration self-contained.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/project/manager.go
```diff
@@ -269,8 +257,27 @@ func (m *Manager) handleTemplate(ctx context.Context, project *v1alpha3.Project)
 	project.SetConditionTrue(v1alpha3.ProjectConditionProjectTemplateFound)
 	project.SetTemplateGeneration(projectTemplate.Generation)
 
+	// A template stored as v1alpha1 with a Helm resourcesTemplate comes up marked and without the
+	// string (see the conversion webhook). Rendering it as the empty structured template it now looks
+	// like would upgrade the release to a lone Namespace and delete every object the Helm string used
+	// to produce, so the project is parked in Error until an administrator rewrites the template and
+	// removes the mark. The release is left exactly as it is.
+	if projectTemplate.Annotations[v1alpha2.TemplateAnnotationLegacyHelm] == "true" {
```
AlwxSin, line 265: 🔴 **Removing the mark alone prunes every object of every project on the template**

This guard is the only thing standing between an empty structured template and a Helm upgrade that deletes everything the Helm string produced — as the PR description states. Nothing enforces that the rewrite happens before the mark goes away, and the condition message just below reads "Rewrite the template with structured fields and remove the ... annotation".

`kubectl annotate projecttemplate X projects.deckhouse.io/legacy-helm-template-` as a separate first step persists the template at v1alpha2 with no structured fields and no mark; the next reconcile renders a lone Namespace and Helm prunes the rest. `kubectl edit` / `kubectl apply` are atomic and safe — it is the two-step order the message invites that bites.

The template webhook is the natural place to refuse dropping the mark while the spec is still structurally empty. Since #23022 is the PR that documents the mark and the procedure in USAGE, the guard arguably belongs there.

`hooks/conversion_projects_test.go` covers the happy path of the mark (set on a non-empty string, not set on an empty one, controller refuses to render a marked template). Neither this ordering hazard nor the unreadability of the original string has a test.

modules/160-multitenancy-manager/images/multitenancy-manager/src/apis/deckhouse.io/v1alpha3/project.go
```diff
@@ -31,13 +31,15 @@ const (
...
-	ProjectConditionHelmOwnership             = "HelmOwnership"
+	ProjectConditionProjectTemplateFound     = "ProjectTemplateFound"
+	ProjectConditionProjectValidated         = "Validated"
+	ProjectConditionProjectResourcesUpgraded = "ResourcesUpgraded"
+	ProjectConditionStandardFieldsApplied    = "StandardFieldsApplied"
+	ProjectConditionHelmOwnership            = "HelmOwnership"
+	// ProjectConditionTemplateRequiresRewrite is False while the project's template carries the
+	// legacy-helm-template mark: the template was a Helm string in v1alpha1, the string is gone, and
+	// rendering the empty structured shape would delete the objects that string produced.
+	ProjectConditionTemplateRequiresRewrite = "TemplateRequiresRewrite"
```
AlwxSin: 🟡 **`TemplateRequiresRewrite` has inverted polarity**

`SetConditionFalse(TemplateRequiresRewrite, ...)` means a rewrite *is* required, and `True` means it is not. This PR simultaneously deletes the module's two other mixed-polarity conditions (`TemplateResourcesFiltered`, `TemplateRolesAllowed`), so this is the moment to settle on one convention: `TemplateRewritten` or `ProjectTemplateUsable` reads correctly with `False` = broken, like every other condition here.

v1alpha2/v1alpha3 are unreleased, so the rename is free now and a compatibility problem later.

modules/160-multitenancy-manager/webhooks/conversion/projecttemplates
Jabejixo: Fixed in a0d6467a83: the conversion now carries the string into the `projects.deckhouse.io/legacy-helm-template-body` annotation beside the mark, and the parked project's condition message points at it. A body over 64 KiB is left out, because all annotations of an object together may not exceed 256 KiB and an object that cannot be written back cannot be fixed either.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/project/manager.go
Jabejixo: Agreed, and fixed in a0d6467a83: the template webhook now refuses an update that removes the mark while the spec still renders nothing (`ProjectTemplateSpec.RendersObjects`), so only the single request that rewrites and unmarks together — what `kubectl edit` and `kubectl apply` send — goes through. Covered by `TestHandle_LegacyMarkRemoval` and by the e2e section that used to do exactly the two-step removal.

modules/160-multitenancy-manager/images/multitenancy-manager/src/apis/deckhouse.io/v1alpha3/project.go
Jabejixo: Renamed to `ProjectTemplateUsable` in a0d6467a83, so `False` means broken like every sibling condition on the Project status. Docs, the e2e harness and the following PR follow the new name.

modules/160-multitenancy-manager/images/multitenancy-manager/src/apis/deckhouse.io/v1alpha2/template.go
```diff
@@ -160,15 +174,23 @@ type ProjectTemplateSpec struct {
...
+//
+// It answers one question: would applying this template to a project that already has a release
+// delete everything in it. The legacy-Helm mark exists for exactly that case, and dropping the mark
+// while this is false is what the template webhook refuses.
+func (p *ProjectTemplateSpec) RendersObjects() bool {
+	if !p.PodSecurityStandard.IsZero() || !p.NodeSelector.IsZero() || !p.Tolerations.IsZero() ||
+		!p.AllowedUIDs.IsZero() || !p.AllowedGIDs.IsZero() {
+		return true
+	}
+	return p.NetworkPolicy != nil || p.NamespaceMetadata != nil || p.Features != nil ||
```
AlwxSin: 🟡 **`RendersObjects()` is satisfied by an empty stanza**

The guard this feeds — `legacyMarkRemoval` — is the only thing left between a half-rewritten template and a Helm upgrade that prunes every object of every project on it. It asks a structural question (is the pointer non-nil) where the one that matters is semantic.

`networkPolicy: {}` makes this return `true` while `networkPolicies()` returns `nil` on an unresolved mode (`internal/render/render.go:237`). The same holds for `namespaceMetadata: {}`, `features: {}`, `logShipping: {}` and `runtimeAudit: {}`: present, structurally, and rendering nothing.

So an administrator who rewrites the template half-way and unmarks it in the same request — exactly the request the guard is built to accept — still gets the bare-namespace render and the prune. Checking the content rather than the pointer (`NetworkPolicy != nil && !NetworkPolicy.Mode.IsZero()`, and so on for the other four) keeps it cheap and closes the hole.

Jabejixo: You are right, and it defeats the guard in exactly the request it is meant to accept. Fixed in 6accf45a3b: every optional stanza is judged by its content now (`NetworkPolicy != nil && !Mode.IsZero()`, and the same for namespaceMetadata, features, logShipping and runtimeAudit), with a table test covering all five empty shells plus a literal and a fromParam reference.

AlwxSin (approved):

z9r5 (approved):

PR #23022, opened 2026-09-14
[multitenancy-manager, user-authz] Fix grant resolution, namespace ownership and admission gaps

Jabejixo:
## Description

Stacked on `feat/mtm-drop-resources-template` (itself on `feat/mtm-namespaces-as-full-projects` → `integration/rbacv2`), so the diff to read is this branch against its base. It fixes the places where the two modules behave differently from what their documentation promises, plus two controller-level problems, and corrects the documentation against the code. Commits are grouped by area and the paragraphs below follow them.

**Leader election on every replica count.** The controller elects a leader always, not only when the module runs in HA mode. A Deployment of one replica still runs two pods during a rolling update; without a lease both reconciled at once, and a rollout of the previous PR gave every project on the template that had just changed a second, byte-identical Helm revision from the pod that was already being terminated. The lease budget is the one user-authz-controller runs with (60 s lease, 40 s renew, 8 s retry) so a short API-server absence does not restart the pod. The webhooks keep serving on every replica; only the controllers wait for the lease. The `HA_MODE` environment variable had no other reader and is removed from the Deployment.

**Grant violations are counted by the controller.** The `ClusterResourceGrantPolicyViolation` alert read a series a module hook produced from its own copy of the availability rules, evaluated one policy at a time. The webhook treats the applicable policies of a project as a union, so an object using a name the second policy allows was admitted and then reported as a violation for the life of the object. The controller now recounts the violations of a project namespace on every reconcile (and at least every two minutes), through the same references, field paths, guards and `Resolve`/`Decide` as `/is-granted`; the series can only name what the webhook would refuse. Metric name and labels are unchanged, so the alert rule and the dashboard are untouched; series are dropped per namespace before each recount and when the namespace stops being a project namespace. The controller's metrics server, previously disabled, listens on 9091 with a PodMonitor. The hook is removed.

**Grant resolver and CRD validation.** A `ClusterResourceGrantPolicy` is matched against the union of the labels of the `Project` object and the labels of the namespace (the namespace wins a shared key), in one place — `resolve.GrantsForNamespace` — that `/is-granted`, `/defaults`, the catalog reconciler and the violation recount all go through. A label on the Project therefore selects its main and additional namespaces, as the usage documentation had described all along, and the catalog reconciler now watches Project label changes to re-evaluate the project's namespaces. A new `SelectorsValid` condition on the policy reports a selector the schema accepts but the selector library refuses (`In` without values), which the resolver used to turn into a silent "matches nothing". The default taken from an annotation (`defaultFrom`) requires the value `true`, so a class marked `is-default-class: "false"` is no longer a default. The CRDs refuse `allowed: []`, an empty `allowedSelector`/`deniedSelector` and `apiGroups: ["*"]` in a reference — the last one was accepted by the schema and silently intercepted nothing.

**Catalog protection and namespace ownership.** The per-project catalog (`AvailableClusterResource`) carries `heritage: multitenancy-manager` — the module's own value, which the protective admission policy binds on — instead of `heritage: deckhouse`, so it is refused to everyone but the controller even while the controller is down; the `/protect` webhook stays as the second guard. A catalog with nothing available is kept as an object with `available: []` and `availableCount: 0` rather than deleted, so a reader can tell "nothing is available here" from "not reconciled yet". A namespace that is being deleted is left alone instead of producing a retry per registration. A second policy, `d8-multitenancy-manager-namespace-ownership`, lets only the controller and Deckhouse set or change the `projects.deckhouse.io/project` label on a Namespace: a hand-labelled namespace looked owned but was neither adopted nor protected. 

**Labels and annotations on adopted namespaces.** Adoption turns every user namespace into a project namespace, and the heritage policy refused every direct edit of it, cluster administrators included — `kubectl label ns … istio-injection=enabled`, a Pod Security label, an Argo CD tracking annotation, every GitOps re-apply of a Namespace manifest. The policy now allows an UPDATE of a project namespace that changes only labels and annotations outside the keys the module owns (`heritage`, `projects.deckhouse.io/*`, the template-driven policy labels, the Helm ownership annotations, the node-selector and tolerations annotations); `spec`, finalizers, the owned keys and deletion stay with the controller. The owned-key list is a CEL literal in the policy (iterating the label map does not fit the CEL cost budget) and a Go slice in `internal/naming`; a test keeps the copies identical, and adoption uses the same list to decide what not to mirror into the Project. One edit the heritage policy cannot see on its own is dropping the `heritage` label itself: its binding selects a Namespace by the labels of the *new* object, so the request falls outside the binding — that was true before this PR as well. The selector-less namespace-ownership policy therefore also holds the owned keys of a namespace the module owned before the request; a run on a dev stand is what surfaced it.

**Template literals.** `podSecurityStandard` and `networkPolicy.mode` take one of a few words or a `{fromParam}` reference; a literal outside the set was accepted, rendered into a pod-security label nothing reads or into no NetworkPolicy, and the project reported Deployed. The template webhook refuses such a literal and names the accepted values. The CRD carries no CEL rule for this: both fields are untyped in the schema (string or object via `x-kubernetes-preserve-unknown-fields`) and the API server builds no CEL declarations for an untyped field, neither on the field nor through its parent.

**user-authz: dictionaries and the exec webhook.** The dictionary ClusterRole (`d8:dict`) is granted to subjects that hold a project role through a ProjectRoleBinding or ClusterProjectRoleBinding, as the README had promised: the dict reconciler now counts the fan-out RoleBindings with a `d8:project:*` roleRef. The exec/attach/port-forward binding of the system-resource webhook excludes at the API server the same four identities its body lets through (two service accounts were missing), and a test pins that the denial names exactly the roles the check accepts. 

**Fan-out pacing and compat aliases.** Each RoleBinding write of a ClusterProjectRoleBinding fan-out takes a token from a bucket (20 QPS, burst 40): two or three projects still fan out within seconds, thousands of namespaces spread their writes over minutes instead of a burst; a `d8_multitenancy_cprb_fanout_duration_seconds` histogram records the fan-out time. The `d8:use:role:*` compatibility aliases carry `rbac.deckhouse.io/disabled-for-direct-use-in-projects`, so a new ProjectRoleBinding or ClusterProjectRoleBinding cannot point at an old role name while existing plain RoleBindings keep working.

**A policy that grants nothing says so; Coerce says what it replaced.** The `clusterroles` registration excludes every ClusterRole without `rbac.deckhouse.io/delegatable: "true"` from all projects, and an exclusion wins over an allow-list, so a `ClusterResourceGrantPolicy` that allows such a role grants nothing — and the usage documentation showed exactly that policy as the way to share a custom role. The policy reconciler now reports an `AllowedEffective` condition: `False` with the entry, the name and the reason whenever an allowed name would be refused by the resolver anyway, `True` once every allowed name can be granted. The `/defaults` webhook in `Coerce` mode used to rewrite a value the author had set without a word; it now returns an admission warning with the object, the field, the original value, the project and the substituted default, which `kubectl` prints. `FillEmpty` stays silent. USAGE and README (RU and EN) describe the condition, the label requirement on allowed roles, and the warning.

**USAGE, README and the user-authz FAQ say what the code does.** Each corrected paragraph was traced to the code that contradicts it: `projectSelector` matches the union of Project and namespace labels; a project administrator holds `bind` on exactly eight roles, listed, and a binding to any other role is refused rather than compared with their permissions; `fromParam` applies to twelve named fields; Project `v1alpha1` is not served; a ProjectTemplate stored as `v1alpha1` with a Helm `resourcesTemplate` is not rendered and parks its projects in `Error` with `TemplateRequiresRewrite`; the heritage policy covers CREATE and does not exempt `system:masters`; a namespace owned by another Helm release is reported through `HelmOwnership`, not taken over. In user-authz: `d8:dict` is also bound for holders of `d8:project:*` roles; the FAQ no longer calls a ClusterAuthorizationRule namespace filter a "hard boundary" (sources are summed, each at its own level, as the module description and the e2e matrix state; the RU answer sat under a stray duplicate heading and is moved); the role mapping table says `Editor` keeps the level but not the scope and `ClusterEditor` has no single counterpart; the identity collision check covers AuthorizationRule, runs on create and on a name change, and warns on delete.

**Review follow-up.** The violation series name the project, the object and the field that was refused, and were served in plaintext to anything that could reach the pod; the controller now keeps its metrics endpoint on loopback and a `kube-rbac-proxy` sidecar publishes it, authorizing every scrape against the `deployments/prometheus-metrics` subresource. The fan-out limiter took a token per namespace walked rather than per namespace written, so an idle resync of a large cluster held the single worker for minutes: the shared upsert helper now reports whether it wrote, only a write spends a token, the histogram is recorded on every return, and the limiter is built when the controller is wired instead of lazily from the reconcile goroutine. A ProjectTemplate could carry inline grant selectors the `ClusterResourceGrantPolicy` schema refuses, and the managed policy built from them was then rejected at admission forever with nothing on the template to say why; the template webhook refuses them up front. Adoption mirrors the scheduler annotations again: no template that adoption can pick declares `nodeSelector` or `tolerations`, so without the mirror an adopted namespace kept its placement in no one's desired state and nobody could change it.

**Design notes.** `PROJECTS_DESIGN` now opens with a status block naming what never shipped (self-service mode switch, control namespace, `namespace-role` label, `ClusterResourceGrant` quota pool, Project-label propagation) and what replaced propagation; `CLUSTER_OBJECT_GRANTS_DESIGN` describes the shipped resolver, the policy reconciler with its two conditions, the Coerce warning, the `defaultFrom` value rule and the violation scan with its metric. RU copies alongside.

**A run on a dev stand surfaced two gaps, fixed in a later commit.** The binding webhooks checked the role-name prefix before looking the role up, so a ProjectRoleBinding to a `d8:use:role:*` alias got the generic prefix refusal and never the deprecation hint; the lookup now comes first and the refusal quotes `rbac.deckhouse.io/deprecated-replaced-by`. The policy reconciler evaluated `AllowedEffective` only on policy events, so labelling a ClusterRole delegatable left the condition `False`; it now watches ClusterRole label changes and definition changes and requeues an ineffective policy on the catalog cadence.

**Stand results on the final image (2026-09-15).** The RBAC end-to-end suite: 877 pass, 1 fail, 3 skip — the one failure was a test fixture (an arbitrary ClusterRole cannot be labelled delegatable; the user-authz webhook refuses it by design), and the section covering this PR's behaviour, re-run with the corrected fixture, gives 147 pass, 0 fail, 1 skip. The access matrix suite: 193 pass, 0 fail. No regression against the baseline taken before this PR. The remaining skip is the `defaultFrom` value check, blocked on the stand by the platform's default-storage-class policy and covered by resolver unit tests.

Most of these are cases where the module's behaviour and its documentation had drifted apart, and the cost fell on working configurations rather than on clarity alone. A grant policy written against a `Project` label, which is how the usage guide shows it, selected nothing. After a namespace became a project, every direct edit of its labels and annotations was refused, cluster administrators included, so Istio injection, a Pod Security label, an Argo CD annotation and any GitOps re-apply of a Namespace manifest stopped working. An object that one of several policies allowed was admitted and then reported as a violation for the rest of its life. A typo in a template literal produced no NetworkPolicy, a pod-security label nothing reads, and a project that reported `Deployed`. A policy that allows a ClusterRole without the delegatable label grants nothing and said so nowhere. Each of these is fixed above, and the documentation is corrected against the code rather than the other way round.

Two changes concern the controller rather than its contract. A lease removes the overlap between two reconcilers of the same Helm release during a rolling update: two concurrent writers produce redundant revisions on every rollout and, in the worst case, race each other on the release secret. The cost is a few seconds of controller idle time while the lease moves, and admission and conversion keep being served by whichever pod is up. The ClusterProjectRoleBinding fan-out is paced so that a cluster with thousands of namespaces spreads its RoleBinding writes over minutes instead of bursting them at the API server.

Not necessarily.

## Checklist
- [x] e2e tests passed (dev stand, 2026-09-15: RBAC suite 877/1/3 with the one failure traced to a test fixture, access matrix 193/0, new-behaviour section 147/0/1 after the fixture fix).

## Changelog entries

```changes
section: multitenancy-manager
type: fix
summary: The controller always elects a leader, so a rolling update no longer runs two reconcilers over the same project releases.
impact_level: low
```

```changes
section: multitenancy-manager
type: fix
summary: The ClusterResourceGrantPolicyViolation metric is computed by the controller with the same union-of-policies rule as the admission webhook, so an object allowed by one of several policies no longer raises a phantom alert.
impact: The `d8_cluster_objects_grant_violated` series is now scraped from the multitenancy-manager controller (port 9091, PodMonitor `multitenancy-manager`) instead of the module hook; the metric name, labels, alert and dashboard are unchanged.
impact_level: default
```

```changes
section: multitenancy-manager
type: fix
summary: ClusterResourceGrantPolicy projectSelector matches the labels of the Project object as well as the namespace labels; a policy written against a Project label now covers the main and the additional namespaces of the project.
impact: |
 The selector is evaluated against the union of Project and namespace labels, the namespace winning a shared key. Existing policies keyed on namespace labels (`projects.deckhouse.io/project`, `projects.deckhouse.io/project-namespace`, `projects.deckhouse.io/project-template`) behave as before.
 The CRDs now refuse `allowed: []`, an empty `allowedSelector`/`deniedSelector` and `apiGroups: ["*"]` in a GrantableClusterResourceReference; such objects were accepted before and either opened the resource or intercepted nothing. Existing objects are validated on their next write.
 `defaultFrom` counts only an annotation with the value `true`; a StorageClass carrying `storageclass.kubernetes.io/is-default-class: "false"` is no longer treated as the default.
 Policies carry a new `SelectorsValid` condition.
impact_level: high
```

```changes
section: multitenancy-manager
type: fix
summary: Labels and annotations that the module does not own can be changed directly on a project namespace again; the project ownership label can be set only by the controller; the per-project catalog is protected fail-closed.
impact: |
 On an adopted (or any project) namespace, `kubectl label`/`annotate` and GitOps re-applies are accepted for keys the module does not own; `heritage`, `projects.deckhouse.io/*`, the template-driven policy labels, the Helm ownership annotations, the node-selector and tolerations annotations, `spec`, finalizers and deletion are still refused and are changed through the Project and its template.
 Creating a namespace with a `projects.deckhouse.io/project` label, or changing that label, is refused for everyone but the controller — cluster administrators included; create a Project or a ProjectNamespace instead.
 `AvailableClusterResource` objects carry `heritage: multitenancy-manager` (relabelled in place) and are refused to everyone but the controller by the admission policy; a project with nothing available keeps an empty catalog object instead of none.
impact_level: high
```

```changes
section: multitenancy-manager
type: fix
summary: A ProjectTemplate with a podSecurityStandard or networkPolicy.mode literal outside the accepted values is refused by the admission webhook instead of rendering into nothing.
impact_level: default
```

```changes
section: user-authz
type: fix
summary: Subjects that hold a project role through ProjectRoleBinding or ClusterProjectRoleBinding receive the read-only dictionaries (d8:dict) like namespace-role holders; the exec webhook binding excludes the aggregation controller and the multitenancy-manager controller at the API server.
impact_level: default
```

```changes
section: multitenancy-manager
type: fix
summary: ClusterProjectRoleBinding fan-out is rate-limited (20 writes/s, burst 40) and measured by the d8_multitenancy_cprb_fanout_duration_seconds histogram.
impact_level: low
```

```changes
section: user-authz
type: fix
summary: The deprecated d8:use:role:* aliases can no longer be the target of a new ProjectRoleBinding or ClusterProjectRoleBinding; existing RoleBindings to them keep working.
impact: A ProjectRoleBinding or ClusterProjectRoleBinding that names a `d8:use:role:*` alias is refused with a pointer to the `d8:namespace:*` role; migrate such bindings to the new names.
impact_level: default
```

```changes
section: multitenancy-manager
type: fix
summary: A ClusterResourceGrantPolicy reports an AllowedEffective condition when an allowed name is excluded by the resource registration (a ClusterRole without the delegatable label); the Coerce defaulting mode returns an admission warning naming the replaced value.
impact: Administrators see in the policy status why a binding to an allowed but non-delegatable ClusterRole is still refused; users see in kubectl output when their StorageClass was replaced with the project default.
impact_level: low
```

```changes
section: multitenancy-manager
type: fix
summary: The controller's metrics endpoint is served through kube-rbac-proxy instead of plaintext.
impact: The d8_cluster_objects_grant_violated series name projects and the objects inside them; the endpoint now requires authentication and the get verb on deployments/prometheus-metrics of the multitenancy-manager deployment. Prometheus is allowed by the module itself; anything else scraping the port directly must be granted that permission.
impact_level: default
```

```changes
section: user-authz
type: fix
summary: The FAQ and README describe how ClusterAuthorizationRule, AuthorizationRule and RoleBinding permissions combine in multitenancy mode, which roles get the d8:dict binding, and what the identity collision check covers; the legacy-to-granular role mapping table is corrected.
impact_level: low
```


modules/160-multitenancy-manager/templates/validation.yaml
```diff
@@ -51,14 +51,54 @@ spec:
...
+      expression: 'variables.isNamespaceUpdate && variables.namespaceBodyUntouched && variables.managedLabelsUntouched && variables.managedAnnotationsUntouched'
   validations:
-    - expression: 'request.userInfo.username == "system:serviceaccount:d8-multitenancy-manager:multitenancy-manager"'
+    # Two doors: the controller, and -- on a project namespace only -- a metadata edit that leaves the
+    # keys the module owns alone. Adoption turns every user namespace into a project namespace, so
+    # without the second door `kubectl label ns` and every GitOps re-apply of a Namespace manifest
+    # would be refused to cluster administrators for the life of the cluster.
+    # An update that drops the heritage label itself is not seen here at all (the binding selects a
+    # Namespace by the labels of the NEW object); the namespace-ownership policy below refuses it.
+    - expression: 'request.userInfo.username == "system:serviceaccount:d8-multitenancy-manager:multitenancy-manager" || variables.namespaceMetadataEdit'
```
AlwxSin, line 94: 🟡 **The grant selector now reads labels that users can set**

Two changes in this PR meet here. This policy gains a second door that accepts any UPDATE touching only labels and annotations outside `managedLabels`/`managedAnnotations`, and `resolve.GrantsForNamespace` (`internal/resolve/resolve.go:75`) matches `projectSelector` against the union of Project and namespace labels.

Before this PR every direct namespace edit was refused, so the selector input was entirely module-controlled. Now anyone with `update` on the Namespace object can add a label that makes another project's `ClusterResourceGrantPolicy` match their namespace. Namespace writes live in `d8:manage:permission:subsystem:kubernetes:manage_resources` — a cluster-level permission, and the `d8:project:*` / `d8:namespace:*` roles only get read on namespaces — so this is not escalation from a tenant, but it does silently drop a defense-in-depth property that used to hold.

Restricting the selector to Project labels plus the module-managed namespace labels keeps the documented behaviour and the old guarantee.

modules/160-multitenancy-manager/templates/multitenancy-manager/podmonitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  jobLabel: app
+  selector:
+    matchLabels:
+      app: multitenancy-manager
+  namespaceSelector:
+    matchNames:
+    - d8-{{ $.Chart.Name }}
+  podMetricsEndpoints:
+  - port: metrics
```
AlwxSin: 🟡 **Controller metrics are served in plaintext, against the module convention**

`:9091` is opened on the pod (`templates/multitenancy-manager/controller.yaml:74`) with no kube-rbac-proxy sidecar, and this PodMonitor scrapes it over plain HTTP. Nearly every sibling module does the opposite — `port: https-metrics`, `scheme: https`, `bearerTokenSecret: prometheus-token` (see `modules/015-admission-policy-engine/templates/podmonitor.yaml:12-18`).

The series here carry `project`, `violating_object_name` and `violating_field` — a per-tenant inventory of what each project uses and what it was refused — readable by any pod that can reach the controller's IP. In a multitenancy module that is the wrong default.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/controllers/clusterprojectrolebinding/reconciler.go
```diff
@@ -126,9 +161,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 
 	// Fan out into every namespace, accumulating per-namespace errors so a single bad namespace
-	// does not block the rest of the cluster (CPRB can span thousands of namespaces).
+	// does not block the rest of the cluster (CPRB can span thousands of namespaces). Each write
+	// takes a token first, so the burst a large cluster would otherwise see is spread out.
+	started := time.Now()
+	limiter := r.limiter()
 	var errs []error
 	for ns, project := range target {
+		if err := limiter.Wait(ctx); err != nil {
```
AlwxSin: 🟡 **The fan-out limiter meters iterations, not writes**

The token is taken before `upsertRoleBinding` on every namespace in `target`, including the overwhelming majority of reconciles where `CreateOrUpdate` finds the RoleBinding already correct and writes nothing. The comment just above says "Each write takes a token first", which is not what this does.

With `MaxConcurrentReconciles: 1` and a 5000-namespace cluster, an idle resync of one CPRB occupies the worker for 250 s, and every other CPRB queues behind it on the same shared bucket. Taking the token inside the mutate path — or only when `CreateOrUpdate` reports a change — keeps the burst protection without the idle cost.

Separately, `fanOutDuration.Observe` is skipped on this error return, so a cancelled fan-out never reaches the histogram.

modules/160-multitenancy-manager/crds/multitenancy.deckhouse.io_clusterresourcegrantpolicies.yaml
```diff
@@ -165,9 +169,13 @@ spec:
                             Set of labels for cluster-wide resources.
 
                             All specified labels must match.
+                          minProperties: 1
                           type: object
                       type: object
                       x-kubernetes-map-type: atomic
+                      x-kubernetes-validations:
+                        - rule: 'has(self.matchLabels) || has(self.matchExpressions)'
+                          message: 'a selector must carry matchLabels or matchExpressions; an empty selector is not accepted'
```
AlwxSin, line 178: 🟡 **ProjectTemplate still accepts selector shapes this CRD now refuses**

The new `minItems`/`minProperties`/CEL rules landed on `ClusterResourceGrantPolicy` only. A ProjectTemplate whose inline `resources[].allowedSelector` (or `deniedSelector`) is `{}` is still accepted by its own schema (`crds/projecttemplate.yaml:157`) and by the template webhook; `templategrants` then materializes a managed policy carrying that empty selector, the create is rejected here at admission, and the reconciler retries forever with nothing in the template status to point at.

`allowed: []` is safe — `Allowed []string` has `omitempty`, so an empty slice never reaches the wire — but the pointer selectors do. Either mirror the rules onto the ProjectTemplate schema, or check them in `validateGrantPolicies`.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/namespace/adopt.go
```diff
@@ -140,16 +142,11 @@ func namespaceMeta(namespace *corev1.Namespace) map[string]any {
...
-	"kubernetes.io/metadata.name",
-	labelPodPolicy,
-	labelExtendedMonitoring,
-	labelSecurityScanning,
-}
+// managedMetaExact are platform-owned keys that must match in full ("heritage" among them: a
+// HasPrefix match would also strip user keys such as heritageSomething). The list is the one the
+// protective admission policy enforces on the namespace, so what adoption refuses to mirror and
+// what the policy refuses to let a user change is the same set.
+var managedMetaExact = append(append([]string{}, naming.ManagedNamespaceLabels...), naming.ManagedNamespaceAnnotations...)
```
AlwxSin: 🟡 **Node-selector and tolerations on an adopted namespace become frozen**

`managedMetaExact` now inherits `naming.ManagedNamespaceAnnotations`, so `scheduler.alpha.kubernetes.io/node-selector` and `.../defaultTolerations` stop being mirrored into `params.namespace.annotations` on adoption. Nothing takes over: those annotations are rendered only from `spec.nodeSelector`/`spec.tolerations` of the template (`internal/render/render.go:171-190`), and `TemplateFor` only ever picks `simple`, `default` or `secure` — none of which declare them. `secure-with-dedicated-nodes` is never chosen automatically.

The live annotation survives the Helm three-way merge, but it is now in no one's desired state, and the namespace-ownership policy added in this PR refuses every user edit of it. After adoption the namespace's default placement can no longer be changed or removed by anyone.

This one is only visible across the stack: #23011 made mirroring the annotations the only path that preserves them (it wired `simple` through `namespaceMetadata.annotations: {fromParam: namespace.annotations}`), and this PR narrows the mirror. Either keep mirroring these two keys, or seed the corresponding template parameters at adoption time.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/naming/managed_keys_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// TestManagedNamespaceKeysMatchThePolicy: the admission policies enforce the module-owned keys as
+// CEL literals (one copy per policy) and the controller filters them in Go; every copy must be the
+// same list as the Go one.
+func TestManagedNamespaceKeysMatchThePolicy(t *testing.T) {
+	path := filepath.Join("..", "..", "..", "..", "..", "templates", "validation.yaml")
+	raw, err := os.ReadFile(path)
+	if err != nil {
+		t.Skipf("policy template not reachable from here (%v); run from the module tree", err)
```
AlwxSin: 🔵 **The copy-sync guard test skips itself instead of failing**

`t.Skipf` on a failed `os.ReadFile` means the whole point of this test — keeping the three copies of the owned-key list in step — evaporates the moment the relative path breaks or the test runs somewhere `templates/` is not present. For a test whose only job is to fail, `t.Fatalf` is the right call.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/webhook/template/hook.go
```diff
@@ -177,3 +186,36 @@ func (v *validator) validateManagedGrantNames(ctx context.Context, template *v1a
 	}
 	return admission.Allowed("")
 }
+
+// literalValues are the values the fixed-set fields accept as literals. The CRD carries the same
```
AlwxSin: 🔵 **`literalValues` comment contradicts the CRD**

"The CRD carries the same sets as CEL rules; this check is the answer for a cluster whose CRD predates them" — the CRD carries no such rules, and the PR description explains why it cannot (both fields are untyped in the schema, so the apiserver builds no CEL declarations for them, neither on the field nor through its parent). The comment sends the next reader looking for validation that does not exist.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/controllers/clusterprojectrolebinding/reconciler.go
```diff
@@ -44,6 +48,37 @@ import (
...
+	Help:    "Wall time of one ClusterProjectRoleBinding fan-out (RoleBinding upserts and prunes across the project namespaces).",
+	Buckets: []float64{0.1, 0.5, 1, 2, 5, 10, 30, 60, 120, 300},
+})
+
+func init() {
+	metrics.Registry.MustRegister(fanOutDuration)
+}
+
+// limiter returns the configured limiter, building the default one on first use.
+func (r *Reconciler) limiter() flowcontrol.RateLimiter {
```
AlwxSin: 🔵 **Lazy limiter init mutates shared state without a lock**

`limiter()` writes `r.Limiter` from the reconcile goroutine. That is safe today only because `MaxConcurrentReconciles: 1` is set ~190 lines away in `SetupWithManager`; raising it turns this into a data race that `-race` will only catch under load. A `sync.Once`, or building the limiter in `SetupWithManager`, removes the implicit coupling.

modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -34,7 +34,7 @@ The following project templates are included in the Deckhouse Kubernetes Platfor
   - `dedicatedNodes.nodeSelector` — the node selector of the project. The node selector of a created pod is **substituted** with this value.
   - `dedicatedNodes.defaultTolerations` — tolerations in the format of the pod's `spec.tolerations`. They are **added** to the created pods of the project.
 
-The `default`, `secure`, and `secure-with-dedicated-nodes` templates are described in the [structured form](#structured-templates) (`deckhouse.io/v1alpha2`); the `simple` template is a minimal legacy (`v1alpha1`) template.
+The `default`, `secure`, and `secure-with-dedicated-nodes` templates are described in the [structured form](#structured-templates) (`deckhouse.io/v1alpha2`); the `simple` template is a minimal structured template that creates only the namespace and sets its labels and annotations from the project parameters.
```
AlwxSin, line 38: 🔵 **`simple` gained a parameter USAGE does not list**

(Anchored here because it is the nearest line this PR changes; the text that needs the edit is the `simple` bullet at `docs/USAGE.md:12-14`.)

`simple` now declares three parameters — #23011 added `requiredRequests` to its `parametersSchema` with `default: false` — but the `simple` bullet lists only `namespace.labels` and `namespace.annotations`. `requiredRequests` is documented under `default` with "(default `true`)", which is correct there and wrong for `simple`. One line in the `simple` bullet closes it.

modules/160-multitenancy-manager/templates/multitenancy-manager/podmonitor.yaml
Jabejixo: Fixed: the controller keeps its metrics endpoint on loopback and a `kube-rbac-proxy` sidecar publishes it, authorizing every scrape against `deployments/prometheus-metrics`, with the Role that lets the scraper through and an https PodMonitor. Verified on a stand — 401 without a token, the full series set with a token of the `d8-monitoring/prometheus` ServiceAccount.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/controllers/clusterprojectrolebinding/reconciler.go
Jabejixo: Fixed: `UpsertServiceRoleBinding` now reports whether it actually wrote, and only a write spends a token, so a resync that changes nothing costs no waiting (pinned by a test). The histogram moved into a `defer`, so the limiter error path records the fan-out time too.

modules/160-multitenancy-manager/crds/multitenancy.deckhouse.io_clusterresourcegrantpolicies.yaml
Jabejixo: Fixed in the webhook rather than the schema: `validateInlineGrantSelectors` refuses an empty selector and one the selector library cannot compile, naming the entry index and the resource. That way the author hears about it instead of the reconciler retrying a rejected policy forever.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/namespace/adopt.go
Jabejixo: Good catch, and it is exactly the cross-PR case we would have missed: adoption mirrors the two scheduler annotations again, so they land in the namespace parameter and stay changeable through the Project, while the policy keeps owning them on the namespace. Pinned by `TestFilterUserMeta_KeepsPlacementAnnotations`.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/naming/managed_keys_test.go
Jabejixo: Fixed: `t.Fatalf`.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/webhook/template/hook.go
Jabejixo: Fixed: the comment now states why the CRD cannot carry those sets — both fields are untyped in the schema, so the API server builds no CEL declarations for them, neither on the field nor through its parent.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/controllers/clusterprojectrolebinding/reconciler.go
Jabejixo: Fixed: the limiter is built in `SetupWithManager`, so nothing writes `r.Limiter` from a reconcile goroutine and `MaxConcurrentReconciles` becomes an independent choice again.

modules/160-multitenancy-manager/docs/USAGE.md
Jabejixo: Fixed: the `simple` bullet now lists `requiredRequests` (default `false`), and the `default` bullet says the default is `true` in that template.

modules/160-multitenancy-manager/templates/validation.yaml
Jabejixo: You are right that the property changed, and we traced the escalation path before deciding: `d8:project:*` and `d8:namespace:*` only read namespaces, a `ProjectNamespace` carries no labels, and writing a `Project` is a manage-level capability — so no tenant can reach the selector input. It needs the same cluster-level namespace write that can already edit the policies themselves.

We would rather not narrow it to the module-managed labels, though. Before #22609 a plain namespace could carry any user label, and a policy keyed on one of those is a legitimate thing to have today; those namespaces become project namespaces on upgrade, so narrowing would silently stop matching them and quietly take granted resources away from running projects. A silent loss of access on upgrade looks worse to us than the defense-in-depth we give up here.

What we did instead is write the property down: USAGE now says the namespace labels take part in the match, and names the permission that decides it. If you think the exposure outweighs the upgrade risk we are happy to revisit — an allow-list of namespace label keys that participate would be the middle ground.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/render/render.go
```diff
@@ -186,7 +187,7 @@ func (r *renderer) namespace(spec *v1alpha2.ProjectTemplateSpec) (map[string]any
 		if sErr != nil {
 			return nil, fmt.Errorf("render nodeSelector: %w", sErr)
 		}
-		annotations["scheduler.alpha.kubernetes.io/node-selector"] = annotation
+		annotations[naming.NodeSelectorAnnotation] = annotation
```
AlwxSin, line 210: 🟡 **The adoption mirror now silently outranks `spec.nodeSelector`**

This assignment is overwritten four lines below: `namespaceMetadata.annotations` is merged after `spec.Tolerations` and `spec.NodeSelector`, so the free-form parameter wins over the dedicated field.

That precedence was harmless while adoption filtered the two scheduler keys out of `params.namespace.annotations`. The fix for the adoption finding (`internal/manager/namespace/adopt.go:156-165`) now puts them there automatically, so the collision happens on its own rather than needing a hand-written parameter.

Adopt a namespace carrying `scheduler.alpha.kubernetes.io/node-selector: x=y`, then move the project to `secure-with-dedicated-nodes` and set `dedicatedNodes.nodeSelector`: the stale mirrored annotation overwrites it, the project reports `Deployed`, and the pods land on the wrong nodes. `TemplateFor` never picks that template itself, but `spec.parameters` survives a template switch and the cumulative schema accepts the `namespace` key.

Merging `namespaceMetadata` first and letting the dedicated fields overwrite makes the more specific declaration win; alternatively, adoption could drop a mirrored key that the target template declares.

Jabejixo: Confirmed and fixed in 64e354dab1: `namespaceMetadata` is merged before the dedicated fields, so `spec.nodeSelector`/`spec.tolerations` win a shared key while the rest of the mirrored annotations survive beside them. A render test pins the template-switch case you describe.

AlwxSin (approved):

Lada7878 (approved):

deckhouse-BOaTswain:
:green_circle: `Deploy web to test7` for `deckhouse:feat/mtm-audit-remediation` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/35570302951) in 6m16s.

Environment URL: <a href="https://deckhouse-7.test.flant.dev/products/kubernetes-platform/documentation/v1/">deckhouse-7.test.flant.dev</a>

z9r5 (approved):

AlwxSin (approved):

PR #23053, opened 2026-09-15
[node-manager] cheap discover_standby_ng: patch status on change, stable filter

borg-z:
## Description

`discover_standby_ng` is the most expensive node-manager hook on main: 2.9 s per run, 64% of all hook CPU during a NodeGroup burst (30 groups without standby), measured 09–10.09 on DVP.

Two causes, both fixed here without changing behaviour:

1. The handler patched `status.standby` of **every** NodeGroup on every run, writing `null` for groups without standby. Each patch is a sequential merge request on `/status`, and the patch phase is counted inside the hook duration. 30 groups × ~97 ms = 2.9 s per run, ~240 no-op requests per burst. Now the current `status.standby` is part of the filter result and the status is patched only when the value differs.
2. The NodeGroup filter returned `maxPerZone`, `zones`, `taints`, `overprovisioningRate` for every group, so any change of those fields on a group without standby changed the filter checksum and woke the hook. Now groups without standby produce a minimal stable result (`name`, `needStandby: false`, current status), and shell-operator suppresses the event.

Side effect: the hook's own status write changes the filter checksum once, causing one extra run that finds nothing to patch. Steady state: zero patches.

3. Second commit: `discover_standby_ng` and `set_ng_priorities` move from their own queues into `/modules/node-manager`, the queue of `get_crds` and every other values writer of the module. addon-operator compares the module values checksum before and after each hook run, so a values commit from a concurrent queue is attributed to the hook running in the module queue at that moment, and that hook queues a second ModuleRun for the same event. Hooks in one queue run one after another and never see a foreign commit inside their own run.

## Tests

- `Test_standbyNodeGroupFilter_StableWithoutStandby`: filter result of a group without standby must not depend on `maxPerZone`. Fails on the old code.
- Ginkgo context "Cluster where NodeGroup statuses already hold the actual standby value": hook must not emit status patches. Fails on the old code (two patches expected empty).

- `TestValuesWritersShareModuleQueue`: pins the queue of every values writer (`discover_standby_ng`, `get_crds`, `set_ng_priorities`). Fails on the old code:

```
values_writers_queue_test.go:44: discover_standby_ng.go runs in queue "/modules/node-manager/discover_standby_ng", values writers share /modules/node-manager
values_writers_queue_test.go:44: set_ng_priorities.go runs in queue "/modules/node-manager/set_priorities", values writers share /modules/node-manager
```

`go test ./modules/040-node-manager/hooks/ -count=1` green.

## Live check

Dev cluster on DVP, deckhouse image `deckhouse-oss:pr23053` against `:main`, same harness, 120 s window per experiment, 30-group scenarios run twice.

| Scenario | Metric | main | pr23053 |
|---|---|---|---|
| create one NodeGroup | `discover_standby_ng` | 1 run, 0.124 s | 1 run, 0.003 s |
| maxPerZone 0→1 on a group without standby | `discover_standby_ng` | 1 run, 0.325 s | 0 runs |
| maxPerZone 0→1 | ModuleRun / helm | 2 / 2 | 1 / 1 |
| maxPerZone 0→1 | all hooks | 50 runs, 3.6 s | 25 runs, 1.05 s |
| delete one group | PATCH `nodegroups/status` | 4 | 0 |
| create 30 groups | `discover_standby_ng` | 6 runs, 17.9 s (2.99 s/run) | 25 runs, 0.35 s (0.014 s/run) |
| create 30 groups | all hooks | 391 runs, 33.9 s | 412 runs, 15.9 s |
| create 30 groups | PATCH `nodegroups/status` | 190 | 76 |
| delete 30 groups | `discover_standby_ng` | 4 runs, 9.0 s | 20 runs, 0.28 s |
| delete 30 groups | PATCH `nodegroups/status` | 69 | 0 |
| maxPerZone on one of 34 groups | `discover_standby_ng` | wakes up | 0 runs |

Live check of the queue move is pending the next image build: changing `cloudInstances.priority` or the size of a group with `standby` must give 1 ModuleRun (2 on main, 6.2 s of the `main` queue for a 1:1 to 2:2 change with 0 changed objects, measured 15.09).

The remaining 76 status patches on 30 groups come from node-controller, not from the hook. The hook runs more often (25 instead of 6) because each run is now 14 ms and the queue no longer coalesces them. The second helm run on min/max changes disappears because the hook no longer changes values from its own queue.

## Changelog entries

```changes
section: node-manager
type: fix
summary: The discover_standby_ng hook patches NodeGroup status only when the standby value changes and ignores changes of NodeGroups without standby.
---
section: node-manager
type: fix
summary: Hooks `discover_standby_ng` and `set_ng_priorities` run in the module queue, so one NodeGroup change no longer queues a second module run.
```


deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:fix/cheap-discover-standby-ng` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/35065887857) in 28m16s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.2.161`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 26m48s.

</details>

090809 (approved):

PR #23055, opened 2026-09-15
[dhctl] Fix master bootstrap Secret dependency during scale-up and replacement

mary-ppv:
## Description

This PR changes how `dhctl` uses `manual-bootstrap-for-master` during mutable master converge.

Regular master scale-up no longer waits for the newly created master IP addresses to appear in the bootstrap Secret. New masters are created using the existing cloud configuration, after which `dhctl` waits only for the Kubernetes Nodes and control-plane components to become Ready.

When a Terraform plan requires destructive replacement of a master VM, `dhctl`:

1. Identifies the surviving master nodes.
2. Resolves their `InternalIP` addresses from Kubernetes Nodes.
3. Verifies that every surviving master is published as a Ready API server endpoint through kube-apiserver Pods or the `default/kubernetes` EndpointSlice.
4. Waits until these addresses are present in `manual-bootstrap-for-master`.
5. Reads the refreshed `cloud-config`.
6. Rebuilds the Terraform variables and infrastructure plan before running the safety hooks and apply.

Bootstrap Secret polling has been replaced with a race-safe Kubernetes `List + Watch` flow with a single absolute timeout. Validation now rejects missing or malformed `apiserverEndpoints`, missing expected hosts, and an empty `cloud-config`. Timeout errors include the last resource version, currently published endpoints, missing hosts, and the last validation or Watch error.

The bootstrap secrets controller now watches kube-apiserver Pods and the Kubernetes EndpointSlice, so endpoint changes trigger an immediate Secret refresh instead of waiting for the periodic 30-minute resync.

API server endpoint discovery has been moved to a shared `dhctl/pkg/kubernetes/apiserver` package. It is used by mutable destructive converge and the existing immutable join flow. Immutable join remains independent of `manual-bootstrap-for-master` and preserves its existing endpoint discovery behavior.

The change affects master/control-plane converge sequencing. It does not initiate additional control-plane restarts by itself; the new logic runs only when the infrastructure plan already requires a master VM replacement.

`manual-bootstrap-for-<NodeGroup>` is not removed. It remains the bootstrap data source for mutable nodes, including manually managed static nodes.

During regular master scale-up, `dhctl` previously waited for the IP addresses of newly created masters to appear in `manual-bootstrap-for-master`.

The Secret is generated asynchronously by node-controller. Its contents could remain unchanged until the next reconciliation or periodic resync, causing an otherwise successful scale-up to block and eventually time out.

At the same time, a destructive master VM replacement still requires fresh cloud configuration. The replacement must be able to connect to the surviving API servers after the old VM is removed.

This PR separates these two cases:

- Regular scale-up does not wait for the Secret to publish the new master addresses.
- Destructive VM replacement keeps the safety check and requires the Secret to contain all surviving Ready API server addresses.
- Terraform variables and the plan are refreshed immediately before the destructive operation, ensuring that the applied plan uses the validated cloud configuration.
- Single-master destructive converge continues to use the existing `1 → 3 → 1` safety workflow.
- Immutable masters continue to build per-node join payloads directly from Kubernetes and do not depend on the manual bootstrap Secret.

Not necessarily.

The change affects explicit `dhctl converge` operations involving master infrastructure. It does not affect normal cluster runtime. It can be backported if the target patch release must address master scale-up timeouts or destructive replacement failures caused by stale bootstrap data.

## Checklist

- [x] e2e tests passed.

## Changelog entries

```changes
section: dhctl
type: fix
summary: Avoid waiting for newly added master endpoints during regular scale-up and refresh Terraform variables only before destructive master VM replacement.
impact_level: default
```

```changes
section: node-manager
type: fix
summary: Refresh bootstrap Secrets when kube-apiserver Pods or the Kubernetes EndpointSlice change.
impact_level: default
```


deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:fix/dhctl-master-bootstrap-secret-dependency-redo` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/34953742902) in 22m10s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.2.184`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 20m40s.

</details>

PR #23066, opened 2026-09-15
[cert-manager] Restore the permissions lost against the upstream chart

Suselz:
## Description

Align the RBAC of the `cert-manager` module with the upstream cert-manager chart of the version
the module ships, v1.20.3. Four rules were lost when the chart was ported into the module.

`d8:cert-manager:cert-manager:ingress-shim` gains `listenersets` and `listenersets/finalizers`
from the `gateway.networking.k8s.io` API group:

```yaml
- apiGroups: ["gateway.networking.k8s.io"]
  resources: ["gateways", "httproutes", "listenersets"]
  verbs: ["get", "list", "watch"]
- apiGroups: ["gateway.networking.k8s.io"]
  resources: ["gateways/finalizers", "httproutes/finalizers", "listenersets/finalizers"]
  verbs: ["update"]
```

`d8:cert-manager:cert-manager:orders` gains the finalizers of `Issuer` and `ClusterIssuer`:

```yaml
- apiGroups: ["cert-manager.io"]
  resources: ["clusterissuers/finalizers", "issuers/finalizers"]
  verbs: ["update"]
```

`d8:cert-manager:cert-manager:issuers` and `d8:cert-manager:cert-manager:clusterissuers` gain
`patch` on the resource and on its status subresource, next to the `update` they already have.

`d8:cert-manager:cainjector` gains `patch` on `validatingwebhookconfigurations`,
`mutatingwebhookconfigurations`, `apiservices` and `customresourcedefinitions`.

The rules are added unconditionally, the same way upstream adds them. A rule in a ClusterRole does
not require the corresponding CRD to be present in the cluster, so tying the `listenersets` rules
to `certManager.enableListenerSet` or to the presence of the `ListenerSet` kind would only add
a way for the role to drift out of sync with the running controller.

The change updates ClusterRoles only. It restarts no component.

**`listenersets`.** The module can already turn on ListenerSet support: `certManager.enableListenerSet`
enables the `ListenerSets` feature gate and, when Gateway API integration is on, passes
`--enable-gateway-api-listenerset=true` to the controller. The controller then starts a watch on
`ListenerSet`, but the ClusterRole granted only `gateways` and `httproutes`. The watch failed on
every attempt and the controller logged an error in a loop:

```text
Failed to watch" err="failed to list *v1.ListenerSet: listenersets.gateway.networking.k8s.io is forbidden: User \"system:serviceaccount:d8-cert-manager:cert-manager\" cannot list resource \"listenersets\" in API group \"gateway.networking.k8s.io\" at the cluster scope"
```

Certificates for domains published through a `ListenerSet` were never issued, so the parameter was
unusable as it stood.

**`issuers/finalizers` and `clusterissuers/finalizers`.** An Order sets an owner reference on the
Issuer or ClusterIssuer that produced it. In a cluster running the `OwnerReferencesPermissionEnforcement`
admission plugin, setting an owner reference requires the `update` verb on the owner's `finalizers`
subresource, so without the rule the Order is rejected on creation.

**`patch` for the issuer controllers and for cainjector.** Both write through server-side apply,
which the API server accounts as `patch` rather than `update`.

None of the three affects a default installation, so the fixes come with the `listenersets` one
rather than separately.

No existing guarantee is relaxed. Every rule covers a resource type the controller already tries
to use, and follows the upstream chart exactly.

The audit found one more difference that this pull request leaves alone: upstream also ships
a `cert-manager-tokenrequest` Role granting `create` on `serviceaccounts/token` for the controller's
own ServiceAccount, used by an `Issuer` that authenticates to Vault through `serviceAccountRef`.
It is a namespaced Role rather than a missing verb in an existing one, so it belongs in its own
change.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: cert-manager
type: fix
summary: Restore the `cert-manager` and `cainjector` permissions that were lost against the upstream chart, including access to `ListenerSet` resources.
```


AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [23075](https://github.com/deckhouse/deckhouse/pull/23075) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

deckhouse-BOaTswain:
Cherry pick PR [23077](https://github.com/deckhouse/deckhouse/pull/23077) to the branch [release-1.76](https://github.com/deckhouse/deckhouse/tree/release-1.76) successful!

PR #23079, opened 2026-09-15
[node-manager] Run values-writing hooks in the module queue

borg-z:
## Description

`discover_standby_ng` and `set_ng_priorities` write node-manager values from their own queues. addon-operator compares the module values checksum before and after every hook run (`RunHookByName`), so a values commit from a concurrent queue is attributed to the hook running in `/modules/node-manager` at that moment, and that hook queues a second ModuleRun for the same event.

Both hooks now run in `/modules/node-manager`, the queue of `get_crds` and every other values writer of the module. Hooks in one queue run one after another, so no hook sees a foreign commit inside its own run.

`TestValuesWritersShareModuleQueue` pins the queue of every values writer and fails on the old code:

```
values_writers_queue_test.go:44: discover_standby_ng.go runs in queue "/modules/node-manager/discover_standby_ng", values writers share /modules/node-manager
values_writers_queue_test.go:44: set_ng_priorities.go runs in queue "/modules/node-manager/set_priorities", values writers share /modules/node-manager
```

Live measurement on a DVP cluster (15.09.2026, addon-operator 1.24.18): changing `minPerZone`/`maxPerZone` of a NodeGroup produced 2 ModuleRuns of node-manager, 6.2 s of the `main` queue, with 0 changed release objects. The second run is the race described above (`values changed` reported by a hook that writes no values). Enabling the autoscaler on a group: 2 ModuleRuns and 2 helm runs on `main`, 1 and 1 with `discover_standby_ng` idle (#23053). One queue for all writers removes the second run for size and priority changes.

One ModuleRun per NodeGroup change that touches values. Check on a live cluster: change `cloudInstances.priority` or the size of a group with `standby`, `deckhouse_module_run` count for node-manager grows by 1.

## Checklist

- [x] Unit tests: `go test ./modules/040-node-manager/hooks/ -run TestValuesWritersShareModuleQueue`, and the specs of both hooks (`-ginkgo.focus='standby|riorit'`)
- [ ] Live check after the image is built

## Changelog entries

```changes
section: node-manager
type: fix
summary: Hooks `discover_standby_ng` and `set_ng_priorities` run in the module queue, so one NodeGroup change no longer queues a second module run.
```


borg-z:
Folded into #23053, same change set.

PR #23093, opened 2026-09-16
feat(user-authn): redesign pages using new branding

alexeyvokin:
## Description

Updates the Dex authentication page to the new Deckhouse branding.

**Logo.** The wordmark on the login card is replaced with the new Deckhouse logo.

**Background.** The page background is now a solid brand color with a large watermark logo instead of `background.png` with a blue gradient. Two variants are shipped and selected by the visitor's system theme via `prefers-color-scheme`:

| Theme | Background color | Watermark |
|---|---|---|
| light | `#EAF0FB` | `deckhouse-background-logo-theme-light.svg` (white) |
| dark | `#004DF2` | `deckhouse-background-logo-theme-dark.svg` |

The background moved from an inline `style` attribute on `<body>` into `main.css` (class `.login-page`), since a media query cannot be expressed inline. Relative `url()` inside the stylesheet resolves against the stylesheet location, so the assets are found correctly regardless of the issuer path prefix — verified on an issuer with a `/dex` prefix.

**Flant logo.** The shipped logo is white and became unreadable on the new light background. Light-theme copies in `#262429` were added, and the footer switches between them with `<picture media="(prefers-color-scheme: dark)">`. The existing files were renamed to `-theme-dark` for symmetry. The unused `flant-logo.svg` (no `-en`/`-ru` suffix, not referenced by any template) is removed.

**Card shadow.** A shadow was added to the login card in the light theme, where the card (`#F6F6FD`) and the background (`#EAF0FB`) are close in lightness and the card edge visually dissolved into the watermark. In the dark theme the shadow is disabled — contrast against `#004DF2` is sufficient.

**Favicons.** Added `favicon.svg` and a PNG set (32/180/192/512) with the new brand mark.

Scope of changes: web assets and templates of the module only. Dex sources, patches, and the pinned version are untouched, so the login flow, connectors, and MFA behavior do not change.

The authentication page is the first screen users see when accessing cluster components, and it still carried the previous visual identity: the old wordmark and the gradient background. The change aligns the page with the current Deckhouse branding and adds light theme support, which the page did not have at all — the background used to be dark for every user regardless of their system settings.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

Verified locally on Dex v2.44.0 built with the module patches, with `frontend.dir` pointed at this directory: login, password, approval, password change, device and error pages render correctly in both themes; every asset referenced from the templates and the stylesheet resolves.

Note for reviewers: rolling out the new image restarts the Dex pods, so authentication is briefly unavailable during the update. Already authenticated sessions are not affected.

## Changelog entries

```changes
section: user-authn
type: feature
summary: Update the Dex authentication page to the new Deckhouse branding, with light and dark themes selected by the user's system settings.
```


AlwxSin (approved):

deckhouse-BOaTswain:
Cherry pick PR [23155](https://github.com/deckhouse/deckhouse/pull/23155) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #23099, opened 2026-09-16
[deckhouse-controller/package] add resources fields to applications

ldmonster:
## Description

Adds `Application.spec.resourceRequests`: a per-workload override of the resource footprint an application package ships — how many replicas a workload runs and what its containers may consume.

```yaml
kind: Application
metadata:
  name: redis-cache
  namespace: my-project
spec:
  packageName: redis
  packageVersion: v7.2.0
  resourceRequests:
    - kind: StatefulSet
      name: redis
      replicas: 3
      containers:
        - name: redis
          resources:
            requests:
              memory: 512Mi
            limits:
              memory: 1Gi
```

A workload is addressed by `kind` + `name`, as a keyed list, so at most one entry exists per pair. A workload or container that is not listed keeps the package defaults. Container resources are merged resource by resource: setting only `memory` leaves the package's `cpu` in place, and a resource the package declares can be given a different value but cannot be removed.

### How the overrides are applied

They are not chart values — a package is under no obligation to expose any — so they are applied as a post-render overlay:

1. The `Application` controller projects `spec.resourceRequests` onto the runtime shape, carrying quantities as canonical strings, and hands it to the package runtime next to settings and maintenance mode.
2. The runtime store keeps it as a pending value alongside the others. A change that touches only the overrides needs no new package context: it stores the value and reschedules with reason `ResourceRequestsChanged`.
3. The Configure task sets it on the package through an optional `resourceSizer` interface — the same path `maintenance` already takes. Modules do not implement it, since the field exists only on `Application`.
4. `nelm.Service.Upgrade` renders the package chart into classified resources, overlays the requested replicas and container resources onto the rendered manifests, and folds the result back into a manifest string **before** the checksum is calculated. The release checksum, the endpoint URLs and the absent-resources monitor therefore all see exactly what will be applied.
5. nelm renders a chart rather than taking manifests, so the overlaid resources are materialised as a synthetic chart in the process temp dir — one template per manifest, `Chart.yaml` carried over from the source minus `dependencies`, template actions escaped so the second render hands the manifest back unchanged — and the release is installed from that chart. It is written only after the upgrade is known to be needed, and removed on the way out.

`Render`, which backs the runtime dump, applies the same overlay, so the dump shows what the package actually gets rather than the chart's own sizing.

### Feature gate

Off by default: `DECKHOUSE_ENABLE_RESOURCE_REQUESTS`, set to `"false"` in `modules/002-deckhouse/templates/deployment.yaml`. With the gate off the field is accepted by the API and ignored, and both the rendered manifests and the chart the release installs from are what they were before this PR. The gate is read at the single point the overlay enters the release path.

It exists because that install path replaces the chart nelm renders with a fully-rendered synthetic one, and anything a chart carries beyond the manifests it produces is dropped: `NOTES.txt`, values schemas, subcharts. Helm hooks survive, because nelm reclassifies them from the annotations on the manifests themselves. Standalone `crds/` CRDs are deliberately left out of the overlay, since `Install` already passes `NoInstallStandaloneCRDs` and carrying them into `templates/` would start applying CRDs the release has never applied.

### Validation and known limits

- `kind` is an enum of the seven workload kinds that carry a pod spec; a CEL rule rejects `replicas` on anything but `Deployment`, `StatefulSet` and `ReplicaSet`.
- Containers are matched across both `containers` and `initContainers`, since a container name is unique across the two within one pod spec. A container the workload does not have is ignored.
- Workloads are matched by kind and name, without namespace: an entry applies to every rendered workload with that kind and name.
- An entry matching no rendered workload is logged as a warning rather than failing the release — a chart may render a workload conditionally, so a dormant entry is legitimate. A typo in `kind` or `name` looks the same from there.

### Refactoring carried along

`Client.Render` is split into `Client.RenderResources`, which returns each rendered resource with the classification nelm gave it (regular release resource or Helm hook), and `JoinManifests`, which flattens the regular ones into the multi-document YAML the callers used to get. `Render` is now a thin wrapper over the two, and the filtering it used to do inline is unchanged.

No critical cluster components are affected: nothing here runs unless both the package system and the new gate are on.

An application package ships one sizing for the workloads it deploys, chosen by whoever authored the package. Clusters differ: a controller sized for a small installation is wrong for a large one, and a replica count that suits production is waste in a sandbox. Until now the only ways to change it were to fork the package or to hope its author exposed the right values — both of which leave the user maintaining something the package should own.

`spec.resourceRequests` moves that decision to the `Application` CR, where the user already declares everything else about the instance. It works the same way for every package, whether or not its chart exposes resource values, and it is expressed in the same shape as a Pod container's `resources`, so there is nothing package-specific to learn.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

Unit tests cover the overlay and the synthetic-chart generation, including a round-trip check that a generated template renders back to the manifest it was written from. Documentation is the CRD field descriptions and their Russian translation; the marketplace user guide is not extended, the same way `spec.maintenance` was shipped.

## Changelog entries

```changes
section: deckhouse-controller
type: feature
summary: Per-workload replicas and container resources of an application can be overridden via `Application.spec.resourceRequests`.
impact_level: low
```


z9r5 (approved):

deckhouse-BOaTswain:
:green_circle: `e2e: AWS` for `deckhouse:feat/add-resource-control-to-applications` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/35839615021) in 34m3s.



AWS-WithoutNAT-Containerd-1.34 - Connection string: `ssh  ec2-user@3.64.198.254`


:green_circle: `e2e: AWS, Containerd, Kubernetes 1.34` succeeded in 32m44s.

</details>

PR #23108, opened 2026-09-16
[user-authn] Remove the kubeconfig-generator web UI

AlwxSin:
## Description

Removes the `kubeconfig-generator` web UI (`dex-k8s-authenticator`) from the `user-authn` module and, with it, the `kubeconfig.<publicDomainTemplate>` hostname.

Deleted:

- the `Deployment`, `Service`, `Ingress`, `HTTPRoute`, `PodDisruptionBudget`, `Certificate` and custom-certificate templates under `templates/kubeconfig-generator/`;
- the `images/kubeconfig-generator/` build directory;
- the `kubeconfig` listeners of the module `ListenerSet` — the hostname disappears entirely, there is no redirect to the console;
- the `https://kubeconfig.<domain>/callback/...` redirect URIs of the dex OAuth2 clients;
- the `Kubeconfig generator` documentation page and its sidebar entry.

Deliberately kept:

- the `userAuthn.kubeconfigGenerator[]` configuration field — it is the data source for the `kubernetes-dex-client-configuration` Secret and is read by the Deckhouse console from the ModuleConfig. It is not deprecated by this change;
- the `kubeconfig-generator` and `kubeconfig-generator-N` dex OAuth2 clients, so that kubeconfigs already issued to users keep working. They are removed in DKP 1.79;
- the legacy `kubeconfig-generator` Secret (`templates/kubeconfig-generator/config.yaml`), because an older console version can land on a newer DKP and still read it through its fallback path;
- the CLI and device-flow redirect URIs (`http://localhost:8000`, `http://localhost:18000`, `/device/callback`) that existing user kubeconfigs depend on.

`template_tests/kubernetes_oauth2client_test.go` now asserts that the `kubeconfig-generator` and `kubeconfig-generator-N` OAuth2Clients render and keep accepting those CLI callbacks, so a later cleanup cannot drop `templates/kubeconfig-generator/` silently.

The web UI is superseded by the Deckhouse console, which reads the `kubernetes-dex-client-configuration` Secret shipped in DKP 1.77. Keeping the old UI means maintaining a separate patched third-party application, its image, its own certificate and Ingress/Gateway plumbing, and a public hostname — all for functionality the console already provides.

This is the first of two phases. Phase two (DKP 1.79) removes the remaining `kubeconfig-generator` / `kubeconfig-generator-N` OAuth2 clients once the kubeconfigs issued by the old UI have aged out.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: chore
summary: Remove the kubeconfig-generator web UI and the kubeconfig.<publicDomainTemplate> hostname.
impact: |
  The kubeconfig generation web UI is removed and the `kubeconfig.<publicDomainTemplate>`
  hostname stops resolving; there is no redirect to a replacement page.
  Kubeconfigs already issued to users keep working: the dex OAuth2 clients behind them
  remain in place until DKP 1.79.
  Users should re-issue their kubeconfig from the Deckhouse console, which provides the
  same functionality.
  The `userAuthn.kubeconfigGenerator` module setting is not affected and keeps working —
  the console reads the API endpoints defined there.
  On clusters that served the UI with `https.mode: CertManager`, the TLS secrets issued for
  the removed hostname (`kubeconfig-ingress-tls`, plus `kubeconfig-httproute-tls` where
  Gateway API was used) remain in the `d8-user-authn` namespace after the upgrade. Their
  Certificates set `certificateOwnerRef: false`, so deleting them does not garbage-collect
  the secrets. The leftovers are unreferenced and no longer renewed; a migration in DKP 1.79
  removes them.
impact_level: high
```


z9r5 (approved):

Jabejixo (commented):
Reviewed the removal end to end. The mechanical part is clean: the `if/else` collapse in `templates/dex/oauth2client.yaml` stays balanced and the `kubernetes` client keeps rendering `redirectURIs: []` (it is an audience holder, its peers do the redirecting); the `ListenerSet` still emits the dex listeners in both branches, so `spec.listeners` can never come out empty; nothing references `sectionName: kubeconfig` or the `kubeconfig-ingress-tls` / `kubeconfig-httproute-tls` secrets any more; no monitoring rule or build index mentions the removed image; the deleted `user/web/kubeconfig.html` page has no inbound links left and the new `user/web/ui.html` targets resolve. The nine spelling-dictionary additions are all required, since the spellchecker validates changed files in full and `KDC`, `SPN`, `ListenerSet`, `ALBInstance`, `unboundedly` and the rest already lived in the `FAQ` / `K8S_API_LB` pages this PR touches.

Four notes, three of them inline. This one has no anchor in the diff.

---

🟡 **`publishAPI.addKubeconfigGeneratorEntry` now documents a UI that no longer exists**

The setting survives this PR and still gates real behaviour: the published-API entry in the `kubernetes-dex-client-configuration` Secret that the console reads, the `kubeconfig-publish-api` trustedPeer, and the matching entry in the legacy Secret. Its description still points at the deleted web UI:

```yaml
# modules/040-control-plane-manager/openapi/config-values.yaml:227-230
              addKubeconfigGeneratorEntry:
                type: boolean
                default: true
                description: 'Setting it to `false` will remove an entry in kubeconfig-generator.'
```

An administrator who wants to hide the published API endpoint from the console's cluster list finds only a reference to a UI that is gone, so the flag reads as dead — and is either left alone or flipped as a presumed no-op, silently dropping the endpoint from the console. The same sentence is in four places:

- `modules/040-control-plane-manager/openapi/config-values.yaml:230`
- `ee/modules/040-control-plane-manager/openapi/config-values.yaml:230`
- `ee/cse/modules/040-control-plane-manager/openapi/config-values.yaml:230`
- `modules/040-control-plane-manager/openapi/doc-ru-config-values.yaml:139` (`будет удалена запись в kubeconfig-generator`)

Suggested wording: the entry is the published API endpoint offered for kubeconfig generation (in the Deckhouse web UI), not an entry in a generator page.

modules/150-user-authn/template_tests/kubernetes_oauth2client_test.go
```diff
@@ -279,18 +279,12 @@ var _ = Describe("Module :: user-authn :: helm template :: kubernetes oauth2clie
...
-			))
-			// Authenticators redirect to their own OAuth2Client, so no tenant-controlled
-			// domain may reach the redirect URIs of the privileged kubernetes client.
-			Expect(uris).NotTo(ContainElement(ContainSubstring("/dex-authenticator/callback")))
+			// Every peer redirects to its own OAuth2Client, so the privileged kubernetes
+			// client needs no redirect URI of its own. In particular, no tenant-controlled
+			// domain and no leftover kubeconfig web UI callback may reach it.
+			redirectURIs := oauth2Client.Field("redirectURIs")
+			Expect(redirectURIs.IsArray()).To(BeTrue(), "redirectURIs must render as a list, not a bare key parsed as null")
+			Expect(redirectURIs.Array()).To(BeEmpty())
```
Jabejixo, line 287: 🟡 **Nothing pins the two compatibility-critical OAuth2Clients**

This rewrite is correct, but the loop right below it walks only `kubeconfig-publish-api`, `kubeconfig-plain` and `kubeconfig-prod-eu` — the clients from `templates/dex/kubeconfig-oauth2clients.yaml`, which this PR does not touch:

```go
for _, c := range []struct{ clientID string }{
    {clientID: "kubeconfig-publish-api"},
    {clientID: "kubeconfig-plain"},
    {clientID: "kubeconfig-prod-eu"},
} {
    ...
    Expect(uris).To(ConsistOf(
        "http://localhost:8000",
        "http://localhost:18000",
        "/device/callback",
    ))
}
```

The clients the PR description promises to keep until 1.79 — `kubeconfig-generator` and `kubeconfig-generator-N` — are asserted nowhere. `kubeconfig_generator_config_test.go` contains no `OAuth2Client` assertion at all, and line 274 above only checks that the string `kubeconfig-generator` appears in the `trustedPeers` list of the `kubernetes` client, not that the client object exists or what it accepts.

`templates/kubeconfig-generator/` is now down to two files, so the next cleanup pass is likely to take the directory as a whole. Every kubeconfig issued by the old UI would then fail on login and on refresh, because dex no longer knows `kubeconfig-generator-N` or its `localhost:8000` callback — with the template suite still green. A few lines extending the loop above with both ids would make that break loudly.

modules/150-user-authn/templates/kubeconfig-generator/oauth2client.yaml
```diff
@@ -11,7 +11,6 @@ id: kubeconfig-generator
 name: kubeconfig-generator
 secret: {{ $context.Values.userAuthn.internal.kubernetesDexClientAppSecret | quote }}
 redirectURIs:
-- https://{{ include "helm_lib_module_public_domain" (list $context "kubeconfig") }}/callback/
```
Jabejixo, line 14: 🔵 **The legacy Secret still advertises the callback removed here**

`templates/kubeconfig-generator/config.yaml` is deliberately kept for the older-console fallback, and it still emits the URI this hunk deletes:

```yaml
# templates/kubeconfig-generator/config.yaml:67
  redirect_uri: "https://{{ include "helm_lib_module_public_domain" (list $context "kubeconfig") }}/callback/{{ $index }}"
```

So one preserved artifact advertises a callback the other no longer accepts, and the tests pin it both ways: `kubeconfig_generator_config_test.go:62` and `:219` still expect `https://kubeconfig.example.com/callback/`.

In practice this is probably only a misleading value: `redirect_uri` is `dex-k8s-authenticator`'s own browser callback, and a kubeconfig consumer does not need it — issued kubeconfigs authenticate through kubelogin on `localhost:8000`, which you keep. It turns into a real failure only in the case the description names as the reason for keeping the Secret: if an older console replays `redirect_uri` from it, dex rejects the authorization request as an invalid redirect URI and kubeconfig issuance fails for exactly the users the shim protects. Worth a confirmation from the console side, then either drop the line from the Secret or leave this URI on the client.

modules/150-user-authn/templates/kubeconfig-generator/certificate.yaml
```diff
@@ -1,44 +0,0 @@
...
-  {{- if and (eq (include "helm_lib_module_ingress_enabled" .) "true") (eq (include "helm_lib_module_https_mode" .) "CertManager") }}
----
-apiVersion: cert-manager.io/v1
-kind: Certificate
-metadata:
-  name: kubeconfig
-  namespace: d8-{{ .Chart.Name }}
-  {{- include "helm_lib_module_labels" (list . (dict "app" "kubernetes-configurator")) | nindent 2 }}
-spec:
-  certificateOwnerRef: false
```
Jabejixo, line 13: 🔵 **The issued TLS secret is left in the namespace after the upgrade**

Both Certificates deleted here opt out of owner references:

```yaml
spec:
  certificateOwnerRef: false
  secretName: {{ include "helm_lib_module_https_secret_name" (list . "kubeconfig-ingress-tls") }}
```

With `certificateOwnerRef: false` cert-manager never sets the owner reference on the issued Secret, so deleting the Certificate does not garbage-collect it — and this holds regardless of `certManager.cleanupOrphanSecrets`, which is `false` by default anyway. On any cluster that ran the UI with `https.mode: CertManager`, Helm removes the Certificate during the upgrade while `kubeconfig-ingress-tls` (and `kubeconfig-httproute-tls` where Gateway API was used) stays in `d8-user-authn`: unowned, unreferenced and no longer renewed, surfacing as an expired certificate in any tooling that inventories TLS secrets. A one-off migration hook deleting both secrets, or a line in the release impact text, closes it out.

modules/150-user-authn/template_tests/kubernetes_oauth2client_test.go
AlwxSin: Good catch — fixed in 5a30199.

You are right that nothing asserted those two clients: lines 244-245 that look like coverage are `ValuesSet` calls feeding the render, and line 274 only checks that the string appears in the `trustedPeers` list of the `kubernetes` client.

Added a dedicated `It` that renders `kubeconfig-generator`, `kubeconfig-generator-0` and `kubeconfig-generator-1`, asserts each OAuth2Client exists under its FNV-encoded CR name, and pins the redirect URIs to exactly `http://localhost:8000`, `http://localhost:18000` and `/device/callback` — the CLI and device-flow callbacks that already-issued kubeconfigs authenticate through.

Verified the assertion actually bites rather than passing vacuously: swapping one id for a non-existent one fails with `legacy OAuth2Client ... must survive the web UI removal`, and restoring it goes green (86 passed).

modules/150-user-authn/templates/kubeconfig-generator/oauth2client.yaml
AlwxSin: Confirmed from the console side, and the failure mode you describe does not occur: the console never replays that `redirect_uri`.

In `modules/console`, `images/backend/internal/k8s/kubeconfig/kubeconfig.go` declares `RedirectURI string `json:"redirect_uri"`` on the legacy struct (line 101), and that declaration is its only occurrence in the whole backend — nothing reads the field. `normalizeLegacyConfig` maps only `client_id`, `client_secret`, `issuer`, `k8s_master_uri`, `name` and `k8s_ca_pem` into the canonical config; `redirect_uri` is dropped on the floor. So an older console reading the preserved Secret builds its authorization request without it, and dex has no invalid-redirect to reject.

That leaves your first reading: a stale value in a Secret nobody reads it from. Leaving it is the safer option — the Secret exists purely as a compatibility shim for older console builds, and changing its schema now would alter the exact payload those builds parse, which is the one thing the shim is supposed to hold still. It goes away together with the Secret and the OAuth2Clients in DKP 1.79.

modules/150-user-authn/templates/kubeconfig-generator/certificate.yaml
AlwxSin: Confirmed, and observed on a live stand: after switching a cluster to this PR image, `kubeconfig-ingress-tls` stayed in `d8-user-authn` with no `ownerReferences`, carrying only the `cert-manager.io/*` annotations, while the Certificate and the Ingress were gone.

Your mechanism is right on both counts — both Certificates set `certificateOwnerRef: false`, and `cert-manager.cleanupOrphanSecrets` defaults to `false` in `modules/101-cert-manager/openapi/config-values.yaml`.

Handled as you suggested, taking the release-notes option now and the migration next: the `impact` block of the changelog entry now names both secrets, the reason they survive and the fact that they are no longer renewed. The one-off cleanup is scheduled for DKP 1.79, where it lands together with the removal of the `kubeconfig-generator` OAuth2Clients and the legacy Secret — keeping this PR free of a migration hook that would run against clusters whose OAuth2Clients are still in use.

Jabejixo:
Checked all three replies against the sources rather than taking them on trust, and they hold:

- the new `It` is not vacuous — `ToFnvLikeDex("kubeconfig-generator")` really is the hardcoded CR name `nn2wezldn5xgm2lhfvtwk3tfojqxi33szpzjzzeeeirsk`, the enclosing context sets `publishAPI.enabled: true` plus both `kubeconfigGenerator` entries so all three clients render, and the `kubeconfigEncodedNames` fixture matches what the hook computes, so the lookups hit real objects. `modules/150-user-authn/template_tests` is green on `5a30199e52`;
- `RedirectURI` occurs exactly once in the console backend, as the struct field declaration, and `normalizeLegacyConfig` carries over only `ID`, `Name`, `MasterURI`, `MasterCA`, `ClientID`, `ClientSecret` and `Issuer`. The invalid-redirect path I described cannot happen, and holding the shim's payload still is the right call;
- the `impact` block now names both secrets, `certificateOwnerRef: false` as the reason and the 1.79 migration.

The red `Tests` job on this head is unrelated: the only `=== FAIL` in the log is `modules/030-cloud-provider-gcp/hooks` `change_host_ip_test.go:74`, a flake in that hook's suite.

---

That leaves the `addKubeconfigGeneratorEntry` note from my review body, and I have to correct myself on it twice.

**It is in seven places, not four, and the two that matter most are the ones I failed to list.** The copies on the user-authn configuration page are the ones this PR's own FAQ links to via `/modules/user-authn/configuration.html#parameters-publishapi`:

```yaml
# modules/150-user-authn/openapi/config-values.yaml:58-61
      addKubeconfigGeneratorEntry:
        type: boolean
        default: true
        description: 'Setting it to `false` will remove an entry in kubeconfig-generator.'
```

Full list at `5a30199e52`:

| File | Line |
|---|---|
| `modules/150-user-authn/openapi/config-values.yaml` | 61 |
| `modules/150-user-authn/openapi/doc-ru-config-values.yaml` | 35 |
| `modules/040-control-plane-manager/openapi/config-values.yaml` | 230 |
| `modules/040-control-plane-manager/openapi/doc-ru-config-values.yaml` | 139 |
| `ee/modules/040-control-plane-manager/openapi/config-values.yaml` | 230 |
| `ee/modules/040-control-plane-manager/openapi/doc-ru-config-values.yaml` | 139 |
| `ee/cse/modules/040-control-plane-manager/openapi/config-values.yaml` | 230 |

None of them is a symlink, so a fix has to be applied seven times.

**And my wording was stronger than the facts.** I said it documents a UI that no longer exists; strictly it does not, because an object named `kubeconfig-generator` still exists — the legacy Secret — and the flag really does still gate an entry inside it (`templates/kubeconfig-generator/config.yaml:43`). So the sentence is not false today. It becomes false in 1.79, when that Secret goes.

The accurate complaint is narrower: after this PR the sentence points at the one artifact a user cannot see and that is scheduled for deletion, while saying nothing about what the flag now visibly does. At `5a30199e52` it gates four things:

```
templates/dex/kubernetes-dex-client-configuration.yaml:1,20   the Secret the console reads, and the publishAPI entry in it
templates/dex/kubeconfig-oauth2clients.yaml:2                 the kubeconfig-publish-api OAuth2Client
templates/dex/oauth2client.yaml:38                            its trustedPeer on the privileged kubernetes client
templates/kubeconfig-generator/config.yaml:1,43               the entry in the legacy Secret
```

So flipping it to `false` now removes the published API endpoint from the list the console offers, together with the OAuth2Client behind it. Something like "Setting it to `false` removes the published API endpoint from the endpoints offered for kubeconfig generation" would survive 1.79 and describe the effect an administrator can actually observe. Entirely fine as a follow-up rather than in this PR — the setting is not part of the removal.

AlwxSin:
Verified your corrected list — seven places, and none of them a symlink:

| File | Line |
|---|---|
| `modules/150-user-authn/openapi/config-values.yaml` | 61 |
| `modules/150-user-authn/openapi/doc-ru-config-values.yaml` | 35 |
| `modules/040-control-plane-manager/openapi/config-values.yaml` | 230 |
| `modules/040-control-plane-manager/openapi/doc-ru-config-values.yaml` | 139 |
| `ee/modules/040-control-plane-manager/openapi/config-values.yaml` | 230 |
| `ee/modules/040-control-plane-manager/openapi/doc-ru-config-values.yaml` | 139 |
| `ee/cse/modules/040-control-plane-manager/openapi/config-values.yaml` | 230 |

Your self-correction is right on the substance too: `templates/kubeconfig-generator/config.yaml:43` does still gate an entry in the legacy Secret, so the sentence is accurate today and only becomes wrong in 1.79. And the flag gates exactly the four things you list — I get the same set grepping the templates.

Taking it as the follow-up you suggested rather than folding it in here, for two reasons beyond scope. Five of the seven files belong to `040-control-plane-manager` and its ee/cse mirrors, so the change reaches a module this PR does not otherwise touch; and fixing only the two user-authn copies would leave the same setting documented differently in two modules, which is worse than the current uniform staleness. Tracked with your suggested wording as the starting point, since "removes the published API endpoint from the endpoints offered for kubeconfig generation" survives 1.79 and names an effect an administrator can actually observe.

Thanks for the end-to-end pass — the `redirectURIs: []` audience-holder check and the `ListenerSet` never-empty check are exactly the two things I was most worried about breaking.

Jabejixo (approved):

nabokihms (approved):

PR #23112, opened 2026-09-16
[deckhouse-controller] fix source change

ldmonster:
## Description

Two fixes to the "registry settings changed" fan-out, one in the `ModuleSource` controller and one in the `PackageRepository` controller. Both controllers stamp an annotation on every deployed object of a source so that it re-applies the new registry settings, and both had a way to lose that work.

**`ModuleSource`** (`module-controllers/source`) annotated each `ModuleRelease` with `client.Update`. The release controller deletes the very same annotation once it has applied the settings, so the two controllers write the same object and collide:

```
failed to sync registry settings for module source
  error: set RegistrySpecChanged annotation to the 'console-v1.63.4' module release:
  Operation cannot be fulfilled on modulereleases.deckhouse.io "console-v1.63.4":
  the object has been modified; please apply your changes to the latest version and try again
```

The annotation is now written with a merge patch. A merge patch carries no `resourceVersion`, so the two controllers cannot collide over the object at all. On top of that the loop no longer stops at the first failure — it reports all of them — and a release deleted meanwhile is no longer treated as an error.

**`PackageRepository`** (`packages/package-repository`) committed the repository checksum *before* checking whether any `Application` had actually been annotated. The checksum is what makes the next reconcile return early, so committing it first defused the "nothing could be annotated" branch below it: the returned error requeued the reconcile, the requeue matched the checksum and returned before the fan-out, and every `Application` stayed on the old registry settings for good. The check now runs before the commit. A partial failure still commits and only warns, which is the existing intent — replaying the fan-out re-annotates the applications that did succeed.

On a cluster running v1.76.14 the main queue filled up in bursts with dozens of duplicate tasks, always the same set of modules, cycling:

```
Queue 'main': length 39, status: 'run first task'
 1. ModuleRun:main:observability:ModuleManager-Update-Module
 2. ModuleRun:main:sds-object:ModuleManager-Update-Module
 3. ModuleRun:main:snapshot-controller:ModuleManager-Update-Module
 ...
```

Every module in the queue was a deployed `ModuleRelease` of the `deckhouse` source, of which there were 20. Sequence:

1. The `deckhouse` `ModuleSource` is rendered by the deckhouse module's own chart (`modules/002-deckhouse/templates/default-module-source.yaml`), so a helm apply that changes `spec.registry` invalidates the stored `modules.deckhouse.io/registry-spec-checksum`.
2. `syncRegistrySettings` annotates all 20 deployed releases — and collides with the release controller on one of them.
3. On that error it returned before storing the new checksum, so the source never converged and the next reconcile replayed the **whole** fan-out, re-annotating the releases that had already succeeded. Measured: 178 `apply new registry settings to module` log records per burst for 20 releases, i.e. roughly 9 full replays.
4. Every re-annotated release pushes a `moduleRun` task, and each `moduleRun` is a helm release.

Observed helm revisions on that cluster: `managed-opensearch` at 937, `deckhouse` at 667, `observability` at 334. The duplicates also ran concurrently and raced each other over the same helm release:

```
ModuleRun failed. Requeue task to retry after delay.
  error: run helm install: install nelm release "managed-opensearch": ...
  create release "managed-opensearch" (namespace: "d8-system"): release: already exists
```

The `PackageRepository` half of this PR is the same class of bug found while fixing the first one; it has not been observed in the field yet, but it silently strands applications on stale registry settings when it fires.

**Note for reviewers:** this removes the cause of the replay, not the mechanism that lets duplicates accumulate. `PushRunModuleTask` in addon-operator is guarded by `queueHasPendingModuleRunTaskWithStartup`, which returns `has && meta.doStartup` while every caller pushes with `doStartup=false` — so the guard never fires and identical tasks stack up. That bug is present both in v1.21.33 (running on the cluster above) and in v1.24.18 (pinned in our `go.mod`), and needs a separate addon-operator PR plus a version bump here.

Yes. This is a live bug on released versions: it converges by luck, blocks the main queue while it lasts, and churns helm releases and etcd writes for every downloaded module in the meantime.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse-controller
type: fix
summary: Stop the module source registry fan-out from replaying and flooding the main queue with duplicate moduleRun tasks.
impact_level: default
---
section: deckhouse-controller
type: fix
summary: Do not commit the package repository registry checksum when no application could be annotated, which left them all on stale registry settings.
impact_level: default
```


deckhouse-BOaTswain:
:green_circle: `e2e: DVP` for `deckhouse:fix/fix-source-change` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/35116327152) in 36m57s.



DVP-WithoutNAT-Containerd-1.34 - Connection string: `ssh -J e2e-user@185.11.73.171 debian@10.12.5.38`


:green_circle: `e2e: DVP, Containerd, Kubernetes 1.34` succeeded in 35m26s.

</details>

Glitchy-Sheep (approved):

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 9bb173af0cbc25273232c12c865c835457a20817 to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/35198514423) for details.

PR #23118, opened 2026-09-16
[service-with-healthchecks] Fixed metric ports

raorudzhov:
## Description
Bind the `agent` and `controller` metrics servers to `127.0.0.1` explicitly instead of relying on the default all-interfaces bind address. Update the `agent` DaemonSet args with `--metrics-bind-address=127.0.0.1:9874` and change the `controller` Deployment arg from `--metrics-bind-address=:8080` to `--metrics-bind-address=127.0.0.1:8080`. Adjust the matching unit test snapshot in `template_tests/module_test.go` to reflect the new argument.

Both containers already run a `kube-rbac-proxy` sidecar that is meant to be the only entry point to the `/metrics` endpoint, authorizing every request against the Kubernetes API. However, the underlying binaries bound their metrics servers to all network interfaces (`:9874` for the agent, `:8080` for the controller) instead of `127.0.0.1`. As a result, anyone with network access to the pod could reach the metrics endpoint directly on the pod IP, bypassing `kube-rbac-proxy` and its authorization checks entirely. Binding the metrics servers to loopback only forces all metrics access through `kube-rbac-proxy`, as intended.

This closes an unauthenticated access path to the metrics endpoints, so it should be backported to patch releases rather than wait for the next minor release.

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
type: fix
summary: Fixed metrics endpoints being reachable without going through kube-rbac-proxy authorization.
impact_level: default
```


AndreyPavlovFlant (approved):

deckhouse-BOaTswain:
Cherry pick PR [23166](https://github.com/deckhouse/deckhouse/pull/23166) to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) successful!

PR #23124, opened 2026-09-17
[user-authz, multitenancy-manager] Close the aggregation-label gap, wait for the namespace on project deletion, warn about the deprecated parameter

Jabejixo:
## Description

Follow-ups to the RBACv2 role model and the multitenancy-manager adoption change, gathered from a review of the two design notes against the shipped code and from the first reports after adoption landed in `main`. Code only; the documentation that goes with it is a separate PR on top of this one.

**A platform aggregation label is refused on anything but a custom capability.** `d8:namespace:admin` aggregates every ClusterRole labelled `rbac.deckhouse.io/aggregate-to-namespace-as: admin`, and the aggregation controller reads nothing else — not the kind label, not the name. A ClusterRole with any name and no framework kind at all could carry that label, pass the `cluster_roles` webhook, and have its rules merged into the platform role for every holder in the cluster. Kubernetes only requires the author to already hold those rules, so this is not self-escalation; it is a way to hand one's own permissions to every namespace admin without a single binding, and nothing in the model saw it. The webhook now refuses such an object unless it declares itself a `custom-capability`, which the remaining checks then hold to the contract. A custom capability may in turn no longer carry an `aggregationRule` — a capability is the leaf of the model, as the platform test already requires of the built-in ones. A new test in `testing/rbacv2` (`TestRBACv2AggregationLabelPlacementValidation`, so that CI's `-run Validation` picks it up) walks every template of every module — subcharts included — and fails on the label anywhere outside `templates/rbacv2` and the compatibility aliases beside it, because the contract test only ever looked inside that directory. A webhook only sees writes, so a hook exports every ClusterRole that already carries the label without being a custom capability as `d8_user_authz_foreign_aggregation_label{name, lineage, kind}`, and `D8UserAuthzForeignAggregationLabel` says what to do with it before its next `UPDATE` is refused; the migration guide (#23180) carries the pre-upgrade query.

**The Project finalizer stays until the namespace is gone.** Deleting a project uninstalled the Helm release and dropped the finalizer at once; the uninstall is not asked to wait, and the namespace takes as long to go as everything inside it. So `kubectl delete project` returned immediately where `kubectl delete ns` used to block until the environment was gone, and scripts that relied on that ordering ran ahead of the cleanup. `Delete` now requeues while the namespace exists and removes the finalizer only once it is gone, so the Project stays in `Terminating` exactly as long as its namespace and kubectl blocks on it as before. The requeue interval grows with the time the namespace has spent terminating (a quarter of it, between 3s and 1m), and the Project carries a `NamespaceDeleted=False` condition naming when the namespace started terminating and what the namespace controller reports as remaining, written only when that changes, so `kubectl describe project` says what a stuck deletion waits for without an etcd write per poll.

**A deprecated parameter is announced before it disappears.** `allowNamespacesWithoutProjects` is deprecated and ignored, kept in the schema for one release, and removed in the next minor — at which point a ModuleConfig that still sets it fails validation and stops applying. A hook exports `d8_mc_deprecated_parameter{module, parameter}` while the key is present with either value (its own name: the documentation module's `d8_mc_deprecated` has a `{module}` label only), and `MultitenancyManagerDeprecatedConfigParameter` asks to remove it, including from the CI or GitOps process that applies the config. Same pattern as the documentation module's renamed ModuleConfig.

**The `virtual` template is kept for the platform's virtual projects.** A user Project with `projectTemplateName: virtual` reached the virtual code path of the project controller (no namespace, `Deployed`) but carried no virtual-project label, so the ClusterProjectRoleBinding fan-out and everything else keyed on the label treated it as an ordinary project and kept trying to bind into a namespace that never exists. Every CPRB in the cluster then failed its reconcile, never reached `Ready` and never updated `boundProjects`, with the error retried forever. Found on the stand during the e2e run of this PR. `Project.IsVirtual()` now recognises a virtual project by the label or by the template name and every label-only check goes through it, and the project webhook refuses the `virtual` template — and the `default` and `deckhouse` names, which the Create branch used to allow outright for any user — on user-created or user-edited projects; the controller keeps creating its two virtual projects.

**`system:masters` no longer bypasses the Project and ProjectTemplate webhooks.** Those two webhooks skipped every request from the `system:masters` group (the ProjectRoleBinding, ProjectNamespace and ClusterProjectRoleBinding webhooks never had an exclusion list and keep validating everyone, `system:sudouser` included: nothing in the platform renders those objects, so no release can deadlock on them). The exclusion list exists so that platform components never depend on the webhook backend; administrators with a master kubeconfig are not platform components, and for them the exclusion only switched validation off (a Project with a reserved name, an out-of-schema parameter or the `virtual` template went in unchecked — which is how the project above appeared). The heritage `ValidatingAdmissionPolicy` meanwhile held `system:masters` to its rules and exempted `system:sudouser`, so the two mechanisms disagreed on who the privileged identity is. Both now agree: cluster components and `system:sudouser` bypass, `system:masters` does not; an administrator who has to get past a check impersonates `system:sudouser` on purpose (`kubectl --as system:sudouser`), and the audit log shows it.

**Tests for the project name rules** the webhook already enforces: reserved `d8-`/`kube-` prefixes and the two-way collision with an existing project's additional-namespace name space.

Also from the earlier review: the `d8:use:role:*` compatibility aliases no longer declare `rules: []` (the field belongs to the aggregation controller), and the deprecated-role alert and the migration guide say what a binding to an old role name actually loses.

The aggregation-label gap is the one item here that changes who has what: a label the platform selects by was accepted from anyone on any ClusterRole. The finalizer change repairs a contract external teams built scripts on — deletion used to block until the environment was gone. The alert gives operators a release to remove a parameter that will otherwise break their ModuleConfig silently. The rest is dead rules and missing tests found while comparing the design notes with the code.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authz
type: fix
summary: A ClusterRole that carries a platform aggregation label without declaring itself a custom capability is now refused; such an object used to be merged into the platform role of that lineage for every holder in the cluster.
impact_level: default
```

```changes
section: multitenancy-manager
type: fix
summary: Deleting a Project now blocks until its namespace is actually gone, as deleting the namespace used to; the finalizer is kept while the namespace exists.
impact_level: default
```

```changes
section: multitenancy-manager
type: feature
summary: The MultitenancyManagerDeprecatedConfigParameter alert fires while a ModuleConfig still sets the deprecated and ignored allowNamespacesWithoutProjects parameter, which is removed in the next minor release.
impact: If your `multitenancy-manager` ModuleConfig sets `allowNamespacesWithoutProjects`, remove the parameter. It has no effect, and in the next minor release a ModuleConfig that still carries it will fail validation and stop applying.
impact_level: high
```

```changes
section: multitenancy-manager
type: fix
summary: A user Project on the reserved `virtual` template is refused, and projects on that template are skipped by the ClusterProjectRoleBinding fan-out, which no longer fails on a namespace that never exists.
impact_level: low
```

```changes
section: multitenancy-manager
type: fix
summary: The Project and ProjectTemplate admission webhooks no longer exempt `system:masters`; cluster components and `system:sudouser` bypass them, as they do the heritage ValidatingAdmissionPolicy.
impact: Requests from `system:masters` to Project and ProjectTemplate are now validated like everyone else's (ProjectRoleBinding, ProjectNamespace and ClusterProjectRoleBinding always were). To bypass a Project or ProjectTemplate check deliberately, impersonate `system:sudouser` (`kubectl --as system:sudouser`).
impact_level: high
```

```changes
section: user-authz
type: fix
summary: A ClusterRole that carries a platform aggregation label without being a custom capability is exported as `d8_user_authz_foreign_aggregation_label` and raises `D8UserAuthzForeignAggregationLabel`, since the webhook only refuses new writes of such a role.
impact_level: low
```


testing/rbacv2/rbacv2_aggregation_labels_test.go
```diff
@@ -0,0 +1,104 @@
...
+	string(filepath.Separator) + "templates" + string(filepath.Separator) + "rbacv2" + string(filepath.Separator),
+	string(filepath.Separator) + "templates" + string(filepath.Separator) + "rbacv2-compat" + string(filepath.Separator),
+}
+
+// TestNoPlatformAggregationLabelOutsideRBACv2 keeps the platform aggregation label out of every
+// template that is not part of the role model. A ClusterRole in a module's rbac-for-us.yaml with
+// rbac.deckhouse.io/aggregate-to-namespace-as: admin would be merged into d8:namespace:admin for
+// every holder in the cluster, and the contract test would never see it: it walks templates/rbacv2
+// only. The admission webhook refuses such an object from a user; this test refuses it from a module.
+func TestNoPlatformAggregationLabelOutsideRBACv2(t *testing.T) {
```
AlwxSin: 🔴 **1. The new aggregation-label guard test never runs in CI**

CI runs the validation suite as `go run gotest.tools/gotestsum@latest -- -tags=validation -run Validation -timeout=… ./testing/...` (`.github/ci_templates/tests.yml:43`, and the same line in `build-and-test_dev/main/pre-release.yml`; `Makefile:200` too). `-run Validation` is an unanchored regex over the test name. The two existing tests in this package match it — `TestRBACv2SensitiveCapabilityLevelsValidation` and `TestRBACv2TemplatesValidation` — but `TestNoPlatformAggregationLabelOutsideRBACv2` contains no `Validation` substring, so it is never executed.

Verified locally at this head: `go test -tags=validation -run Validation -v ./testing/rbacv2/...` runs exactly those two tests and reports `ok`; the new test only appears with `-run TestNoPlatformAggregationLabelOutsideRBACv2` (where it passes). The PR description presents this test as the control that stops a module from shipping the label outside `templates/rbacv2` — as wired, the control does not execute.

Renaming it to something ending in `Validation` (e.g. `TestRBACv2AggregationLabelPlacementValidation`) is enough.

testing/rbacv2/rbacv2_aggregation_labels_test.go
```diff
@@ -0,0 +1,104 @@
...
+			continue
+		}
+		err := filepath.WalkDir(base, func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if d.IsDir() {
+				// Module sources, test fixtures and vendored charts are not templates the platform renders.
+				switch d.Name() {
+				case "images", "testdata", "charts", "docs", "crds", "openapi", "hooks", "monitoring":
```
AlwxSin: 🟡 **2. The walker skips `charts/`, where Helm still renders templates**

`filepath.SkipDir` on `charts` is justified in the comment as "vendored charts are not templates the platform renders", but a subchart under `charts/<name>/templates/` is rendered by Helm exactly like the parent chart's templates — a ClusterRole there reaches the cluster the same way.

Verified by planting the same offending object in two places on top of this head:

* `modules/140-user-authz/templates/evil/x.yaml`
* `modules/015-admission-policy-engine/charts/constraint-templates/templates/evil/x.yaml`

The test reports one offender, not two. Today `015-admission-policy-engine` is the only module with a subchart and it is clean, so nothing is currently hidden — but the blind spot is exactly the one this test exists to close.

Side note: `crds`, `openapi`, `hooks` and `monitoring` are redundant in that switch — the `strings.Contains(path, "/templates/")` filter below already excludes them. `images`, `testdata` and `docs` are the ones doing real work.

modules/140-user-authz/webhooks/validating/cluster_roles.py
```diff
@@ -266,6 +279,21 @@ def validate(ctx: DotMap) -> Optional[str]:
...
 
+    # A label of a platform lineage on the object itself is what makes the aggregation controller pour
+    # its rules into the platform role of that lineage -- d8:namespace:admin selects exactly
+    # rbac.deckhouse.io/aggregate-to-namespace-as: admin. The API server does not care what else the
+    # object carries, so without this check any ClusterRole with such a label, whatever its name and
+    # with no framework kind at all, joins every holder of that role in the cluster. Only an object
+    # that declares itself a custom capability may carry one; the checks below then hold it to the
+    # rest of the contract.
+    platform_lineages = _platform_lineages_on(labels)
+    if platform_lineages and kind_label != "custom-capability":
```
AlwxSin, line 290: 🟡 **3. Existing offending ClusterRoles are never detected — only new writes are gated**

The webhook's `CONFIG` binds `operations: ["CREATE", "UPDATE"]`, so this check only sees writes. A ClusterRole that is already in the cluster carrying `rbac.deckhouse.io/aggregate-to-namespace-as: admin` without the `custom-capability` kind label keeps being merged into the platform role for every holder after the upgrade — no metric, no alert, no migration hook, nothing surfaces it. The PR ships a hook, an alert and a doc block for a deprecated *config parameter*, and nothing at all for the leftover privilege escalation that is the headline of the change.

The upgrade path also bites from the other side: a third-party or GitOps-managed ClusterRole that keeps the label now gets every `UPDATE` denied, so an operator learns about it from a red Argo sync rather than beforehand.

A one-shot hook exporting something like `d8_rbacv2_orphan_aggregation_label{name}`, or at minimum a migration-guide section with the `kubectl get clusterroles -l …` query to run before upgrading, would close both sides.

modules/160-multitenancy-manager/templates/admission/validation.yaml
```diff
@@ -6,15 +6,22 @@
...
   Excluding system writers makes that impossible regardless of backend health, while ordinary users are
   still validated. Mirrors the same exclusion on the cluster-objects grant webhooks.
+
+  People are not system writers: system:masters goes through these webhooks like everyone else, so a
+  Project an administrator creates is validated the same way as one a user creates. The one human
+  identity that bypasses them is system:sudouser -- the same identity the heritage
+  ValidatingAdmissionPolicy exempts -- so an administrator who has to bypass a check does it on
+  purpose, by impersonating that user, and the bypass is visible in the audit log.
 */ -}}
 {{- $excludeSystemWriters := `matchConditions:
```
AlwxSin, line 24: 🟡 **4. The changelog promises a change to five resources; only two webhooks have `matchConditions` at all**

`$excludeSystemWriters` is interpolated into exactly two webhooks in this file — `projects` and `projecttemplates`. That is 3 occurrences of the variable (one definition, two usages), identical on `main`. The `projectrolebindings`, `projectnamespaces` and `clusterprojectrolebindings` webhooks carry no `matchConditions` block at all, so they never skipped `system:masters` in the first place and nothing changes for them here.

The changelog entry, marked `impact_level: high`, tells operators:

> Requests from `system:masters` to Project, ProjectTemplate, ProjectRoleBinding, ProjectNamespace and ClusterProjectRoleBinding are now validated like everyone else's.

Three of the five are false. Either narrow the entry to Project and ProjectTemplate, or add the `matchConditions` to the other three webhooks so the entry becomes true.

modules/160-multitenancy-manager/templates/admission/validation.yaml
```diff
@@ -6,15 +6,22 @@
   a namespace) would fail and addon-operator would retry it forever, locking the module's queue.
   Excluding system writers makes that impossible regardless of backend health, while ordinary users are
   still validated. Mirrors the same exclusion on the cluster-objects grant webhooks.
```
AlwxSin: 🟡 **5. Four comments now assert the opposite of what the code does**

This line still ends with "Mirrors the same exclusion on the cluster-objects grant webhooks" — but those webhooks keep `system:masters` in their exclusion (`hooks/configure_grant_validation_webhook.go:93`, the `exclude-cluster-admins-and-nodes` condition), so after this change it is no longer mirrored.

Three more comments make the same now-stale claim:

* `images/multitenancy-manager/src/internal/webhook/project/hook.go:60-61` — "system:masters never reaches this handler at all: the webhook's matchConditions skip it at the API server (templates/admission/validation.yaml)". Falsified by this very PR, three lines above the new `virtual`-template check.
* `images/multitenancy-manager/src/internal/webhooks/is_granted.go:44-45` and `:124-125` — "In-cluster, matchConditions already skip system:masters before any handler runs."

Keeping the exemption on the grant webhooks is defensible on its own — they intercept every grantable cluster-resource write into project namespaces, a much larger deadlock surface than Project CRUD. The defect is that four comments now document a rule the code no longer follows, and `hook.go:60` in particular sits directly above logic that depends on the opposite being true.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/project/manager.go
```diff
@@ -377,6 +382,26 @@ func (m *Manager) Delete(ctx context.Context, project *v1alpha3.Project) (ctrl.R
...
+	// exactly as long as its namespace does, which is what the old contract promised.
+	namespace := new(corev1.Namespace)
+	switch err := m.client.Get(ctx, client.ObjectKey{Name: project.Name}, namespace); {
+	case err == nil:
+		if namespace.DeletionTimestamp.IsZero() {
+			// Not even terminating: the uninstall did not reach it (a foreign release owns it, or the
+			// delete was dropped). Retrying the uninstall is what the next reconcile does.
+			m.logger.Info("the project namespace is not terminating yet, waiting", "project", project.Name)
+		}
+		return ctrl.Result{RequeueAfter: namespaceDeletionPoll}, nil
```
AlwxSin: 🟡 **6. `Delete` requeues at a flat 3s forever, silently, and re-runs the uninstall each time**

`ctrl.Result{RequeueAfter: namespaceDeletionPoll}` is an unconditional fixed-interval requeue, not the workqueue's rate-limited backoff. A namespace stuck in `Terminating` — a dangling `APIService`, a CR finalizer, or a foreign Helm release that owns the namespace so this module's uninstall never touches it (the case the comment right above names explicitly) — means the Project can never be deleted, at roughly 20 reconciles per minute per project, forever. Each pass re-runs `deleteStandardFields` plus a Helm storage read; `helmClient.Delete` is safe to repeat (`uninstall.IgnoreNotFound = true`) but not free.

It is also silent in exactly that case: the `"the project namespace is not terminating yet, waiting"` log only fires when `namespace.DeletionTimestamp.IsZero()`, i.e. never once the namespace is actually stuck terminating.

Suggest a growing interval (or returning the error path so the controller-runtime backoff applies) plus a condition on `Project.status`, so `kubectl describe project` says what the deletion is waiting for.

modules/160-multitenancy-manager/hooks/alert_deprecated_config.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+	},
+}, alertOnDeprecatedConfigParameters)
+
+// filterDeprecatedConfigParameters keeps only what the alert needs: the names of the deprecated
+// parameters this ModuleConfig still sets. Presence is what matters, not the value.
+func filterDeprecatedConfigParameters(obj *unstructured.Unstructured) (go_hook.FilterResult, error) {
+	// The name selector above already narrows the watch; the check is repeated here so the filter
+	// stays correct on its own, and a same-named key in another module's config is never counted.
+	if obj.GetName() != "multitenancy-manager" {
```
AlwxSin: 🔵 **7. Dead guard in the hook filter**

`NameSelector.MatchNames: []string{"multitenancy-manager"}` already guarantees the object's name, so this branch is unreachable. It is also the only path that returns a nil `FilterResult`, a shape `SnapshotIter[[]string]` in the handler never actually sees — so the "stays correct on its own" claim in the comment is untested.

The `otherModule` test case passes because the name selector drops that object before the hook runs, not because of this check.

modules/160-multitenancy-manager/hooks/alert_deprecated_config.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// allowNamespacesWithoutProjects is deprecated and ignored: every user namespace outside a project
+// is adopted into a project of its own, whatever the parameter says. The schema keeps it for one
+// release so an existing ModuleConfig still applies; the next minor release drops it from the
+// schema, and a ModuleConfig that still carries it would then fail validation and stop applying.
+// This hook is the warning in between: as long as the key is present in spec.settings -- with any
+// value, since neither value does anything -- a metric is exported and an alert asks to remove it.
+const (
+	deprecatedConfigMetric      = "d8_mc_deprecated"
```
AlwxSin: 🔵 **8. `d8_mc_deprecated` now carries two different label shapes**

`modules/810-documentation/hooks/deprecated_mc.go:56` exports the same metric name with `{module}` only; this hook adds `{module, parameter}`. Both alerts filter by `module`, so nothing breaks today, but a shared metric with an inconsistent label set is a trap for whoever writes the next query — `max by (parameter) (d8_mc_deprecated)` yields `parameter=""` for the documentation series.

Either add `parameter` to the documentation hook as well, or use a distinct metric name here.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/webhook/project/hook.go
```diff
@@ -62,6 +62,16 @@ func (v *validator) Handle(ctx context.Context, req admission.Request) admission
...
+	// keyed on the virtual-project label, so the PRB/CPRB fan-out kept failing on a namespace that
+	// never exists and no ClusterProjectRoleBinding in the cluster could reach Ready.
+	if project.Spec.ProjectTemplateName == projectmanager.VirtualTemplate && !privileged &&
+		project.Name != projectmanager.DefaultProjectName && project.Name != projectmanager.DeckhouseProjectName {
+		return admission.Denied(fmt.Sprintf("the %q project template is reserved for the platform's virtual projects", projectmanager.VirtualTemplate))
+	}
+
 	if req.Operation == admissionv1.Create {
 		// pass virtual projects
 		if project.Name == projectmanager.DefaultProjectName || project.Name == projectmanager.DeckhouseProjectName {
```
AlwxSin: 🔵 **9. The `virtual`-template guard is bypassable by project name (pre-existing)**

The new check above exempts `default` and `deckhouse` by name, and this `Create` branch returns `admission.Allowed("")` unconditionally for those same two names, before every other check, for any user. So "the `virtual` template is reserved for the platform's virtual projects" holds exactly as far as "nobody creates a project named `default` or `deckhouse`".

The blanket allow predates this PR, but the new guard inherits its hole, and `TestHandle_VirtualTemplateIsReserved` only exercises those two names from the controller service account. Gating them on `privileged` instead of on the name would close it.

testing/rbacv2/rbacv2_aggregation_labels_test.go
Jabejixo: Done in 4b1e63fb34: renamed to `TestRBACv2AggregationLabelPlacementValidation`; `go test -tags=validation -run Validation ./testing/rbacv2/` runs it now.

testing/rbacv2/rbacv2_aggregation_labels_test.go
Jabejixo: Done in 4b1e63fb34: the walker no longer skips `charts/`, and the redundant skips (`crds`, `openapi`, `hooks`, `monitoring`) are gone — only `images`, `testdata`, `docs` remain.

modules/140-user-authz/webhooks/validating/cluster_roles.py
Jabejixo: Done in 4b1e63fb34: a hook exports every non-platform ClusterRole carrying the label without the `custom-capability` kind as `d8_user_authz_foreign_aggregation_label{name, lineage, kind}`, `D8UserAuthzForeignAggregationLabel` fires on it with the fix-or-relabel steps, and the migration guide (#23180) gets the pre-upgrade query.

modules/160-multitenancy-manager/templates/admission/validation.yaml
Jabejixo: Reworked after finding 10: the exclusions stay on the Project and ProjectTemplate webhooks only (the three additions are reverted in db94551053), and the changelog entry and `impact` are narrowed to those two resources.

modules/160-multitenancy-manager/templates/admission/validation.yaml
Jabejixo: Done in 9ce4103c0d: the header names the grant webhooks as the ones that still skip `system:masters` (deliberately — larger deadlock surface), `hook.go` says masters do reach the handler, and the two `is_granted.go` notes say which webhooks' matchConditions they mean.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/project/manager.go
Jabejixo: Done in cdcf3e5d22: the interval grows with the time the namespace has spent terminating (a quarter of it, clamped to [3s, 1m]), the Project gets a `NamespaceDeleted=False` condition with the elapsed time and the namespace controller's remaining-content report, and the terminating case is logged with the next check time.

modules/160-multitenancy-manager/hooks/alert_deprecated_config.go
Jabejixo: Done in cdcf3e5d22: the guard and the `otherModule` case are gone — the framework ignores the NameSelector, so the case only ever passed because of the guard.

modules/160-multitenancy-manager/hooks/alert_deprecated_config.go
Jabejixo: Done in cdcf3e5d22: the metric is `d8_mc_deprecated_parameter{module, parameter}` now; rule and `deckhouse-alerts.yml` updated.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/webhook/project/hook.go
Jabejixo: Done in 9ce4103c0d: both the two names and the template are refused unless the request is the controller's; the early allow is gated on `privileged`, and the test covers a user taking the names with an ordinary template.

modules/160-multitenancy-manager/templates/admission/validation.yaml
```diff
@@ -65,6 +74,7 @@ webhooks:
     admissionReviewVersions:
       - v1
     matchPolicy: Equivalent
+    {{- $excludeSystemWriters | nindent 4 }}
```
AlwxSin: 🟡 **10. This exempts platform service accounts from ProjectRoleBinding / ProjectNamespace / CPRB validation, which they were never exempt from before**

Thanks for closing the five-vs-two gap — but of the two ways out of that finding (narrow the changelog, or extend `matchConditions`), this one has a side effect worth a second look. Same for the two sibling additions at lines 101 and 126.

`rolebindingwebhook.Validate` (`internal/webhook/rolebinding/validator.go:83`) treats exactly two identities as privileged: the multitenancy controller SA and the deckhouse SA. Everything else went through the full PRB/PN/CPRB validation until now — the managed-by check at `validator.go:86` and the privilege-escalation check at `validator.go:155`. With `$excludeSystemWriters` applied here, the API server no longer calls these three webhooks at all for:

* `system:sudouser`
* `system:serviceaccount:d8-user-authz:controller`
* every SA in `system:serviceaccounts:d8-system`, `system:serviceaccounts:kube-system`, `system:serviceaccounts:d8-user-authz`
* `system:nodes`

And the anti-deadlock rationale that justifies the exclusion in the first place does not apply to these three resources: nothing in the repo renders them.

```
git grep -lE 'kind: (ProjectRoleBinding|ClusterProjectRoleBinding|ProjectNamespace)' -- '*/templates/*' '*/hooks/*'
# (empty)
```

No module's Helm release applies a ProjectRoleBinding, a ClusterProjectRoleBinding or a ProjectNamespace, so there is no admission call that could fail and lock addon-operator's queue — unlike Project/ProjectTemplate, where the argument holds. The exclusion buys nothing here and drops a real check.

The practical risk is modest (`system:sudouser` is cluster-admin anyway, and a tenant cannot place a ServiceAccount in `kube-system` or `d8-system`), but it is a silent widening of the trust boundary, and the changelog entry mentions only that `system:masters` is now validated. Either revert the three additions and narrow the changelog to Project and ProjectTemplate, or keep them and spell out in `impact` who no longer passes through these webhooks.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/project/manager.go
```diff
@@ -377,6 +411,42 @@ func (m *Manager) Delete(ctx context.Context, project *v1alpha3.Project) (ctrl.R
...
+		}
+		// Terminating: Kubernetes is purging the contents, or something in there holds a finalizer
+		// and the namespace never finishes. Say so where "kubectl describe project" shows it, and
+		// check less and less often the longer it takes.
+		terminatingFor := time.Since(namespace.DeletionTimestamp.Time)
+		summary := namespaceTerminationSummary(namespace)
+		project.SetConditionFalse(v1alpha3.ProjectConditionNamespaceDeleted,
+			fmt.Sprintf("waiting for the '%s' namespace to be deleted (terminating for %s); %s",
+				project.Name, terminatingFor.Round(time.Second), summary))
+		if err := m.updateProjectStatus(ctx, project); err != nil {
```
AlwxSin: 🔵 **11. The `NamespaceDeleted` condition writes to etcd on every poll, forever**

The backoff and the status condition are a good answer to the previous finding; one detail follows from them.

`SetConditionFalse` (`apis/deckhouse.io/v1alpha3/project.go:427`) unconditionally sets `LastProbeTime = metav1.Now()`, and `updateProjectStatus` (`internal/manager/project/utils.go:162`) unconditionally issues `Status().Update()` — neither compares against the current state. On top of that the message embeds a live `terminatingFor`, which changes on every pass. So a namespace stuck in `Terminating` now costs one etcd write per poll, indefinitely: the backoff brought the reconcile rate down from ~20/min to 1/min, but each of those ticks now carries a write that did not exist before.

Nothing breaks — the controller's own predicate keys on `generation`, so there is no self-retrigger, and `ProjectFanoutChanged` already ignores status writes. It is just cheap to avoid: update the status only when `Status`/`Reason` actually change, or round the duration in the message to the poll interval so an unchanged situation produces an identical message and no new revision.

modules/160-multitenancy-manager/templates/admission/validation.yaml
Jabejixo: Agreed — reverted in db94551053: the exclusion stays on the Project and ProjectTemplate webhooks only, the changelog entry and `impact` are narrowed to those two, and the docs (#23180) now say the PRB/PN/CPRB webhooks skip nobody.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/manager/project/manager.go
Jabejixo: Done in db94551053: the message names the moment the namespace started terminating instead of the elapsed time, and the status is written only when the condition actually changes (a second poll in the same state leaves the resourceVersion alone — covered by the test).

AlwxSin (approved):

z9r5 (approved):

PR #23127, opened 2026-09-17
[user-authz] RBACv2 migration tooling: let a module serve both role models from one branch

Jabejixo:
## Description

The RBACv2 migration script (`rbacv2-migrate-module.sh`) used to rewrite the legacy
`manage`/`use` object away. It now keeps it and puts the migrated object beside it, both behind a
version gate, so one branch of a module serves clusters on either side of DKP 1.78:

```
{{- if eq (include "<module>.rbacv2_new_scheme" .) "true" }}
<the migrated object>
{{- else }}
<the legacy object, unchanged>
{{- end }}
```

The gate lands in `templates/_rbacv2_compat.tpl`, written once per module. It reads
`global.deckhouseVersion` — the content of `/deckhouse/version`, put into the global values by a
startup hook — and answers `true` from 1.78 on. Four details in it are load-bearing:

- that value is `dev` on a development build and `unknown` when the version file is missing, and
  `semverCompare` raises on both — a raise takes down the render of the whole module, so anything
  that is not a version has to answer without comparing;
- it answers `true`, and that is a choice about the direction of failure, not about which branch is
  likelier. The version file holds `CI_COMMIT_TAG`, so a build off any branch reports `dev`, and a
  dev stand built from `release-1.77` is indistinguishable from one built from `main`. Only one of
  the two mistakes is dangerous: the new object on a 1.77 cluster is inert, since nothing there
  selects `aggregate-to-namespace-as`, and the module's permissions are merely missing; the legacy
  object on a 1.78 cluster is still aggregated by `d8:subsystem:kubernetes:<level>`, which is granted
  through a `ClusterRoleBinding`, so rules meant for one namespace become cluster-wide. The helper
  carries the one line a dev stand below 1.78 flips to get the legacy objects there;
- the comparison is on major and minor only, because semver orders `1.78.0-rc.1` below `1.78.0`
  and a release candidate of 1.78 would otherwise be served the legacy object;
- the regex uses `[.]`, since a backslash escape inside a Go template string literal is a parse
  error;
- `.Values.global` is parenthesized, since a chart rendered with no global values at all otherwise
  fails on a nil pointer instead of falling back to `dev`.

`--replace` restores the previous behaviour — the new object alone — which is what an in-tree
module wants, since it ships with the platform it targets. The script stays idempotent: a file
that already carries the gate is reported as untouched, and a helper that was deleted by hand is
written again.

The agent prompt gains a matching step (3b), an inventory check for an existing gate and for a
`module.yaml` that already requires 1.78, and an acceptance section that renders the chart at
1.77.5, 1.78.0 and `dev` instead of once. Both migration guides describe the same shape, say where
the value comes from, and point out the one thing outside `templates/` the gate affects: a template
test that looks a ClusterRole up by name now finds it only in the branch its render selects, so it
has to pin the version it means.

An external module is installed on clusters of more than one platform version, and the two RBACv2
role models do not see each other: an object of one model is invisible to the other. With a
rewrite-in-place migration such a module had to pick a side — migrate and lose its permissions on
everything below 1.78, or stay on the legacy scheme and lose them from 1.78 on. The gate lets one
artifact serve both.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

**Against `user-authn` as it stood immediately before the RBACv2 merge (`e122f18c1a^`)**, which
still shipped the four legacy `templates/rbacv2/{manage,use}/{view,edit}.yaml`:

- both branches the script writes are **byte-identical** to what `main` shipped on either side of
  that commit — the `else` branch to the pre-merge file, the `if` branch to the migrated one;
- `--replace` reproduces the in-tree result exactly, and writes no gate;
- `--dry-run` writes nothing; a second run reports all four files as untouched; a hand-deleted
  helper is written again on the next run;
- the module's own helm harness (`modules/150-user-authn/template_tests`, ten added specs) renders
  the legacy objects at `1.77.5` and `v1.77.11` and the new ones at `v1.78.0`, `1.78.0-rc.1`,
  `1.80.2`, `dev`, `unknown` and with the version unset, with the labels, the capability marker,
  the i18n annotations and the rules checked on both sides.

The one thing that does change for a migrated module is its own tests: `cluster_roles_test.go`
asserts the legacy names and fails once the gate renders the new scheme by default. That is the
documented consequence, and the reason the guides now tell a module to pin the version in such a
test.

**Against the twelve external modules installed in a live cluster** — `extended-monitoring`,
`ingress-nginx`, `log-shipper`, `monitoring-applications`, `monitoring-custom`,
`monitoring-kubernetes`, `monitoring-ping`, `operator-prometheus`, `pod-reloader`, `prometheus`,
`prometheus-metrics-adapter`, `upmeter`, every one of them declaring `requirements.deckhouse`
between `>= 1.68` and `>= 1.76`, so every one of them needs both models. Their real
`templates/rbacv2` trees migrate and render correctly: only legacy objects at `1.77.5`, only the
new ones at `v1.78.0`, no render error on `dev`.

The values plumbing was checked on that cluster rather than assumed: `global.deckhouseVersion` is
present in the global values and reaches an external module's render (`upmeter` already builds its
user-agent from it). A module delivered as a `ModulePackage` gets it in the same place; only an
`Application` package reads global values as `.Platform.<path>`, and RBACv2 capabilities are not
shipped that way.

## Changelog entries

```changes
section: user-authz
type: chore
summary: The RBACv2 migration script keeps the legacy role objects behind a version gate, so a module can serve both role models from one branch.
impact_level: low
```


modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
```diff
@@ -120,9 +133,61 @@ TEXTS = {
...
+
+def gate_open(module):
+    return '{{- if eq (include "%s.rbacv2_new_scheme" .) "true" }}\n' % module
+
+
+def chart_root(path):
+    """The directory holding templates/ for this file, i.e. where the helper belongs."""
+    marker = os.sep + "templates" + os.sep
+    index = path.find(marker)
+    return path[:index] if index != -1 else os.path.dirname(path)
```
AlwxSin: 🟡 **Single-file invocation crashes after the templates are already rewritten**

`chart_root()` searches for `os.sep + "templates" + os.sep`, so a bare relative path has no leading separator, `find()` returns -1, and the `os.path.dirname(path)` fallback points at `templates/rbacv2/use`:

```
$ rbacv2-migrate-module.sh templates/rbacv2/use/*.yaml
FileNotFoundError: 'templates/rbacv2/use/templates/_rbacv2_compat.tpl'   # exit 1
```

The per-file loop writes the YAML before the helper loop runs, so the templates already carry the gate while the helper they call does not exist — and a missing `define` makes the render fail outright (`no template "<module>.rbacv2_new_scheme" associated with template ...`), it does not fall back to the legacy branch.

Both the usage line and `--help` document a single template file as valid input, and a glob over `templates/rbacv2/use/*.yaml` is the natural way to pass several. Rerunning from the chart root does recover (the skip branch at line 302 re-registers the chart), but only if someone notices the traceback.

Minimal fix: `path = os.path.join(".", path)` at the top of `chart_root`, or run it on `os.path.abspath(path)`.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
```diff
@@ -209,6 +274,18 @@ def migrate(path, text):
     return text
 
 
+def wrap(path, legacy, new, module):
```
AlwxSin, line 303: 🟡 **A file with two legacy objects leaks an unmigrated one into the new-scheme branch**

`migrate()` rewrites the name with `count=1`, but `wrap()` wraps the whole file text. Two concatenated legacy objects produce this inside the `{{- if }}` branch — the one a 1.78 cluster reads:

```yaml
  name: d8:namespace-capability:service-with-healthchecks:view   # migrated
---
  {{- include "helm_lib_module_labels" (list . (dict "rbac.deckhouse.io/kind" "use")) | nindent 2 }}
  name: d8:use:capability:module:service-with-healthchecks:edit  # untouched legacy
```

The label-stripping `re.sub` calls in `migrate()` carry no `count=1`, so the second object also loses its `aggregate-to-kubernetes-as`. On 1.78 the module therefore ships a dead legacy-named ClusterRole and silently drops that permission entirely. Nothing warns, because `aggregation_labels()` found labels on the first object.

The half-migration of multi-object files predates this change; what is new is that the result now lands in the branch the 1.78 cluster reads, while `RBACV2_MODULE_MIGRATION_PROMPT.md` asserts the opposite ("do not leave a file where both branches carry the same object") and the guides describe the legacy object as "what the cluster below 1.78 reads". The prompt tells a human to stop on "several objects in one file"; the script should too — `if len(LEGACY_NAME.findall(before)) > 1: warn(...)` here.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
```diff
@@ -222,20 +299,48 @@ def unified_diff(path, before, after):
...
     migrated.append(path)
     if DRY_RUN:
         sys.stdout.write(unified_diff(path, before, after))
     else:
         with open(path, "w", encoding="utf-8") as handle:
             handle.write(after)
 
+gate_files = []
+for root, module in sorted(gates):
+    gate_path = os.path.join(root, "templates", "_rbacv2_compat.tpl")
```
AlwxSin, line 359: 🟡 **Two module names in one chart silently clobber the helper**

`gates` is keyed on `(chart_root, module)`, but every entry maps to the same `templates/_rbacv2_compat.tpl`. With two module names under one chart the second write wins, the first module's `define` is gone, and its templates then reference an undefined template — which takes the whole chart's render down, not just that file.

Reproduced on a chart holding `other-module` and `service-with-healthchecks` capabilities:

```
version gate: 2
  ./templates/_rbacv2_compat.tpl
  ./templates/_rbacv2_compat.tpl      # same path twice - the only hint

$ grep define templates/_rbacv2_compat.tpl
{{- define "service-with-healthchecks.rbacv2_new_scheme" -}}   # other-module's is gone
```

The duplicated output line is the entire diagnostic. Either warn when one chart root maps to more than one module, or append the defines instead of overwriting the file.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
```diff
@@ -222,20 +299,48 @@ def unified_diff(path, before, after):
...
+    body = GATE_TEMPLATE % {"module": module}
+    if os.path.exists(gate_path):
+        with open(gate_path, encoding="utf-8") as handle:
+            if handle.read() == body:
```
AlwxSin: 🟡 **The script reverts the hand-edit its own comment prescribes**

The helper is rewritten whenever its content differs from `GATE_TEMPLATE`. The one documented hand-edit is flipping `true` to `false` for a dev stand below 1.78 — stated in the helper's own comment, in the comment above `GATE_TEMPLATE`, and in both migration guides. The next run undoes it:

```
before: 14:    false
run:    version gate: 1
          ./templates/_rbacv2_compat.tpl
after:  14:    true
```

It is reported identically to a fresh write, so nothing tells the user an edit was discarded. Either leave an existing helper that differs only in that line alone, or label it "overwritten" in the report.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
```diff
@@ -120,9 +133,61 @@ TEXTS = {
...
+#     namespace become cluster-wide.
+# One direction loses access, the other hands it out. A dev stand on a release branch below 1.78
+# that needs the legacy objects flips the "true" below to "false" in that build.
+GATE_TEMPLATE = """{{- /*
+  Generated by rbacv2-migrate-module.sh. Tells the RBACv2 templates of this module which role model
+  the cluster they are rendered for speaks: the model of DKP 1.78 and later, or the legacy
+  manage/use one. Remove it once the module no longer supports clusters below 1.78.
+*/ -}}
+{{- define "%(module)s.rbacv2_new_scheme" -}}
+  {{- $raw := (.Values.global).deckhouseVersion | default "dev" -}}
```
AlwxSin: 🔵 **`regexFind` on a non-string kills the whole render**

`deckhouseVersion: 1.78` unquoted in a values file is a YAML float, and `default "dev"` passes it straight through:

```
error calling include: executing "<module>.rbacv2_new_scheme" at <$raw>:
wrong type for value; expected string; got float64
```

`--set global.deckhouseVersion=1.78` is safe (helm's strvals keeps it a string); a values file is not. The guides tell template-test authors to pin the version explicitly, which is exactly where this lands.

`| toString` after `default "dev"` closes it — the same token would need to go into the two guides and the agent prompt, which all quote the helper verbatim.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
```diff
@@ -222,20 +299,48 @@ def unified_diff(path, before, after):
 for path in sys.argv[1:]:
     with open(path, encoding="utf-8") as handle:
         before = handle.read()
+    if GATE_MARKER in before:
+        skipped.append(f"{path}: already carries the version gate")
+        # Still remember the chart: the helper the gate calls may have been deleted, and a file
+        # that asks for a template nobody defines renders empty instead of failing.
```
AlwxSin: 🔵 **The comment's rationale is inverted**

> a file that asks for a template nobody defines renders empty instead of failing

It fails: `error calling include: template: no template "x" associated with template "y"`.

The behaviour this justifies — re-registering the chart on a skipped file so a deleted helper gets rewritten — is right, and in fact more necessary than the comment claims. Only the stated reason is wrong, and it is the kind of line a future reader will take on faith.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
```diff
@@ -222,20 +299,48 @@ def unified_diff(path, before, after):
 for path in sys.argv[1:]:
     with open(path, encoding="utf-8") as handle:
         before = handle.read()
+    if GATE_MARKER in before:
```
AlwxSin, line 328: 🔵 **`--replace` cannot undo a gate**

This check runs before the `DUAL` branch, so an already-gated file is skipped whatever the flag. `--replace` over a migrated tree reports every file as untouched and changes nothing; going from dual back to new-only is a manual edit.

The guides mention removing the gate by hand only in the "module drops support for clusters below 1.78" paragraph, while `--replace`'s help text (line 44 and the `USAGE` heredoc) reads as though it would do the job. One sentence in the help text would cover it.

modules/140-user-authz/docs/internal/RBACV2_MODULE_MIGRATION.md
```diff
@@ -158,13 +158,64 @@ finds, and reports what it did not dare to touch. It replays the platform migrat
...
+- `.Values.global` is parenthesized, so a chart rendered with no global values at all falls back to
+  `dev` instead of failing on a nil pointer.
+
+The module's own template tests need attention, because a test that looks up a ClusterRole by name
+finds it only in the branch its render selects, and a harness that leaves `global.deckhouseVersion`
+unset renders the new scheme. Set the version explicitly in such a test and cover both branches; that
+is also the cheapest place to prove the gate works.
+
+A module whose `module.yaml` already requires 1.78 or newer does not need the legacy branch:
+`--replace` writes the new object alone, which is also what in-tree modules use, since they ship with
```
AlwxSin, line 223: 🔵 **In-tree modules must use `--replace`, not merely prefer it**

`testing/rbacv2/rbacv2_templates_validation_test.go` parses every file under `templates/rbacv2/**` raw — `yaml.Unmarshal` for plain files, `helmNameRe`/`plainLabelRe` regexes for `helm_lib_module_labels` ones. A gated file breaks both: it is not valid YAML, and `helmNameRe` matches two names. That is exactly why the in-tree compatibility layer lives in `templates/rbacv2-compat/` rather than `templates/rbacv2/`, as its own header comment explains.

So for an in-tree module `--replace` is a CI requirement, not a preference. Worth stating here and in the RU guide.

External modules are unaffected — dmt renders the chart through helm's engine (`internal/module/controller.go:RunRender` -> `helm.Renderer.RenderChartFromRawValues`) and lints the resulting objects, so it only ever sees one branch.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
Jabejixo: Confirmed and fixed: `chart_root` now prefixes a bare relative path with `./` so the marker matches, and nothing is written until every helper is in place — the file writes were moved after the helper loop, so a run that stops can no longer leave a gated template calling a define that does not exist. Verified: `rbacv2-migrate-module.sh templates/rbacv2/use/view.yaml` from the chart root now writes `./templates/_rbacv2_compat.tpl` and the chart renders.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
Jabejixo: Confirmed, and worse than reported — with the same level on both objects there was no warning at all, only the level-disagreement one fired. The script now refuses such a file outright (`len(LEGACY_NAME.findall(before)) > 1`): it is left untouched, listed under "untouched" and reported under "needs a human" with the instruction to split it one ClusterRole per file.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
Jabejixo: Confirmed — the second write won and `helm template` failed for the whole chart. `gates` is now a map from chart root to the set of module names, and the helper is written once per chart carrying a define per module. On your two-module fixture both defines are present and the chart renders.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
Jabejixo: Confirmed. A rerun now compares the existing helper with the generated one after masking the dev answer line, and leaves it alone when that is the only difference; it is reported separately as "version gate kept as edited" rather than silently. Verified: flipped to `false`, reran, the edit survived and `dev` renders the legacy objects.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
Jabejixo: Confirmed — `deckhouseVersion: 1.78` in a values file is a float and the render dies. Added `| toString` after `default "dev"`, and the same token went into the helper the agent prompt quotes verbatim plus a bullet in both guides. Verified with an unquoted 1.78 and 1.77 in a values file: new and legacy objects respectively, no error.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
Jabejixo: Right, it fails. Comment corrected to say the render of the whole chart fails, which is what makes the re-registration necessary.

modules/140-user-authz/docs/internal/rbacv2-migrate-module.sh
Jabejixo: Confirmed: the gate check runs before the DUAL branch, so `--replace` reports every gated file as untouched. Documented in both the header comment and the `--help` text — going back to the new scheme alone is a hand edit of the else branch and the helper.

modules/140-user-authz/docs/internal/RBACV2_MODULE_MIGRATION.md
Jabejixo: Confirmed empirically: gating an in-tree file makes `go test -tags validation ./testing/rbacv2/` fail with `label rbac.deckhouse.io/kind must be "role" or "capability", got "use"` — `plainLabelRe` merges the labels of both branches. Stated as a requirement in both guides and in the script's `--help`, together with your note that dmt renders through helm and so only ever sees one branch.

AlwxSin (approved):

z9r5 (approved):

PR #23131, opened 2026-09-17
[user-authn] Keep an authenticator of a dead namespace out of the release

Jabejixo:
## Description

`get_dex_authenticator_crds.go` copied every DexAuthenticator of its snapshot into
`userAuthn.internal.dexAuthenticatorCRDs`, and the chart renders a Deployment, a Service, a Secret,
an Ingress and a PodDisruptionBudget into the namespace each of them names. Nothing checked that the
namespace can still take objects.

The hook now also watches namespaces, keeping only the name and whether the namespace is
terminating, and skips an authenticator whose namespace accepts no objects. A skipped authenticator
is exported as `d8_user_authn_dex_authenticator_skipped{name, namespace}` (an expiring gauge, the
module's usual pattern) and `D8DexAuthenticatorNamespaceUnusable` fires on it after 10 minutes with
the steps to finish the namespace deletion; the log line about it is a detail at Info level, since
the hook runs on every event and would otherwise repeat a warning forever.

An empty namespace snapshot means "do not filter" rather than "nothing is alive". In a cluster the
snapshot is never empty — shell-operator lists every binding synchronously before the
Synchronization run, and a cluster cannot lose its last namespace afterwards — so the guard is
insurance for a broken assumption, not handling of a runtime state: should the snapshot ever come
empty, the failure mode is "no filtering" instead of "delete the objects of every healthy
namespace". The pre-existing specs declare the namespaces they use, so they exercise the filtering
branch the way a cluster does; one entry pins the guard.

No values contract changes, no RBAC changes, no user-facing configuration.

The API server refuses every create in a namespace that is gone or terminating, so a single such
authenticator fails the whole module release:

```
create resource: server-side apply resource "Ingress/x-dex-authenticator (namespace=y)":
ingresses.networking.k8s.io "x-dex-authenticator" is forbidden: unable to create new content in
namespace y because it is being terminated
```

The failing ModuleRun then sits at the head of the main queue and retries with a growing backoff,
which stops every other module from being applied. The release is all or nothing, so one unusable
namespace is enough to stall the queue.

It clears itself only when the authenticator leaves the snapshot, which is exactly what does not
happen while the namespace is terminating: a finalizer on the authenticator holds both the object
and the namespace indefinitely. (A namespace never disappears with objects still in it; the only
"namespace gone, authenticator present" state is a momentary one between the two informers, which
the hook handles the same way.) Clusters that create and destroy short-lived
environments hit this: the namespace disappears, the values keep the authenticator, and the queue
stops until someone recreates the namespace by hand.

Reproduced on a stand before the change: a DexAuthenticator held by a finalizer in a terminating
namespace put `ModuleRun:main:user-authn` at the head of the main queue with a growing failure
count, another module queued behind it, and the queue only moved once the object was released.

The remaining race, where the namespace dies between this hook and the release, now costs one failed
module run instead of a stalled queue: deleting a namespace wakes the hook, the values lose the
authenticator, and the next run succeeds.

Not necessarily.

## Checklist
- [ ] e2e tests passed.
- [x] Documentation updated according to the changes — the new alert is in `deckhouse-alerts.yml` (generated); no other documented behaviour changes.
- [ ] Changes were tested in the Kubernetes cluster manually — the failure was reproduced on a
  stand; verifying the fix needs an image of this branch.

## Changelog entries

```changes
section: user-authn
type: fix
summary: A DexAuthenticator whose namespace is terminating no longer stalls the Deckhouse queue; the `D8DexAuthenticatorNamespaceUnusable` alert points at it.
impact: |
  Such an authenticator is left out of the module values, so the user-authn release stops failing on
  objects that the API server cannot create, and the main queue no longer stops behind the retrying
  module. The authenticator is exported as `d8_user_authn_dex_authenticator_skipped` and the
  `D8DexAuthenticatorNamespaceUnusable` alert fires after 10 minutes with the steps to finish the
  namespace deletion. The authenticator is deleted together with its namespace; to keep it, recreate
  it in a live namespace.
```


modules/150-user-authn/hooks/get_dex_authenticator_crds_test.go
```diff
@@ -500,6 +500,94 @@ spec:
 			inputDexAuthenticatorClientSecret{omitExisting: true},
 		),
 	)
+
+	// An authenticator whose namespace cannot take objects must not reach the values. The API
+	// server refuses every create in a namespace that is gone or terminating, so one such entry
+	// fails the whole module release, and the failing ModuleRun then blocks the main queue.
+	DescribeTable("A DexAuthenticator whose namespace cannot take objects",
```
AlwxSin, line 540: 🟡 **The filtering path is covered by two table entries; every other test runs the fallback branch**

None of the 118 pre-existing specs in this file declare a `Namespace` object, so all of them execute the `namespacesKnown == false` branch of `get_dex_authenticator_crds.go:230` and never reach the lookup they now depend on. Verified by mutation: changing that early return to `return nil, true, nil` breaks 15 pre-existing specs that were never written with namespaces in mind.

So the branch that actually runs in a cluster is exercised only by the two new entries here, while the safety branch is pinned by accident across the whole suite. Anyone who later touches the guard gets a wall of unrelated red and no obvious signal about what broke.

Adding a `Namespace` with `status.phase: Active` to the shared `f.KubeStateSet` fixtures of the existing contexts costs a few lines and makes the default fixture look like a real cluster.

modules/150-user-authn/hooks/get_dex_authenticator_crds.go
```diff
@@ -187,6 +212,40 @@ func sharedKubernetesClientSecret(input *go_hook.HookInput) (string, error) {
...
+// main queue and retries, which stops every other module from being applied.
+//
+// The second return value is false when the snapshot carries no namespace at all. A cluster always
+// has namespaces, so that means the snapshot is not populated rather than that nothing is alive,
+// and filtering against it would drop every authenticator in the cluster and delete the objects of
+// namespaces that are perfectly healthy.
+func renderableNamespaces(input *go_hook.HookInput) (map[string]struct{}, bool, error) {
+	snapshots := input.Snapshots.Get("namespaces")
+	if len(snapshots) == 0 {
+		return nil, false, nil
```
AlwxSin, line 241: 🔵 **The empty-snapshot guard cannot trigger in a live cluster — the comment above oversells it**

`EnableKubernetesBindings` loops over every binding and calls `AddMonitor` → `CreateInformers` → `createSharedInformer` → `loadExistedObjects`, i.e. the initial List runs synchronously per binding; only after the whole loop are the Synchronization execution infos returned and turned into tasks (shell-operator v1.20.5, `pkg/hook/controller/kubernetes_bindings_controller.go:99-129` and `pkg/kube_events_manager/resource_informer.go:152`). The `namespaces` snapshot is therefore populated before this hook's first execution, and afterwards it can only shrink through real DELETE events — a cluster cannot reach zero namespaces.

Keeping the guard as cheap insurance is fine. But the doc comment (and the matching paragraph in the PR description) present it as protection against a runtime state that does not occur; its only live consumer is the test suite. Worth trimming the rationale to what it actually is, so the next reader does not go looking for the cluster condition it describes.

modules/150-user-authn/hooks/get_dex_authenticator_crds.go
```diff
@@ -220,6 +284,17 @@ func getDexAuthenticator(_ context.Context, input *go_hook.HookInput) error {
 			return fmt.Errorf("cannot convert dex authenticaor: failed to iterate over 'authenticators' snapshot: %w", err)
 		}
 
+		if _, alive := renderable[dexAuthenticator.Namespace]; namespacesKnown && !alive {
+			// An authenticator outlives its namespace for as long as the object is still being
+			// deleted, and one left behind by a namespace that never finished terminating stays
+			// here for good. Either way the release must not carry it.
+			input.Logger.Warn("Skipping DexAuthenticator of a namespace that is gone or terminating",
```
AlwxSin: 🔵 **A skipped authenticator is log-only, and this warning repeats on every hook run**

The hook fires on every DexAuthenticator, credentials-Secret and Namespace event, so an authenticator held by a finalizer in a namespace stuck in `Terminating` emits this same line indefinitely. Meanwhile the condition itself — an authenticator silently missing from the release — is not exposed as a metric, so nothing alerts on it.

This module already has the idiomatic pattern: `alert_on_long_id_token_ttl.go:76,100` uses `input.MetricsCollector.Expire(group)` plus `Set` per offending object. One expiring gauge would make the state visible in Prometheus and reduce this log line to a one-shot detail instead of a permanent stream.

modules/150-user-authn/hooks/get_dex_authenticator_crds_test.go
```diff
@@ -500,6 +500,94 @@ spec:
...
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: dead-ns
+status:
+  phase: Terminating
+`,
+			[]string{"alive@live-ns"},
+		),
+		Entry("is dropped when its namespace is gone and other namespaces are known",
```
AlwxSin: 🔵 **"Namespace is gone" is not a reachable state — this entry and the changelog `impact` are built on it**

A namespaced object cannot outlive a fully deleted namespace: the namespace only disappears once everything in it is gone. The real case is `Terminating` plus a finalizer, which the previous entry already covers. This entry is a fine unit test of the map lookup, but its name reads as coverage of a distinct cluster failure that cannot happen.

The same assumption produced an `impact` line that promises something the API cannot do: *"The authenticator returns to the release by itself if its namespace comes back."* A namespace never goes `Terminating` → `Active`, and if it was really deleted the authenticator went with it. In practice recovery means recreating both — worth rewording the changelog accordingly.

modules/150-user-authn/hooks/get_dex_authenticator_crds_test.go
Jabejixo: Done in fa5c0547ab: the six populated fixtures now share an `activeNamespaces` block (test, d8-monitoring, d8-dashboard, d8-user-authn), so the pre-existing specs run the lookup; breaking the guard now fails exactly the one spec that pins it.

modules/150-user-authn/hooks/get_dex_authenticator_crds.go
Jabejixo: Agreed, verified against shell-operator v1.20.4. The comment (and the PR description) now call the guard what it is: insurance for a broken assumption, with "no filtering" as the failure mode instead of dropping every authenticator.

modules/150-user-authn/hooks/get_dex_authenticator_crds.go
Jabejixo: Done in fa5c0547ab: the skip is exported as the expiring gauge `d8_user_authn_dex_authenticator_skipped{name, namespace}`, `D8DexAuthenticatorNamespaceUnusable` fires on it after 10m with the steps to finish the namespace deletion, and the log line is an Info detail now.

modules/150-user-authn/hooks/get_dex_authenticator_crds_test.go
Jabejixo: Right. The entry is renamed for the state it really covers — the Namespace informer running ahead of the DexAuthenticator informer — and the changelog no longer promises that the authenticator comes back: it is deleted with its namespace, and recovery is recreating it in a live namespace.

AlwxSin (approved):

z9r5 (approved):

PR #20410, opened 2026-06-03
[control-plane] disable kube-api qps for kcm

sprait:
## Description

Set  `--kube-api-qps=-1` to the `kube-controller-manager` static pod manifest, which disables the client-side rate limit on requests from `kube-controller-manager` to `kube-apiserver`

## Why do we need it, and what problem does it solve?

`kube-controller-manager` uses a default QPS limit of 20 req/s when talking to `kube-apiserver`. On clusters with a large number of nodes or objects, this limit causes request queuing inside KCM, which slows down reconciliation loops (node lifecycle, garbage collection, endpoint updates, etc.) and can produce visible latency spikes or stalled controllers.

Setting `--kube-api-qps=-1` removes the artificial client-side cap, allowing KCM to issue requests at the rate the apiserver can actually serve. Excessive load is still regulated server-side through the `kube-apiserver` `FlowSchema` and Priority and Fairness (APF) subsystem, which provides fair queuing across all clients without relying on per-client static limits.

## Why do we need it in the patch release (if we do)?

Not necessarily

## Checklist
- [ ] The code is covered by unit tests.
- [x] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [ ] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: control-plane-manager
type: chore
summary: Disabled the kube-api QPS rate limit for kube-controller-manager.
impact_level: default
```

PR #23146, opened 2026-09-17
[control-plane-manager] Make kube-controller-manager worker counts configurable

GureevLeonid:
## Description

Adds three optional parameters to the `control-plane-manager` ModuleConfig that set the matching `kube-controller-manager` worker-count flags:

| ModuleConfig setting | kube-controller-manager flag | upstream default |
|---|---|---|
| `controllerManager.concurrentDeploymentSyncs` | `--concurrent-deployment-syncs` | 5 |
| `controllerManager.concurrentReplicaSetSyncs` | `--concurrent-replicaset-syncs` | 5 |
| `controllerManager.concurrentHorizontalPodAutoscalerSyncs` | `--concurrent-horizontal-pod-autoscaler-syncs` | 5 |

Unset parameters render no flag, so every existing cluster keeps the upstream defaults. Values are bounded 1–100 by the schema. Changing one restarts `kube-controller-manager` on the masters (one at a time, as for any manifest change).

Plumbing follows `nodeMonitorGracePeriodSeconds` exactly: `hooks/arguments.go` copies the settings into `controlPlaneManager.internal.arguments`, `templates/daemonset.yaml` already passes that map to the control-plane manifest templates, and `candi/control-plane/kube-controller-manager.yaml.tpl` renders one flag per present key. Schema added to CE and EE/CSE `config-values.yaml`, RU descriptions, internal `values.yaml`, openapi case tests (positive + negative), hook unit tests and Helm template tests (rendered `kube-controller-manager.yaml.tpl` with/without the flags).

#20410 (`--kube-api-qps=-1`, v1.77) removed the client-side rate limit on KCM. The per-controller worker count is the other half of the same bottleneck, and today there is **no supported way** to set it: the ModuleConfig has no field, `kubeadm` patches placed in `/etc/kubernetes/deckhouse/kubeadm/patches/` are deleted by the manager (`removeOrphanFiles`), and a hand-edited static pod manifest is overwritten on the next converge.

### Evidence (Deckhouse EE v1.76.7, Kubernetes 1.34, 3 masters × 32 cores, 74 nodes, 2 284 Deployments, 6 054 pods, 718 HPAs)

The cluster runs a per-tenant CDP deployment: a release bump upgrades ~430 Helm releases and rolls up to 15 Deployments per tenant × ~140 tenants inside ~15 minutes. Measured with KCM's own metrics during three such waves on 2026-09-16/17:

| metric (`job="kube-controller-manager"`) | idle | during the wave |
|---|---|---|
| `workqueue_depth{name="deployment"}` | 0 | **150–240 for 5–8 min** (peaks 232, 240, 197) |
| `workqueue_queue_duration_seconds{name="deployment"}` p90 | 88 µs | **≥ 10 s** (histogram ceiling) |
| `workqueue_work_duration_seconds{name="deployment"}` p90 | 9 ms | **550–890 ms** |
| `workqueue_adds_total{name="deployment"}` rate | 11/s | 13–31/s |
| `rest_client_rate_limiter_duration_seconds` p90 (KCM client) | 0.2 s | 0.35–0.41 s |
| KCM request rate | 70/s | 95–233/s |
| API server: `apiserver_flowcontrol_rejected_requests_total` | 0 | **0** |
| API server: CPU (3 masters summed) / GET p99 on deployments/replicasets/pods | 6–9 cores / 70 ms | 16 cores / 96 ms |

Reading: a Deployment sync that costs 9 ms at idle costs ~850 ms during the wave — the per-controller 20 qps client bucket (each sync makes several API writes, each waiting ~0.35 s) — and with **5 workers** that is ~6 syncs/s of capacity against 13–31 adds/s, so the queue backs up and stays there for as long as the rollout phase lasts. The API server is not the limit (APF rejects nothing, p99 under 100 ms). The visible effect on the workload: the gap between "replica 1 Ready" and "replica 2 created" for a 2-replica Deployment went from p50 1 s early in the wave to **p50 25 s / p90 50 s** mid-wave — pure controller-manager queueing — which is ~27 % of the whole rollout time of every Deployment in that wave.

`--kube-api-qps=-1` (v1.77) fixes the 100× sync-time inflation. It does not change the worker count, which stays the cap once syncs are fast again on clusters with bursty rollout patterns. Both knobs are needed to size KCM to the workload, which is why this PR exposes the concurrency as a setting rather than changing a default for everyone.

The HPA controller shows the chronic version of the same problem on the same cluster, unrelated to waves: `workqueue_depth{name="horizontalpodautoscaler"}` sits at **400–700 permanently** (718 HPA objects, sync p50 0.55 s, 5 workers ≈ 19 syncs/s vs 48/s needed for a 15 s period), so every HPA is evaluated roughly every 37 s instead of every 15 s, and `workqueue_queue_duration_seconds{name="horizontalpodautoscaler"}` p90 is pinned at the 10 s ceiling. `--concurrent-horizontal-pod-autoscaler-syncs` is the upstream-documented remedy.

Why explicit settings instead of a generic `extraArgs`: it keeps the schema validated (integers, bounded), documented per flag, and matches how the module already exposes `nodeMonitorGracePeriodSeconds` / `failedNodePodEvictionTimeoutSeconds`. Happy to extend the set (StatefulSet, DaemonSet, Job, endpoint controllers) if maintainers prefer a broader list.

Not necessarily.

## Checklist
- [ ] e2e tests passed.

Tests run locally in `golang:1.26` with the repo mounted at `/deckhouse`: `modules/040-control-plane-manager/hooks` (259 specs, 2 new), `modules/040-control-plane-manager/template_tests` (76 specs, 3 new), `testing/openapi_cases` (58 specs; verified the new negative cases are rejected and a deliberately broken positive case fails the run). Docs are generated from the schema (`<!-- SCHEMA -->`).

## Changelog entries

```changes
type: feature
summary: Added `controllerManager.concurrentDeploymentSyncs`, `controllerManager.concurrentReplicaSetSyncs` and `controllerManager.concurrentHorizontalPodAutoscalerSyncs` settings to tune kube-controller-manager worker counts.
impact_level: default
```

https://claude.ai/code/session_01ThjTVu6Ji3SseuMauVCnt2


Nikita-Voskresenskiy:
Thanks for contributing! Reviewing this PR

modules/040-control-plane-manager/openapi/config-values.yaml
```diff
@@ -537,6 +537,34 @@ properties:
...
+    properties:
+      concurrentDeploymentSyncs:
+        type: integer
+        minimum: 1
+        maximum: 100
+        description: |
+          The number of Deployment objects reconciled concurrently (`--concurrent-deployment-syncs`). Upstream default: 5.
+      concurrentReplicaSetSyncs:
+        type: integer
+        minimum: 1
```
z9r5, line 558:
```suggestion
        minimum: 1
        x-doc-defaut: 5
```
How about we do it this way and remove the mention of “default” from the text? Then “default” will be rendered separately in the dock.

sprait:
@GureevLeonid thanks for the detailed measurements! We'd be glad to see the same measurements repeated on 1.77 (if you get the chance), to see what the picture looks like once the client-side limiter is gone.

On the PR itself, we'd rather not expose these as ModuleConfig settings - in our target design the platform picks the right values itself, so users don't have to reason about worker counts. So we're not ready to merge this as it stands, but we'll keep it in mind as we think about how the control plane should be sized.

PR #23160, opened 2026-09-18
[admission-policy-engine] Alert on an unavailable admission webhook

Suselz:
## Description

Adds four alerts on the availability of the module's own components, and documents why the component is critical and how to recover from its outage.

The alerts are defined in `modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml`:

- `D8AdmissionPolicyEngineWebhookUnavailable`, severity 2, `for: 2m`. Fires when `gatekeeper-controller-manager` has no available replica.
- `D8AdmissionPolicyEngineWebhookNotServing`, severity 2, `for: 5m`. Fires when the API server itself reports that it cannot call a webhook of the module, through `apiserver_admission_webhook_rejection_count` with `error_type="calling_webhook_error"`. This covers the outage the replica counters cannot see: a replica that is `Ready` while its certificate no longer matches the `caBundle`, or while the service has no endpoints.
- `D8AdmissionPolicyEngineWebhookDegraded`, severity 3, `for: 10m`. Fires when part of the replicas is unavailable and the webhook has no redundancy left.
- `D8AdmissionPolicyEngineAuditUnavailable`, severity 4, `for: 20m`. Fires when `gatekeeper-audit` has no available replica.

The documentation gains a section on the availability of the components in the module overview, and a recovery procedure in the FAQ. Both are updated in English and in Russian.

Nothing is restarted and no object of the module changes. The pull request adds alert rules and documentation only.

`gatekeeper-controller-manager` serves a ValidatingWebhookConfiguration whose main webhook has `failurePolicy: Fail`. While the deployment has no available replica, the API server rejects every request the webhook intercepts: creation, modification and deletion of pods and of the controllers that create them, of Role, RoleBinding and Gatekeeper constraints, as well as `kubectl exec` and `kubectl attach` in `d8-*` and `kube-*` namespaces.

Until now nothing reported that state:

- The module shipped four alerts. Three report policy violations and are built on the metrics of `gatekeeper-audit`, so they go silent instead of firing when the audit is down. The fourth, `D8AdmissionPolicyEngineNotBootstrapped`, covers the case where the module never bootstrapped, and its metric stays at zero in a cluster that bootstrapped once.
- The generic `KubernetesDeploymentReplicasUnavailable` and `KubernetesPodIsNotReady` alerts are enabled per namespace by the `extended-monitoring.deckhouse.io/enabled` label, which `d8-admission-policy-engine` does not carry and does not gain here. The alerts of this pull request are what covers the module instead, with the severity, the description and the grouping the component needs.

The result was an outage of cluster-wide admission that raised no alert at all. Recovering from it also required knowledge that was nowhere in the documentation.

On `for: 2m` of the first alert: a single-replica installation raises it during a VPA recreate and during a drain of the node that carries the replica, because `PodDisruptionBudget.minAvailable` is `0` outside HA mode. Those are not false positives — in a single-replica installation admission is genuinely blocked in exactly those windows, and the alert is the only thing that says so. A rollout is not among them: outside HA mode `helm_lib_deployment_strategy_and_replicas_for_ha` emits no `strategy` block, so the default `maxSurge` of 25% rounds up to 1 and the new pod is ready before the old one goes away.

`D8AdmissionPolicyEngineAuditUnavailable` is severity 4 rather than lower. An unavailable audit is not only a loss of reporting: the `constraint-exporter` container of the same deployment feeds the ConfigMap from which Deckhouse builds the `rules` section of the webhook configuration, so while it is down, resources matched only by newly added policies are not sent to the webhook and those policies are not enforced on admission.

Not necessarily. Every released version has the same gap, so a backport is useful, but nothing in the change depends on the release it lands in.

## Checklist
- [ ] e2e tests passed.

The rules were validated with `promtool check rules`. Every command in the alerts and in the FAQ was run against a cluster, which is how two of them turned out to be wrong and were replaced.

## Changelog entries

The level is left at the default. Nothing is restarted, no permission or default changes, and nothing has to be audited before the update, while the entries of "Know before update" are restarts and behaviour changes an operator has to prepare for.

```changes
type: feature
summary: Report the availability of the module components with alerts.
```


modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
```diff
@@ -0,0 +1,144 @@
+- name: admission-policy-engine.availability
```
AlwxSin, line 1: 🟡 **Nothing alerts when the webhook is up but not serving**

All three alerts read only the deployment's replica counters. A replica that is Ready while the webhook is not actually reachable — a stale `gatekeeper-webhook-server-cert` versus the `caBundle` in the ValidatingWebhookConfiguration, broken Service/endpoints, a scrape target that stopped answering — yields `kube_deployment_status_replicas_available > 0` and therefore silence, while the API server keeps rejecting everything because of `failurePolicy: Fail`. That is the same cluster-blocked outage this PR exists to report, in its most insidious form.

The companion alert is cheap here: `templates/podmonitor.yaml` already scrapes both deployments with `jobLabel: app`, so `sum(up{job="gatekeeper", namespace="d8-admission-policy-engine"}) == 0 or absent(...)` is one block away. `D8UserAuthzControllerTargetDown` in `modules/140-user-authz/monitoring/prometheus-rules/user-authz-controller.yaml:39` is the precedent in this repo.

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
```diff
@@ -0,0 +1,144 @@
...
+        and
+        (
+          kube_deployment_spec_replicas{
+            namespace="d8-admission-policy-engine",
+            deployment="gatekeeper-controller-manager"
+          } > 0
+        )
+      for: 2m
+      labels:
+        severity_level: "1"
```
AlwxSin: 🟡 **S1 is unprecedented in the platform — worth confirming the routing before it lands**

`severity_level: "1"` appears nowhere else in the tree. The lowest currently in use is `"2"`, in exactly two rules: `modules/002-deckhouse/monitoring/prometheus-rules/cni-checks.yaml:7` and `ee/cse/modules/040-control-plane-manager/monitoring/prometheus-rules/signature.yaml:12`. The docs site does carry `.alertslist__badge_severity-s1` styling (`docs/site/assets/css/docs.scss:2761`), so rendering is fine — the open question is what S1 means to the on-call routing, which lives outside this repository.

Related, on `for: 2m`: the PR description justifies it by saying a rollout of a single-replica installation can raise the alert. It can't. In non-HA `helm_lib_deployment_strategy_and_replicas_for_ha` emits no `strategy` block, so the default RollingUpdate applies (`maxSurge` 25%→1, `maxUnavailable` 25%→0) and the new pod comes up before the old one goes away. The real false-positive sources are the VPA with `updateMode: InPlaceOrRecreate` evicting the single replica, and a node drain against `PodDisruptionBudget.minAvailable: 0` (`templates/gatekeeper-deployment.yaml:222`).

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
```diff
@@ -0,0 +1,144 @@
...
+            namespace="d8-admission-policy-engine",
+            deployment="gatekeeper-controller-manager"
+          } == 0
+        )
+        and
+        (
+          kube_deployment_spec_replicas{
+            namespace="d8-admission-policy-engine",
+            deployment="gatekeeper-controller-manager"
+          } > 0
```
AlwxSin: 🔵 **The `spec_replicas > 0` guard protects against nothing and costs a real state**

When the module is disabled, Deckhouse deletes the Deployment object outright, so both series disappear and `kube_deployment_status_replicas_available == 0` matches nothing anyway — the guard is not what suppresses the alert in that case.

What it does suppress is `kubectl scale deploy gatekeeper-controller-manager --replicas=0`, a classic emergency move that leaves the ValidatingWebhookConfiguration in place (it is gated on `internal.bootstrapped`, not on the replica count) and the cluster fully blocked, with no alert until Deckhouse reconciles the replica count back. Dropping the clause gives a shorter expression and one more covered state. Same clause in the audit alert at lines 111-116.

modules/015-admission-policy-engine/docs/README.md
```diff
@@ -642,3 +641,30 @@ The module allows you to use the [Gatekeeper Custom Resources](gatekeeper-cr.htm
...
+- Creation and modification of Role, RoleBinding and Gatekeeper constraints.
+- `kubectl exec` and `kubectl attach` in namespaces whose names start with `d8-` and `kube-`. These are intercepted by a separate webhook that also has `failurePolicy: Fail`, so the outage narrows the ways to diagnose itself.
+
+The mutating webhook is configured differently: its `failurePolicy` is `Ignore`, and an unavailable deployment only means that mutations are not applied.
+
+Disabling the module unblocks the cluster, but the control goes with it: no policy is enforced any more, and an object a policy used to forbid is created without hindrance. The FAQ describes [what to do](faq.html#what-to-do-if-the-admission-webhook-is-unavailable) while the webhook is unavailable.
+
+### Why only the webhook pods are excluded from validation
+
+One exclusion in the webhook configuration covers the pods of `gatekeeper-controller-manager` and nothing else: every webhook excludes objects carrying the `gatekeeper.sh/operation: webhook` label through its `objectSelector`. The exclusion exists to break a circular dependency, not to relax the policies for the module.
```
AlwxSin: 🔵 **The new exclusion section overstates the exclusion**

"every webhook excludes objects carrying the `gatekeeper.sh/operation: webhook` label through its `objectSelector`" is not accurate. `deny-exec-heritage.admission-policy-engine.deckhouse.io` (`templates/validatingwebhookconfiguration.yaml:118`) has no `objectSelector` at all — it matches `pods/exec` and `pods/attach` in every `d8-*` and `kube-*` namespace, `d8-admission-policy-engine` included.

So `exec` into the gatekeeper pods themselves is blocked precisely during the outage this section describes. That is the operationally interesting part, and it is why the FAQ correctly reaches for `logs` and `delete pod` rather than `exec`. Same wording in `docs/README_RU.md:673` («каждый вебхук исключает через `objectSelector`»).

modules/015-admission-policy-engine/docs/README.md
```diff
@@ -642,3 +641,30 @@ The module allows you to use the [Gatekeeper Custom Resources](gatekeeper-cr.htm
...
+
+A webhook that validates the pods serving it cannot recover from its own outage. Once the last replica is gone, the API server has nowhere to deliver the request, so it rejects the creation of the replacement pod and the deployment can never return to a running replica on its own. The label, which Gatekeeper sets on its own pods, is the narrowest exclusion that breaks the cycle: the deployment can always create a pod, while every other object of the namespace is validated as usual.
```
AlwxSin, line 670: 🔵 **The same section leaves out what the exclusion does not cover**

"the deployment can always create a pod, while every other object of the namespace is validated as usual" is accurate, but stops one step short of the consequence a reader of this exact section needs.

`gatekeeper-audit` pods carry `gatekeeper.sh/operation: audit`, so the `objectSelector` does not exempt them, and the namespace carries `security.deckhouse.io/enable-security-policy-check: "true"` (`templates/namespace.yaml:4`), which is exactly what `security-policy-exception.admission-policy-engine.deckhouse.io` selects on. While `gatekeeper-controller-manager` is down, the audit deployment therefore cannot get a replacement pod — so an outage of the webhook can drag `D8AdmissionPolicyEngineAuditUnavailable` along with it, and the two new alerts are not independent. Worth saying so here. Same in `docs/README_RU.md:675`.

modules/015-admission-policy-engine/templates/namespace.yaml
```diff
@@ -1,4 +1,5 @@
 {{- define "labels" }}
+extended-monitoring.deckhouse.io/enabled: ""
```
AlwxSin, line 2: 🔵 **Custom alerts and the new extended-monitoring label overlap — worth deciding consciously**

This label is the `impact_level: high` part of the change, and it turns on the generic workload alerts for both deployments — the same condition the two new deployment-replica alerts already report. I could not check the generic alerts' severities: `extended-monitoring` is not in this repository, it ships as a separate module.

Better decided now than discovered on the first incident: either accept the duplication (the generic alerts also bring image-availability and pod-not-ready coverage the custom ones lack), or suppress the per-deployment generic alert. `extended-monitoring.deckhouse.io/enabled: "false"` on `modules/040-node-manager/templates/standby-holder/deployment.yaml:8` is the mechanical precedent, though weaker support than it looks — standby-holder is a permanently-degraded-by-design workload rather than a deduplication case.

modules/015-admission-policy-engine/templates/namespace.yaml
```diff
@@ -1,4 +1,5 @@
 {{- define "labels" }}
```
AlwxSin, line 1: 🔵 **The namespace label change has no test**

(Anchored here because the real location, `modules/015-admission-policy-engine/template_tests/`, is not touched by this PR.)

The label added on the next line is the only behavioural change in the diff — the one the changelog marks `impact_level: high` — and nothing in `template_tests/` asserts the namespace's labels today. That also leaves it unprotected against a future edit to this `{{- define "labels" }}` block. A three-line assertion that the rendered Namespace carries `extended-monitoring.deckhouse.io/enabled` locks it in.

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
Suselz: Fixed in 12b8c6ab3c, but on a different metric than the one suggested.

`up` does not cover the scenario you describe. The webhook listener is port 8443 with `gatekeeper-webhook-server-cert`, while the PodMonitor scrapes kube-rbac-proxy on 10354, so a `caBundle` that no longer matches the certificate leaves `up == 1`. The other case a TargetDown rule would catch, a wedged process, already fails the readiness probe on `/readyz` and is reported by `D8AdmissionPolicyEngineWebhookUnavailable`.

The new `D8AdmissionPolicyEngineWebhookNotServing` reads `apiserver_admission_webhook_rejection_count{error_type="calling_webhook_error", name=~".*admission-policy-engine\\.deckhouse\\.io"}` instead. That is the API server reporting that its own call failed, independently of the replica counters. Verified on a live cluster that the series exists with `name` set to each webhook of the module, and that it survives the scrape: the `kube-apiserver` job carries a single `metric_relabel_configs` entry, which drops buckets of `apiserver_request_duration_seconds_bucket`.

One more point in favour of not using `up` here: with `jobLabel: app` both PodMonitors produce `job="gatekeeper"`, so a TargetDown rule could not have told the webhook apart from the audit.

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
Suselz: Lowered to `severity_level: "2"` in 12b8c6ab3c. You are right that 1 is used by no rule in the tree and that what it routes to is decided outside this repository; 2 is the most severe level the platform actually uses, so it routes for certain.

On `for: 2m`, the PR description was wrong and is corrected. Your reading of the rollout is right: outside HA mode no `strategy` block is emitted, so the default `maxSurge` of 25% rounds up to 1 and the new pod is ready before the old one goes away.

The two sources you name stay as firing conditions on purpose. A VPA recreate and a drain against `minAvailable: 0` each leave a single-replica installation without a serving webhook, and with `failurePolicy: Fail` admission is blocked for that window. They are short outages rather than false positives, and this alert is the only thing that reports them.

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
Suselz: Fixed in 12b8c6ab3c, dropped from both alerts. Your reading holds on both counts: a disabled module deletes the Deployment, so both series vanish and the guard suppresses nothing there, while a Deployment scaled to zero by hand leaves the webhook configuration in place and the cluster blocked.

modules/015-admission-policy-engine/docs/README.md
Suselz: Fixed in 12b8c6ab3c, in both languages. The sentence now attributes the exclusion to the webhooks that intercept the creation of objects, and adds that `deny-exec-heritage` carries no `objectSelector`, so exec into the pods of the module is blocked during an outage along with everything else.

modules/015-admission-policy-engine/docs/README.md
Suselz: Not adding this one. The overview keeps the two questions it answers — why an unavailable webhook blocks the cluster, and why the pods of the webhook are the only ones exempt from validation — and the mechanics of which pods carry which value of `gatekeeper.sh/operation` are below that line.

The dependency you describe is real and your reading of it is accurate, including the point that the two alerts are not independent. It is left out as detail a reader of the overview does not act on, not because it was missed.

modules/015-admission-policy-engine/templates/namespace.yaml
Suselz: Acted on in 88d4df2bb1: the label is dropped, the four alerts of this module stay as the single source.

Your framing of the choice is what settled it. The generic alerts would report the same two deployment-replica conditions with a severity decided in another repository and without the module-specific description, while the coverage they add on top — pod-not-ready and image availability — is not worth carrying the duplicate for a component whose failure mode is already covered here in detail.

Removing the label also takes the last object change out of the branch. Nothing in a cluster is touched on update now, so the changelog entry carries no `impact` note and the level is the default rather than `high`.

modules/015-admission-policy-engine/templates/namespace.yaml
Suselz: The namespace is no longer touched by this branch, so there is nothing left for a template test to pin.

An assertion on the rendered Namespace was added in 12b8c6ab3c and removed again in 88d4df2bb1, together with the `extended-monitoring.deckhouse.io/enabled` label it was protecting. The branch now adds alert rules and documentation only, and `templates/` is unchanged.

modules/015-admission-policy-engine/templates/namespace.yaml
Suselz: Reversed in 88d4df2bb1: the label is dropped, the four alerts of this module stay as the single source.

Your framing of the choice is what settled it. The generic alerts would report the same two deployment-replica conditions with a severity decided in another repository and without the module-specific description, while the coverage they add on top — pod-not-ready and image availability — is not worth carrying the duplicate for a component whose failure mode is already covered here in detail.

Removing the label also takes the last object change out of the branch. Nothing in a cluster is touched on update now, so the changelog entry no longer needs an `impact` note, and the level is the default rather than `high`.

modules/015-admission-policy-engine/templates/namespace.yaml
Suselz: Follow-up: the assertion is removed again in 88d4df2bb1, together with the label it was protecting. With the namespace unchanged, the branch adds alert rules and documentation only, and there is no rendered object left for a template test to pin.

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
```diff
@@ -0,0 +1,166 @@
...
+          rate(
+            apiserver_admission_webhook_rejection_count{
+              job="kube-apiserver",
+              type="validating",
+              error_type="calling_webhook_error",
+              name=~".*admission-policy-engine\\.deckhouse\\.io"
+            }[5m]
+          )
+        ) > 0
+      for: 5m
```
AlwxSin: 🟡 **This alert pages at S2 on transient call failures**

The metric and the selector check out — I followed `type="validating"` into `k8s.io/apiserver` (verified identical in 0.33, 0.34 and 0.35: it is a hardcoded literal in `pkg/admission/plugin/webhook/validating/dispatcher.go`, not the `stepValidate` constant), and `error_type="calling_webhook_error"` is only recorded under `!ignoreClientCallFailures`, so normal policy denials land under `no_error` and stay out. The `job="kube-apiserver"` relabeling and the single `metricRelabelings` entry are as you described in `modules/040-control-plane-manager/templates/kube-apiserver-service-monitor.yaml`. The rule itself is what I want to raise.

`rate(...[5m]) > 0` combined with `for: 5m` has no tolerance. A burst of errors from T to T+30s keeps the rate elevated until roughly T+5m30s, so the `for` clock completes at T+5m and the alert fires — any burst wider than a scrape interval or two is enough.

The code-verified way to get there without an outage is a webhook timeout. `timeoutSeconds: 15` is set on every webhook in `templates/validatingwebhookconfiguration.yaml`, and the dispatcher excludes only `context.Canceled` from the counter, never `context.DeadlineExceeded` — so gatekeeper taking longer than 15s to evaluate constraints on a large object is counted as `calling_webhook_error` exactly like an unreachable pod. (A rollout window where the API server still reaches a terminating replica is a second candidate, but I could not confirm it: the API server resolves the webhook through the Service and endpoint propagation is usually sub-second. Not making that claim.)

S2 is the most severe level the platform actually routes, and `D8AdmissionPolicyEngineWebhookUnavailable` already carries it. A handful of 15-second timeouts is a genuine user-visible rejection, but it is not the same event as "no replicas left". A nonzero threshold or a longer `for` separates the two — which one is your call. Both S2 alerts firing together during a real outage is fine, they share `plk_grouped_by__d8_admission_policy_engine_unavailable`.

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
```diff
@@ -0,0 +1,166 @@
...
+
+          The alert is independent of the replica counters: it fires while the pods are `Ready` as well. A replica that is running but unreachable to the API server produces the same outage, and only this alert reports it. Typical causes are a `gatekeeper-webhook-server-cert` that no longer matches the `caBundle` of the ValidatingWebhookConfiguration, an empty `gatekeeper-webhook-service` endpoint list, and a network policy that blocks the API server.
+
+          Read the reason the API server reports, then check the certificate and the endpoints:
+
+          ```bash
+          d8 k -n d8-admission-policy-engine logs deploy/gatekeeper-controller-manager -c manager --all-pods=true --tail=200
+          d8 k -n d8-admission-policy-engine get endpoints gatekeeper-webhook-service
+          d8 k get validatingwebhookconfiguration d8-admission-policy-engine-config -o jsonpath='{.webhooks[*].clientConfig.caBundle}' | base64 -d | openssl x509 -noout -subject -dates
```
AlwxSin: 🔵 **This runbook command emits a spurious error**

`jsonpath='{.webhooks[*].clientConfig.caBundle}'` returns one value per webhook, space-separated, and a default install renders three of them (`admission-policy-engine`, `deny-exec-heritage`, `security-policy-exception`).

I checked what `base64 -d` does with that input: it decodes the first value correctly and then prints `base64: invalid input` on stderr. So the command does produce the right certificate — it just reads as if it failed, which is the wrong thing to hand someone working through a runbook mid-incident. `{.webhooks[0].clientConfig.caBundle}` is the whole fix.

The secret command just below is fine: `.data.ca\.crt` is a single value and decodes cleanly.

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
Suselz: Fixed in 1f6c2f1b83 by raising `for` to 15 minutes and leaving the threshold at zero.

Your arithmetic is what decided the shape of the fix. A burst keeps `rate(...[5m])` above zero for roughly its own duration plus the window, so a threshold is not what separates the two events — duration is. Fifteen minutes is longer than any burst can hold the rate, while the state this alert exists for, a replica the API server cannot reach at all, does not stop by itself. That also avoids inventing an errors-per-second number, which would have to be recalibrated against the request rate of each cluster.

The cost is that a webhook which is up but unreachable is now reported at T+15m rather than T+5m. The case that needs to be fast, losing the replicas, is still reported at T+2m by `D8AdmissionPolicyEngineWebhookUnavailable`.

The timeout path you traced is now named in the description as a cause of the metric that is not a cause of the alert, so an operator who sees the counter move without the alert firing knows why.

modules/015-admission-policy-engine/monitoring/prometheus-rules/availability.yaml
Suselz: Fixed in 1f6c2f1b83, the runbook now reads `{.webhooks[0].clientConfig.caBundle}`.

Reproduced the behaviour you describe: GNU `base64 -d` on two space-separated values prints the first decoded value and then `base64: invalid input` on stderr. Worth adding that BSD `base64` on macOS fails the whole input with `error decoding base64 input stream` and decodes nothing, so on a workstation the command was not merely noisy but wrong.

AlwxSin (approved):

PR #23165, opened 2026-09-18
[deckhouse] reconcile health monitor

ipaqsa:
## Description

Adds a periodic, level-triggered reconcile under the existing edge-triggered health/status pipeline, so a package's published status converges on its own.

**1. Health informer resync — `internal/packages/health/monitor`**

The shared informer factory is built with a `defaultResyncPeriod` of `5m` instead of `0`. A resync replays the **local cache** into the existing handlers (it is not a LIST), arriving through `UpdateFunc` with `old == new`, so every known package key is re-enqueued and re-reduced from cache. `Add`/`Update`/`Delete`, the relabel A→B double-enqueue and the `DeletedFinalStateUnknown` tombstone path are untouched.

Worth being precise about what this buys: `health.Service.reconcile` still dedupes against `lastHealth`, which only `reconcile` writes and only from that same cache — so a replay recomputes what is already there and fires no callback. It bounds how long any future divergence between cache and reduced state could last; it is not what makes the published status converge. That is (2).

**2. Status resync — `internal/packages/status`**

`StartResync` runs one goroutine that every `defaultResyncPeriod` (`10m`) snapshots the registered package keys under `s.mu`, **releases the lock**, and routes each key through the existing `queueFor` into `appQueue`/`moduleQueue`. This bypasses every change-detection gate, including `health.Service`'s transition dedup — which is the point. Packages buffered in `pendingHealth` are skipped: they carry no status yet, and `NewStatus` enqueues them when it drains the buffer.

One goroutine serves every package, and the workqueue's set semantics cap a tick at one item per package regardless of the number of events. The queue remains the retry and dedup mechanism; consumers still `Forget` on success and `AddRateLimited` on failure. Nothing is logged per successfully synced package.

**3. Lifecycle** — `Runtime.Run` starts the resync, `Runtime.Stop` → `status.Shutdown` stops it and **waits for it to exit before shutting the queues down**, so no key can be added to a queue that is already closing.

**4. No redundant status PATCH** — both status writers now compare the fetched status against the computed one with `apiequality.Semantic.DeepEqual` and skip `Status().Patch` when they match, so a steady-state resync tick costs a GET and no `resourceVersion` bump. The `AddRateLimited` retry on a failed patch is unchanged.

`health.Monitor` watches `Deployment`/`StatefulSet` objects labelled `health.deckhouse.io/package`, reduces them to a package health state, and hands transitions to `status.Service`, which publishes `ConditionScaled` onto the `Module`/`Application` resource.

Every hop in that chain was edge-triggered and deduplicated: the informer had `resyncPeriod = 0`, `health.Service` fired its callback only on a state transition, and `status.Service` enqueued only when a condition actually changed. That is fast, but it has no self-healing property — a single dropped notification, or a status write lost to a transient API error, left the published `ConditionScaled` stale **until the workload next changed on its own**, which for a healthy steady-state workload can be never.

After this change the same event path still provides the fast reaction, and the periodic reconcile is the backstop: a stale `ConditionScaled` is repaired within one status-resync period plus retry time, with no workload change required, and a transient Kubernetes API error can no longer cause an update to be lost permanently.

This is level-triggered reconciliation and eventual consistency — not exactly-once delivery.

## Notes for reviewers

- **Both periods are unexported package-level consts.** Nothing needs to vary them, so neither is a constructor option.
- **`StartResync` and `Shutdown` are called exactly once**, from `Runtime.Run` and `Runtime.Stop`, and the pairing is now unconditional — there is deliberately no `sync.Once` and no disabled-resync escape hatch. A second `Shutdown` would panic on the double `close(resyncStop)`, and a `Shutdown` with no preceding `StartResync` would block on `resyncDone`. `Run` returns early on a `startAPIServers` error before reaching `StartResync`, so if `Runtime.Stop` ever gains a caller (it has none today), that pairing needs re-checking.
- **The writers read through the manager's cache.** `original` can therefore be stale, and the `DeepEqual` check can skip a patch based on a stale read. This is intended: computing a status from a stale read and writing it would be worse, and the resync is what retries once the cache catches up.
- **The skip relies on `computeAndApplyConditions` being deterministic.** It is: `meta.SetStatusCondition` stamps `LastTransitionTime` only when `Status` actually flips, `ObservedGeneration` tracks `.Generation`, `Tracking` is left untouched when the progress report is empty, and `json.Marshal` sorts map keys.
- Out of scope, unchanged: the `Scaled`/`Reconciling`/`Degraded`/`Unknown` computation rules, the `health.deckhouse.io/package` label, the workqueues and the condition mappers.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: deckhouse
type: fix
summary: Module and Application status conditions now re-sync periodically, so a stale Scaled condition is repaired automatically instead of persisting until the workload changes.
```


ldmonster (approved):

deckhouse-BOaTswain:
Backport failed ⚠️. Failure: cherry-pick commit 6c65d8960f023031b16c709a85687dff70baa825 to the branch [release-1.77](https://github.com/deckhouse/deckhouse/tree/release-1.77) failed. See [Job](https://github.com/deckhouse/deckhouse/actions/runs/35627568397) for details.

PR #21313, opened 2026-07-13
[user-authn] Add CPU limit to self-signed-generator init container

IlshatKhisamov:
Init container lacked `limits.cpu`, causing pod admission failures in namespaces with ResourceQuota or required-resources policies.

## Description

Added CPU and memory limits to the `self-signed-generator` init container in `dex-authenticator` (and aligned memory limit in `user-api`).

For `DexAuthenticator`, init limits are set to the **sum** of the main containers’ limits (`dex-authenticator` + `redis`) from `spec.resources` / `spec.resources.redis`. If resources are not set, defaults are used: `20m` CPU and `50Mi` memory (10m+10m / 25Mi+25Mi).

This matches ResourceQuota accounting: `max(init limits, sum of app container limits)`, so the init container does not inflate quota usage beyond the main containers.

## Why do we need it, and what problem does it solve?

When `ResourceQuota` is enabled in a namespace (or admission policies require CPU limits on all containers), Kubernetes rejects pods where any container — including init containers — lacks `limits.cpu`.

`DexAuthenticator` already allows configuring resources for `dex-authenticator` and `redis`, but the `self-signed-generator` init container had only `requests.cpu` and `limits.memory`. As a result, pods failed admission and `DexAuthenticator` could not start in quota-restricted namespaces.

## Why do we need it in the patch release (if we do)?

Not necessarily.

## Checklist
- [x] The code is covered by unit tests.
- [ ] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Fixed DexAuthenticator pod creation under ResourceQuota by setting init container CPU/memory limits to the sum of main container limits.
impact_level: default
```

PR #23170, opened 2026-09-18
[user-authn] Fix the init container limits of a DexAuthenticator

Suselz:
## Description

The limits of the `self-signed-generator` init container of a DexAuthenticator Deployment were the
sum of the `dex-authenticator` and `redis` limits taken from the custom resource, while the requests
of that container are hardcoded as `10m` and `10Mi`. Whenever that sum landed below the requests the
API server rejected the Deployment.

This pull request gives the init container fixed resources instead, the same literals the identical
init container of `user-api` has carried since the sum was introduced:

```yaml
          requests:
            memory: 10Mi
            cpu: 10m
          limits:
            memory: 25Mi
            cpu: 10m
```

No value from the custom resource reaches the init container any more, so no input can put its
limits below its requests.

It also moves `LIB_HELM_VERSION` from `1.72.21` to `1.72.24`, which carries the quantity-conversion
fixes other modules depend on.

Only DexAuthenticator Deployments are re-rendered. No control plane component, ingress controller or
Prometheus instance is disturbed.

The release of `user-authn` never completes, and the `main` queue of Deckhouse stops on the first
rejected object, which blocks every subsequent task in the cluster:

```
Deployment.apps "<name>-dex-authenticator" is invalid:
spec.template.spec.initContainers[0].resources.requests: Invalid value: "10m":
must be less than or equal to cpu limit of 0
```

Two inputs reach that state. A fractional quantity such as `cpu: "0.5"` converted to `0` through
the `atoi` in the helm_lib helpers, and a limit that is simply smaller than the requests, such as a
single `cpu: 5m`, produced a sum below them honestly. Both are valid input: the CRD accepts a
fractional quantity through the standard Kubernetes quantity pattern, and `0.5` and `500m` mean the
same thing.

## Why literals rather than a fixed sum

The first version of this change fixed the conversion and clamped the sum to the requests. Review
showed the sum earns nothing.

A ResourceQuota counts a pod as `max(max(init), sum(app))`, so literals are accounted exactly like
the sum in every case:

- the main containers ask for more than `10m` — the quota counts them, either way;
- they ask for less — the quota counts the init container, and `10m` is what the clamp produced too;
- `spec.resources` is unset — the main containers carry no limits at all, so a limits-scoped quota
  rejects the pod whatever the init container says.

Under a LimitRange `maxLimitRequestRatio` literals are strictly better: the ratio against
`requests: 10m` is 1 rather than `sum(app)/10m`. What the sum bought was runtime headroom for the
init container, and `user-api` shows `10m` / `25Mi` is enough for the same binary doing the same
work.

## About the library bump

The bump is no longer what fixes this module, since the module no longer converts anything. It
still carries the conversion fixes for `istio` and `cni-cilium`, which reach the same helpers
through `resourcesManagement`, and for any module built on the library.

The vendored chart cannot be patched in place: `make generate-lib-helm` re-syncs it from upstream
and the `go_generate` job fails on the drift.

It also carries deckhouse/lib-helm#216, which renders an extra ClusterRoleBinding from
`helm_lib_module_controller_rbac`. No module in this repository uses that helper, so nothing here
changes because of it:

```
$ grep -rl "helm_lib_module_controller_rbac" modules/ ee/ | wc -l
0
```

The version is `1.72.24` rather than `1.72.23`. `1.72.23` narrowed the quantity helpers so that a
valid quantity — `+1`, `1.`, `100u` for CPU, `100m` and `+1Gi` for memory — aborted the rendering
instead of converting, and put the `E` and `Ei` suffixes on the CPU path where the result no longer
fits an int64. deckhouse/lib-helm#218 restores them, and this branch was moved to it, so the
narrowed release never reaches a patch release. Rendered against the vendored chart:

```
cpu "+1"    -> 1000       cpu "1E"     -> 0
cpu "-1"    -> -1000      cpu "10P"    -> 0
cpu "1."    -> 1000       cpu "abc"    -> 0
cpu "100u"  -> 1          mem "100m"   -> 1
cpu "1k"    -> 1000000    mem "+1Gi"   -> 1073741824
```

## Test coverage of the helpers

The converters are tested upstream in `deckhouse/lib-helm`, not here, and `1.72.24` covers every
quantity form, including a bare fractional `memory: "0.5"`, the sign, the trailing dot, the
exponent, every suffix and the int64 bound. It also carries the assertion that was missing when the
zero was introduced: a fractional `vpa.cpu.min` with a `limitRatio` has to size the limits rather
than collapse them below the requests. That suite previously only asserted the rendered block was
non-empty, which `limits.cpu: 0m` satisfies.

In this repository the tests now assert what this module actually promises: the init container
carries the same resources for a fractional CR, a tiny CR, a whole-unit CR and no `resources` at
all, and the main containers still receive the CR limits untouched.

The defect reached `v1.76.14` and `v1.77.0`, and it stops the `main` queue of an affected cluster
until the custom resource is edited by hand. A cluster that provisions DexAuthenticator objects from
a template hits it again on every new object.

## Cluster check

Ran on a cluster with the build of this branch. Six DexAuthenticator objects, covering a fractional
CPU, a fractional memory suffix, a bare fractional memory, a sum below the requests, whole
millicores and no `resources` at all. Every one of them rendered the same init container:

```
OBJECT       INIT_LIMITS                     INIT_REQUESTS                   MAIN_LIMITS
frac-cpu     {"cpu":"10m","memory":"25Mi"}   {"cpu":"10m","memory":"10Mi"}   {"cpu":"500m","memory":"512Mi"}
frac-mem     {"cpu":"10m","memory":"25Mi"}   {"cpu":"10m","memory":"10Mi"}   {"cpu":"1500m","memory":"512Mi"}
tiny         {"cpu":"10m","memory":"25Mi"}   {"cpu":"10m","memory":"10Mi"}   {"cpu":"5m","memory":"8Mi"}
millicores   {"cpu":"10m","memory":"25Mi"}   {"cpu":"10m","memory":"10Mi"}   {"cpu":"500m","memory":"512Mi"}
defaults     {"cpu":"10m","memory":"25Mi"}   {"cpu":"10m","memory":"10Mi"}   none
bare-frac    {"cpu":"10m","memory":"25Mi"}   {"cpu":"10m","memory":"10Mi"}   {"cpu":"2","memory":"500m"}
```

The `self-signed-generator` container completed with exit code 0 in all six, and the `main` queue
stayed empty throughout.

That run used the build carrying `1.72.23`. The module does not call the quantity helpers, so the
move to `1.72.24` does not change anything it renders, and the template tests of `istio` and
`cni-cilium` cover what the bump does touch.

`bare-frac` is the case that used to take down the rendering of the whole module through the `fail`
branch of the memory helper. It now renders, and the API server only warns:
`fractional byte value "500m" is invalid, must be an integer`. Its `dex-authenticator` container
does not start, because half a byte of memory is not a usable limit, but that is the value the
object asks for. `tiny` stays at 1/2 for the same reason, with `8Mi`.

## Checklist
- [ ] e2e tests passed.

The e2e run on this branch predates the switch to literals, so that box stays unticked.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Render a valid Deployment for a DexAuthenticator whose `resources` limits are fractional or below the init container requests.
impact: |
  The `self-signed-generator` init container now carries the fixed `10m` / `25Mi` it was always
  meant to have, the same as the identical init container of `user-api`, instead of the sum of the
  main container limits. A DexAuthenticator that set large limits no longer gets that sum as burst:
  generating the self-signed certificate then takes 3 to 4 seconds instead of under a second. It
  happens once per pod start and needs no action.
```

```changes
section: helm_lib
type: fix
summary: Size the resources of a module whose `resourcesManagement` names a fractional quantity, instead of rendering a zero limit.
```


deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:user-authn-fractional-cpu-init-limits` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/35430485615) in 40m53s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@111.88.245.212`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 39m31s.

</details>

helm_lib/charts/deckhouse_lib_helm/templates/_resources_management.tpl
```diff
@@ -94,46 +94,63 @@ updatePolicy:
...
+    {{- $mantissa = trimSuffix "m" $units -}}
+    {{- $factor = 1 -}}
+  {{- end -}}
+  {{- if regexMatch "^[0-9]+$" $mantissa -}}
+    {{- $mantissa | atoi | mul $factor -}}
+  {{- else if regexMatch "^[0-9]*[.]?[0-9]+([eE][+-]?[0-9]+)?$" $mantissa -}}
+    {{- $mantissa | float64 | mulf (float64 $factor) | ceil | int64 -}}
   {{- else -}}
-    {{- atoi $units | mul 1000 -}}
+    {{- cat "ERROR: unknown cpu format:" $units | fail -}}
```
AlwxSin: 🟡 **The CPU helper gained a `fail` branch it never had, on CRD-valid input**

The old CPU helper had no `fail` path at all — every input produced a number. This one aborts on anything outside the two regexes above. Measured against sprig v3.3.0:

```
"+1"    -> FAIL: ERROR: unknown cpu format: +1      (was 1000m — rendered correctly)
"-1"    -> FAIL                                      (was -1000m)
"1."    -> FAIL                                      (was 0)
"100u"  -> FAIL                                      (was 0)
"100n"  -> FAIL                                      (was 0)
"1k"    -> FAIL                                      (was 0)
```

All six pass the DexAuthenticator CRD pattern (`modules/150-user-authn/crds/dex-authenticator.yaml:291`) — the standard Kubernetes quantity regex, which allows a leading `+`/`-`, a trailing `.`, and the `n`/`u`/`k`/`M`/`G` suffixes. There is no Quantity canonicalization for CRD `additionalProperties`, so the value reaches the template verbatim.

`fail` inside a template aborts the render of the whole module and stops the `main` queue. That is the same failure this PR exists to eliminate — the description lists it as fixed defect #3 ("the helm_lib helper calls `fail` and takes down the rendering of the whole module") — reached through a different input, and now in a shared helper rather than one module. `+1` is the sharp case: it rendered correctly before and kills the release now.

`^[+-]?` plus `[0-9]+\.?` in the second regex covers the sign and trailing-dot forms; `n` and `u` want their own factor branches alongside `m`.

modules/150-user-authn/template_tests/authenticator_test.go
```diff
@@ -369,6 +369,40 @@ var _ = Describe("Module :: user-authn :: helm template :: dex authenticator", f
...
+      ingressSecretName: test
+    resources:
+      limits:
+        cpu: "0.5"
+        memory: "0.5Gi"
+      redis:
+        limits:
+          cpu: "0.5"
+          memory: "512Mi"
+- name: test-tiny-resources
```
AlwxSin, line 391: 🟡 **A shared helper changed, tested only through one module's rendered output**

The whole test budget for this change is two DexAuthenticator render assertions. Nothing exercises the converters directly, and nothing covers the cni-cilium / istio `resourcesManagement` + `limitRatio` path that the same change now also affects — which is where the identical `limits.cpu: 0m` against `requests.cpu: 0.5` was sitting unnoticed. No tests ship with the vendored chart either (`find helm_lib -name '*test*'` is empty); they live upstream in `deckhouse/lib-helm`.

The gap that bites soonest: a bare fractional memory like `memory: "0.5"` — the case the PR description calls out by name as one of three fixed, and the one that previously made the helper `fail` — is still unexercised. `test-fractional-resources` covers `0.5` CPU and `0.5Gi` memory; `test-tiny-resources` covers the clamp. The third is asserted in prose only, and it is the nastiest of the three: a module-wide render failure rather than one bad object.

If the upstream `deckhouse/lib-helm` 1.72.23 release carries tests for these two helpers, saying so in the PR description closes this. If it does not, a table-driven case per quantity form is worth more than another render assertion here.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
```diff
@@ -39,8 +41,8 @@ memory: 25Mi
       {{- if $redisMem }}{{ $memBytes = add $memBytes (include "helm_lib_resources_management_memory_units_to_bytes" $redisMem | int) }}{{ end -}}
     {{- end -}}
   {{- end -}}
-cpu: {{ printf "%dm" $cpuMillicores }}
-memory: {{ div $memBytes 1048576 }}Mi
+cpu: {{ printf "%dm" (max $cpuMillicores 10) }}
+memory: {{ printf "%dMi" (max (div $memBytes 1048576) 10) }}
```
AlwxSin: 🔵 **The module truncates against the helper's new explicit promise**

Both helpers now state, in `_resources_management.tpl:98` and `:120`: *"A fractional result is rounded up, so a limit is never smaller than the value asked for."* This line then does `div $memBytes 1048576`, truncating to whole Mi — `0.7Gi` lands at `716Mi` against a real `716.8Mi`.

Pre-existing and harmless in effect (staying under the sum is the documented trade-off, since a quota counts `max(init, sum(app))`), but the promise is now written down and the only in-tree caller quietly breaks it. Either round up here too, or scope the helper comment to the helper.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
```diff
@@ -10,6 +10,8 @@ memory: 25Mi
 
 {{- /* Init limits = sum of main container limits when set in CR (ResourceQuota: max(init, sum(app))). */ -}}
 {{- /* Defaults 20m / 50Mi when resources are not set = sum of dex+redis default requests. */ -}}
+{{- /* The result never drops below the init container requests (10m / 10Mi): a lower limit makes the */ -}}
+{{- /* API server reject the Deployment. Staying under the sum is safe, a quota counts max(init, sum(app)). */ -}}
 {{- define "dex_authenticator_init_limits" -}}
```
AlwxSin: 🔵 **The sum machinery is still redundant next to its own sibling**

Not a blocker for this hotfix — follow-up material.

`dex_authenticator_init_limits` survived the rewrite at ~40 lines. Commit 62814b0a78 (#21313) — the one that introduced this defect — gave the *same* `self-signed-generator` init container in `modules/150-user-authn/templates/user-api/deployment.yaml:85-87` a literal `cpu: 10m / memory: 25Mi` in the same change.

A ResourceQuota counts a pod as `max(max(init), sum(app))`, so literals produce an identical result in every case:

- `sum(app) >= 10m` → the quota counts `sum(app)` either way;
- `sum(app) < 10m` → the quota counts `10m`, which is exactly what the new clamp yields;
- `spec.resources` unset → the app containers carry no limits at all, so a limits-scoped quota rejects the pod either way and the `20m`/`50Mi` default is decoration.

Under a LimitRange `maxLimitRequestRatio` literals are strictly better: the ratio against `requests: 10m` is 1 instead of `sum(app)/10m`. The only thing the sum buys is runtime headroom for the init container, which this PR already gives up in the clamped case.

helm_lib/charts/deckhouse_lib_helm/templates/_module_controller.tpl
```diff
@@ -635,6 +644,31 @@ roleRef:
...
+       (TokenReview) and whether they hold get on the prometheus-metrics subresource
+       (SubjectAccessReview). Both verbs live in d8:rbac-proxy, the ClusterRole
+       Deckhouse ships for exactly this. The binding is rendered here rather than
+       left to the module because the sidecar it serves is rendered by this helper's
+       twin: a module that sets controllerMetricsProxyPort in one config and forgets
+       the binding in the other gets a proxy that rejects Prometheus and no other
+       sign of it. */}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
AlwxSin, line 659: 🔵 **The chart bump sweeps 1.72.22 into a patch-release hotfix**

The repo was pinned at `1.72.21`, so `→ 1.72.23` also pulls `1.72.22`, which adds this `d8:rbac-proxy` ClusterRoleBinding to `helm_lib_module_controller_rbac`.

It is gated on `controllerMetricsProxyPort`, and no module in this repo sets that key (`grep -rn controllerMetricsProxyPort modules/ ee/` only hits `_module_controller.tpl` itself), so the binding renders for nobody and is inert here. Worth a line in the PR description so a release reviewer does not have to establish that themselves before signing off on a new ClusterRoleBinding in a patch release.

helm_lib/charts/deckhouse_lib_helm/templates/_resources_management.tpl
Suselz: Confirmed, and the same defect sits in the memory helper, which the two regexes narrowed the same way:

```
mem "+1Gi"   FAIL   was 1073741824
mem "abcGi"  FAIL   was 0
```

Fixed in deckhouse/lib-helm#218, released as `1.72.24`. Both helpers now parse the grammar they are given: an optional sign, a decimal mantissa with either side of the point optional, a decimal exponent, and the `n`, `u`, `m`, `k`, `M`, `G`, `T`, `P`, `E` and `Ki`...`Ei` suffixes. Exact integer arithmetic still covers an integer mantissa, so no value changes the result it has today. Measured on your inputs:

```
cpu "+1"    -> 1000       cpu "1k"     -> 1000000
cpu "-1"    -> -1000      cpu "abc"    -> 0
cpu "1."    -> 1000       mem "+1Gi"   -> 1073741824
cpu "100u"  -> 1          mem "abcGi"  -> 0
cpu "100n"  -> 1
```

Nothing that used to produce a number aborts any more. An unparsable CPU value yields `0` as it did before `1.72.23`, and a memory value with a known suffix and an unparsable mantissa does the same; the `fail` of the memory helper survives only where it already was.

This branch pins `1.72.23` today and has to move to `1.72.24` before it merges. The description says so.

Independently, 37bfb5e removes the caller: the init container now carries literal resources, so nothing on the DexAuthenticator path converts a quantity at all.

modules/150-user-authn/template_tests/authenticator_test.go
Suselz: Partly. The bare fractional memory case is covered upstream already: `1.72.23` asserts that `helm_lib_resources_management_memory_units_to_bytes "0.5"` is `1`, alongside `0.5Gi`, `1.5Gi` and `0.25Mi`. The CPU suite covers `0.5`, `1.5`, `.25`, an unquoted `0.5` and `0.5m`.

The gap you point at is real one level up. `helm_lib_resources_management_original_pod_resources` only asserted that the rendered block was non-empty, which a `limits.cpu: 0m` against `requests.cpu: 0.5` satisfies — that is why the defect sat there unnoticed. deckhouse/lib-helm#218 adds the exact assertion:

```yaml
- equal: {path: result_fractional_limits_cpu, value: "750m"}   # 0.5 cores * 1.5
- equal: {path: result_fractional_requests_cpu, value: "0.5"}
```

plus a case per quantity form, including the sign, the trailing dot, the exponent and every suffix.

In this repository the module no longer calls the converters at all (37bfb5e), so its tests now assert what the module does promise: the init container carries the same resources for a fractional CR, a tiny CR, a whole-unit CR and no `resources` at all, and the main containers still receive the CR limits untouched. The description now names where the helper tests live.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
Suselz: Gone with the define. 37bfb5e removes `dex_authenticator_init_limits`, so there is no `$memBytes` and no `div` left — the module does not convert a quantity any more, and nothing here can contradict what the helper promises.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
Suselz: Done in 37bfb5e, with the literals `user-api` already carries: requests `10m` / `10Mi`, limits `10m` / `25Mi`.

Your accounting is what decided it. `max(max(init), sum(app))` makes the literals and the sum indistinguishable to a quota in all three cases, and under a `maxLimitRequestRatio` the literals are better. The runtime headroom the sum bought is the one real loss, and `user-api` runs the same binary doing the same work on those numbers.

Removing the define also removed the two findings above it: there is no conversion left to break on a valid quantity, and no `div` to truncate against the helper's promise.

helm_lib/charts/deckhouse_lib_helm/templates/_module_controller.tpl
Suselz: The description carries that line under "About the library bump", with the same `grep` and the same conclusion, so a release reviewer does not have to establish it. Left as is.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
```diff
@@ -179,13 +144,17 @@ spec:
...
+        {{- /* container limits, and they cannot land below the requests whatever the CR asks for. */}}
         resources:
           requests:
             memory: 10Mi
             cpu: 10m
             {{- include "helm_lib_module_ephemeral_storage_only_logs" . | nindent 12 }}
           limits:
-            {{- include "dex_authenticator_init_limits" ($crd.spec.resources | default dict) | nindent 12 }}
+            memory: 25Mi
+            cpu: 10m
```
AlwxSin, line 157: 🟡 **Every existing DexAuthenticator loses its init container's CPU headroom, and the changelog is silent**

The quota reasoning in the description is right, and I checked it: `max(max(init), sum(app))` accounts these literals identically to the old sum in all three cases you list. No objection to the approach — it is the correct simplification.

But the change is not limited to broken objects. A DexAuthenticator with `cpu: 300m` on the main container had an init limit of `300m` and now has `10m`. `self-signed-generator` generates a self-signed certificate, RSA keygen is CPU-bound, and capping the burst at 1% of a core stretches wall-clock at pod start by a large factor. Requests were always `10m`, so nothing was guaranteed — but the headroom was real and every existing object loses it on the next rollout.

`user-api` carrying `cpu: 10m` limits for the same binary since #21313 is a genuinely good precedent and I would not block on this. What is missing is the paper trail: the `user-authn` changelog entry describes only the bugfix, while the earlier revision of this PR carried a full `impact` block for a strictly narrower behavioural change.

Either a measured startup number on a DexAuthenticator that previously had a large sum, or an `impact` line noting that init container startup may take longer.

AlwxSin (commented):
🔵 **The description's opening contradicts its own "About the library bump" section**

Paragraph 3 says:

> It also moves `LIB_HELM_VERSION` from `1.72.21`, which carries the quantity-conversion fixes **the main containers** and other modules depend on.

The main containers do not depend on it — they receive the CR limits through `toYaml` verbatim (`deployment.yaml`, `containers[0]`/`containers[1]` `resources`), with no conversion anywhere on that path. The later section states it correctly: *"The bump is no longer what fixes this module, since the module no longer converts anything."* The opening is the paragraph a release reviewer reads first, so it is worth correcting there.

No anchor — this one is about the description, not a line of code.

modules/150-user-authn/templates/dex-authenticator/deployment.yaml
Suselz: Measured, and you are right that it is not free.

Ran the `self-signed-generator` image from this cluster twice at each limit, identical pods differing only in the init container's CPU limit:

```
certgen-10m     limit=10m    4s
certgen-10m-b   limit=10m    3s
certgen-300m    limit=300m   0s
certgen-300m-b  limit=300m   0s
```

So three to four seconds against under a second, once per pod start.

Both changelog entries now carry that, in an `impact` block with the numbers. The level stays the default rather than `high`, on the argument that #21313 gave the identical init container of `user-api` a literal `cpu: 10m / memory: 25Mi` in the same commit: `10m` / `25Mi` was the intended size for `self-signed-generator` all along, and the sum was chosen for the ResourceQuota accounting its comment describes, not to hand the init container burst. The headroom was a side effect of that choice, not a documented or configurable guarantee, so removing it reads as a defect going away rather than a default changing under an operator.

The paper trail you asked for is there either way — the entry states what changes and by how much, it just does not go into "Know Before Update".

On the inconsistency you spotted: the earlier revision carried an `impact` block and I removed it, because the clamp then only affected objects the API server was already rejecting. Switching to literals moved that line, so the block is back.

AlwxSin (approved):

deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:user-authn-fractional-cpu-init-limits` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/35616424936) in 1h16m3s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@93.77.188.96`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 48m26s.

</details>

ldmonster (approved):

PR #23171, opened 2026-09-18
2c

W1seKappa:
## Description

* **Every join attempt is checked against the API.** Bootstrap and the rejoin loop share one attempt:
  1. a consistent `GET` of the own Node checks that the API answers, that the NodeGroup label is right and that the UID is unchanged;
  2. parallel `GET`s of up to three candidate Nodes: two peers that gossip does not list and one that it does;
  3. one `memberlist.Join` per validated seed.

  Seed addresses come only from these fresh answers, never from the informer cache. A NotFound, a foreign label or a UID mismatch refuses the join. Every failed attempt doubles the delay. The rejoin loop no longer reads the fallback heartbeat verdict.
* **Second rejoin trigger.** The loop also runs while a peer's `FencingFailedNodeState` records this node as failed and the node still holds gossip quorum. Only records created during the agent's current life count.
* **Writer pause.** A peer's failed record about the agent's own node can stay for one takeover step (`rejoin.maxInterval`) under one UID. The agent's fencing state writer then stops creating, marking and removing records. The only exception is its own verdicts about peers it sees alive. On resume, every open incident restarts its `detectedAt` and its takeover clock.
* **Fresh `detectedAt` for re-created records.** A record can disappear while the failure persists. It is then re-created with the time of the write, and the takeover order counts from the disappearance.
* **memberlist timings derived from the SLA profile.**

  | Timing | Formula |
  |---|---|
  | TCP timeout | `max(fallback.kubernetesAPITimeout, 1s)` |
  | push/pull interval | `max(gossipToTheDeadTime, 2 × TCP timeout)` |
  | dead node reclaim time | `gossipToTheDeadTime` |
  | leave timeout | `min(max(2 × retransmitMult × gossipInterval, 250ms), 4s)` |

  Before, every profile ran with 10s / 30s / 10s / 3s. The derived values are logged in the startup line:

  | Profile | TCP timeout | push/pull | reclaim | leave |
  |---|---|---|---|---|
  | Critical | 1s | 2s | 2s | 800ms |
  | Medium | 2s | 5s | 5s | 1.6s |
  | Moderate | 5s | 10s | 10s | 2.4s |
  | Slow | 10s | 20s | 20s | 4s |
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
type: fix | feature | chore
```


z9r5 (approved):
Will review on merging to the main branch

PR #23180, opened 2026-09-21
[user-authz][multitenancy-manager] Document RBACv2 and project follow-ups

Jabejixo:
## Description

Documentation follow-ups for the RBACv2 review nits and the multitenancy-manager project follow-ups in #23124. This PR is stacked on #23124 and carries only documentation, so that the code PR stays free of paired-document changes.

- **RBACv2 migration guide (EN + RU)**: a new section on what the compatibility aliases do *not* carry, so module authors know which parts of the old role model stop working under the aliases and have to be migrated explicitly.
- **user-authz README (EN + RU)**: a note for module authors that the system-resource marker is a **label**, not an annotation. The protection is applied by the API server through admission `matchConditions`, which read labels; an object marked with an annotation is silently left writable.
- **multitenancy-manager USAGE (EN + RU)**: a section on how `ClusterResourceGrantPolicy.projectSelector` selects projects. The selector is matched per namespace against the union of the Project labels and the namespace labels (namespace wins on conflicts); there is no separate `namespaceSelector`.
- **multitenancy-manager USAGE (EN + RU)**: the adoption section now says how to remove an adopted environment: delete the Project. `kubectl delete project <name>` deletes the namespace as well and returns only once the namespace is gone, as `kubectl delete ns` used to.

- **user-authz migration guide (EN + RU)**: the pre-upgrade query for ClusterRoles that carry a platform aggregation label without being a custom capability, with the metric and alert #23124 adds for them.
- **Spelling wordlist**: the terms these pages use (`ClusterResourceGrantPolicy`, `superadmin`, a few Russian derivations, …) are added to `tools/docs/spelling/wordlist` and the dictionary is regenerated with `make docs-spellcheck-generate-dictionary`, so the spellcheck job passes on whole-file checks instead of being skipped by label.

The behaviour described here is already in the code (#23124 and its base) but was not documented: users asked whether a `namespaceSelector` appears alongside `projectSelector`, module authors have used an annotation instead of the marker label, and the change in what `kubectl delete project` waits for was invisible without reading the controller.

Documentation matches the behaviour shipped in #23124. No code changes.

## Checklist
- [x] The code is covered by unit tests. (n/a, documentation only)
- [x] e2e tests passed. (n/a, documentation only)
- [x] Changes were tested in the Kubernetes cluster manually. (n/a, documentation only)

## Changelog entries

```changes
section: user-authz
type: fix
summary: The RBACv2 migration guide describes what the compatibility aliases do not carry, and the README tells module authors that the system-resource marker is a label, not an annotation.
impact_level: low
```

```changes
section: multitenancy-manager
type: fix
summary: The usage guide explains how `ClusterResourceGrantPolicy.projectSelector` matches projects (union of Project and namespace labels) and how to remove an adopted environment by deleting the Project.
impact_level: low
```


modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -593,6 +593,10 @@ The following components are used for this:
 1. `ValidatingAdmissionPolicyBinding`: Defines which objects the validation applies to:
    - Uses `namespaceSelector` and `objectSelector` to select resources by the label `heritage: multitenancy-manager`.
 
+### Who bypasses admission
+
+The module's admission webhooks (on Project, ProjectTemplate, ProjectRoleBinding, ProjectNamespace and ClusterProjectRoleBinding) and the `ValidatingAdmissionPolicy` above follow one rule: requests from cluster components (the API server, the Deckhouse and module service accounts, kubelets) and from the `system:sudouser` identity are not checked; everyone else is, `system:masters` included. An administrator who has to get past a check does it deliberately, by impersonating that identity — `kubectl --as system:sudouser …` — and the bypass is then visible in the audit log.
```
AlwxSin: 🔴 **1. The documented break-glass command does not work for three of the five webhooks named here**

This sentence names five webhooks and then prescribes `kubectl --as system:sudouser …` as the way past them. That only holds for two of them.

In `templates/admission/validation.yaml` the `$excludeSystemWriters` block is applied at `:50` (projects) and `:151` (projecttemplates) only. `projectrolebindings` (`:59`), `projectnamespaces` (`:83`) and `clusterprojectrolebindings` (`:106`) carry no `matchConditions` at all, and their handler exempts just two service accounts:

```go
// internal/webhook/rolebinding/validator.go:82-83
user := req.UserInfo.Username
privileged := user == ControllerServiceAccount || user == DeckhouseServiceAccount
```

`git grep sudouser` over `modules/160-multitenancy-manager/images/` returns zero hits. An administrator following this on a ProjectRoleBinding gets denied, and the page offers no alternative.

Either narrow the list to Project and ProjectTemplate, or state explicitly that there is no bypass for the other three.

modules/140-user-authz/docs/internal/RBACV2_MODULE_MIGRATION.md
```diff
@@ -273,6 +273,21 @@ d8 k get clusterrole d8:namespace:viewer -o json | jq '[.rules[] | select(.apiGr
...
+
+- **Namespaced projection.** A ClusterRoleBinding to a `d8:manage:*` alias no longer produces the
+  RoleBindings that carried it into the module namespaces, and the ones the previous release created
+  are removed. Cluster-scoped rights stay; logs, `describe` and `exec` inside `d8-*` namespaces do
+  not. Recreating the binding on the new name restores them.
+- **`serviceaccounts/token` and `impersonate`**, which moved from admin to superadmin by design.
+
+Check your own documentation as well: examples that name `d8:manage:*` or `d8:use:role:*` keep
+working only for this release, and inside a project namespace a RoleBinding to an alias is already
+refused, because aliases are not delegatable.
```
AlwxSin: 🔴 **2. This is backwards for `d8:use:role:*` — and so is the reason given**

The `d8:use:role:*` aliases are marked delegatable precisely so that a RoleBinding to them inside a project namespace is *not* refused:

```
# templates/rbacv2-compat/legacy-role-aliases.yaml:43-49
These are marked delegatable, exactly like the d8:namespace:<level> roles they alias. Without the
marker … a RoleBinding to an alias inside a project namespace is refused by
cluster-objects-grants-validator. … the compatibility layer would be dead on arrival
```

`:62-63` sets `"delegatable" true`, `:99-101` renders `rbac.deckhouse.io/delegatable: "true"`, and `grantable-resources.yaml:155-159` excludes only roles that are *not* `"true"`.

What actually gets refused is a **new ProjectRoleBinding/ClusterProjectRoleBinding** pointing at an old name, via `rbac.deckhouse.io/disabled-for-direct-use-in-projects` (`internal/webhook/rolebinding/validator.go:132-138`) — a different object, a different reason. For `d8:manage:*` the sentence is technically right, but nobody binds those with a RoleBinding.

Likely source: the stale header comment at `legacy-role-aliases.yaml:13-15` ("Aliases are also not `delegatable`…"), which contradicts `:43-49` of the same file. Worth fixing that comment too.

modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -669,13 +673,23 @@ The module allows you to manage project access to cluster-wide resources such as
...
+
 The following sections provide common scenarios for configuring and using the mechanism.
 
 ### For cluster administrators
 
-The following examples show how to configure project access to cluster-wide resources using ClusterResourceGrantPolicy.
-
-`projectSelector` is matched against the union of the labels of the Project object and the labels of each namespace of the project; when the same key is set on both, the namespace value wins. A label on the Project therefore selects its main namespace and all its additional namespaces. The examples below put the label on the Project, which is the only place a project's own labels are set: a ProjectNamespace does not carry labels, and the labels a template adds through `namespaceMetadata.labels` are the same in every namespace of the project.
+The following examples show how to configure project access to cluster-wide resources using ClusterResourceGrantPolicy. They put the label on the Project, which selects every namespace of the project; the labels a template adds through `namespaceMetadata.labels` are the same in every namespace of the project and can be used the same way.
```
AlwxSin: 🔴 **3. `namespaceMetadata.labels` cannot be used to select a project's namespaces**

"… and can be used the same way" does not hold. Those labels only ever reach the main namespace.

`internal/helm/renderer.go:142-147` keeps only the Namespace whose name equals the project name, so template-rendered labels never reach additional namespaces. Additional namespaces are built by `internal/controllers/projectnamespace/reconciler.go`, which stamps three ownership labels and inherits a fixed allowlist:

```go
// reconciler.go:138-144
var inheritedNamespaceLabels = []string{
	"security.deckhouse.io/pod-policy",
	"extended-monitoring.deckhouse.io/enabled",
	"security-scanning.deckhouse.io/enabled",
	v1alpha3.ResourceLabelTemplate,
}
```

Anything outside that list is actively deleted on update (`:183-188`). A `projectSelector` written against an arbitrary `namespaceMetadata` label matches the main namespace and silently skips the rest — in an access-control mechanism.

The "same in every namespace" half predates this PR, but rewriting the sentence to add "can be used the same way" turns a stale factoid into actionable wrong advice. Same line in `USAGE_RU.md:698`.

modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -593,6 +593,10 @@ The following components are used for this:
 1. `ValidatingAdmissionPolicyBinding`: Defines which objects the validation applies to:
    - Uses `namespaceSelector` and `objectSelector` to select resources by the label `heritage: multitenancy-manager`.
 
+### Who bypasses admission
```
AlwxSin, line 596: 🟡 **4. "Follow one rule" is not accurate, and `system:masters` is exempt on the module's other webhooks**

Two problems with this section's framing.

**The two lists differ.** The VAP additionally exempts `system:kube-controller-manager`, `system:kube-scheduler`, `system:volume-scheduler`, `dhctl`, `observability`, plus a rollout-restart exemption. The webhooks additionally exempt `system:serviceaccount:d8-user-authz:controller` and the `system:serviceaccounts:d8-user-authz` group.

**`system:masters` is not universally checked.** The claim is made about "the module's admission webhooks" generally, but the grant webhooks `/is-granted` and `/defaults` exempt it outright:

```go
// hooks/configure_grant_validation_webhook.go:91-94
Name:       "exclude-cluster-admins-and-nodes",
Expression: `!request.userInfo.groups.exists(g, g == "system:masters" || g == "system:nodes")`,
```

Those three webhooks (`/is-granted`, `/defaults`, `/protect`) aren't listed here either.

Also worth reconciling with the existing line six above (`USAGE.md:591`: "only operations on behalf of the controller's service account are allowed"), which this section now contradicts.

modules/140-user-authz/docs/README.md
```diff
@@ -275,6 +275,8 @@ What is forbidden at the `admin` level and allowed only at the `superadmin` leve
...
 - **Connecting to system pods** — `kubectl exec`, `kubectl attach`, and `kubectl port-forward` into a pod labelled `deckhouse.io/system-resource: "true"` are available to `superadmin` only. This protects against reading foreign secrets and interfering with platform components from inside their pods.
 
+  For module authors: the marker is a **label**, `deckhouse.io/system-resource: "true"`, on the object itself — not an annotation. The objects to protect are selected on the API server side, in the admission `matchConditions`, and those read labels; an object marked with an annotation is not seen by the webhook and stays writable, with nothing to say so. Put the label on every namespaced object your module places into user namespaces (pods, services, PVCs, secrets, custom resources): it then becomes writable by `superadmin` and cluster administrators only, and its pods reachable by `exec`, `attach` and `port-forward` only for them.
```
AlwxSin: 🟡 **5. The `matchConditions` explanation covers only half the protection**

True for the edit binding (`webhooks/validating/system_resources.py:205-219`, condition `only-marked-objects`). Not true for the exec binding: its `matchConditions` (`:230-240`) contain no marker filter at all, the rule matches CONNECT on every pod, and the marker is resolved by reading the pod live. The file says why:

```python
# system_resources.py:75-77
# The exec path cannot use matchConditions (the admission object is PodExecOptions, not the Pod),
# so it reads that one pod live with kubectl
```

The paragraph's conclusion still holds for both paths — `is_system_pod` (`:414-422`) also reads `metadata.labels`, so an annotation is ignored either way. It's the stated mechanism that's wrong for exec. A one-clause caveat covers it.

modules/140-user-authz/docs/README_RU.md
```diff
@@ -281,6 +281,8 @@ RoleBinding, которые `multitenancy-manager` создаёт из ProjectRo
...
 - **Подключение к системным подам** — `kubectl exec`, `kubectl attach` и `kubectl port-forward` в под с лейблом `deckhouse.io/system-resource: "true"` доступны только `superadmin`. Это защищает от чтения чужих секретов и вмешательства в работу платформенных компонентов изнутри их подов.
 
+  Авторам модулей: маркер — это **лейбл** `deckhouse.io/system-resource: "true"` на самом объекте, а не аннотация. Объекты для защиты отбираются на стороне API-сервера, в `matchConditions` admission-конфигурации, а они читают лейблы; объект, помеченный аннотацией, вебхук не увидит, и он останется доступен на запись — без какого-либо сообщения об этом. Поставьте лейбл на каждый объект, который ваш модуль размещает в пользовательских неймспейсах (поды, сервисы, PVC, секреты, кастомные ресурсы), — и менять его смогут только `superadmin` и администраторы кластера, а `exec`, `attach` и `port-forward` в его поды будут доступны только им.
```
AlwxSin: 🟡 **6. Spellcheck CI is red and will block the merge** *(anchored here for visibility — the real fix is in `tools/docs/spelling/wordlist`, which this PR does not touch)*

`Docs Validation (spellcheck)` fails. `tools/docs/spelling/check_diff.sh` calls `file_check_spell ${file_name}` — it checks the **whole file**, not the changed lines — so touching these four files pulls in every pre-existing unknown word: `superadmin`, `labelled`, `ClusterProjectRoleBinding`, `ClusterResourceGrantPolicy`, `GrantableClusterResourceDefinition`, `AvailableClusterResource`, `подсистемная`, `апгрейда`, `адресно` and roughly fifteen more. None are in the wordlist.

#23124 carries `skip/documentation-validation`; this PR does not. Putting that label on a PR whose entire deliverable *is* documentation seems like the wrong lever — better to add the terms to `tools/docs/spelling/wordlist` and run `make docs-spellcheck-generate-dictionary` (the wordlist alone will trip `validate_dictionary_sync.sh`). Otherwise the next docs PR hits exactly this.

For context: `Tests DMT lint` is also red, but it fails identically on #23124 with no findings in either module's docs — pre-existing, not this PR.

modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -669,13 +673,23 @@ The module allows you to manage project access to cluster-wide resources such as
 
 For a description of the mechanism, the resources it uses, and the cluster-wide resources registered by the platform, refer to the [module description](./#managing-access-to-cluster-wide-resources).
 
+### How a policy selects its projects
```
AlwxSin, line 686: 🟡 **7. EN and RU place this heading differently; the EN intro sentence gets orphaned**

Here the heading is inserted *before* "The following sections provide common scenarios for configuring and using the mechanism." (`:690`), so that intro to the parent `## Managing access to cluster-wide resources` section now renders as the closing line of this subsection.

`USAGE_RU.md` puts the heading *after* the equivalent sentence (`:680` then `:682`), which is correct. Move this block down one paragraph to match.

modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -254,7 +254,7 @@ A namespace created directly (for example, `d8 k create ns my-app`) becomes a pr
 
 - the template is picked from what the namespace already carries: `secure` if it has the `security-scanning.deckhouse.io/enabled` label, `default` if it has `security.deckhouse.io/pod-policy` or `extended-monitoring.deckhouse.io/enabled`, and `simple` otherwise;
 - the project parameters are filled in from the current state of the namespace, so nothing inside it changes: the network policy stays unrestricted and the Pod Security Standard keeps the value the namespace already had;
-- from then on the project is the source of truth and is edited like any other project. Deleting the namespace no longer deletes the project — the project recreates the namespace.
+- from then on the project is the source of truth and is edited like any other project. Deleting the namespace no longer deletes the project — the project recreates the namespace. To remove the environment, delete the Project: `kubectl delete project <name>` removes the namespace with it and, like `kubectl delete ns` did, returns only once the namespace is gone.
```
AlwxSin: 🔵 **8. `kubectl` where the file uses `d8 k`**

`USAGE.md` uses `d8 k` 23 times; the only two `kubectl` occurrences in the file are the ones this PR adds (here and `:598`). `kubectl delete ns` as a historical comparison reads fine, but `kubectl delete project <name>` is a command the reader is meant to run. Same in `USAGE_RU.md:257`.

modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -669,13 +673,23 @@ The module allows you to manage project access to cluster-wide resources such as
 
 For a description of the mechanism, the resources it uses, and the cluster-wide resources registered by the platform, refer to the [module description](./#managing-access-to-cluster-wide-resources).
 
+### How a policy selects its projects
+
+A ClusterResourceGrantPolicy has one selector, `projectSelector`, and it is evaluated **per namespace**, not per Project object. For every namespace of a project the module merges two label sets — the labels of the Project object and the labels of that namespace — and matches the selector against the union. When the same key is set on both, the namespace value wins.
```
AlwxSin, line 688: 🔵 **9. The section is silent about the nil selector**

`internal/resolve/resolve.go:117-118` skips any policy with `ProjectSelector == nil`, and the CRD already documents it: "A nil selector matches no projects; an explicit empty selector matches all."

That isn't a corner case — `ProjectTemplate.resourceGrantPolicies` references library policies created deliberately without a selector. In a section titled "How a policy selects its projects", "has one selector" invites the opposite reading (that a policy without one applies everywhere).

modules/140-user-authz/docs/README.md
```diff
@@ -275,6 +275,8 @@ What is forbidden at the `admin` level and allowed only at the `superadmin` leve
 - **Modifying and deleting system resources in user namespaces.** Some platform components place their objects (for example, Dex authenticator pods, or pods and disks of virtual machines) directly in application namespaces. Such objects carry the `deckhouse.io/system-resource: "true"` label. Only `superadmin` may modify or delete them; for everyone else these operations are rejected at the API server level with an explanation.
 - **Connecting to system pods** — `kubectl exec`, `kubectl attach`, and `kubectl port-forward` into a pod labelled `deckhouse.io/system-resource: "true"` are available to `superadmin` only. This protects against reading foreign secrets and interfering with platform components from inside their pods.
```
AlwxSin, line 277: 🔵 **10. The module-author note is nested under the wrong bullet**

The new paragraph is indented two spaces, so it renders as a continuation of this bullet ("Connecting to system pods") only. Its content — and specifically the instruction "put the label on every namespaced object your module places into user namespaces" — applies equally to the bullet above it ("Modifying and deleting system resources in user namespaces").

A module author scanning for "how do I protect my objects" reads that bullet and stops. Unindenting the paragraph so it follows the list would cover both. Same in `README_RU.md:284`.

modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -669,13 +673,23 @@ The module allows you to manage project access to cluster-wide resources such as
...
+
+Which means, in practice:
+
+- a label on the **Project** selects the project's main namespace and all of its additional namespaces at once — this is the place to label a project as a whole, because a ProjectNamespace carries no labels of its own;
+- a label on a **Namespace** selects that namespace only — a per-namespace override, or a way to reach a namespace by a label the platform set (for example `projects.deckhouse.io/project=<name>`, which the module puts on every project namespace);
+- there is no separate `namespaceSelector`; both label sources feed the same `projectSelector`.
+
+A user interface that counts matching *Projects* by their own labels undercounts a policy written against a namespace label — the policy still applies, because the match is decided on the namespace.
```
AlwxSin: 🔵 **11. This documents a defect in an unnamed UI, in end-user docs**

The statement is accurate, but it's a bug report about some console aimed at readers who can't act on it. The actionable half — that the match is decided on the namespace — is already said two paragraphs up (`:678`). This feels like it belongs in an issue rather than USAGE. Same in `USAGE_RU.md:692`.

modules/160-multitenancy-manager/docs/USAGE.md
Jabejixo: Reworked after finding 10 on #23124: the exclusions apply to the Project and ProjectTemplate webhooks only, so the section (bade02d30c) now says the PRB/PN/CPRB webhooks skip nobody and that the `system:sudouser` door exists for Project, ProjectTemplate and the policy.

modules/140-user-authz/docs/internal/RBACV2_MODULE_MIGRATION.md
Jabejixo: Done in dba9113b28: the paragraph now says the `d8:use:role:*` aliases are delegatable, a plain RoleBinding to them works, and what is refused is a new PRB/CPRB to an old name. The stale header comment in `legacy-role-aliases.yaml` is fixed in #23124 (4b1e63fb34).

modules/160-multitenancy-manager/docs/USAGE.md
Jabejixo: Done in dba9113b28: the sentence now says `namespaceMetadata.labels` reach the main namespace only (additional namespaces get the ownership labels and the fixed inherited set), so label the Project when a policy has to cover the whole project. Same in RU.

modules/160-multitenancy-manager/docs/USAGE.md
Jabejixo: Done in dba9113b28: the section gives the two lists separately (webhooks vs VAP, including the control-plane identities and the rollout-restart exemption), names the grant webhooks that still skip `system:masters`, and the VAP bullet above no longer says the controller is the only allowed writer.

modules/140-user-authz/docs/README.md
Jabejixo: Done in dba9113b28: one clause added — for `exec`/`attach`/`port-forward` the admission object is the connect request, so the webhook reads the pod itself; the conclusion (labels, not annotations) stays. Same in RU.

modules/140-user-authz/docs/README_RU.md
Jabejixo: Done in dba9113b28: the terms are added to `tools/docs/spelling/wordlist` and `dev_OPS.dic` is regenerated with `make docs-spellcheck-generate-dictionary` (2911 lines both).

modules/160-multitenancy-manager/docs/USAGE.md
Jabejixo: Done in dba9113b28: the block now follows the intro sentence, as in RU.

modules/160-multitenancy-manager/docs/USAGE.md
Jabejixo: Done in dba9113b28: `d8 k delete project <name>` (and in RU); the historical `kubectl delete ns` comparison stays.

modules/160-multitenancy-manager/docs/USAGE.md
Jabejixo: Done in dba9113b28: a bullet says a policy without a selector matches no project on its own (a library policy applied through `resourceGrantPolicies`), and `projectSelector: {}` matches every project.

modules/140-user-authz/docs/README.md
Jabejixo: Done in dba9113b28: the paragraph is unindented and follows the list, in EN and RU.

modules/160-multitenancy-manager/docs/USAGE.md
Jabejixo: Done in dba9113b28: the UI sentence is removed (EN and RU); the Console side is tracked separately.

modules/160-multitenancy-manager/docs/USAGE.md
```diff
@@ -588,11 +588,21 @@ The following components are used for this:
...
    - Applies to all resources and API groups.
 1. `ValidatingAdmissionPolicyBinding`: Defines which objects the validation applies to:
    - Uses `namespaceSelector` and `objectSelector` to select resources by the label `heritage: multitenancy-manager`.
 
+### Who bypasses admission
+
+Two exclusion lists are in force, and they are not identical:
+
+- The module's admission webhooks — on Project, ProjectTemplate, ProjectRoleBinding, ProjectNamespace and ClusterProjectRoleBinding — skip requests from platform components only: the API server, the service accounts of Deckhouse, of this module and of `user-authz` (with the `d8-system`, `kube-system` and `d8-user-authz` service-account groups), kubelets — and from the `system:sudouser` identity. Everyone else is validated, `system:masters` included.
+- The `ValidatingAdmissionPolicy` above lets through this module's controller, the control-plane components (`system:kube-controller-manager`, `system:kube-scheduler`, `system:volume-scheduler`, `dhctl`, `observability`), the same service-account groups and kubelets, `system:sudouser`, and a rollout restart of a project workload. Everyone else is refused, `system:masters` included.
```
AlwxSin: 🔵 **13. "The same service-account groups" — they differ**

The rewrite is right that the two lists aren't identical, which is exactly why the enumeration should be exact. Two small gaps in this bullet:

**The groups aren't the same.** The webhook bullet above lists `d8-system`, `kube-system` and `d8-user-authz`. The VAP has only two:

```yaml
# templates/validation.yaml:28
- name: 'exclude-groups'
  expression: '!(["system:nodes", "system:serviceaccounts:kube-system", "system:serviceaccounts:d8-system"].exists(e, (e in request.userInfo.groups)))'
```

No `system:serviceaccounts:d8-user-authz`.

**`system:apiserver` is missing.** It's in the VAP's `exclude-users` (`templates/validation.yaml:30`) but isn't covered by the control-plane list as enumerated here.

Same bullet in `USAGE_RU.md:602` ("те же группы service account'ов").

Jabejixo: Done in bade02d30c: each mechanism now gets its own list verbatim — the VAP bullet names `system:apiserver` and the two service-account groups (not `d8-user-authz`), and the PRB/PN/CPRB webhooks are listed separately as skipping nobody.

AlwxSin (approved):

PR #23181, opened 2026-09-21
[user-authn] Fix the init container limits of a DexAuthenticator

Suselz:
## Description

Backport of #23170 to `release-1.76`, touching the module only.

The limits of the `self-signed-generator` init container of a DexAuthenticator Deployment were the
sum of the `dex-authenticator` and `redis` limits taken from the custom resource, while the requests
of that container are hardcoded as `10m` and `10Mi`. Whenever that sum landed below the requests the
API server rejected the Deployment.

The init container now carries fixed resources, the same literals the identical init container of
`user-api` has had since the sum was introduced:

```yaml
          requests:
            memory: 10Mi
            cpu: 10m
          limits:
            memory: 25Mi
            cpu: 10m
```

No value from the custom resource reaches the init container any more, so no input can put its
limits below its requests.

Only DexAuthenticator Deployments are re-rendered. No control plane component, ingress controller or
Prometheus instance is disturbed.

The release of `user-authn` never completes, and the `main` queue of Deckhouse stops on the first
rejected object, which blocks every subsequent task in the cluster:

```
Deployment.apps "<name>-dex-authenticator" is invalid:
spec.template.spec.initContainers[0].resources.requests: Invalid value: "10m":
must be less than or equal to cpu limit of 0
```

Two inputs reach that state. A fractional quantity such as `cpu: "0.5"` converts to `0` through the
`atoi` in the helm_lib helpers, and a limit that is simply smaller than the requests, such as a
single `cpu: 5m`, produces a sum below them honestly. Both are valid input: the CRD accepts a
fractional quantity through the standard Kubernetes quantity pattern, and `0.5` and `500m` mean the
same thing.

## Why literals rather than a fixed conversion

A ResourceQuota counts a pod as `max(max(init), sum(app))`, so literals are accounted exactly like
the sum in every case:

- the main containers ask for more than `10m` — the quota counts them, either way;
- they ask for less — the quota counts the init container;
- `spec.resources` is unset — the main containers carry no limits at all, so a limits-scoped quota
  rejects the pod whatever the init container says.

What the sum bought was runtime headroom for the init container, and `user-api` shows `10m` /
`25Mi` is enough for the same binary doing the same work.

This also keeps the backport inside the module. The conversion defect itself belongs to helm_lib and
is fixed upstream, but this branch vendors `deckhouse_lib_helm` `1.71.8`, so taking that fix would
mean moving a shared library across two minor versions inside a patch release. Since the init
container no longer converts anything, none of that is needed here.

The defect reached `v1.76.14`, the current release of this branch, and it stops the `main` queue of
an affected cluster until the custom resource is edited by hand. A cluster that provisions
DexAuthenticator objects from a template hits it again on every new object. Until this ships, the
only remedy is to rewrite every fractional value by hand.

## Verification

The template tests assert what the module promises: the init container carries the same resources
for a fractional CR, a tiny CR, a whole-unit CR and no `resources` at all, and the main containers
still receive the CR limits untouched.

Checked on a `v1.76.12` cluster with the build of this branch. Six DexAuthenticator objects,
covering a fractional CPU, a fractional memory suffix, a bare fractional memory, limits below the
init container requests, whole millicores and no `resources` at all. All six rendered, and all six
carry the same init container:

```
OBJECT       INIT_LIMITS                      INIT_REQUESTS
frac-cpu     {"cpu":"10m","memory":"25Mi"}    {"cpu":"10m","memory":"10Mi"}
frac-mem     {"cpu":"10m","memory":"25Mi"}    {"cpu":"10m","memory":"10Mi"}
tiny         {"cpu":"10m","memory":"25Mi"}    {"cpu":"10m","memory":"10Mi"}
millicores   {"cpu":"10m","memory":"25Mi"}    {"cpu":"10m","memory":"10Mi"}
defaults     {"cpu":"10m","memory":"25Mi"}    {"cpu":"10m","memory":"10Mi"}
bare-frac    {"cpu":"10m","memory":"25Mi"}    {"cpu":"10m","memory":"10Mi"}
```

Every Deployment was applied, so the release of `user-authn` completed, which is what stopped
happening before this change. `self-signed-generator` terminated with exit code 0 in the pods that
started, and `frac-cpu`, `frac-mem` and `defaults` reached 2/2.

Three objects do not reach a running state, none of them for a reason this change introduces:

- `tiny` is `OOMKilled` on its own `memory: 8Mi`;
- `bare-frac` cannot start with its own half a byte — the runtime reports
  `unable to set memory limit to 1`;
- `millicores` stays `Pending` on `Insufficient cpu`, because the test objects declare limits
  without requests and Kubernetes defaults the requests to them.

On `v1.76.14` the first two would have no Deployment at all: `tiny` was rejected for a limit below
its requests, and `bare-frac` took down the rendering of the whole module.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: user-authn
type: fix
summary: Render a valid Deployment for a DexAuthenticator whose `resources` limits are fractional or below the init container requests.
impact: |
  The `self-signed-generator` init container now carries the fixed `10m` / `25Mi` it was always
  meant to have, the same as the identical init container of `user-api`, instead of the sum of the
  main container limits. A DexAuthenticator that set large limits no longer gets that sum as burst:
  generating the self-signed certificate then takes 3 to 4 seconds instead of under a second. It
  happens once per pod start and needs no action.
```


deckhouse-BOaTswain:
:green_circle: `e2e: Yandex.Cloud` for `deckhouse:user-authn-init-limits-1.76` [succeeded](https://github.com/deckhouse/deckhouse/actions/runs/35604659581) in 1h7m56s.



Yandex.Cloud-WithoutNAT-Containerd-1.34 - Connection string: `ssh  cloud-user@111.88.249.58`


:green_circle: `e2e: Yandex.Cloud, Containerd, Kubernetes 1.34` succeeded in 1h6m4s.

</details>

AlwxSin (approved):

PR #23216, opened 2026-09-22
[multitenancy-manager] Scope GrantableClusterResourceReference field paths by resource

AlwxSin:
## Description

`GrantableClusterResourceReference.spec.fieldPaths[]` entries could be scoped by API group and version only, and `SelectFieldPath` returned the first scoped entry that matched. Paths that differ by **resource** within one group and version were therefore indistinguishable.

- core/v1: `pods` hold the name at `$.spec.priorityClassName`, `replicationcontrollers` at `$.spec.template.spec.priorityClassName`
- batch/v1: `jobs` at `$.spec.template.spec.priorityClassName`, `cronjobs` at `$.spec.jobTemplate.spec.template.spec.priorityClassName`

The first entry won for both members of each pair, with two separate consequences: the validating webhook read a path absent from the other resource's schema, got no values and skipped the check silently; the defaults webhook, for an entry with `FillEmpty` or `Coerce`, would have written the name into a path the object does not have. Describing such a cluster resource required one `GrantableClusterResourceReference` per Kubernetes resource.

This PR adds an optional `resources` scope to `FieldPath` and changes `SelectFieldPath` to pick the **most specific** matching entry instead of the first one:

- `resources` (weight 4) outranks `apiGroups` (2), which outranks `apiVersions` (1); the weights add up;
- equal weights go to the earliest entry in the list, so the choice stays deterministic;
- an empty dimension still matches anything, so an unscoped entry remains the fallback.

The CRD and its `doc-ru` counterpart, the deepcopy, the design note and the usage docs are updated along with the code.

Cluster resources referenced from a pod template — `PriorityClass`, `RuntimeClass` — are used by every workload kind, and the path to the reference depends on the kind. Without a resource scope, each of them needs its own reference object, and the collisions inside `core/v1` and `batch/v1` above cannot be expressed at all: whichever entry comes first silently decides for the other resource too. This is groundwork for registering those cluster resources as grantable.

The field is optional and additive. Every reference shipped by the module (`templates/cluster-objects-controller/grantable-resources.yaml`) carries a single unscoped `fieldPaths` entry and keeps selecting it. The only behaviour change is on objects that already have two matching **scoped** entries, where the more specific one now wins instead of the first in the list; no such object exists in the repository.

Not necessarily.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: multitenancy-manager
type: feature
summary: A GrantableClusterResourceReference field path can be scoped to specific resources, so paths that differ by resource within one API group and version no longer need a separate reference object.
impact_level: low
```


modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/engine/engine.go
```diff
@@ -45,29 +45,46 @@ func RuleMatches(rule v1alpha1.UsageRule, group, version, resource string) bool
...
-			return *fp, true // scoped entry wins
+		if len(fp.APIGroups) > 0 && !listMatches(fp.APIGroups, group) {
+			continue
+		}
+		if len(fp.APIVersions) > 0 && !listMatches(fp.APIVersions, version) {
+			continue
+		}
+		weight := 0
+		if len(fp.Resources) > 0 {
+			weight += 4
```
Jabejixo: 🟡 **`*` in a scope dimension counts as the most specific entry**

A `*` in a scope dimension is counted as if it were a specific name: `resources: ["*"]` scores 4 and outranks an explicit `apiGroups: [batch], apiVersions: [v1]` (3), and ties with an explicit `pods` on list order. Semantically `*` is the absence of a restriction, and the CRD text says the most specific entry wins and the unrestricted one is the fallback.

Suggest scoring a dimension only when its list has no `*`, and giving the new field the same CEL guard `rule.apiGroups` already has (`!('*' in self)`); the field is new, so the guard breaks nothing. For the existing `apiGroups`/`apiVersions` inside `fieldPaths` only the weight change is safe.

Nit: 4/2/1 as named constants would make the intent readable without the comment.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/engine/engine_test.go
```diff
@@ -53,19 +53,100 @@ func TestSelectFieldPath(t *testing.T) {
...
+		{Resources: []string{"pods"}, Path: "$.byResource"},
+		{Path: "$.unscoped"},
+	}
+	if fp, ok := SelectFieldPath(fallback, "", "v1", "services"); !ok || fp.Path != "$.unscoped" {
+		t.Fatalf("unscoped fallback: path = %q ok=%v", fp.Path, ok)
+	}
+	// Equal weights → the earliest entry wins.
+	tie := []v1alpha1.FieldPath{
+		{Resources: []string{"pods"}, Path: "$.first"},
+		{Resources: []string{"*"}, Path: "$.second"},
```
Jabejixo: 🔵 **The tie test pins the `*` behaviour; a few more cases would help**

This case pins the `*` behaviour discussed above (`["pods"]` vs `["*"]` as a tie). If `*` stops scoring, the tie case needs two explicit names. Worth adding alongside: a list of two resources matching both; a resource-scoped entry whose group matches but resource does not, falling back to the unscoped one; an empty non-nil `Resources` behaving as unscoped.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/engine/engine.go
```diff
@@ -45,29 +45,46 @@ func RuleMatches(rule v1alpha1.UsageRule, group, version, resource string) bool
...
+// SelectFieldPath returns the FieldPath entry that applies to the given resource/group/version. An
+// empty dimension on an entry matches anything, so the unscoped entry is the fallback. Among the
+// matching entries the most specific one wins: Resources (weight 4) outranks APIGroups (2), which
+// outranks APIVersions (1), and the weights add up. Ties go to the earliest entry in the list, so the
+// choice is deterministic. The bool is false when no entry matches.
+//
+// The resource scope is what separates paths that collide within one group and version, e.g. core/v1
+// pods ($.spec.priorityClassName) versus replicationcontrollers
+// ($.spec.template.spec.priorityClassName).
+func SelectFieldPath(fieldPaths []v1alpha1.FieldPath, group, version, resource string) (v1alpha1.FieldPath, bool) {
```
Jabejixo, line 75: 🟡 **Entry scopes are not checked against `rule`; a typo is fail-open**

Not introduced here, but this is where it shows: the scope of an entry is never checked against `spec.rule`, and an unscoped fallback is recommended, not required. A typo (`pod`, `Pods`) with no fallback entry gives `ok=false`, and all three callers `continue`: `/is-granted` lets the request through, `/defaults` skips, the violations scan skips. That is fail-open, silently.

The validating webhook of #23227 is the natural place to refuse an entry whose `resources`/`apiGroups`/`apiVersions` are not a subset of `rule` (or to require an unscoped entry). Could be folded into that PR or a follow-up.

modules/160-multitenancy-manager/crds/multitenancy.deckhouse.io_grantableclusterresourcereferences.yaml
```diff
@@ -136,6 +136,14 @@ spec:
 
                         The path can also point to an annotation, for example, `$.metadata.annotations['cert-manager.io/cluster-issuer']`.
                       type: string
+                    resources:
```
Jabejixo, line 139: 🔵 **`resources` wants `minLength` and the same `*` guard as `rule.apiGroups`**

Suggest `items: {type: string, minLength: 1}` and an `x-kubernetes-validations` rule `!('*' in self)` mirroring `rule.apiGroups` above, see the comment on `SelectFieldPath`. `[""]` is a dead entry and `["*"]` currently ranks as the most specific scope.

Jabejixo (commented):
Reviewed together with #23227 and #23236 (the whole stack). Ran `go vet`, `go test ./...` and `-race` on the affected packages at the top of the stack: all green. Two behavioural points and a few doc/test gaps inline. Two things that belong to the description rather than a line:

- "the design note and the usage docs are updated": `docs/USAGE.md` / `USAGE_RU.md` are not in the diff, and `USAGE.md:1033` still sends the reader to the CRD for "resources with multiple API versions" only. Either add the resource case there or drop the claim.
- The behaviour change for objects that already carry two scoped entries (e.g. `[{apiVersions: [v1]}, {apiGroups: [batch]}]`: the first used to win, now the `apiGroups` one does) lives only in the commit message. It is not in the repository today, but it is a semantic change for external references, so the changelog entry should say it.

---
🔴 blocking · 🟡 should be fixed before merge · 🔵 nit or note

Jabejixo (approved):
Verified on the new head: `*` no longer adds weight, the CRD cost test is a good addition. USAGE follows after #23180, as agreed.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/engine/engine.go
AlwxSin: Fixed in 4a95d932f1. A dimension now adds weight only when its list is non-empty and has no `*`, so `*` counts as unrestricted everywhere — for `resources` and for the existing `apiGroups`/`apiVersions` alike, where only the weight changed and matching did not. The weights are named constants.

One deliberate difference from the suggestion: the new field has no CEL guard against `*`, see the thread on the CRD for why. With `*` weighing nothing, `resources: ["*"]` behaves exactly like an omitted field, so the guard is not needed for correctness.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/engine/engine_test.go
AlwxSin: Done in 4a95d932f1. The tie case now uses two explicit names. Added: `resources: ["*"]` loses to an explicit `apiGroups` + `apiVersions`; `apiGroups: ["*"]` ties with the unscoped entry in both orders; a list of two resources matches both; a resource-scoped entry whose group matches but resource does not falls back to the unscoped one; an empty non-nil `Resources` behaves as unscoped.

modules/160-multitenancy-manager/images/multitenancy-manager/src/internal/engine/engine.go
AlwxSin: Folded into #23227 (6ba156817c) as suggested, as two webhook checks rather than a mandatory fallback:

- *subset*: every value in an entry scope must be in the matching dimension of `spec.rule` (`*` in the rule admits anything, `*` in the entry restricts nothing);
- *coverage*: every (group, version, resource) the rule matches must be selected by some entry. A `*` in a rule dimension is replaced by a placeholder no explicit list contains, so only an entry unrestricted in that dimension covers it.

Coverage instead of a required fallback because scoped entries that together cover the rule are legitimate — `rule.resources: [jobs, cronjobs]` with one entry per resource needs no fallback — while a hole is exactly the silent fail-open. The same checks run in the reference reconciler and are reported as a `FieldPathsValid` condition, so an object stored before the webhook or while it was down says what is wrong with it. Both are ratcheted on UPDATE.

modules/160-multitenancy-manager/crds/multitenancy.deckhouse.io_grantableclusterresourcereferences.yaml
AlwxSin: `minLength: 1` added in 4a95d932f1. The CEL guard is deliberately not there. We added it first, and the API server refused the whole CRD on a test cluster:

```
fieldPaths.items.properties[resources].x-kubernetes-validations[0].rule:
Forbidden: estimated rule cost exceeds budget by factor of more than 100x
```

A rule on a list nested inside the `fieldPaths` list has no size bound to estimate against; the same rule on `rule.apiGroups` sits on a top-level list and fits. It blocked the Deckhouse main queue on `ModuleEnsureCRDs`, and neither unit tests nor CI noticed. Bounding it would have meant `maxItems` on `fieldPaths` — a new restriction on an existing field for a guard that is no longer needed, since `*` there now behaves like an omitted field.

To catch this class of problem earlier, 4a95d932f1 adds `modules/160-multitenancy-manager/crds/crds_test.go`: every CRD of the module goes through the API server's own `ValidateCustomResourceDefinition` (structural schema, CEL compilation, estimated cost against the server's limits). With the guard restored it fails with the same message the cluster gave.

AlwxSin:
Thanks for the review — every thread is answered inline. The two points from the summary of this PR:

- **USAGE claim.** The description no longer says the usage docs are updated. They are held back on purpose: the spellcheck job checks a changed page as a whole and fails on custom-resource names `USAGE.md` already carries, which #23180 adds to the dictionary. The usage text for the stack follows once it lands.
- **Behaviour change.** Now in the changelog entry as `impact`: where several scoped entries match, the most specific one applies instead of the first, and a dimension listing `*` counts as unrestricted.

The whole stack (4a95d932f1, 6ba156817c, 099a563470) was also run on a cluster on the `pr23236` image: the CRDs apply, and the new refusals, the coverage check, `FieldPathsValid` and the catalog sweep behave as described. That run is what caught the CEL cost problem described in the CRD thread.

z9r5 (approved):

Suselz (approved):

